PR #3, opened 2026-06-07
Add RESP3 protocol support with HELLO version negotiation

fatal10110:
## Summary
- Track per-session protocol version via `ClientSession.setProtocolVersion`/`protocolVersion`, negotiated through `HELLO` and reset to RESP2 by `RESET`
- Implement RESP3 encoding (`encodeResp3`) covering map, set, push, double, boolean, big-number, verbatim string, and null types, wired into `Resp2SessionAdapter` based on the session's negotiated version
- Update `HELLO` to return a RESP3 map when version 3 is negotiated and a flat array for RESP2
- Refresh README with Table of Contents, client connection examples (ioredis/node-redis), unit-test usage, and API reference; minor architecture-review doc updates
- Fix `buildRedisCluster` to start replica nodes with empty slot ranges instead of inheriting master slots

## Test plan
- [x] `npm test` — 87/87 unit tests pass (includes new RESP3 encoder coverage)
- [x] `tsc --noEmit` — type checks clean
- [ ] Manual smoke test of `HELLO 3` against a RESP3-aware client

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ In-memory Redis-compatible server implementation in JavaScript. Useful for testi
   - [As a CLI](#as-a-cli)
   - [CLI Options](#cli-options)
 - [Connecting Clients](#connecting-clients)
+  - [Protocol Version (RESP2 / RESP3)](#protocol-version-resp2--resp3)
   - [Connecting with ioredis](#connecting-with-ioredis)
   - [Connecting with node-redis](#connecting-with-node-redis)
   - [Using in Unit Tests](#using-in-unit-tests)
@@ -30,7 +31,7 @@ In-memory Redis-compatible server implementation in JavaScript. Useful for testi
 
 ## Features
 
-- **Redis-compatible protocol** - Works with any Redis client (ioredis, node-redis, etc.)
+- **RESP2 and RESP3 protocols** - Per-session version negotiation via `HELLO`
 - **Standalone and Cluster modes** - Run a single server or a full cluster
 - **Lua scripting support** - Execute Redis Lua scripts via WebAssembly
 - **No external dependencies** - Pure JavaScript, no Redis installation needed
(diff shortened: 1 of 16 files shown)


fatal10110:
## Code Review Findings

### 🔴 Bug — HELLO inside MULTI not rejected (`src/commands/connection.ts:392`)

`helloCommand` has no guard against being queued in a MULTI block. When EXEC runs a queued `HELLO 3`, `setProtocolVersion(3)` fires mid-transaction; `executeTransaction` then returns and the session-adapter encodes the EXEC array reply using the now-updated `session.protocolVersion=3`, sending RESP3 to a client still expecting RESP2 — corrupting the connection.

Real Redis returns `ERR Command not allowed inside a transaction`.

**Fix:** Check `ctx.session.mode === 'transaction'` at the top of `execute` and throw, or add a no-multi flag if the command framework supports one.

---

### 🔴 Bug — `CLIENT LIST`/`CLIENT INFO` hardcode `resp=2` (`src/commands/connection.ts:213`)

```ts
'resp=2',   // always 2, even after HELLO 3
```

`session.protocolVersion` is available but never read here. After `HELLO 3`, `CLIENT INFO` and `CLIENT LIST` report the wrong protocol version. node-redis reads this field to verify connection state.

**Fix:** `` `resp=${session.protocolVersion}` ``

---

### 🟡 Bug — `encodeResp3VerbatimString` no 3-byte format guard (`src/core/resp-encoder.ts:156`)

RESP3 verbatim string spec requires the format prefix to be **exactly 3 bytes**. No validation:

```ts
function encodeResp3VerbatimString(format: string, value: Buffer): Buffer {
  const payload = Buffer.concat([Buffer.from(`${format}:`), value])
  // if format.length !== 3, receiver misparses the colon boundary
```

A non-3-byte format string desynchronizes the receiver's stream parser.

**Fix:** Add `if (format.length !== 3) throw new Error(\`verbatim format must be 3 bytes, got ${format.length}\`)` at the top.

---

### 🟡 Bug — HELLO always reports `role: master` (`src/commands/connection.ts:416`)

```ts
[value('role'), value('master')],   // hardcoded
```

With replica support (PR #4), clients connecting directly to a replica and issuing `HELLO` receive `role=master`. RESP3-aware clients that use the HELLO `role` field for topology or routing decisions will misidentify the node.

Requires passing the current node's role through `RedisExecutionContext` — no quick fix without a small interface change.

---

_Reviewed by automated multi-angle analysis (line-by-line diff scan, removed-behavior audit, cross-file tracer). Findings verified before reporting._

PR #6, opened 2026-06-09
feat: stream commands MVP (XADD, XLEN, XRANGE, XREVRANGE, XDEL)

fatal10110:
## Summary

Implements the first Streams slice, backing the previously stubbed `RedisStreamData` (`{ type: 'stream' }`) with real storage and the core read/write/range commands.

## Commands

- **XADD** — explicit ids, `*` (auto ms+seq), and `<ms>-*` (auto seq). Enforces id > 0-0 and strict monotonicity vs the stream's retained last id.
- **XLEN**
- **XRANGE** / **XREVRANGE** — `-`/`+` open ends, exclusive `(` bounds, bare-`ms` seq defaulting (0 for start, max for end), and `COUNT`.
- **XDEL** — exact ids; the stream is **kept when emptied** and its last id is **retained**, matching Redis.

## Implementation

- `state/data-types.ts`: `RedisStreamData` holds ordered `entries` + `lastId`; new `StreamId` / `RedisStreamEntry`, `createStreamData`, and deep clone.
- `state/database.ts`: `getStream` / `updateStream` typed accessors.
- `core/redis-error.ts`: `InvalidStreamIdError`, `StreamIdEqualOrSmallerError`, `StreamIdNotGreaterThanZeroError` (messages match Redis).
- `commands/streams.ts`: the five commands + id parse/compare/format helpers. Mutations go through `updateStream` (note: `db.get` returns a defensive clone, so reads can't persist edits).

## Tests

`tests-integration/ioredis/stream-integration.test.ts` — 12 cases through the `TEST_BACKEND` switch (mock + real). Covers ids/monotonicity, range bounds + COUNT, reverse, XDEL keep-empty/retain-last-id, WRONGTYPE, and missing-key behavior.

- 89 unit + 164 mock integration pass. Real-backend run needs a Redis cluster up (designed to exact real-Redis semantics; not run here — no local cluster).

## Deferred to later slices

XREAD/blocking, consumer groups (XGROUP/XREADGROUP/XACK/XPENDING), XINFO, XTRIM/MAXLEN, NOMKSTREAM.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,116 +47,99 @@ npm run clean:redis
 
 ## Architecture Overview
 
+For full diagrams and a request-lifecycle walkthrough, see [docs/ARCHITECTURE.md](docs/ARCHITECTURE.md) — keep it in sync with any structural change.
+
 ### Core Architecture Pattern
 
-This is a Redis-compatible server implementation that supports both standalone and cluster modes. The architecture uses a **commander pattern** where commands are executed through a pipeline:
+Redis-compatible server (standalone + cluster modes) built as a layered pipeline. The **same** `CommandExecutor` pipeline drives standalone mode, cluster mode, `MULTI`/`EXEC` transactions, and Lua `EVAL` alike, so routing, queueing, and command semantics never diverge:
 
-1. **Transport Layer** - Handles RESP protocol encoding/decoding
-2. **Commander Layer** - Routes commands and manages execution context
-3. **DB Layer** - Manages in-memory data structures
-4. **Command Layer** - Individual command implementations
+1. **Transport Layer** - Frames RESP bytes on/off the wire (`SocketConnectionTransport` / `InMemoryConnectionTransport`)
+2. **Session Layer** - Per-connection state: selected DB, RESP version, transaction queue, `WATCH`ed keys (`ClientSession`)
+3. **Execution Layer** - Looks up commands, parses args, extracts routing keys, runs composable policies (`CommandExecutor`, `CommandRegistry`, `ExecutionPolicy`)
+4. **Command Layer** - Pure `(args, ctx) → RedisResult` implementations grouped by data type ([src/commands/](src/commands/))
+5. **State Layer** - In-memory keyspace, mutation events, cluster topology, script cache, pub/sub ([src/state/](src/state/))
 
 ### Key Components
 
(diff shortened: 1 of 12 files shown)


fatal10110:
## Review Findings

Three bugs in `src/commands/streams.ts`:

### 1. Wrong error type for non-integer COUNT (line ~516)

`createCountSchema` throws `InvalidStreamIdError` for a non-integer COUNT value. Real Redis returns `ERR value is not an integer or out of range`.

```typescript
// current (wrong)
if (!Number.isInteger(value)) {
  throw new InvalidStreamIdError()
}

// fix
if (!Number.isInteger(value) || value < 0) {
  throw new RedisSyntaxError()
}
```

### 2. Negative COUNT not rejected in XRANGE/XREVRANGE (line ~514)

`createCountSchema` has no `< 0` guard. `XRANGE key - + COUNT -1` silently returns all entries (the `count >= 0` condition in `rangeReply` is false, so the limit is skipped). Real Redis returns `ERR COUNT is not valid`.

Compare with XREAD schema (~line 424) which correctly validates: `if (!Number.isInteger(n) || n < 0) throw new RedisSyntaxError()` — the same logic belongs in `createCountSchema`.

Both findings 1 and 2 are fixed together with the snippet above.

### 3. seq overflow in `nextAutoId` / `nextSeqForMs` (lines 67, 73)

When `lastId.seq === MAX_UINT64` and the clock hasn't advanced, `lastId.seq + 1n` produces `2^64` — not a valid uint64. The entry is stored and serialized as `<ms>-18446744073709551616`. Real Redis handles this by advancing to `{ms: lastId.ms + 1, seq: 0}`.

```typescript
function nextAutoId(lastId: StreamId): StreamId {
  const now = BigInt(Date.now())
  if (now > lastId.ms) return { ms: now, seq: 0n }
  if (lastId.seq === MAX_UINT64) return { ms: lastId.ms + 1n, seq: 0n }  // add this
  return { ms: lastId.ms, seq: lastId.seq + 1n }
}

function nextSeqForMs(ms: bigint, lastId: StreamId): StreamId {
  if (ms > lastId.ms) return { ms, seq: 0n }
  if (ms === lastId.ms && lastId.seq === MAX_UINT64) return { ms: ms + 1n, seq: 0n }  // add this
  return { ms, seq: lastId.seq + 1n }
}
```

PR #7, opened 2026-06-09
feat: implement BLPOP, BRPOP, and XREAD BLOCK

fatal10110:
## Summary

- **BLPOP / BRPOP**: blocking list pop commands using `ctx.park()` to release the database turn while waiting, subscribing to key mutation events to wake on writes, retry-loop on wakeup to handle thundering herd, nil-array on timeout
- **XREAD BLOCK**: extends existing XREAD with optional `BLOCK <ms>` option; resolves `$` to the stream's current last ID at call time so only entries added _after_ the command started are returned; null bulk on timeout
- All three have `noscript` flag — blocking commands cannot run inside Lua scripts; the sync path would reject the returned Promise naturally, but the flag makes it explicit

## Implementation notes

The park infrastructure (`ctx.park`, `createTurnAwareParkHandler`, `SerialTurnQueue.suspend`) was already fully wired in the codebase — no plumbing changes needed. The blocking commands are the first consumers.

Pattern used by all three:
1. Try to pop/read immediately (synchronous, Lua-safe path)
2. If no data: subscribe to `db.subscribeKey` write events on all watched keys, park with optional timeout
3. On wakeup: unsubscribe, check for null (timeout → return nil) or retry pop/read
4. Loop if another client won the race

## Test plan

- [ ] `npm test` — 117 unit tests (13 new blocking command tests)
- [ ] `npm run test:integration:mock` — 188 integration tests (10 new blocking integration tests)
- [ ] `npm run test:integration:real` — verify against real Redis cluster (Docker Compose)

### DIFF
--- a/src/commands/index.ts
+++ b/src/commands/index.ts
@@ -122,6 +122,8 @@ export {
 } from './scan'
 export {
   listsCommands,
+  blpopCommand,
+  brpopCommand,
   lindexCommand,
   llenCommand,
   lpopCommand,
--- a/src/commands/lists.ts
+++ b/src/commands/lists.ts
@@ -385,7 +385,13 @@ async function blockingListPop(
       timeoutMs: remaining,
       signal: ctx.signal,
     })
-    for (const unsub of unsubs) unsub()
+    for (const unsub of unsubs) {
+      try {
+        unsub()
+      } catch {
+        // ignore errors from individual unsubscribers so all are attempted
+      }
+    }
 
     if (woken === null) return RedisResult.create(RedisValue.nullArray())
 
--- a/src/commands/lists.ts
+++ b/src/commands/lists.ts
@@ -380,16 +380,20 @@ async function blockingListPop(
       }),
     )
 
-    const woken = await ctx.park({
-      waitFor,
-      timeoutMs: remaining,
-      signal: ctx.signal,
-    })
-    for (const unsub of unsubs) {
-      try {
-        unsub()
-      } catch {
-        // ignore errors from individual unsubscribers so all are attempted
+    let woken: boolean | null
+    try {
+      woken = await ctx.park({
+        waitFor,
+        timeoutMs: remaining,
+        signal: ctx.signal,
+      })
+    } finally {
+      for (const unsub of unsubs) {
+        try {
+          unsub()
+        } catch {
+          // ignore errors from individual unsubscribers so all are attempted
+        }
       }
     }
 
--- a/src/commands/lists.ts
+++ b/src/commands/lists.ts
@@ -2,7 +2,15 @@ import { defineCommand } from '../core/command-definition'
 import { t } from '../core/command-schema'
 import { integer, bulk, ok, array } from './helpers'
 import { RedisValue } from '../core/redis-value'
-import { IndexOutOfRangeError, NoSuchKeyError } from '../core/redis-error'
+import { RedisResult } from '../core/redis-result'
+import {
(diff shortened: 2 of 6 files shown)


fatal10110:
Quick summary / recommendation
- This PR cleanly implements BLPOP, BRPOP and XREAD ... BLOCK (with unit + integration tests). The design uses the existing park/subscribe infrastructure and covers the main success, wake, and timeout paths. I recommend merging after addressing a few small correctness/clarity items and adding one extra test for thundering‑herd behavior.

What I looked at
- Changed files: `src/commands/lists.ts`, `src/commands/streams.ts`, `src/commands/index.ts`, `tests/commands-blocking.test.ts`, `tests-integration/ioredis/blocking-integration.test.ts`  
- Focus: blocking logic, subscribe/unsubscribe flow, timeout handling, resolving `$` for XREAD, `noscript` flags, and tests.

High-level notes
- `lists.ts`
  - Added `tryListPop`, `blockingListPop`, `blpop`, `brpop`. `blockingListPop` subscribes to key write events, parks the execution context, unsubscribes after wake, and retries in a loop (good for races).
- `streams.ts`
  - XREAD schema extended to accept `COUNT` and `BLOCK` (in any order). `BLOCK` is ms (integer).
  - `readStreamEntries` and `blockingXread` implement the same subscribe/park/retry pattern.
  - `xread` resolves `$` to the stream’s `lastId` at call time so only later entries are returned.
- Tests
  - Unit + integration tests cover immediate return, block/unblock on producers, fractional-second timeouts for lists, XREAD BLOCK semantics and COUNT.

Risks / items to address
1) Thundering‑herd / race behavior  
   - Add a test where multiple clients block on the same key and a single producer `LPUSH`/`RPUSH` occurs. Assert only one client returns the value and others continue waiting (or re-enter park correctly). This verifies the retry-loop under contention.

2) Defensive unsubscribe  
   - Wrap each `unsubscribe()` call in `try/catch` (or otherwise guard) so a thrown error from one unsubscriber doesn’t prevent others from being unsubscribed. This prevents leaking subscriptions if an unsubscribe unexpectedly throws.

3) XREAD COUNT semantics  
   - Current implementation treats `COUNT` as a per‑stream cap (break inside each stream’s loop). Confirm whether `COUNT` should be per-stream (current) or global across all streams. Add a test for `COUNT` across multiple streams to assert intended behavior and update logic if global `COUNT` is desired.

4) `subscribeKey` contract & error-handling  
   - Ensure `subscribeKey` always returns a valid unsubscribe function even on transient failures. Consider guarding subscribe and handling cases where subscription fails.

Minor notes / clarifications
- `BLPOP`/`BRPOP` accept seconds (floats allowed) and code converts to ms via `Math.ceil(timeoutSecs * 1000)` — tests already use fractional seconds, so this looks intentional/covered.
- `XREAD BLOCK` expects integer milliseconds (`BLOCK n`). `BLOCK 0` is treated as block-forever (implementation uses an undefined deadline). Consider documenting the difference in units between `BLPOP` (seconds) and `XREAD BLOCK` (ms) if not already in docs.
- Consider a brief comment in `streams.ts` expanding why `$` is resolved at call time (there is already a comment — maybe expand slightly for future readers).

Recommended critical actions before merge
- Add a thundering‑herd test (unit or integration) to exercise multiple simultaneous blockers on the same key.  
- Make unsubscribe defensive (try/catch around unsubscribe).  
- Add/clarify tests for `XREAD + COUNT` across multiple streams and adjust logic if `COUNT` should be global.

PR #121, opened 2026-06-10
fix: SELECT inside MULTI switches DB for later queued commands (#94)

### LINKED ISSUE #94: Bug: SELECT inside MULTI executes against wrong database — noBlockCtx.db frozen before EXEC loop
## Description

`executeTransaction` in `src/core/client-session.ts` builds a single `noBlockCtx` **before** the plan loop:

```typescript
const noBlockCtx = { ...this.createExecutionContext(), park: async () => null }
for (const plan of plans) {
  const result = await this.executor.executePlan(plan, noBlockCtx)
  ...
}
```

`createExecutionContext()` returns an object literal with `db: this.db` — a snapshot of the current database reference at call time. When a queued `SELECT N` runs, it calls `ctx.session.selectDatabase(N)`, correctly updating `this.selectedDatabaseId`, but `noBlockCtx.db` is already frozen to the pre-SELECT database object. Every subsequent plan in the same EXEC still reads and writes to the original database.

## Root Cause

`noBlockCtx` is created once. `db` is captured by value (not by getter) from `this.db` at that moment. SELECT mutates `selectedDatabaseId` but does not retroactively update `noBlockCtx.db`.

## Expected behaviour (real Redis, standalone mode)

```
MULTI
SELECT 1   → +QUEUED
SET k v    → +QUEUED
EXEC       → [OK, OK]    ← SET lands in DB 1
GET k      → "v"         (on DB 1)
```

## Actual behaviour (mock, standalone mode)

SET lands in DB 0 (the pre-SELECT database captured in `noBlockCtx.db`) even though SELECT ran first in EXEC.

## Notes

- Cluster mode is not affected: SELECT is not permitted in cluster mode and returns `-ERR SELECT is not allowed in cluster mode` during EXEC, so no DB switch occurs.
- Related: issue #82 (closed) confirmed SELECT is correctly queued (matching real Redis). This is the follow-on consequence: the queued SELECT's DB switch is silently ignored for all later commands in the same EXEC.

## Fix

Either re-call `this.createExecutionContext()` once per plan iteration (rebuilding `db` from live `this.selectedDatabaseId` each time), or replace the snapshot `db` in `noBlockCtx` with a getter proxy that always reads `this.selectedDatabaseId`.

fatal10110:
## Summary

Fixes #94. A queued `SELECT N` inside `MULTI` updated `selectedDatabaseId` mid-`EXEC`, but `executeTransaction` had snapshotted `db` by value into a single `noBlockCtx` built before the plan loop. Every command queued after the `SELECT` kept writing to the pre-`SELECT` database.

## Fix

Expose `noBlockCtx.db` as a live getter delegating to `this.db` (which resolves `selectedDatabaseId` on each access), so commands queued after a `SELECT` see the newly selected database. Matches real Redis standalone semantics; cluster mode is unaffected (`SELECT` is rejected there).

```
MULTI
SELECT 1   → +QUEUED
SET k v    → +QUEUED
EXEC       → [OK, OK]   ← SET now lands in DB 1, not DB 0
```

## Tests

- **Unit:** regression test in `tests/commands-transactions.test.ts`.
- **Integration (new):** `SELECT` is rejected in cluster mode, so the existing cluster-only harness couldn't reach this. Added a standalone path to `TestRunner` (`setupIoredisStandalone`) — mock backend spins an in-process `Resp2Server` with 16 DBs; real backend spawns a real `redis-server` child on a free port. New ioredis test `tests-integration/ioredis/select-multi.test.ts` asserts the queued `SELECT` switches the DB for later `EXEC` commands.
- Both tests verified red→green against the pre-fix `client-session.ts`.
- Full suite green: unit (122), mock integration (191), **real** Redis integration (191).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,12 +86,16 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Start Redis cluster
+      - name: Start Redis (cluster + standalone)
         run: docker compose -f docker-compose.test.yml up -d --wait
 
       - name: Run integration tests (real backend)
+        env:
+          # Points the harness at the standalone service for SELECT/multi-DB
+          # tests that cluster mode cannot serve (see test-config.ts).
+          REDIS_STANDALONE_PORT: 6399
         run: npm run test:integration:real
 
-      - name: Stop Redis cluster
+      - name: Stop Redis (cluster + standalone)
         if: always()
         run: docker compose -f docker-compose.test.yml down -v
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -25,3 +25,18 @@ services:
       timeout: 5s
       retries: 30
       start_period: 20s
+
+  # Standalone (non-cluster) Redis. Cluster mode rejects SELECT and exposes a
+  # single logical database, so integration tests that need multiple databases
+  # or SELECT (e.g. SELECT inside MULTI/EXEC) connect here instead. The harness
+  # uses it when REDIS_STANDALONE_PORT is set; see tests-integration/test-config.ts.
+  redis-standalone:
+    image: redis:7.2
+    ports:
+      - '6399:6379'
+    healthcheck:
+      test: ['CMD', 'redis-cli', '-p', '6379', 'ping']
+      interval: 5s
+      timeout: 5s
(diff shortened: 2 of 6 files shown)


fatal10110:
@qodo-code-review review

PR #148, opened 2026-06-12
fix: flush valid pipelined responses before protocol error (#79)

### LINKED ISSUE #79: Bug: a bad frame in a pipeline drops all preceding frames' responses and closes the connection
## Description

When a client sends multiple commands in one TCP write (pipelining) and one of the frames is malformed, all responses for the valid preceding frames are discarded and the connection is closed after a single error reply.

## Root Cause

`decoder.push()` in `src/core/transports/resp2/decoder.ts` throws synchronously on a bad frame, abandoning the already-collected `frames` array. The session adapter catches the throw, sends one `-ERR` reply, and closes the connection. The good frames never reach their command handlers.

## Expected behaviour (real Redis)

Real Redis sends one response per successfully-parsed frame in order, then sends `-ERR Protocol error` for the bad frame, then closes the connection. The client can correlate which commands succeeded before the error.

## Impact

Clients using pipelining lose all responses in the same TCP chunk as a bad frame — they cannot determine which prior commands executed, making error recovery impossible.

fatal10110:
## Summary
- A bad frame in a RESP2 pipeline dropped the responses for all valid commands that preceded it and closed the connection.
- Root cause: `Resp2CommandDecoder.push()` threw synchronously on the malformed frame, discarding the `frames` array already collected for the valid earlier commands. `Resp2SessionAdapter` caught the throw, sent a single `-ERR Protocol error`, and closed — so the client never saw replies for the commands that succeeded and couldn't correlate which ones did.
- Fix: `push()` now returns `{ frames, error? }` (new `Resp2DecodeResult` type) instead of throwing — it preserves the valid frames and surfaces the parse error alongside them. `Resp2SessionAdapter.run()` answers the valid frames first, then writes the protocol error and closes, matching real Redis (one reply per valid frame → `-ERR Protocol error...` → close).

Closes #79

## Test plan
- [x] New unit test in `tests/core-transport-session.test.ts` feeds a pipeline of two valid commands + a malformed frame; asserts `+OK\r\n$5\r\nvalue\r\n-ERR Protocol error: invalid bulk length\r\n` and that the connection closed (red before the fix — only the error was written).
- [x] Unit (125) + mock integration (210) + real integration (210) all pass via `npm run test:all`.
- [x] `npm run build` and `npm run lint` clean (1 pre-existing unrelated warning).

> Note: this is a raw-bytes protocol bug — real clients (ioredis/node-redis) never emit malformed frames and the mock backend uses an in-memory transport (no socket), so it's covered at the transport/decoder unit level rather than via the client integration suites.

### DIFF
--- a/src/core/transports/resp2/decoder.ts
+++ b/src/core/transports/resp2/decoder.ts
@@ -3,6 +3,11 @@ export type Resp2CommandFrame = {
   args: Buffer[]
 }
 
+export type Resp2DecodeResult = {
+  frames: Resp2CommandFrame[]
+  error?: Resp2ParseError
+}
+
 export class Resp2ParseError extends Error {
   constructor(message: string) {
     super(message)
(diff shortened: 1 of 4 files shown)


fatal10110:
Rebased onto `main` (now includes the raw-TCP harness from #149) and added a **raw-TCP integration test** for this fix: `tests-integration/raw-tcp/protocol-errors.test.ts`.

It opens a bare socket, writes `SET … / GET … / *1\r\n$abc\r\n` in one TCP write, and asserts the server answers `+OK` and `$5\r\nvalue` first, then sends `-ERR Protocol error: invalid bulk length` and closes. It passes against **both** the mock server and **real Redis** — confirming real Redis behaves identically and that the bug was ours.

So #79 now has both the transport-level unit test and a client-impossible integration test. `npm run test:all`: unit 125/125, mock integration 216/216, real integration 216/216.

PR #174, opened 2026-06-13
Add Redis Pub/Sub support

### LINKED ISSUE #17: Feature missing: Pub/Sub system (SUBSCRIBE, PUBLISH, PSUBSCRIBE, PUBSUB) completely absent
## Description

`RedisPubSubBroker` in `src/state/pubsub-broker.ts` is a 5-line stub. No pub/sub commands are registered. Any client attempting pub/sub workflows fails silently or with unknown-command errors.

## Current state

```ts
// src/state/pubsub-broker.ts — entire file
export class RedisPubSubBroker {
  publish(_channel: Buffer, _message: Buffer): number {
    return 0   // always 0, no subscribers, no delivery
  }
}
```

## Missing Commands

- `SUBSCRIBE channel [channel ...]`
- `UNSUBSCRIBE [channel ...]`
- `PSUBSCRIBE pattern [pattern ...]`
- `PUNSUBSCRIBE [pattern ...]`
- `PUBLISH channel message`
- `PUBSUB CHANNELS [pattern]`
- `PUBSUB NUMSUB [channel ...]`
- `PUBSUB NUMPAT`
- `PUBSUB SHARDCHANNELS` / `PUBSUB SHARDNUMSUB` (Redis 7.0+)

## Impact

- `PUBLISH` succeeds and returns `0` silently — published messages are silently dropped with no error
- `SUBSCRIBE` returns unknown command error — any queue/event pattern breaks
- `ClientSession.mode` has a `'subscribed'` variant (defined but never set) — dead code
- `CLIENT LIST` `sub` and `psub` fields are hardcoded to `0`

## Notes

Subscribed mode has specific connection semantics in Redis: once subscribed, the client can only issue `SUBSCRIBE`, `UNSUBSCRIBE`, `PSUBSCRIBE`, `PUNSUBSCRIBE`, `PING`, `RESET`, and `QUIT`. All other commands must return an error. The session adapter needs to enforce this mode once the first `SUBSCRIBE`/`PSUBSCRIBE` is processed.

### LINKED ISSUE #92: Bug: pub/sub subscribed-mode command gating is not enforced
## Description

Real Redis restricts a connection in subscribed mode to only seven commands: `SUBSCRIBE`, `UNSUBSCRIBE`, `PSUBSCRIBE`, `PUNSUBSCRIBE`, `PING`, `RESET`, and `QUIT`. All other commands must be rejected with `-ERR Command not allowed in subscribed state`.

The mock has no `ExecutionPolicy` that checks subscribed-mode state. Because SUBSCRIBE is unimplemented (#17), no session enters subscribed mode today — but the enforcement layer is entirely absent. When pub/sub is implemented, every arbitrary command will silently execute in subscribed mode.

## Fix

Add an execution policy (or a check in `CommandExecutor`) that rejects commands when `session.mode === 'subscribed'` unless the command is in the allowed set.

### LINKED ISSUE #98: Pub/Sub: SUBSCRIBE/UNSUBSCRIBE/PSUBSCRIBE/PUNSUBSCRIBE + subscribed-mode session enforcement
Part of #17.

## Scope
- `SUBSCRIBE channel [channel ...]`
- `UNSUBSCRIBE [channel ...]`
- `PSUBSCRIBE pattern [pattern ...]`
- `PUNSUBSCRIBE [pattern ...]`
- `ClientSession.mode` 'subscribed' variant — currently defined but never set
- Enforce subscribed-mode command gating: while subscribed, only `SUBSCRIBE`/`UNSUBSCRIBE`/`PSUBSCRIBE`/`PUNSUBSCRIBE`/`PING`/`RESET`/`QUIT` are allowed; everything else must error
- Also resolves #92 (pub/sub subscribed-mode command gating not enforced)

## Notes
This is the foundational session/connection-state slice — `PUBLISH` delivery and `PUBSUB` introspection build on top of the subscriber registry introduced here.

### LINKED ISSUE #99: Pub/Sub: PUBLISH + RedisPubSubBroker fan-out implementation
Part of #17.

## Scope
- Replace the 5-line `RedisPubSubBroker` stub in `src/state/pubsub-broker.ts` (currently `publish()` always returns 0, no subscribers, no delivery)
- Maintain a channel subscriber registry (exact match) and pattern subscriber registry (glob match) fed by SUBSCRIBE/PSUBSCRIBE
- `PUBLISH channel message` delivers to all matching channel + pattern subscribers and returns the number of clients that received the message
- Update `CLIENT LIST`/`CLIENT INFO` `sub`/`psub` fields to reflect real subscription counts (currently hardcoded to `0`)

## Depends on
Subscriber registry from the SUBSCRIBE/UNSUBSCRIBE sub-issue.

### LINKED ISSUE #100: Pub/Sub: PUBSUB introspection commands (CHANNELS, NUMSUB, NUMPAT, SHARDCHANNELS, SHARDNUMSUB)
Part of #17.

## Scope
- `PUBSUB CHANNELS [pattern]`
- `PUBSUB NUMSUB [channel ...]`
- `PUBSUB NUMPAT`
- `PUBSUB SHARDCHANNELS` (Redis 7.0+)
- `PUBSUB SHARDNUMSUB [shardchannel ...]` (Redis 7.0+)

## Depends on
Subscriber registry from the SUBSCRIBE/UNSUBSCRIBE sub-issue.

fatal10110:
## Summary

- Add Redis Pub/Sub command support for `PUBLISH`, channel subscriptions, pattern subscriptions, and `PUBSUB` introspection.
- Track per-session subscription state, enforce Redis subscribed-mode command restrictions, and deliver server-initiated Pub/Sub frames through the RESP adapter.
- Document the implemented Pub/Sub surface and remaining shard subscription gap.

## Sub-Issues Covered

- Fixes #98: `SUBSCRIBE`/`UNSUBSCRIBE`/`PSUBSCRIBE`/`PUNSUBSCRIBE` and subscribed-mode session enforcement.
- Fixes #99: `PUBLISH` plus `RedisPubSubBroker` channel/pattern fan-out and `CLIENT INFO`/`CLIENT LIST` subscription counts.
- Fixes #100: `PUBSUB CHANNELS`, `NUMSUB`, `NUMPAT`, `SHARDCHANNELS`, and `SHARDNUMSUB` introspection.
- Fixes #92: subscribed-mode command gating for non-Pub/Sub commands.

## Root Cause

`RedisPubSubBroker` was a stub that always returned zero subscribers, and no Pub/Sub commands were registered. `ClientSession` had a `subscribed` mode type but no subscription lifecycle, outbound push queue, or subscribed-mode policy.

## Validation

- `TEST_BACKEND=mock node --enable-source-maps --import tsx --no-warnings --test ./tests-integration/ioredis/pubsub-integration.test.ts`
- `TEST_BACKEND=real node --enable-source-maps --import tsx --no-warnings --test-concurrency 1 --test-timeout 30000 --test ./tests-integration/ioredis/pubsub-integration.test.ts`
- `TEST_BACKEND=mock node --enable-source-maps --import tsx --no-warnings --test ./tests-integration/raw-tcp/pubsub-protocol.test.ts`
- `TEST_BACKEND=real node --enable-source-maps --import tsx --no-warnings --test-concurrency 1 --test-timeout 30000 --test ./tests-integration/raw-tcp/pubsub-protocol.test.ts`
- `npm run build`
- `npm test`
- `npm run lint` (passes with existing `src/types/respjs.d.ts` `no-explicit-any` warning)

Fixes #17

### DIFF
--- a/src/commands/connection.ts
+++ b/src/commands/connection.ts
@@ -324,10 +324,7 @@ export const pingCommand = defineCommand({
   flags: ['readonly', 'fast', 'subscribed'],
   keys: () => [],
   execute: (args, ctx) => {
-    if (
-      ctx.session.mode === 'subscribed' &&
-      ctx.session.protocolVersion === 2
-    ) {
+    if (ctx.session.usesSubscribedReplyMode) {
       return RedisResult.create(
         RedisValue.push('pong', [
           RedisValue.bulkString(Buffer.from(args.message ?? '')),
--- a/src/commands/connection.ts
+++ b/src/commands/connection.ts
@@ -201,9 +201,9 @@ function formatClientLine(
     `db=${ctx.session.selectedDatabase}`,
     'age=0',
     'idle=0',
-    'flags=N',
-    'sub=0',
-    'psub=0',
+    `flags=${session.mode === 'subscribed' ? 'P' : 'N'}`,
+    `sub=${session.pubsubChannelCount}`,
+    `psub=${session.pubsubPatternCount}`,
     'multi=-1',
     'qbuf=0',
     'qbuf-free=0',
@@ -321,9 +321,17 @@ export const pingCommand = defineCommand({
   schema: t.object({
     message: t.optional(t.bulk()),
   }),
-  flags: ['readonly', 'fast'],
+  flags: ['readonly', 'fast', 'subscribed'],
   keys: () => [],
-  execute: args => {
+  execute: (args, ctx) => {
+    if (ctx.session.mode === 'subscribed') {
+      return RedisResult.create(
+        RedisValue.push('pong', [
+          RedisValue.bulkString(Buffer.from(args.message ?? '')),
+        ]),
+      )
+    }
+
     if (args.message) {
       return bulk(args.message)
     }
@@ -335,7 +343,7 @@ export const pingCommand = defineCommand({
 export const quitCommand = defineCommand({
   name: 'quit',
   schema: t.object({}),
-  flags: ['readonly', 'fast'],
+  flags: ['readonly', 'fast', 'subscribed'],
   keys: () => [],
   execute: () =>
     RedisResult.create(RedisValue.simpleString('OK'), { close: true }),
@@ -530,12 +538,13 @@ export const authCommand = defineCommand({
 export const resetCommand = defineCommand({
   name: 'reset',
   schema: t.object({}),
-  flags: ['admin'],
+  flags: ['admin', 'subscribed'],
   keys: () => [],
   execute: (_args, ctx) => {
     clientNames.delete(ctx.session)
     clientLibraryNames.delete(ctx.session)
     clientLibraryVersions.delete(ctx.session)
+    ctx.session.resetPubSub()
     ctx.session.discardTransaction()
     ctx.session.unwatch()
     ctx.session.selectDatabase(0)
--- a/src/commands/connection.ts
+++ b/src/commands/connection.ts
@@ -324,7 +324,10 @@ export const pingCommand = defineCommand({
   flags: ['readonly', 'fast', 'subscribed'],
   keys: () => [],
   execute: (args, ctx) => {
-    if (ctx.session.mode === 'subscribed') {
+    if (
+      ctx.session.mode === 'subscribed' &&
+      ctx.session.protocolVersion === 2
+    ) {
       return RedisResult.create(
         RedisValue.push('pong', [
           RedisValue.bulkString(Buffer.from(args.message ?? '')),
--- a/src/commands/pubsub.ts
+++ b/src/commands/pubsub.ts
@@ -1,0 +1,285 @@
+import { defineCommand } from '../core/command-definition'
+import { t } from '../core/command-schema'
+import {
+  RedisCommandError,
+  WrongNumberOfArgumentsError,
+} from '../core/redis-error'
+import { RedisResult } from '../core/redis-result'
+import { RedisValue } from '../core/redis-value'
+import type { RedisExecutionContext } from '../core/redis-context'
+import type { ResponseStream } from '../core/response-stream'
+import { array, integer } from './helpers'
+import { commandSubcommandInfo } from './introspection'
+
+type PubSubArgs = {
+  subcommand: string
+  args: Buffer[]
+}
+
+export const subscribeCommand = defineCommand({
+  name: 'subscribe',
+  schema: t.object({
+    channels: t.variadic(t.bulk(), { min: 1 }),
+  }),
+  flags: ['pubsub', 'noscript', 'subscribed'],
+  introspection: {
+    arity: -2,
+    flags: ['pubsub', 'noscript', 'loading', 'stale'],
+    firstKey: 0,
+    lastKey: 0,
+    keyStep: 0,
+    categories: ['@pubsub', '@slow'],
+  },
+  keys: () => [],
+  execute: (args, ctx) =>
+    framesResult(ctx.session.subscribePubSubChannels(args.channels)),
+})
+
+export const unsubscribeCommand = defineCommand({
+  name: 'unsubscribe',
+  schema: t.object({
+    channels: t.variadic(t.bulk()),
+  }),
+  flags: ['pubsub', 'noscript', 'subscribed'],
+  introspection: {
+    arity: -1,
+    flags: ['pubsub', 'noscript', 'loading', 'stale'],
+    firstKey: 0,
+    lastKey: 0,
+    keyStep: 0,
+    categories: ['@pubsub', '@slow'],
+  },
+  keys: () => [],
+  execute: (args, ctx) =>
+    framesResult(ctx.session.unsubscribePubSubChannels(args.channels)),
+})
+
+export const psubscribeCommand = defineCommand({
+  name: 'psubscribe',
+  schema: t.object({
+    patterns: t.variadic(t.bulk(), { min: 1 }),
+  }),
+  flags: ['pubsub', 'noscript', 'subscribed'],
+  introspection: {
+    arity: -2,
+    flags: ['pubsub', 'noscript', 'loading', 'stale'],
+    firstKey: 0,
+    lastKey: 0,
+    keyStep: 0,
+    categories: ['@pubsub', '@slow'],
+  },
+  keys: () => [],
+  execute: (args, ctx) =>
+    framesResult(ctx.session.subscribePubSubPatterns(args.patterns)),
+})
+
+export const punsubscribeCommand = defineCommand({
+  name: 'punsubscribe',
+  schema: t.object({
+    patterns: t.variadic(t.bulk()),
+  }),
+  flags: ['pubsub', 'noscript', 'subscribed'],
+  introspection: {
+    arity: -1,
+    flags: ['pubsub', 'noscript', 'loading', 'stale'],
+    firstKey: 0,
+    lastKey: 0,
+    keyStep: 0,
+    categories: ['@pubsub', '@slow'],
+  },
+  keys: () => [],
+  execute: (args, ctx) =>
+    framesResult(ctx.session.unsubscribePubSubPatterns(args.patterns)),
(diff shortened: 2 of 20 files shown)


src/core/execution-policies/subscribed-policy.ts
```diff
@@ -0,0 +1,22 @@
+import type { ExecutionPolicy } from './index'
+import { RedisResult } from '../redis-result'
+
+export function createSubscribedModePolicy(): ExecutionPolicy {
+  return {
+    name: 'subscribed-mode',
+    beforeExecute(plan, ctx) {
+      if (ctx.session.mode !== 'subscribed') {
```
fatal10110: This policy also applies after the client negotiates RESP3, but Redis does not keep RESP3 clients in the RESP2 command-restricted subscribed mode. Repro against Redis 7.2.1: `HELLO 3`, `SUBSCRIBE ch`, then `GET missing` returns RESP3 null (`_`). With this PR the same sequence returns `ERR Can't execute 'get'...`, so RESP3-aware clients that keep using the subscribed connection will break. The restriction needs to be scoped to RESP2 sessions, or RESP3 sessions need to bypass it.

src/commands/connection.ts
```diff
@@ -321,9 +321,17 @@ export const pingCommand = defineCommand({
   schema: t.object({
     message: t.optional(t.bulk()),
   }),
-  flags: ['readonly', 'fast'],
+  flags: ['readonly', 'fast', 'subscribed'],
   keys: () => [],
-  execute: args => {
+  execute: (args, ctx) => {
+    if (ctx.session.mode === 'subscribed') {
```
fatal10110: This subscribed-mode `PING` reply shape is only correct for RESP2. Under RESP3, Redis still allows commands while subscribed and `PING hello` returns the normal bulk reply (`$5\r\nhello\r\n`), not a push frame. With this branch the PR emits `>2 ... pong ... hello`, which changes the wire protocol for RESP3 subscribers. Please gate the push-style `pong` response on `ctx.session.protocolVersion === 2`.

fatal10110 (commented):
Found two RESP3 Pub/Sub compatibility issues during review. Both reproduce after `HELLO 3` followed by `SUBSCRIBE`.

src/core/execution-policies/subscribed-policy.ts
fatal10110: **RESP3: subscribed-mode gating should not apply.** This rejects every non-pubsub command whenever `mode === 'subscribed'`, regardless of protocol version. Real Redis only enforces the restricted command set when `c->resp == 2`; under RESP3 a subscribed client may run *any* command. Trigger: `HELLO 3` → `SUBSCRIBE ch` → `GET k` returns `ERR Can't execute 'get'...` instead of executing the GET. Also gate on `ctx.session.protocolVersion === 2`.

src/commands/connection.ts
fatal10110: **RESP3: PING should return `+PONG`, not the push array.** This returns the `["pong", msg]` multi-bulk whenever subscribed, ignoring protocol version. Redis only returns the array form when `c->resp == 2 && CLIENT_PUBSUB`; under RESP3 it replies with the normal `+PONG` (or bulk of the message). Trigger: `HELLO 3` → `SUBSCRIBE ch` → `PING` yields a `>2 pong ...` push instead of `+PONG`. Add a `ctx.session.protocolVersion === 2` condition.

src/state/pubsub-broker.ts
```diff
@@ -1,5 +1,289 @@
...
+    }
+    return count
+  }
+}
+
+function pubsubKey(value: Buffer): string {
+  return value.toString('hex')
+}
+
+function redisGlobMatch(
```
fatal10110: **Duplicated glob matcher.** `redisGlobMatch`, `matchCharacterClass`, and the `STAR`/`OPEN_BRACKET`/... constants are a near-verbatim ~76-line copy of the same functions already in `src/commands/scan.ts` (used by SCAN/HSCAN MATCH). The two copies have already diverged: this one uses `nextValueCursor <= value.length` (line 165) where scan.ts uses `<`. Extract one shared helper (e.g. `src/core/glob.ts`) and import it in both places so the pattern-match semantics can't drift apart.

src/commands/pubsub.ts
```diff
@@ -0,0 +1,285 @@
...
+  subcommand: string
+  args: Buffer[]
+}
+
+export const subscribeCommand = defineCommand({
+  name: 'subscribe',
+  schema: t.object({
+    channels: t.variadic(t.bulk(), { min: 1 }),
+  }),
+  flags: ['pubsub', 'noscript', 'subscribed'],
```
fatal10110, line 24: **SUBSCRIBE/UNSUBSCRIBE/PSUBSCRIBE/PUNSUBSCRIBE are not rejected inside MULTI.** They carry no `no-multi` marker, so inside a transaction they are queued and executed at `EXEC` — emitting push frames and flipping the session into subscribed mode mid-EXEC. Real Redis marks these `CMD_NO_MULTI` and replies `ERR SUBSCRIBE is not allowed in transactions` at queue time. Consider a flag the transaction policy can reject on.

src/core/execution-policies/subscribed-policy.ts
fatal10110: Fixed in 55fdc22 by scoping the subscribed-mode command restriction to RESP2 sessions only. RESP3 subscribed clients now bypass the restriction, and the raw protocol test covers HELLO 3 -> SUBSCRIBE -> GET missing returning RESP3 null.

fatal10110: Fixed in 55fdc22. The subscribed-mode policy now gates the RESP2-only restriction with ctx.session.protocolVersion === 2, and the raw protocol test covers HELLO 3 -> SUBSCRIBE -> GET missing executing normally.

src/commands/connection.ts
fatal10110: Fixed in 55fdc22 by limiting the Pub/Sub push-style pong reply to RESP2 subscribed sessions. RESP3 subscribed PING now returns the normal PONG / message reply shape, covered by the raw protocol regression test.

fatal10110: Fixed in 55fdc22. The subscribed PING special-case is now RESP2-only, so RESP3 subscribed clients get the normal PONG or message bulk reply. Added raw protocol coverage for HELLO 3 -> SUBSCRIBE -> PING and PING hello.

src/state/pubsub-broker.ts
fatal10110: Fixed in 55fdc22 by extracting the matcher to src/core/glob.ts and importing it from both SCAN and the Pub/Sub broker. Added tests/core-glob.test.ts coverage for wildcards, character classes, reversed ranges, and escaped glob tokens.

src/commands/pubsub.ts
fatal10110: I checked this before changing it, and Redis 7.2.1 does not appear to reject these at queue time. On a single connection, MULTI -> SUBSCRIBE ch -> EXEC replies OK, QUEUED, then the subscribe acknowledgement. COMMAND INFO SUBSCRIBE on the same Redis 7.2.1 build reports pubsub, noscript, loading, stale, with no no_multi flag. I left this behavior unchanged to stay aligned with Redis 7.2.1; happy to adjust if this PR should target different Redis behavior or there is another repro I should validate.

fatal10110: You're right, and my repro on Redis 7.2 agrees with yours — conceding this one.

```
COMMAND INFO subscribe -> denyoom pubsub noscript loading stale   (no no-multi flag)
MULTI / SUBSCRIBE testchan / EXEC -> OK, QUEUED, [subscribe, testchan, 1]
PSUBSCRIBE behaves the same
```

Subscribe/psubscribe are queued and executed inside MULTI, not rejected. My original note was based on a stale memory of a `CMD_NO_MULTI` flag that does not apply to these commands in 7.2. The PR's queue-and-execute behavior is the correct match for real Redis. No change needed.

PR #175, opened 2026-06-14
Add stream consumer group commands

### LINKED ISSUE #19: Feature missing: Stream consumer groups (XGROUP, XREADGROUP, XACK, XPENDING, etc.) entirely absent
## Description

Only 7 basic stream commands are implemented. All consumer-group commands are missing. Consumer groups are the primary use case for Redis Streams (reliable at-least-once message delivery with acknowledgement).

## Missing Commands

| Command | Description |
|---------|-------------|
| `XGROUP CREATE` | Create a consumer group |
| `XGROUP SETID` | Set last-delivered ID for a group |
| `XGROUP DESTROY` | Delete a consumer group |
| `XGROUP CREATECONSUMER` | Explicitly create a consumer |
| `XGROUP DELCONSUMER` | Delete a consumer |
| `XREADGROUP GROUP group consumer STREAMS key id` | Read as a consumer group member |
| `XACK key group id [id ...]` | Acknowledge processed messages |
| `XPENDING key group [IDLE min-idle-time] min max count [consumer]` | Inspect pending messages |
| `XCLAIM key group consumer min-idle-time id [id ...]` | Transfer ownership of pending messages |
| `XAUTOCLAIM key group consumer min-idle-time start [COUNT count]` | Auto-claim idle messages |
| `XINFO GROUPS key` | List consumer groups |
| `XINFO CONSUMERS key group` | List consumers in a group |
| `XINFO STREAM key [FULL [COUNT count]]` | Full stream introspection |

## Impact

Any application using streams for:
- Task queues with worker acknowledgement
- Event sourcing with consumer groups
- At-least-once delivery guarantees

…is completely broken. These use cases represent the overwhelming majority of Redis Streams adoption.

## Data Model Changes Needed

`RedisStreamData` in `src/state/data-types.ts` needs to track:
- Consumer groups with last-delivered ID
- Pending Entry List (PEL) per group per consumer
- Consumer last-seen timestamps and idle time

## Related

`XREAD BLOCK` is also missing (tracked separately).

fatal10110:
## Summary
- Add stream consumer-group state to `RedisStreamData`, including groups, consumers, pending entries, delivery counts, and stream metadata for `XINFO`.
- Implement `XGROUP`, `XREADGROUP`, `XACK`, `XPENDING`, `XCLAIM`, `XAUTOCLAIM`, and `XINFO STREAM|GROUPS|CONSUMERS` through the existing command executor pipeline.
- Add Redis-validated ioredis integration coverage for group lifecycle, delivery/acknowledgement, pending inspection, claim/autoclaim, introspection, and representative error paths.
- Update stream command documentation and architecture/guidance notes for the new state model.

## Root Cause
Issue #19 was caused by the stream data model only tracking entries and generated IDs; there was no durable consumer-group or pending-entry-list state for stream group commands to operate on.

## Validation
- `TEST_BACKEND=mock node --enable-source-maps --import tsx --no-warnings --test ./tests-integration/ioredis/stream-integration.test.ts`
- `TEST_BACKEND=real node --enable-source-maps --import tsx --no-warnings --test-concurrency 1 --test-timeout 30000 --test ./tests-integration/ioredis/stream-integration.test.ts`
- `npm run build`
- `npm test`
- `npm run lint` (passes with existing `src/types/respjs.d.ts` no-explicit-any warning)

Fixes #19

### DIFF
--- a/src/commands/streams.ts
+++ b/src/commands/streams.ts
@@ -294,6 +294,13 @@ function entryToReply(id: StreamId, fields: Buffer[]): RedisValue {
   ])
 }
 
+function deletedEntryToReply(id: StreamId): RedisValue {
+  return RedisValue.array([
+    RedisValue.bulkString(Buffer.from(formatStreamId(id))),
+    RedisValue.bulkString(null),
+  ])
+}
+
 type FieldList = Buffer[]
 
 // Parses the trailing `field value [field value ...]` of XADD into a flat
@@ -1108,8 +1115,11 @@ function readGroupEntries(
           if (compareStreamId(pending.id, id) <= 0) continue
 
           const entry = findEntry(stream, pending.id)
-          if (!entry) continue
-          replies.push(entryToReply(entry.id, entry.fields))
+          replies.push(
+            entry
+              ? entryToReply(entry.id, entry.fields)
+              : deletedEntryToReply(pending.id),
+          )
 
           if (count !== null && count > 0 && replies.length >= count) break
         }
@@ -1118,7 +1128,7 @@ function readGroupEntries(
       return replies
     })
 
-    if (entries.length > 0) {
+    if (entries.length > 0 || id !== '>') {
       results.push([bulkString(key), RedisValue.array(entries)])
     }
   }
@@ -1839,17 +1849,17 @@ function streamInfoReply(
     return fields
   }
 
-  const entryLimit = count ?? stream.entries.length
+  const fullCount = count ?? 10
   fields.push(
     RedisValue.array(
       Array.from(stream.groups.values(), group =>
-        fullGroupInfoReply(stream, group),
+        fullGroupInfoReply(stream, group, fullCount),
       ),
     ),
     bulkString('entries'),
     RedisValue.array(
       stream.entries
-        .slice(0, entryLimit)
+        .slice(0, fullCount)
         .map(entry => entryToReply(entry.id, entry.fields)),
     ),
   )
@@ -1877,6 +1887,7 @@ function groupInfoReply(stream: RedisStreamData) {
 function fullGroupInfoReply(
   stream: RedisStreamData,
   group: RedisStreamConsumerGroup,
+  count: number,
 ): RedisValue {
   return RedisValue.array([
     bulkString('name'),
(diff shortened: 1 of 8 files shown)


src/commands/streams.ts
```diff
@@ -589,6 +753,1200 @@ export const xreadCommand = defineCommand({
...
+
+          if (count !== null && count > 0 && replies.length >= count) break
+        }
+      } else {
+        for (const pending of pendingEntriesSorted(group)) {
+          if (pending.consumerId !== consumerId) continue
+          if (compareStreamId(pending.id, id) <= 0) continue
+
+          const entry = findEntry(stream, pending.id)
+          if (!entry) continue
```
fatal10110: **Bug: XREADGROUP history read drops entries that were `XDEL`'d from the stream.** In the explicit-ID branch (`id !== '>'`), a PEL entry whose stream entry no longer exists is silently skipped (`if (!entry) continue`). Real Redis (verified on 7.2.14) still returns it as `[id, nil]` so the client can see and `XACK`/`XCLAIM` it.

Repro vs real Redis:
```
XADD k 1-1 f v ; XADD k 2-2 f v
XGROUP CREATE k g 0
XREADGROUP GROUP g c COUNT 10 STREAMS k >
XDEL k 1-1
XREADGROUP GROUP g c STREAMS k 0
# real:  1) 1-1 -> (nil)   2) 2-2 -> [f v]
# here:  only 2-2
```
Fix: emit a nil-fields reply for the deleted entry instead of `continue`.

src/commands/streams.ts
```diff
@@ -589,6 +753,1200 @@ export const xreadCommand = defineCommand({
...
+          replies.push(entryToReply(entry.id, entry.fields))
+
+          if (count !== null && count > 0 && replies.length >= count) break
+        }
+      }
+
+      return replies
+    })
+
+    if (entries.length > 0) {
```
fatal10110: **Bug: explicit-ID (history) read returns nil instead of an empty per-key result.** `readGroupEntries` only adds a stream to `results` when it produced ≥1 entry, so a consumer with an empty PEL yields `null` and XREADGROUP replies nil. For `>` reads nil is correct, but for an explicit ID real Redis returns the key with an empty entry list.

Verified (7.2.14):
```
XREADGROUP GROUP g zzz STREAMS k 0
# real: 1) 1) "k"  2) (empty array)
# here: (nil)
```
Only the `>` path should collapse empty results to nil; explicit-ID streams should always appear in the reply.

src/commands/streams.ts
```diff
@@ -589,6 +753,1200 @@ export const xreadCommand = defineCommand({
...
+      integerValue(stream.groups.size),
+      bulkString('first-entry'),
+      firstEntry ? entryToReply(firstEntry.id, firstEntry.fields) : nullBulk(),
+      bulkString('last-entry'),
+      lastEntry ? entryToReply(lastEntry.id, lastEntry.fields) : nullBulk(),
+    )
+    return fields
+  }
+
+  const entryLimit = count ?? stream.entries.length
```
fatal10110: Nit: `XINFO STREAM FULL` without an explicit `COUNT` returns **all** entries (and all PEL rows) here; real Redis defaults the FULL count to 10. Low impact for a mock, but worth matching if a test ever asserts on a large stream.

Issue #38, opened 2026-06-09
Bug: HDEL on non-existent key creates and persists a phantom empty hash


artur-ma:
## Description

`HDEL` on a non-existent (or already-empty) key silently creates and persists an empty hash entry in the keyspace.

## Root Cause

In `src/commands/hashes.ts`, `HDEL` calls `keyspace.update()` which unconditionally creates the key when it is missing. The mutator runs, finds zero fields to delete, and the post-delete cleanup guard `(deleted > 0)` is `false` — so the empty hash is never removed. The spurious key then persists in the store.

## Impact

Any of the following will observe the phantom key:
- `TYPE key` → returns `hash` instead of `none`
- `EXISTS key` → returns `1` instead of `0`
- `DBSIZE` → inflated by one per phantom key
- `KEYS *` / `SCAN` → phantom key appears in results

## Expected behaviour (real Redis)

```
> DEL myhash
> HDEL myhash field1
(integer) 0
> EXISTS myhash
(integer) 0
> TYPE myhash
none
```

## Actual behaviour (mock)

```
> DEL myhash
> HDEL myhash field1
(integer) 0
> EXISTS myhash
(integer) 1   ← phantom key
> TYPE myhash
hash          ← wrong
```

PR #177, opened 2026-06-14
fix: centralize empty-collection cleanup in keyspace.update (#124)

### LINKED ISSUE #124: Bug: RedisKeyspace.update() leaves ghost/empty entries on mutator throw, and never auto-deletes emptied collections
## Description

`RedisKeyspace.update()` (`src/state/keyspace.ts:131-154`) inserts a freshly-created entry into `this.entries` **before** invoking the mutator:

```typescript
update<TValue extends RedisDataValue, TResult>(
  key: Buffer,
  expectedType: TValue['type'],
  createValue: () => TValue,
  mutator: (value: TValue) => TResult,
): TResult {
  let entry = this.getLiveEntry(key)

  if (!entry) {
    entry = {
      key: Buffer.from(key),
      value: createValue(),
    }
    this.entries.set(keyId(key), entry)   // <-- inserted before mutator runs
  }

  if (entry.value.type !== expectedType) {
    throw new WrongRedisTypeError(expectedType, entry.value.type)
  }

  const result = mutator(entry.value as TValue)   // <-- if this throws, entry is already in the map
  this.emitWrite(entry)
  return result
}
```

## Impact

1. **Mutator throws → ghost entry persists.** If `mutator` throws (e.g. a validation error inside a multi-pair command like `ZADD`), the empty collection created by `createValue()` is already committed to `this.entries` with no `write` event emitted. The key now exists (`EXISTS`/`TYPE`/`KEYS` see an empty hash/list/set/zset) when it shouldn't, and `WATCH` never saw it appear.
2. **No automatic cleanup of emptied collections.** Even on the success path, `update()` never deletes `entry` if the mutator leaves the collection empty (e.g. `HDEL` removes the last field, `SREM` empties a set). Each command currently has to remember to call `this.delete(key)` itself. #38 ("HDEL on non-existent key creates and persists a phantom empty hash") is one instance of this broader gap — the root cause is in `update()`, not just `HDEL`.

## Fix

- Only insert the new entry into `this.entries` *after* `mutator` returns successfully (or roll back the insert on throw).
- After a successful mutation, check whether the underlying collection is now empty and call `this.entries.delete(keyId(key))` + emit a `delete` event, centralizing the "delete collection when empty" rule.

## Files
- `src/state/keyspace.ts:131-154`

Related: #38

fatal10110:
## Summary
- `RedisKeyspace.update()` inserted a freshly-created entry into the keyspace **before** running the mutator and always emitted a `write` event. Two latent bugs:
  1. **Ghost on throw** — if the mutator threw, the empty collection from `createValue()` was already committed, leaving a key `EXISTS`/`TYPE`/`KEYS` could see, that `WATCH` never observed appearing.
  2. **No auto-delete of emptied collections** — every command had to remember to call `db.delete(key)` after emptying a hash/list/set/zset (HDEL, SREM, ZREM, LTRIM, ...). Scattered, inconsistent cleanup; root cause behind #38.
- Fix centralizes both rules in `update()`:
  - New keys mutate a **not-yet-committed** entry → a throwing mutator rolls back for free (no ghost).
  - After a successful mutation, if the collection is empty, delete the key and emit a single `delete` event.
  - Strings (`""` is a real value) and empty streams (`XGROUP CREATE MKSTREAM`) are never auto-deleted.
- Per-command self-delete calls become harmless no-ops (`db.delete` on an already-gone key returns false, emits nothing) — left in place to keep this change minimal.

Closes #124

## Test plan
- [x] `tests/keyspace-update.test.ts` (unit) reproduces both facets directly — red before the fix (every shipped command masks them at the wire level, so this is the meaningful repro), green after.
- [x] `tests-integration/ioredis/keyspace-empty-cleanup.test.ts` guards wire behavior end-to-end (no-op mutation on a missing key doesn't disturb a WATCH; emptying a key removes it; a real emptying still dirties a WATCH) — green on mock **and** real Redis.
- [x] `npm run build` + `npm run lint` clean.
- [x] `npm run test:all` passes — 132 unit / 262 mock / 262 real.

### DIFF
--- a/src/state/keyspace.ts
+++ b/src/state/keyspace.ts
@@ -134,21 +134,37 @@ export class RedisKeyspace {
     createValue: () => TValue,
     mutator: (value: TValue) => TResult,
   ): TResult {
-    let entry = this.getLiveEntry(key)
+    const existing = this.getLiveEntry(key)
 
-    if (!entry) {
-      entry = {
-        key: Buffer.from(key),
-        value: createValue(),
-      }
-      this.entries.set(keyId(key), entry)
+    if (existing && existing.value.type !== expectedType) {
+      throw new WrongRedisTypeError(expectedType, existing.value.type)
     }
 
-    if (entry.value.type !== expectedType) {
-      throw new WrongRedisTypeError(expectedType, entry.value.type)
+    // For a new key, mutate a not-yet-committed entry: if the mutator throws,
+    // the keyspace is left untouched (no ghost empty collection persists).
+    const entry: KeyspaceEntry = existing ?? {
+      key: Buffer.from(key),
+      value: createValue(),
     }
 
     const result = mutator(entry.value as TValue)
+    const id = keyId(key)
+
+    // Centralized "delete the key when its collection is empty" rule, so each
+    // command no longer has to remember to clean up emptied hashes/lists/etc.
+    if (isEmptyCollection(entry.value)) {
+      if (existing) {
+        this.entries.delete(id)
+        this.mutations.emit({
+          type: 'delete',
+          database: this.database,
+          key: entry.key,
+        })
+      }
+      return result
+    }
+
+    this.entries.set(id, entry)
     this.emitWrite(entry)
     return result
   }
@@ -231,4 +247,23 @@ export class RedisKeyspace {
 
 function keyId(key: Buffer): string {
   return key.toString('hex')
+}
+
+// A collection-typed value is "empty" when it holds no elements; such keys are
+// deleted from the keyspace (matching real Redis). Strings are always a real
+// value (even ""), and empty streams persist (e.g. XGROUP CREATE MKSTREAM), so
+// neither is ever auto-deleted here.
+function isEmptyCollection(value: RedisDataValue): boolean {
+  switch (value.type) {
+    case 'hash':
+      return value.fields.size === 0
+    case 'list':
+      return value.values.length === 0
+    case 'set':
+    case 'zset':
+      return value.members.size === 0
+    case 'string':
+    case 'stream':
+      return false
+  }
 }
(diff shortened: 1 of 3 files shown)


src/state/keyspace.ts
```diff
@@ -232,3 +248,22 @@ export class RedisKeyspace {
...
+// neither is ever auto-deleted here.
+function isEmptyCollection(value: RedisDataValue): boolean {
+  switch (value.type) {
+    case 'hash':
+      return value.fields.size === 0
+    case 'list':
+      return value.values.length === 0
+    case 'set':
+    case 'zset':
+      return value.members.size === 0
```
fatal10110, line 264: This helper is now the central owner for empty-key cleanup across all collection types, but the new tests only exercise an emptied hash and set. Please add coverage for the new `list` and `zset` branches too, either with direct `RedisKeyspace.update` unit cases or wire coverage around commands like `LTRIM` and `ZREM`, so regressions in these branches are caught.

tests-integration/ioredis/keyspace-empty-cleanup.test.ts
```diff
@@ -0,0 +1,114 @@
+import { Cluster } from 'ioredis'
+import { after, before, describe, it } from 'node:test'
```
fatal10110: The repo test guidance asks new test cases to use `test()` from `node:test`, not `it()`. This file imports `it` and uses it for all four cases; please switch the import and usages to `test` so the new coverage follows the project convention.

fatal10110 (commented):
Reviewed the keyspace cleanup change. I didn't find a production correctness regression; left two small inline comments on test coverage/style.

src/state/keyspace.ts
fatal10110: Done in 4268085 — added direct `RedisKeyspace.update()` unit cases for the `list` branch (LTRIM-style `values.length = 0`) and `zset` branch (ZREM-style, removing the last member), each asserting the key is deleted and a single `delete` event fires. Now all four collection arms of `isEmptyCollection` are exercised.

tests-integration/ioredis/keyspace-empty-cleanup.test.ts
fatal10110: Done in 4268085 — switched the import and all four cases from `it()` to `test()`.

PR #184, opened 2026-06-14
Add MONITOR response stream

### LINKED ISSUE #152: Feature missing: MONITOR via ResponseStream
## Context

The architecture refactor acceptance criteria call out `SUBSCRIBE/MONITOR` working via `ResponseStream` without per-command transport capture shims. Pub/Sub has dedicated tracking issues (#98, #99, #100, #126), but `MONITOR` did not have its own issue.

`MONITOR` is related to the streaming transport work, but it has different semantics from Pub/Sub: it streams a textual feed of executed commands from other clients and should keep the monitoring connection open until it disconnects.

## Scope

- Add a `MONITOR` command implemented through the new `defineCommand` API.
- Return a long-lived `ResponseStream` from `MONITOR`.
- Broadcast command events through server-level state without commands writing directly to transports.
- Ensure the monitoring connection can disconnect cleanly.
- Add unit and integration coverage for raw TCP / client-visible behavior.

## Out of scope

- Full Redis formatting perfection for every edge case in the first slice.
- Pub/Sub subscription delivery; covered by #98, #99, #100.
- Keyspace notifications; covered by #87.

## Sub-issues

Sub-issues will track the event source, command stream, and integration coverage slices.

### LINKED ISSUE #153: MONITOR: add server-level command event feed
Parent: #152

## Scope

Add the internal event source needed by `MONITOR` without coupling normal commands to transports.

- Define a server-level command event shape with timestamp, selected DB, client address/identity where available, command name, and raw arguments suitable for Redis-style monitor formatting.
- Emit events from the execution/session boundary so every normal RESP command can be observed consistently.
- Avoid emitting recursive monitor output for the monitor stream itself if Redis compatibility requires it.
- Keep the event feed independent from Pub/Sub and keyspace notifications.

## Acceptance criteria

- No normal command imports transport/session adapter internals just to support monitoring.
- Tests prove events are emitted for ordinary commands and not emitted for connection setup noise unless intentionally supported.

### LINKED ISSUE #154: MONITOR: implement command as a ResponseStream
Parent: #152

## Scope

Implement the client-visible `MONITOR` command on top of the server-level event feed.

- Register `MONITOR` with `defineCommand` and appropriate flags/introspection metadata.
- Return a long-lived `ResponseStream` that sends monitor lines as command events arrive.
- Clean up subscriptions when the monitoring session disconnects, resets, or the stream closes.
- Ensure `MONITOR` is rejected where Redis would reject it, including Lua/script contexts if applicable.

## Depends on

- #153
- #126 for adapter behavior that allows long-lived streams without blocking the connection loop.

### LINKED ISSUE #155: MONITOR: add raw TCP and integration coverage
Parent: #152

## Scope

Add compatibility coverage for the `MONITOR` stream behavior.

- Cover the initial `OK`/stream activation behavior expected by Redis clients.
- Verify commands from another connection are delivered to the monitor stream.
- Verify disconnect/cleanup does not leak monitor subscribers.
- Cover RESP2/RESP3 behavior if replies differ.
- Prefer raw TCP tests for exact stream framing and add client integration tests when useful.

## Depends on

- #154

fatal10110:
## Summary
- Add a server-level RedisMonitorFeed and emit cloned command events from ClientSession for MONITOR subscribers.
- Implement MONITOR as a noscript ResponseStream with Redis-style simple-string event lines and cleanup on close, disconnect, and RESET.
- Add raw TCP mock/real compatibility coverage plus unit coverage for feed cloning and stream cleanup.

## Root cause
MONITOR had no command definition or server-level event source, so command observation would have required transport-specific shims instead of flowing through the shared command/session pipeline.

## Validation
- TEST_BACKEND=mock node --enable-source-maps --import tsx --no-warnings --test tests/state-core.test.ts tests/commands-foundation.test.ts tests-integration/raw-tcp/monitor-protocol.test.ts
- TEST_BACKEND=real node --enable-source-maps --import tsx --no-warnings --test-concurrency 1 --test-timeout 30000 --test tests-integration/raw-tcp/monitor-protocol.test.ts
- npm run build
- npm test
- npm run test:integration:real
- npm run lint (passes with existing src/types/respjs.d.ts no-explicit-any warning)

Fixes #152
Fixes #153
Fixes #154
Fixes #155


src/core/client-session.ts
```diff
@@ -733,6 +762,30 @@ export class ClientSession implements RedisClientSession {
...
+      return
+    }
+
+    this.server.monitorFeed.publish({
+      timestampMs: Date.now(),
+      database,
+      clientId: this.id,
+      clientAddress: this.clientAddress,
+      command,
+      args: rawArgs.map(arg => Buffer.from(arg)),
```
fatal10110: **Credential leak: `AUTH` args are not redacted in the MONITOR feed.**

Real Redis redacts sensitive arguments before feeding monitors:
```
1781488754.245892 [2 127.0.0.1:52967] "auth" "(redacted)"
1781488818.772889 [0 127.0.0.1:53027] "AUTH" "(redacted)"
```
This implementation publishes `rawArgs` verbatim, so a `MONITOR` client (or anything tailing its output) sees plaintext passwords for `AUTH`, `HELLO ... AUTH`, `CONFIG SET requirepass/masterauth`, `MIGRATE ... AUTH`, `ACL SETUSER ... >pass`, etc. Verified against Redis 7.x via redis-cli MONITOR. Suggest redacting the known sensitive arg positions per command before publishing.

src/core/client-session.ts
```diff
@@ -603,7 +623,15 @@ export class ClientSession implements RedisClientSession {
     this.activeTurnAccess = turnAccess
     try {
       const ctx = this.createExecutionContext(turnAccess)
-      return await this.executor.executeRaw(rawCommand, rawArgs, ctx)
+      const execution = await this.executor.executeRawWithPlan(
+        rawCommand,
+        rawArgs,
+        ctx,
+      )
+      if (execution.plan) {
```
fatal10110: **MONITOR hook is too shallow: transaction-replayed and Lua `redis.call` commands never reach the feed.**

The publish point is here in `ClientSession.execute()`, the top-level client entry. But commands replayed during `EXEC` (`executeTransaction` → `executor.executePlan`) and commands run from Lua (`redis.call` → `executePlanSync`) bypass `execute()` entirely, so they are never published.

Real Redis feeds at the `call()` level, so all of these appear. Verified:
```
# MULTI/EXEC — inner SET shows between multi and exec
... "multi"
... "set" "a" "1"
... "exec"
# EVAL — inner calls show with a "lua" source
... "eval" "redis.call(...); return redis.call(...)" "1" "luakey"
[0 lua] "set" "luakey" "v"
[0 lua] "get" "luakey"
```
This mock currently emits only `multi`/`exec` and `eval` (no inner commands). Consider moving the publish into the executor pipeline (e.g. `executePlan`/`executePlanSync`) so every execution path is covered, rather than hooking a single caller.

src/core/client-session.ts
```diff
@@ -733,6 +762,30 @@ export class ClientSession implements RedisClientSession {
...
+    rawCommand: Buffer | string,
+    rawArgs: readonly Buffer[],
+    database: number,
+  ): void {
+    if (this.server.monitorFeed.subscriberCount === 0) {
+      return
+    }
+
+    const command = monitorCommandBuffer(rawCommand)
+    if (equalsAscii(command, 'monitor')) {
```
fatal10110: **Admin commands leak into MONITOR — real Redis excludes `CMD_ADMIN`/`CMD_SKIP_MONITOR`.**

Only the literal `monitor` command is filtered here. Real Redis skips the whole admin class (the feed check is `!(c->cmd->flags & (CMD_SKIP_MONITOR|CMD_ADMIN))`). Verified — `CONFIG GET/SET`, etc. never appear in a real MONITOR stream:
```
# config get maxmemory / config set maxmemory 0 / config set requirepass secretpw
# → none of these appear in MONITOR output
```
This mock will surface `CONFIG`, `DEBUG`, `SLAVEOF/REPLICAOF`, `SAVE/BGSAVE`, `FAILOVER`, etc. Filtering should key off a command flag (e.g. the existing `admin` flag) instead of a hardcoded name. Note `monitor` itself is admin, so a flag-based filter subsumes this special-case.

src/core/transports/resp2/server.ts
```diff
@@ -104,3 +105,26 @@ export class Resp2Server {
...
+    return address
+  }
+
+  return `${address}:${socket.remotePort}`
+}
+
+function normalizeSocketAddress(
+  address: string | undefined,
+): string | undefined {
+  if (address === '::1') {
```
fatal10110: **IPv6 client addresses are rewritten to a fake IPv4 address.**

Mapping `::1` → `127.0.0.1` (and stripping the `::ffff:` prefix) misreports a genuinely IPv6-connected client. Real Redis emits the actual address with bracket form, e.g. `[0 [::1]:6379]`. This looks like a band-aid to satisfy the test's `(?:\d{1,3}\.){3}\d{1,3}:\d+` regex rather than correct address formatting. The `::ffff:` unwrap is reasonable (true IPv4-mapped), but the `::1` rewrite changes the reported family. Consider formatting IPv6 as `[addr]:port` and loosening the test regex instead.

src/core/client-session.ts
```diff
@@ -752,3 +805,13 @@ function createAbortError(): Error {
   err.name = 'AbortError'
   return err
 }
+
+function monitorCommandBuffer(rawCommand: Buffer | string): Buffer {
```
fatal10110: **Dead ternary — both branches are identical.**

```ts
function monitorCommandBuffer(rawCommand: Buffer | string): Buffer {
  return typeof rawCommand === 'string'
    ? Buffer.from(rawCommand)
    : Buffer.from(rawCommand)
}
```
Both arms call `Buffer.from(rawCommand)`. Collapse to `return Buffer.from(rawCommand)` (the whole helper could be inlined).

src/core/command-executor.ts
```diff
@@ -270,10 +318,234 @@ export class CommandExecutor {
...
+): boolean {
+  if (ctx.monitor?.disabled) {
+    return false
+  }
+
+  if (ctx.server.monitorFeed.subscriberCount === 0) {
+    return false
+  }
+
+  if (plan.flags.includes('admin')) {
```
fatal10110: **Over-broad: filtering on the `admin` flag now hides commands real Redis *does* show in MONITOR.**

Reusing the `admin` command flag as the monitor-skip predicate excludes more than real Redis does. In this repo both `info` and `client` carry `admin` (`connection.ts`), but real Redis feeds both to monitors:
```
1781490929.082578 [0 127.0.0.1:60437] "info" "server"
1781490929.089162 [0 127.0.0.1:60438] "client" "getname"
```
Real Redis only excludes `CMD_SKIP_MONITOR | CMD_ADMIN`, and there `info`/`client info`/`client getname` are *not* admin (only `CONFIG`, `CLUSTER`, `MONITOR`, etc. are). So this fix correctly hides CONFIG/CLUSTER/MONITOR but regresses INFO and CLIENT. The existing test only covers `CONFIG GET`, so the divergence is uncaught — and it would fail against the real backend if asserted.

Suggest a dedicated `skip-monitor` flag (set on the genuinely-excluded commands) instead of overloading `admin`, or narrow the predicate to the real exclusion set.

src/core/command-executor.ts
```diff
@@ -270,10 +318,234 @@ export class CommandExecutor {
...
+
+  if (!(result instanceof RedisResult)) {
+    return true
+  }
+
+  if (isQueuedTransactionCommand(plan, ctx, result)) {
+    return false
+  }
+
+  return result.value.kind !== 'error' || result.value.code !== 'MOVED'
```
fatal10110: **Only `MOVED` is skipped, but real Redis skips *all* cluster redirects/errors.**

Cluster redirection (`MOVED`/`ASK`) and cluster errors (`CROSSSLOT`, `CLUSTERDOWN`, `TRYAGAIN`) are all produced before `call()` in real Redis, so none of them are ever fed to monitors. This predicate special-cases only `MOVED`, so a `CROSSSLOT`/`CLUSTERDOWN` command still leaks into the monitor feed. Minor, but to match real Redis the skip should cover the whole cluster-redirect/error class (or, more robustly, skip any command short-circuited by the cluster policy before execution).

src/core/command-executor.ts
```diff
@@ -270,10 +318,234 @@ export class CommandExecutor {
...
+      continue
     }
+
+    redactArg(args, i + 1)
+    redactArg(args, i + 2)
+    i += 2
   }
 }
 
+function redactConfigArgs(args: Buffer[]): void {
```
fatal10110: Nit: `redactConfigArgs` is unreachable. `CONFIG` has the `admin` flag, so `shouldPublishMonitorEvent` returns false before any event is built — `CONFIG SET requirepass …` never reaches redaction. Harmless (and would matter if CONFIG ever became monitorable), but worth a note so it isn't mistaken for live protection.

PR #217, opened 2026-06-16
Fix no-op WATCH invalidation

### LINKED ISSUE #125: Bug: keyspace.update() emits a write event even when the mutator changes nothing — spurious WATCH invalidation
## Description

`RedisKeyspace.update()` (`src/state/keyspace.ts:131-154`) unconditionally calls `this.emitWrite(entry)` after running the mutator:

```typescript
const result = mutator(entry.value as TValue)
this.emitWrite(entry)
return result
```

`emitWrite` emits a `'write'` mutation event (`src/state/keyspace.ts:221-229`), which `ClientSession.watch()` subscribes to per-key (`src/core/client-session.ts:300-302`) and uses to mark the session's WATCH dirty (`dirtyWatches.add(id)`).

## Impact

Commands whose mutator can be a true no-op — e.g. `HDEL` on a field that doesn't exist, `SREM` of a member not in the set, `ZADD XX` on a member that doesn't exist, `LREM` removing zero elements — still emit a `write` event. Any client that `WATCH`ed that key will have its transaction unconditionally fail on the next `EXEC` (`isWatchDirty()` true), even though the value is byte-for-byte identical to what was watched. Real Redis only signals modification when the dataset actually changes.

## Fix

Have `mutator` return whether a change occurred (e.g. `{ result: TResult; changed: boolean }`), and only call `emitWrite` when a real mutation happened.

## Files
- `src/state/keyspace.ts:131-154`

### LINKED ISSUE #133: Core: keyspace.update() mutator contract gains a 'changed' signal; emitWrite only fires on real mutation
Sub-issue of #125. Prerequisite for the per-data-type sub-issues below.

## Description
`RedisKeyspace.update()` (`src/state/keyspace.ts:131-154`) currently calls `this.emitWrite(entry)` unconditionally after every mutator invocation, even when the mutator made no actual change (e.g. `HDEL` on a missing field). This causes `WATCH` to be invalidated for genuinely no-op writes (#125).

## Fix
Change the mutator contract so it can signal "no change occurred", e.g.:

```typescript
update<TValue extends RedisDataValue, TResult>(
  key: Buffer,
  expectedType: TValue['type'],
  createValue: () => TValue,
  mutator: (value: TValue) => { result: TResult; changed: boolean },
): TResult
```

(or a simpler sentinel/symbol-based approach — pick whichever keeps call sites terse). `emitWrite` is only called when `changed` is true. Update `RedisDatabase.updateHash/updateList/updateSet/updateSortedSet/updateStream` (`src/state/database.ts:108-141`) signatures accordingly.

## Scope
- `src/state/keyspace.ts` (`update()` signature + emitWrite gating)
- `src/state/database.ts` (`updateTyped` + the 5 typed wrappers)
- Do NOT update individual command call sites here — that's the per-data-type sub-issues (#<hash>, #<list>, #<set>, #<zset>, #<stream>). To avoid breaking the build mid-refactor, either land this with a default `changed: true` shim that the per-type issues remove, or land all sub-issues as one coordinated PR.

### LINKED ISSUE #134: Hashes: HSET/HSETNX/HDEL/HMSET/HINCRBY/HINCRBYFLOAT report 'changed' to suppress no-op WATCH invalidation
Sub-issue of #125. Depends on the core `update()` contract change (see #125's core sub-issue).

## Scope
Update the `updateHash` mutators in `src/commands/hashes.ts` to return `{ result, changed }`:

- `hset` (`hashes.ts:44`) — `changed` if any field was added or its value changed.
- `hsetnx` (`hashes.ts:63`) — `changed` only if the field didn't already exist (i.e. only when it actually sets).
- `hdel` (`hashes.ts:93`) — `changed` only if at least one field was actually removed.
- `hmset` (`hashes.ts:111`) — same as `hset`.
- `hincrby` (`hashes.ts:214`) — always `changed: true` (value always changes, even `INCRBY 0` writes the same numeric string — confirm against real Redis; if `INCRBY 0` is a true no-op when field already holds that exact value, treat as changed only if the stored bytes differ).
- `hincrbyfloat` (`hashes.ts:240`) — same as above.

## Test
Add a unit test: `WATCH key`, `HDEL key nonexistent-field`, `EXEC` with a no-op queued command must succeed (not return nil) when nothing else touched `key`.

### LINKED ISSUE #135: Lists: LPUSH/RPUSH/LPOP/RPOP/LSET/LREM/LTRIM/LPUSHX/RPUSHX/RPOPLPUSH report 'changed' to suppress no-op WATCH invalidation
Sub-issue of #125. Depends on the core `update()` contract change (see #125's core sub-issue).

## Scope
Update the `updateList` mutators in `src/commands/lists.ts` to return `{ result, changed }`:

- `lpush`/`rpush` (`lists.ts:59,78`) — always `changed: true` (always pushes ≥1 element; arity guarantees ≥1 value).
- `lpop`/`rpop` (`lists.ts:99,119`) — `changed` only if the list was non-empty (an element was actually removed).
- `lset` (`lists.ts:199`) — always `changed: true` if it succeeds (errors on missing key/index, so any success is a real write).
- `lrem` (`lists.ts:219`) — `changed` only if `removed count > 0`.
- `ltrim` (`lists.ts:241`) — `changed` only if the list's length actually decreased (trimming to the same bounds it already has is a no-op).
- `lpushx`/`rpushx` (`lists.ts:271,293`) — `changed` only if the key existed (push happened); these no-op (return 0) when key missing.
- `rpoplpush` (`lists.ts:318,325` — two `updateList` calls on source/destination) — `changed` only if an element actually moved (source non-empty).

## Test
Add a unit test: `WATCH key`, `LPOPX`/`LPOP` on empty/missing list (no-op), `EXEC` with another queued command must succeed.

### LINKED ISSUE #136: Sets: SADD/SREM/SPOP/SMOVE/SDIFFSTORE/SINTERSTORE/SUNIONSTORE report 'changed' to suppress no-op WATCH invalidation
Sub-issue of #125. Depends on the core `update()` contract change (see #125's core sub-issue).

## Scope
Update the `updateSet` mutators to return `{ result, changed }`:

- `sadd` (`sets.ts:81`) — `changed` only if `added > 0`.
- `srem` (`sets.ts:108`) — `changed` only if at least one member was actually removed.
- `spop` (`sets.ts:182`) — `changed` only if the set was non-empty (members actually removed).
- `smove` (`sets.ts:328` source, `sets.ts:341` destination) — `changed` for the source only if the member existed and was removed; destination only mutated when source removal succeeded.
- `sdiffstore`/`sinterstore`/`sunionstore` (`sets.ts:354,378,402` → `sets.ts:61` shared `updateSet(destKey, ...)` helper) — `changed` if the resulting set's contents differ from the destination's previous contents (or simpler: `changed` if result set is non-empty OR destination key previously existed — i.e. any case where the destination key's existence/contents could differ from before).

## Test
Add a unit test: `WATCH key`, `SREM key nonexistent-member`, `EXEC` with another queued command must succeed.

### LINKED ISSUE #137: Sorted sets: ZADD/ZREM/ZINCRBY/ZREMRANGEBYSCORE/ZPOPMIN/ZPOPMAX report 'changed' to suppress no-op WATCH invalidation
Sub-issue of #125. Depends on the core `update()` contract change (see #125's core sub-issue).

## Scope
Update the `updateSortedSet` mutators in `src/commands/zsets.ts` to return `{ result, changed }`:

- `zadd` (`zsets.ts:80`) — `changed` only if a member was added or an existing member's score changed (not a true no-op when `XX`/`NX`/`GT`/`LT` reject every pair, or re-adding a member with an identical score).
- `zrem` (`zsets.ts:100`) — `changed` only if at least one member was actually removed.
- `zincrby` (`zsets.ts:182`) — always `changed: true` (score always changes unless increment is 0 *and* member already exists with that score — edge case, treat `INCRBY 0` on existing member as no-op if scores match).
- `zremrangebyscore` (`zsets.ts:282`) — `changed` only if at least one member was removed.
- `zpopmin`/`zpopmax` (`zsets.ts:330,359`) — `changed` only if the zset was non-empty (a member was actually popped).

## Test
Add a unit test: `WATCH key`, `ZREM key nonexistent-member`, `EXEC` with another queued command must succeed.

### LINKED ISSUE #138: Streams: XADD/XDEL/XTRIM report 'changed' to suppress no-op WATCH invalidation
Sub-issue of #125. Depends on the core `update()` contract change (see #125's core sub-issue).

## Scope
Update the `updateStream` mutators in `src/commands/streams.ts` to return `{ result, changed }`:

- `xadd` (`streams.ts:234`) — always `changed: true` (an entry is always appended on success).
- `xdel` (`streams.ts:368`) — `changed` only if `deleted > 0`.
- `xtrim` (`streams.ts:398`) — `changed` only if `removed > 0`.

## Test
Add a unit test: `WATCH key`, `XDEL key nonexistent-id`, `EXEC` with another queued command must succeed.

fatal10110:
## Summary
- Add an explicit `{ result, changed }` contract for in-place keyspace updates so `WATCH` is dirtied only by real mutation events.
- Update hash, list, set, sorted-set, and stream mutators to report true no-op removals/options without suppressing Redis-compatible write invalidation for commands Redis still treats as writes.
- Add ioredis WATCH regression coverage against no-op HDEL/HSETNX/LREM/SREM/ZREM/ZADD XX/XDEL/XTRIM paths and document the keyspace update contract.

## Root Cause
`RedisKeyspace.update()` emitted a `write` mutation after every successful mutator call, even when the mutator left the stored value unchanged. `ClientSession.watch()` subscribes to those mutation events, so no-op commands such as `HDEL` on a missing field could spuriously make `EXEC` return nil.

## Validation
- `npm test`
- `npm run build`
- `TEST_BACKEND=mock node --enable-source-maps --import tsx --no-warnings --test tests-integration/ioredis/watch.test.ts`
- `TEST_BACKEND=real node --enable-source-maps --import tsx --no-warnings --test-concurrency 1 --test-timeout 30000 --test tests-integration/ioredis/watch.test.ts`
- `npm run clean:redis`
- `npm run lint`

Fixes #125
Fixes #133
Fixes #134
Fixes #135
Fixes #136
Fixes #137
Fixes #138

### DIFF
--- a/src/commands/hashes.ts
+++ b/src/commands/hashes.ts
@@ -134,10 +134,13 @@ export const hsetCommand = defineCommand({
       let count = 0
       for (const { field, value } of args.pairs) {
         const hex = field.toString('hex')
-        if (!hash.fields.has(hex)) count++
+        const existing = hash.fields.get(hex)
+        if (!existing) {
+          count++
+        }
         hash.fields.set(hex, { field, value })
       }
-      return count
+      return { result: count, changed: true }
     })
     return integer(added)
   },
@@ -151,9 +154,9 @@ export const hsetnxCommand = defineCommand({
   execute: (args, ctx) => {
     const set = ctx.db.updateHash(args.key, hash => {
       const hex = args.field.toString('hex')
-      if (hash.fields.has(hex)) return false
+      if (hash.fields.has(hex)) return { result: false, changed: false }
       hash.fields.set(hex, { field: args.field, value: args.value })
-      return true
+      return { result: true, changed: true }
     })
     return integer(set ? 1 : 0)
   },
@@ -186,7 +189,7 @@ export const hdelCommand = defineCommand({
       for (const field of args.fields) {
         if (hash.fields.delete(field.toString('hex'))) deleted++
       }
-      return hash.fields.size
+      return { result: hash.fields.size, changed: deleted > 0 }
     })
     if (remaining === 0) {
       ctx.db.delete(args.key)
@@ -203,8 +206,10 @@ export const hmsetCommand = defineCommand({
   execute: (args, ctx) => {
     ctx.db.updateHash(args.key, hash => {
       for (const { field, value } of args.pairs) {
-        hash.fields.set(field.toString('hex'), { field, value })
+        const hex = field.toString('hex')
+        hash.fields.set(hex, { field, value })
       }
+      return { result: undefined, changed: true }
     })
     return ok()
   },
@@ -339,7 +344,7 @@ export const hincrbyCommand = defineCommand({
       const next = current + args.increment
       const valueBuf = Buffer.from(String(next))
       hash.fields.set(hex, { field: args.field, value: valueBuf })
-      return next
+      return { result: next, changed: true }
     })
     return integer(result)
   },
@@ -374,7 +379,7 @@ export const hincrbyfloatCommand = defineCommand({
       }
       const valueBuf = Buffer.from(formatted)
       hash.fields.set(hex, { field: args.field, value: valueBuf })
-      return valueBuf
+      return { result: valueBuf, changed: true }
     })
     return bulk(result)
   },
--- a/src/commands/hashes.ts
+++ b/src/commands/hashes.ts
@@ -133,14 +133,9 @@ export const hsetCommand = defineCommand({
     const added = ctx.db.updateHash(args.key, hash => {
       let count = 0
       for (const { field, value } of args.pairs) {
-        const hex = field.toString('hex')
-        const existing = hash.fields.get(hex)
-        if (!existing) {
-          count++
-        }
-        hash.fields.set(hex, { field, value })
+        if (hash.setField(field, value, { forceDirty: true }).added) count++
       }
-      return { result: count, changed: true }
+      return count
     })
     return integer(added)
   },
@@ -153,10 +148,7 @@ export const hsetnxCommand = defineCommand({
   keys: args => [args.key],
   execute: (args, ctx) => {
     const set = ctx.db.updateHash(args.key, hash => {
-      const hex = args.field.toString('hex')
-      if (hash.fields.has(hex)) return { result: false, changed: false }
-      hash.fields.set(hex, { field: args.field, value: args.value })
-      return { result: true, changed: true }
+      return hash.setFieldIfAbsent(args.field, args.value)
     })
     return integer(set ? 1 : 0)
   },
@@ -187,9 +179,9 @@ export const hdelCommand = defineCommand({
     let deleted = 0
     const remaining = ctx.db.updateHash(args.key, hash => {
       for (const field of args.fields) {
-        if (hash.fields.delete(field.toString('hex'))) deleted++
+        if (hash.deleteField(field)) deleted++
       }
-      return { result: hash.fields.size, changed: deleted > 0 }
+      return hash.size
     })
     if (remaining === 0) {
       ctx.db.delete(args.key)
@@ -206,10 +198,8 @@ export const hmsetCommand = defineCommand({
   execute: (args, ctx) => {
     ctx.db.updateHash(args.key, hash => {
       for (const { field, value } of args.pairs) {
-        const hex = field.toString('hex')
-        hash.fields.set(hex, { field, value })
+        hash.setField(field, value, { forceDirty: true })
       }
-      return { result: undefined, changed: true }
     })
     return ok()
   },
@@ -331,8 +321,7 @@ export const hincrbyCommand = defineCommand({
   keys: args => [args.key],
   execute: (args, ctx) => {
     const result = ctx.db.updateHash(args.key, hash => {
-      const hex = args.field.toString('hex')
-      const entry = hash.fields.get(hex)
+      const entry = hash.getField(args.field)
       let current = 0
       if (entry) {
         const raw = entry.value.toString()
@@ -343,8 +332,8 @@ export const hincrbyCommand = defineCommand({
       }
       const next = current + args.increment
       const valueBuf = Buffer.from(String(next))
-      hash.fields.set(hex, { field: args.field, value: valueBuf })
-      return { result: next, changed: true }
+      hash.setField(args.field, valueBuf, { forceDirty: true })
+      return next
     })
     return integer(result)
   },
@@ -357,8 +346,7 @@ export const hincrbyfloatCommand = defineCommand({
   keys: args => [args.key],
   execute: (args, ctx) => {
     const result = ctx.db.updateHash(args.key, hash => {
-      const hex = args.field.toString('hex')
-      const entry = hash.fields.get(hex)
+      const entry = hash.getField(args.field)
       let current = 0
       if (entry) {
         const raw = entry.value.toString()
@@ -378,8 +366,8 @@ export const hincrbyfloatCommand = defineCommand({
         formatted = next.toFixed(17).replace(/\.?0+$/, '')
       }
       const valueBuf = Buffer.from(formatted)
-      hash.fields.set(hex, { field: args.field, value: valueBuf })
-      return { result: valueBuf, changed: true }
+      hash.setField(args.field, valueBuf, { forceDirty: true })
+      return valueBuf
     })
     return bulk(result)
   },
--- a/src/commands/lists/access.ts
+++ b/src/commands/lists/access.ts
@@ -78,6 +78,7 @@ export const lsetCommand = defineCommand({
 
     ctx.db.updateList(args.key, list => {
       list.values[idx] = args.value
+      return { result: undefined, changed: true }
     })
     return ok()
   },
@@ -98,7 +99,10 @@ export const lremCommand = defineCommand({
 
     const result = ctx.db.updateList(args.key, list => {
       const removed = listRemove(list.values, args.count, args.element)
-      return { removed, empty: list.values.length === 0 }
+      return {
+        result: { removed, empty: list.values.length === 0 },
+        changed: removed > 0,
+      }
     })
     if (result.empty) ctx.db.delete(args.key)
     return integer(result.removed)
@@ -129,7 +133,10 @@ export const ltrimCommand = defineCommand({
         list.values.splice(0, start)
         list.values.splice(stop - start + 1)
       }
-      return { empty: list.values.length === 0 }
+      return {
+        result: { empty: list.values.length === 0 },
+        changed: true,
+      }
     })
     if (result.empty) ctx.db.delete(args.key)
     return ok()
--- a/src/commands/lists/access.ts
+++ b/src/commands/lists/access.ts
@@ -3,7 +3,7 @@ import { t } from '../../core/command-schema'
 import { IndexOutOfRangeError, NoSuchKeyError } from '../../core/redis-error'
 import { RedisValue } from '../../core/redis-value'
 import { array, bulk, integer, ok } from '../helpers'
-import { listRemove, resolveIndex } from './helpers'
+import { resolveIndex } from './helpers'
 
 export const llenCommand = defineCommand({
   name: 'llen',
@@ -77,8 +77,7 @@ export const lsetCommand = defineCommand({
     if (idx < 0 || idx >= list.values.length) throw new IndexOutOfRangeError()
 
     ctx.db.updateList(args.key, list => {
-      list.values[idx] = args.value
-      return { result: undefined, changed: true }
+      list.setAt(idx, args.value)
     })
     return ok()
   },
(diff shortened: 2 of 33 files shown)


src/commands/sets.ts
```diff
@@ -66,12 +66,18 @@ function storeSetResult(
     db.delete(destKey)
     return 0
   }
+  const existing = db.getSet(destKey)
+  const changed =
+    !existing ||
+    existing.members.size !== hexSet.size ||
+    Array.from(hexSet).some(hex => !existing.members.has(hex))
```
fatal10110: **Real-Redis incompatibility (new WATCH divergence introduced by this fix).**

`SINTERSTORE`/`SUNIONSTORE`/`SDIFFSTORE` gate `changed` on whether the stored result differs from the destination's prior contents. But real Redis signals the destination as modified (dirties `WATCH`) on **every** successful store-write, even when the new value is identical to what the key already held — it does an unconditional `dbDelete(dst)` + `dbAdd(dst)`, each of which signals.\n\nVerified against `redis-server` (port 6399):\n```\nSADD s1 a b c; SINTERSTORE dst s1      # dst = {a,b,c}\nconnA: WATCH dst\nconnB: SINTERSTORE dst s1               # identical re-store\nconnA: MULTI / SET marker 1 / EXEC      # => (nil)  -> EXEC ABORTED\n```\nEXEC returned nil → the identical re-store invalidated the watch. With this PR, `changed` is `false`, so `keyspace.update()` early-returns before `emitWrite`, the watch is NOT dirtied, and EXEC would run. (Controls confirmed matching: `HDEL` missing field and `SADD` of existing members do *not* dirty — so the no-op direction is right; only the STORE path regresses.)\n\nFix: for the non-empty store branch, just return `changed: true` (the empty-result `db.delete` path already dirties correctly). Same issue in the zset store helper.

src/commands/zsets/zsetops.ts
```diff
@@ -171,9 +171,18 @@ function storeResult(
     db.delete(destination)
     return 0
   }
+  const existing = db.getSortedSet(destination)
+  const changed =
+    !existing ||
+    existing.members.size !== members.size ||
+    Array.from(members).some(([hex, member]) => {
+      const current = existing.members.get(hex)
+      return !current || current.score !== member.score
```
fatal10110: **Real-Redis incompatibility (same class as the set store path).**

`ZUNIONSTORE`/`ZINTERSTORE`/`ZDIFFSTORE` gate `changed` on the result differing from the destination's prior contents, but real Redis dirties `WATCH` on every successful store-write, even when the result is identical.\n\nVerified against `redis-server`:\n```\nZADD zs1 1 a 2 b; ZINTERSTORE zdst 1 zs1\nconnA: WATCH zdst\nconnB: ZINTERSTORE zdst 1 zs1            # identical re-store\nconnA: MULTI / SET marker 1 / EXEC       # => (nil)  -> EXEC ABORTED\n```\nWith this PR `changed` is `false`, so the watch is not dirtied and EXEC would run — diverging from real Redis. Recommend `changed: true` for the non-empty store branch (empty-result delete path already dirties).

fatal10110 (commented):
Design/code-quality review focused on the new mutation-tracking contract. I’m not duplicating the existing STORE compatibility comments; the main concern is that the current `{ result, changed }` API pushes dirty-state bookkeeping into every command. I’d like this moved toward tracked data-type wrappers/helpers before this becomes the pattern all future commands copy.

src/state/keyspace.ts
```diff
@@ -21,6 +21,11 @@ export type SetOptions = {
   keepTtl?: boolean
 }
 
+export type KeyspaceUpdateResult<TResult> = {
```
fatal10110: This contract fixes the immediate emission bug, but it makes every command manually decide and propagate `changed`. That is the part I would push out of command implementations: `RedisKeyspace.update()` or the typed update helpers can hand the mutator a tracked value and derive structural dirtiness from wrapper operations. We still need an explicit `markChanged()`/`forceWrite()` escape hatch for Redis semantics where a successful write dirties `WATCH` even if the final value is equal, but the default path should not be `return { result, changed }` in every command.

src/state/database.ts
```diff
@@ -107,35 +108,35 @@ export class RedisDatabase {
 
   updateHash<TResult>(
     key: Buffer,
-    mutator: (hash: RedisHashData) => TResult,
+    mutator: (hash: RedisHashData) => KeyspaceUpdateResult<TResult>,
```
fatal10110: Because these typed update APIs still expose raw `RedisHashData`/`RedisSetData`/`RedisSortedSetData`/`RedisListData`/`RedisStreamData`, the state layer cannot know which operations actually changed the value. I’d split this by data type: wrap hash/set/zset first with tracked Map-backed accessors (`setField`, `deleteField`, `addMember`, `deleteMember`, `setScore`), then add tracked list helpers around `set`/`splice`/`shift`/`pop`/`push`/`trim`. Streams are more complex because entries, group state, pending entries, and scalar metadata form one domain value, so they should get stream-specific mutation helpers instead of exposing the whole raw object.

src/commands/hashes.ts
```diff
@@ -134,10 +134,13 @@ export const hsetCommand = defineCommand({
...
         const hex = field.toString('hex')
-        if (!hash.fields.has(hex)) count++
+        const existing = hash.fields.get(hex)
+        if (!existing) {
+          count++
+        }
         hash.fields.set(hex, { field, value })
       }
-      return count
+      return { result: count, changed: true }
```
fatal10110: This illustrates why command-local booleans will be hard to maintain: the command now has to know both reply semantics (`count`) and keyspace dirtiness (`changed`). A tracked hash wrapper could make this `const added = hash.setField(field, value)` and leave dirty tracking to state, while the command only returns the Redis reply. The same pattern applies to `hsetnx`/`hdel` and the set/zset Map-backed commands.

src/commands/lists/access.ts
```diff
@@ -78,6 +78,7 @@ export const lsetCommand = defineCommand({
 
     ctx.db.updateList(args.key, list => {
       list.values[idx] = args.value
+      return { result: undefined, changed: true }
```
fatal10110: Lists need the same treatment, but probably as operation helpers rather than generic raw-array mutation. A tracked list API (`setAt`, `removeMatching`, `pop`, `trim`, `move`) can mark dirty when its operation actually removes, replaces, or appends, and the command no longer has to remember `changed: true` or `removed > 0` at each call site.

src/commands/streams/xdel.ts
```diff
@@ -31,7 +31,7 @@ export const xdelCommand = defineCommand({
           count++
```
fatal10110: Streams should not stay as raw object mutation either, but I would not try to solve them with only a Map wrapper. `entries`, `lastId`, `entriesAdded`, `maxDeletedEntryId`, `groups`, `pending`, and consumer metadata form one domain value; stream-specific helpers like `deleteEntries`, `trim`, `ack`, `claim`, and `createGroup` can return the Redis command result and update a tracked dirty flag internally.

src/state/keyspace.ts
fatal10110: Agree with this direction — and the STORE compat bug above is the evidence for it: that bug is a command hand-deriving `changed` from a final-state comparison, which is exactly the bookkeeping this proposal removes.

One implementation caveat that I think is load-bearing: the wrapper-derived dirtiness must be **operation-based (any effective mutating op flips the flag), not a before/after value diff.** The trap:
```
storeSetResult: set.clear(); for (m of result) set.addMember(m)
```
If dirtiness is computed by diffing final vs initial value, an identical re-store reads as "unchanged" and the STORE bug comes right back. If `clear()` (an effective op on a non-empty collection) flips dirty, it stays correct for free. The current PR fell into the diff version, so worth calling out explicitly in the helper contract.

With op-based tracking, commands split into two clean buckets that match Redis `signalModifiedKey`:

- **Gated — automatic, from the op result:** `deleteField`/`deleteMember` dirty iff the map actually removed; `addMember`/`setScore` dirty iff new-or-different; list `pop`/`shift`/`removeMatching` dirty iff they had an effect. Covers HDEL/SREM/ZREM/SADD/HSETNX/SMOVE-dest/ZADD/LREM — the error-prone family — and the command can no longer get it wrong.
- **Always-dirty — explicit `forceWrite()`:** HSET/HMSET/HINCRBY (a write to an equal value still dirties in Redis), LSET, LTRIM, pushes. Trivial and obviously correct because there is no computation.

Note the same structural op needs a different policy per type — hash `setField` is always-dirty (HSET), zset `setScore` is gated (ZADD) — which is the whole point: encode the Redis semantics once in the typed helper instead of at every call site.

The escape hatch is genuinely needed, not just for value-equal writes: `LTRIM key 0 -1` removes nothing (zero effective ops) yet real Redis still dirties the key, so it has to call `forceWrite()` even though no tracked op fired. Good case to bake into the helper docs.

For sequencing: the STORE fix is a correctness bug and can land as the one-line `changed: true` now; this wrapper refactor is the larger design change. Happy to do it as a stacked follow-up if you would rather ship the WATCH fix first, or fold it in here if you want the pattern right before it gets copied — your call as maintainer.

PR #238, opened 2026-06-17
feat: publish keyspace notifications to pub/sub channels (#87)

### LINKED ISSUE #87: Feature missing: keyspace notifications not published to pub/sub channels
## Description

Real Redis publishes keyspace and keyevent notifications to `__keyevent@{db}__:{event}` and `__keyspace@{db}__:{key}` channels on every key mutation (when `notify-keyspace-events` is configured). The mock's `MutationBus` has no bridge to `PubSubBroker`, so these channels are always empty.

## Impact

Any application that relies on keyspace notifications (e.g. expiry events via `__keyevent@0__:expired`, set events via `__keyevent@0__:set`) will receive no messages from the mock — silent failure with no error.

## Note

This is a feature-level gap, not a crasher. Implementing it requires:
1. `PubSubBroker` to be functional (see #17)
2. `MutationBus` to translate mutation events into the correct channel/message format and call `broker.publish()`
3. A config option to toggle which event types are published (real Redis `notify-keyspace-events` flag)

fatal10110:
## Summary
Real Redis publishes keyspace/keyevent notifications on key mutations when `notify-keyspace-events` is configured; the mock's `MutationBus` had no bridge to the `PubSubBroker`, so `__keyspace@<db>__:<key>` / `__keyevent@<db>__:<event>` channels were always empty (silent failure for apps relying on e.g. `expired` cache-invalidation events).

This wires a `KeyspaceNotifier` from each database's mutation bus to the broker:

- **Lifecycle events** (`del`, `expire`, `persist`, `expired`) derive purely from the mutation type — always correct, no command knowledge needed. `expired` fires when a key is lazily evicted on access.
- **Write event names** (`set`, `lpush`, `hset`, `sadd`, `zadd`, `incrby`, `append`, …) are named after the originating command. The mutation bus doesn't carry the command, so the `CommandExecutor` records the active command on the `RedisDatabase` around `execute` (saved/restored for Lua nesting and parked commands). Override table matches real Redis: `INCR`/`DECR` → `incrby`, `LPUSHX` → `lpush`, `ZINCRBY` → `zincr`, `RENAME`/`RENAMENX` → `rename_from`+`rename_to`, `COPY` → `copy_to`.
- **Config** `notify-keyspace-events` lives on `RedisServerState`; `CONFIG GET`/`SET` parse, validate (exact Redis error on bad class char) and normalize the flag string (`KEA` → `AKE`). Class gating (`K`/`E`/`A`/`g`/`$`/`l`/`s`/`h`/`z`/`x`/`e`/`t`/…) matches Redis. No-op commands emit no mutation, hence no notification — matching Redis.

All event names, config normalization, and the invalid-char error were verified against a real Redis instance.

Known gaps (documented in COMMANDS.md): `SETEX`/`SET ... EX` emit only `set` (real Redis also emits a secondary `expire`); cross-DB `COPY` doesn't name the destination event; notifications are process-local to a `RedisServerState`.

Closes #87

## Test plan
- [x] New integration test `tests-integration/ioredis/keyspace-notifications-integration.test.ts` (set/del/expire/persist, type-specific write events, `expired`, RENAME, disabled, class gating, CONFIG normalization)
- [x] Test passes against **real** Redis (proves assertions match ground truth)
- [x] Test passes on **mock** (proves the implementation matches)
- [x] Unit test `tests/keyspace-notifier.test.ts` for flag parse/normalize edge cases
- [x] `npm run test:all` passes (unit 177, mock 456, real 456) + build + lint clean

### DIFF
--- a/src/state/keyspace-notifier.ts
+++ b/src/state/keyspace-notifier.ts
@@ -27,7 +27,8 @@ export type KeyspaceNotifyFlags = {
 }
 
 // The classes that 'A' (NOTIFY_ALL) expands to — everything except the
-// key-miss (m), new-key (n) and module (d) classes, matching Redis.
+// key-miss (m) and new-key (n) classes, matching Redis. Note `module` (d) IS
+// part of 'A': `Ad` collapses to `A`, while `g$lshzxet` (no d) stays expanded.
 const ALL_CLASS_KEYS = [
   'generic',
   'string',
@@ -38,6 +39,7 @@ const ALL_CLASS_KEYS = [
   'expired',
   'evicted',
   'stream',
+  'module',
 ] as const satisfies readonly (keyof KeyspaceNotifyFlags)[]
 
 function emptyFlags(): KeyspaceNotifyFlags {
@@ -124,9 +126,13 @@ export function parseKeyspaceNotifyFlags(value: string): KeyspaceNotifyFlags {
 }
 
 /**
- * Render flags back to Redis' canonical string form. When every class that 'A'
- * covers is set, those classes collapse to a single 'A' (matching Redis'
- * `keyspaceEventsFlagsToString`), e.g. parsing `"KEA"` normalizes to `"AKE"`.
+ * Render flags back to Redis' canonical string form, matching Redis 7.2's
+ * `keyspaceEventsFlagsToString` exactly. When every class 'A' covers
+ * (`g$lshzxetd`) is set they collapse to a single 'A'; otherwise the type
+ * classes (including `d`) plus `n` are emitted, then `K`/`E`, then `m` last.
+ * Note `n` is only emitted when the type classes are NOT collapsed to 'A'
+ * (e.g. `KEn` → `nKE`, but `And` → `A`), while `m` is always emitted last
+ * (e.g. `Adm` → `Am`). Examples: `KEA` → `AKE`, `KEgnd$` → `g$dnKE`.
  */
 export function keyspaceNotifyFlagsToString(
   flags: KeyspaceNotifyFlags,
@@ -144,12 +150,12 @@ export function keyspaceNotifyFlagsToString(
     if (flags.expired) result += 'x'
     if (flags.evicted) result += 'e'
     if (flags.stream) result += 't'
+    if (flags.module) result += 'd'
+    if (flags.newKey) result += 'n'
   }
-  if (flags.keyMiss) result += 'm'
-  if (flags.newKey) result += 'n'
-  if (flags.module) result += 'd'
   if (flags.keyspace) result += 'K'
   if (flags.keyevent) result += 'E'
+  if (flags.keyMiss) result += 'm'
   return result
 }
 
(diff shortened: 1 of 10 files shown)


src/state/keyspace-notifier.ts
```diff
@@ -0,0 +1,359 @@
...
+    if (flags.string) result += '$'
+    if (flags.list) result += 'l'
+    if (flags.set) result += 's'
+    if (flags.hash) result += 'h'
+    if (flags.zset) result += 'z'
+    if (flags.expired) result += 'x'
+    if (flags.evicted) result += 'e'
+    if (flags.stream) result += 't'
+  }
+  if (flags.keyMiss) result += 'm'
```
fatal10110: **Canonical form diverges from real Redis for the `m` / `n` / `d` classes.** Verified against `redis-server` 7.2.14 (`CONFIG SET notify-keyspace-events <X>; CONFIG GET notify-keyspace-events`):

| input | real Redis | this PR |
|---|---|---|
| `Km` | `Km` | `mK` |
| `KEm` | `KEm` | `mKE` |
| `Ad` | `A` | `Ad` |
| `g$lshzxet` | `g$lshzxet` | `A` |

Two root causes:

1. `ALL_CLASS_KEYS` (line 31) omits `module`. Real `NOTIFY_ALL` = `g$lshzxet` **+ `d`**. So `g$lshzxet` alone must NOT collapse to `A` (real keeps it expanded), and `Ad` MUST collapse to `A`. Here it is reversed.
2. The trailing emit order here is `m n d K E`. Real Redis `keyspaceEventsFlagsToString` is: `[A | g$lshzxet d]` then `n`, `K`, `E`, `m` — `n` before `K`/`E`, `m` last, `d` inside the type group (and part of the `A` collapse).

The unit + integration normalization cases only cover inputs without `m`/`n`/`d`, so this is uncaught. The PR claim that normalization was "verified against a real Redis instance" only held for the simple cases.

src/state/keyspace-notifier.ts
```diff
@@ -0,0 +1,359 @@
...
+  decrby: 'incrby',
+  lpushx: 'lpush',
+  rpushx: 'rpush',
+  hmset: 'hset',
+  hsetnx: 'hset',
+  zincrby: 'zincr',
+  rename: 'rename_to',
+  renamenx: 'rename_to',
+  copy: 'copy_to',
+  move: 'move_to',
```
fatal10110: `move` is dead code — `MOVE` is not implemented (`docs/COMMANDS.md:183` lists it as `[ ] MOVE`), so both the `move: 'move_to'` here and `move: 'move_from' in `DELETE_EVENT_OVERRIDES` never fire. And if MOVE is added later it is inherently cross-DB: the executor sets `activeNotifyCommand` only on `ctx.db` (the source), so the destination DB sees `activeNotifyCommand === null` and `move_to` can never be emitted — the same limitation already documented for cross-DB `COPY`. Drop the `move` entries, or document the cross-DB caveat alongside COPY.

src/state/keyspace-notifier.ts
fatal10110: Fixed in 69c4dcb. `ALL_CLASS_KEYS` now includes `module` (d), so `Ad`→`A` and `g$lshzxetd`→`A` while `g$lshzxet` (no d) stays expanded. Emit order corrected to `[A | g$lshzxet d]` then `n` (only when not collapsed — `KEn`→`nKE`, `And`→`A`), then `K`/`E`, then `m` last (`Adm`→`Am`). Verified with a 254-case differential fuzz vs redis-server 7.2.14 (0 divergences); added m/n/d cases to the unit test.

src/state/keyspace-notifier.ts
fatal10110: Fixed in 69c4dcb — dropped the `move` entries from `WRITE_EVENT_OVERRIDES`, `DELETE_EVENT_OVERRIDES` and `GENERIC_WRITE_COMMANDS`. MOVE is unimplemented (dead code), and if added it would hit the same cross-DB limitation as COPY (already documented).

PR #245, opened 2026-06-18
Add sharded Pub/Sub commands

### LINKED ISSUE #215: Feature: Sharded Pub/Sub — SSUBSCRIBE, SUNSUBSCRIBE, SPUBLISH
## Description

Implement Redis 7.0 sharded Pub/Sub:

- `SSUBSCRIBE shardchannel [shardchannel ...]` — subscribe to one or more shard channels
- `SUNSUBSCRIBE [shardchannel ...]` — unsubscribe from shard channels
- `SPUBLISH shardchannel message` — publish a message to a shard channel

Shard channels are routed by cluster slot (unlike regular pub/sub which fans out to all nodes). In standalone mode, shard channels behave like regular channels scoped to the local node.

`PUBSUB SHARDCHANNELS` and `PUBSUB SHARDNUMSUB` are already registered as stubs; they should be wired to real state after this is implemented.

## Acceptance Criteria

- `SSUBSCRIBE` / `SUNSUBSCRIBE` enforce subscribed-mode command restrictions
- `SPUBLISH` delivers to all shard-channel subscribers on the same node
- `PUBSUB SHARDCHANNELS` / `PUBSUB SHARDNUMSUB` return correct counts
- Integration test against real Redis 7.0+

fatal10110:
## Summary

- implement Redis 7 sharded Pub/Sub commands: `SSUBSCRIBE`, `SUNSUBSCRIBE`, and `SPUBLISH`
- track shard-channel subscriptions separately from regular channels/patterns while sharing subscribed-mode counts and push delivery
- wire `PUBSUB SHARDCHANNELS` and `PUBSUB SHARDNUMSUB` to real shard subscription state
- route sharded Pub/Sub commands by shard channel slot in cluster mode

## Root Cause

`PUBSUB SHARDCHANNELS` and `PUBSUB SHARDNUMSUB` were registered as compatibility stubs, but there was no shard-channel broker/session state and no command definitions for `SSUBSCRIBE`, `SUNSUBSCRIBE`, or `SPUBLISH`.

## Validation

- `npm run build`
- focused mock + real raw Pub/Sub protocol tests
- focused mock + real sharded Pub/Sub cluster tests
- focused mock + real ioredis Pub/Sub tests
- `npm test`
- `npm run lint`

Fixes #215

### DIFF
--- a/src/commands/pubsub.ts
+++ b/src/commands/pubsub.ts
@@ -205,13 +205,7 @@ export const pubsubCommand = defineCommand({
       }),
     ],
   },
-  keys: args => {
-    if (args.subcommand.toLowerCase() === 'shardnumsub') {
-      return args.args
-    }
-
-    return []
-  },
+  keys: () => [],
   execute: (args, ctx) => {
     const subcommand = args.subcommand.toLowerCase()
 
--- a/src/commands/pubsub.ts
+++ b/src/commands/pubsub.ts
@@ -54,6 +54,44 @@ export const unsubscribeCommand = defineCommand({
     framesResult(ctx.session.unsubscribePubSubChannels(args.channels)),
 })
 
+export const ssubscribeCommand = defineCommand({
+  name: 'ssubscribe',
+  schema: t.object({
+    channels: t.variadic(t.key(), { min: 1 }),
+  }),
+  flags: ['pubsub', 'noscript', 'subscribed'],
+  introspection: {
+    arity: -2,
+    flags: ['pubsub', 'noscript', 'loading', 'stale'],
+    firstKey: 1,
+    lastKey: -1,
+    keyStep: 1,
+    categories: ['@pubsub', '@slow'],
+  },
+  keys: args => args.channels,
+  execute: (args, ctx) =>
+    framesResult(ctx.session.subscribePubSubShardChannels(args.channels)),
+})
+
+export const sunsubscribeCommand = defineCommand({
+  name: 'sunsubscribe',
+  schema: t.object({
+    channels: t.variadic(t.key()),
+  }),
+  flags: ['pubsub', 'noscript', 'subscribed'],
+  introspection: {
+    arity: -1,
+    flags: ['pubsub', 'noscript', 'loading', 'stale'],
+    firstKey: 1,
+    lastKey: -1,
+    keyStep: 1,
+    categories: ['@pubsub', '@slow'],
+  },
+  keys: args => args.channels,
+  execute: (args, ctx) =>
+    framesResult(ctx.session.unsubscribePubSubShardChannels(args.channels)),
+})
+
 export const psubscribeCommand = defineCommand({
   name: 'psubscribe',
   schema: t.object({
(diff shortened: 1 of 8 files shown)


src/core/client-session.ts
```diff
@@ -163,12 +164,20 @@ export class ClientSession implements RedisClientSession {
...
+
   get pubsubPatternCount(): number {
     return this.pubsubPatterns.size
   }
 
   get pubsubSubscriptionCount(): number {
-    return this.pubsubChannelCount + this.pubsubPatternCount
+    return (
+      this.pubsubChannelCount +
+      this.pubsubShardChannelCount +
```
fatal10110: **Real-Redis incompatibility: reply counts are wrong once a client mixes shard and regular subscriptions.**

This folds shard channels into the single `pubsubSubscriptionCount`, which is then used verbatim as the count integer in *every* reply frame — `subscribe`/`unsubscribe`/`psubscribe`/`punsubscribe` (lines 435/457/475/590/612/630) **and** `ssubscribe`/`sunsubscribe` (lines 512/534/551). Real Redis keeps **two independent counters**: regular frames report `channels + patterns` (shard excluded); shard frames report shard channels only.

Verified against `redis-server 7.2.1`, one connection:
```
SUBSCRIBE ch1   -> subscribe   ch1  :1
SSUBSCRIBE sch1 -> ssubscribe  sch1 :1   # shard-only, NOT 2
SUBSCRIBE ch2   -> subscribe   ch2  :2   # excludes sch1
PSUBSCRIBE p*   -> psubscribe  p*   :3   # ch+pat, excludes sch1
SSUBSCRIBE sch2 -> ssubscribe  sch2 :2   # shard-only
```
With this PR the mock emits the combined total instead: `SSUBSCRIBE sch1` after `SUBSCRIBE ch1` replies `:2` (should be `:1`), and `SUBSCRIBE ch2` after `SSUBSCRIBE sch1` replies `:3` (should be `:2`).

Fix: keep the combined `pubsubSubscriptionCount` for `refreshPubSubMode` (mode detection is fine), but give shard frames a shard-only count and regular/pattern frames a `channels + patterns` count. The new raw-tcp test only passes because it never mixes shard + regular subs on one connection — worth adding that case.

src/commands/pubsub.ts
```diff
@@ -147,7 +205,13 @@ export const pubsubCommand = defineCommand({
       }),
     ],
   },
-  keys: () => [],
+  keys: args => {
+    if (args.subcommand.toLowerCase() === 'shardnumsub') {
+      return args.args
```
fatal10110: **Real-Redis incompatibility (cluster): `PUBSUB SHARDNUMSUB` should not be slot-routed.** Real Redis's `PUBSUB` container has no key specs, so it is never routed:
```
$ redis-cli COMMAND GETKEYS PUBSUB SHARDNUMSUB foo bar
ERR The command has no key arguments
```
By returning the channel args from `keys()`, `ClusterPolicy` computes a slot, so in cluster mode `PUBSUB SHARDNUMSUB ch1 ch2` spanning two slots throws `CROSSSLOT`, and a channel in a foreign slot throws `MOVED`. Real Redis answers locally instead — returning `0` for shard channels the node does not track, exactly like `NUMSUB`/`subscriberCount`.

It is also unnecessary: each cluster node owns its own `pubsubBroker`, so `shardSubscriberCount` already returns the correct local value (0 on the wrong node) with no redirect. Return `[]` like the other PUBSUB subcommands. The cluster test passes only because it queries `SHARDNUMSUB` with a single channel on its owner node.

PR #261, opened 2026-06-19
Fix active expiry sweep

### LINKED ISSUE #86: Bug: no active expiry sweep — DBSIZE/KEYS inflated and memory grows unboundedly
## Description

Expiry is purely passive: `getLiveEntry()` evicts a key only when it is read. No background sweep runs. This has two consequences:

1. **Memory**: In any long-running mock instance with many short-lived keys, heap usage grows without bound because expired entries are never collected unless every key is eventually read.
2. **Correctness**: `DBSIZE`, `KEYS *`, and `SCAN` all operate on the raw keyspace snapshot. Expired-but-unread keys are included in the count and returned in the results — inconsistent with real Redis.

## Real Redis behaviour

Real Redis runs an active expiry cycle (~10 times/second) that randomly samples volatile keys and evicts expired ones. This bounds memory and keeps DBSIZE/KEYS/SCAN consistent.

## Minimal fix for correctness

Filter expired entries in `DBSIZE`, `KEYS`, and `SCAN` (already somewhat done for SCAN via `getLiveEntry`, but `DBSIZE`/`KEYS` need the same treatment). A full background sweep is optional for a mock but should be documented as absent.

fatal10110:
## Summary
- Add a RedisServerState active-expiry sweep that removes expired entries without waiting for a later read.
- Run sweep work under each database SerialTurnQueue and clean up the timer through RedisServerState.close()/Resp2Server.close().
- Keep lazy eviction as a fallback and document the updated expiration lifecycle.

## Root cause
Expired entries were only evicted by read/snapshot paths. Keys that expired and were never touched again stayed allocated, and active expiry notifications were never emitted.

## Validation
- Red check: mock active-expiry notification test timed out before the fix.
- Focused unit active-expiry state test.
- Focused mock and real keyspace-notification active-expiry tests.
- npm run build
- npm test
- npm run test:integration:real (518 passed, 1 skipped, 0 failed)
- npm run lint
- git diff --check
- Pre-push hook: npm test

Fixes #86

### DIFF
--- a/src/core/transports/resp2/server.ts
+++ b/src/core/transports/resp2/server.ts
@@ -62,6 +62,7 @@ export class Resp2Server {
           reject(err)
           return
         }
+        this.state.close()
         resolve()
       }),
     )
--- a/src/core/transports/resp2/server.ts
+++ b/src/core/transports/resp2/server.ts
@@ -58,11 +58,11 @@ export class Resp2Server {
   close(): Promise<void> {
     return new Promise((resolve, reject) =>
       this.server.close(err => {
+        this.state.close()
         if (err) {
           reject(err)
           return
         }
-        this.state.close()
         resolve()
       }),
     )
--- a/src/state/server-state.ts
+++ b/src/state/server-state.ts
@@ -46,8 +46,7 @@ export class RedisServerState {
    */
   notifyKeyspaceEvents = ''
   private readonly clientSessions = new Set<RedisClientSession>()
-  private activeExpiryTimer: ReturnType<typeof setInterval> | null = null
-  private activeExpiryRunning = false
+  private activeExpiryTimer: ReturnType<typeof setTimeout> | null = null
   private closed = false
   private luaRuntimePromise: Promise<RedisLuaRuntime> | null = null
 
@@ -136,7 +135,7 @@ export class RedisServerState {
   close(): void {
     this.closed = true
     if (this.activeExpiryTimer) {
-      clearInterval(this.activeExpiryTimer)
+      clearTimeout(this.activeExpiryTimer)
       this.activeExpiryTimer = null
     }
   }
@@ -151,39 +150,50 @@ export class RedisServerState {
       throw new Error(`Invalid active expiry interval ${resolvedIntervalMs}`)
     }
 
-    const timer = setInterval(() => {
-      void this.runActiveExpiry()
-    }, resolvedIntervalMs)
-    ;(
-      timer as ReturnType<typeof setInterval> & { unref?: () => void }
-    ).unref?.()
+    this.scheduleActiveExpiry(resolvedIntervalMs)
+  }
+
+  private scheduleActiveExpiry(intervalMs: number): void {
+    const timer = setTimeout(() => {
+      this.activeExpiryTimer = null
+      void this.runActiveExpiryTick(intervalMs)
+    }, intervalMs)
+
+    ;(timer as ReturnType<typeof setTimeout> & { unref?: () => void }).unref?.()
     this.activeExpiryTimer = timer
   }
 
+  private async runActiveExpiryTick(intervalMs: number): Promise<void> {
+    try {
+      await this.runActiveExpiry()
+    } catch {
+      // Active expiry is best-effort; command reads still lazily expire keys.
+    } finally {
+      if (!this.closed) {
+        this.scheduleActiveExpiry(intervalMs)
+      }
+    }
+  }
+
   private async runActiveExpiry(): Promise<void> {
-    if (this.closed || this.activeExpiryRunning) {
+    if (this.closed) {
       return
     }
 
-    this.activeExpiryRunning = true
-    try {
-      const now = Date.now()
-      for (const database of this.databases) {
-        if (this.closed) {
-          return
-        }
-
-        const turn = await database.turnQueue.waitTurn()
-        try {
-          if (!this.closed) {
-            database.sweepExpired(now)
-          }
-        } finally {
-          turn.release()
-        }
+    const now = Date.now()
+    for (const database of this.databases) {
+      if (this.closed) {
+        return
       }
-    } finally {
-      this.activeExpiryRunning = false
+
+      const turn = await database.turnQueue.waitTurn()
+      try {
+        if (!this.closed) {
+          database.sweepExpired(now)
+        }
+      } finally {
+        turn.release()
+      }
     }
   }
 }
--- a/src/state/server-state.ts
+++ b/src/state/server-state.ts
@@ -18,12 +18,19 @@ export type RedisServerStateOptions = {
   pubsubBroker?: RedisPubSubBroker
   scriptCache?: RedisScriptCache
   /**
+   * Interval for the background active-expiry sweep. Set to `false` to disable
+   * the timer for tests that need full manual control of expiration.
+   */
+  activeExpiryIntervalMs?: number | false
+  /**
    * Optional server password (Redis `requirepass`). When set, connections start
    * unauthenticated and must `AUTH` before running most commands. When unset,
    * the default user is `nopass` and no authentication is enforced.
    */
   requirepass?: string
 }
+
+const DEFAULT_ACTIVE_EXPIRY_INTERVAL_MS = 100
 
 export class RedisServerState {
   readonly databases: RedisDatabase[]
@@ -39,6 +46,9 @@ export class RedisServerState {
    */
   notifyKeyspaceEvents = ''
   private readonly clientSessions = new Set<RedisClientSession>()
+  private activeExpiryTimer: ReturnType<typeof setInterval> | null = null
+  private activeExpiryRunning = false
+  private closed = false
   private luaRuntimePromise: Promise<RedisLuaRuntime> | null = null
 
   constructor(options?: RedisServerStateOptions) {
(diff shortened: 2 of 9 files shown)


src/state/server-state.ts
```diff
@@ -71,6 +81,8 @@ export class RedisServerState {
         ),
       )
     }
+
+    this.startActiveExpiry(options?.activeExpiryIntervalMs)
```
fatal10110, line 84: **Replica nodes now run active expiry independently — diverges from Redis replica semantics.**

`startActiveExpiry` is called unconditionally for every `RedisServerState`, and the cluster builder constructs replica node states with the default interval (`src/cluster.ts:233` — `new RedisServerState({ databaseCount, clusterTopology })`, no `activeExpiryIntervalMs`). Replica databases hold replicated keys *with* their `expiresAt` (`syncDatabase`/`applyReplicationEvent`, `src/cluster.ts:274` and `:318`), so each replica will now delete expired keys and emit its own `evict` → `expired` keyspace notification on its own ~100ms clock.

Real Redis replicas never expire keys on their own: a replica keeps a logically-expired key in memory (serving it as nil on read) and only physically deletes it + fires the `expired` notification when the master propagates the DEL/UNLINK. Self-expiry on the replica breaks two things this repo deliberately models:

- `replicaUpdateDelayMs` replication-lag semantics — the replica now drops keys on its own schedule regardless of the configured delay.
- Notification fidelity — the replica emits `expired` independently of (and can double-emit alongside) the master’s replicated `evict`→`delete`.

No current test catches this (default `replicasPerMaster` is 0), so it is latent, but it is a real divergence in an area the codebase models carefully.

Suggested fix: build replica states with the option this PR already added — `activeExpiryIntervalMs: false` when `node.role === 'replica'` — or gate `runActiveExpiry` on master role.

src/state/server-state.ts
```diff
@@ -112,4 +124,66 @@ export class RedisServerState {
...
+      return
+    }
+
+    const resolvedIntervalMs = intervalMs ?? DEFAULT_ACTIVE_EXPIRY_INTERVAL_MS
+    if (!Number.isInteger(resolvedIntervalMs) || resolvedIntervalMs < 1) {
+      throw new Error(`Invalid active expiry interval ${resolvedIntervalMs}`)
+    }
+
+    const timer = setInterval(() => {
+      void this.runActiveExpiry()
```
fatal10110: **Fire-and-forget sweep can surface as an unhandledRejection.**

`void this.runActiveExpiry()` discards the promise. The sweep chain (`sweepExpired` → `evictIfExpired` → mutation emit → `KeyspaceNotifier.handle` → `broker.publish` → subscriber listener) runs here with no surrounding error handling — unlike the lazy path, which runs inside a command’s turn where the session/adapter contains any throw. If a listener or the notifier throws during a sweep, this becomes an unhandled promise rejection fired from a timer callback, which fails the node test runner and can crash the process in production.

Recommend not discarding it, e.g. `void this.runActiveExpiry().catch(...)` (the class has no logger, so at minimum route/swallow the error rather than leaving it unhandled).

src/state/server-state.ts
fatal10110, line 84: Following up on my earlier comment here — I reproduced this against a real Redis 7.2.1 master+replica pair, which confirms the divergence and warrants bumping it to **high**.

With **no read on the replica**, a real replica keeps the expired key physically present and never emits `expired` on its own clock — it only emits `del` (class `g`) if/when the master propagates the DEL:

- Real replica, key `SET ... PX 1200` on the master, 3s past TTL, zero reads: `DBSIZE` on the replica still returns `1` (key physically retained), `DEBUG OBJECT` still finds it, `GET` returns nil (read-time logical expiry only). The replica’s `__keyevent@0__:*` stream contained only `set` and `expire` — **no `expired`, no `del`**.
- When the master was then allowed to propagate the delete, the replica emitted `__keyevent@0__:del` (not `expired`), while the master emitted `__keyevent@0__:expired`.

This PR makes every replica `RedisServerState` (built at `src/cluster.ts:233`) run its own 100ms sweep, so a mock replica will instead physically delete the key and emit `__keyevent@0__:expired` on its own clock — wrong event name, wrong class, wrong timing, and it breaks `replicaUpdateDelayMs` lag semantics.

Master-side active expiry is correct (verified: a real master emits `__keyevent@0__:expired` with no forcing read, matching the `evict`→`expired` / class `x` mapping). Fix: suppress the sweep on replicas — thread `activeExpiryIntervalMs: false` for replica states in `createClusterNodeStates`, leaving replica expiry driven solely by the master’s propagated `evict`→`replica.delete()`→`del` path that `applyReplicationEvent` already implements.

src/core/transports/resp2/server.ts
```diff
@@ -62,6 +62,7 @@ export class Resp2Server {
           reject(err)
           return
         }
+        this.state.close()
```
fatal10110: **`close()` skips `state.close()` on the `server.close()` error path, leaking the active-expiry timer.**

`this.state.close()` only runs on the success branch; if `this.server.close()` calls back with an error it `reject`s and the interval is never cleared. This is reachable: `buildRedisCluster` constructs every node’s `RedisServerState` — and the constructor starts the 100ms `setInterval` immediately — *before* `cluster.listen()`. If a cluster is built and torn down without a successful `listen()` (an error between build and listen, or a test that builds then aborts), each `server.close()` invokes its callback with `ERR_SERVER_NOT_RUNNING`, so `state.close()` is skipped and the timer leaks. `RedisCluster.close()` compounds it via `Promise.all`, where the first rejection discards the remaining `close()`s, leaking every node’s timer.

Timers are `unref`’d so the process can still exit, but the lifecycle is inconsistent — the timer starts unconditionally in the constructor yet is only disposed on a clean close after a successful listen. Clear it regardless of the error, e.g. run `this.state.close()` in a `finally` (or before the `if (err)` branch).

src/state/server-state.ts
```diff
@@ -112,4 +124,66 @@ export class RedisServerState {
...
+    if (intervalMs === false) {
+      return
+    }
+
+    const resolvedIntervalMs = intervalMs ?? DEFAULT_ACTIVE_EXPIRY_INTERVAL_MS
+    if (!Number.isInteger(resolvedIntervalMs) || resolvedIntervalMs < 1) {
+      throw new Error(`Invalid active expiry interval ${resolvedIntervalMs}`)
+    }
+
+    const timer = setInterval(() => {
```
fatal10110: **Prefer a self-scheduling `setTimeout` over `setInterval` for this async sweep (style/robustness).**

`setInterval` fires on a fixed cadence regardless of whether the previous async tick finished. You have correctly defended the overlap that would otherwise cause with the `activeExpiryRunning` guard — so this is not a correctness bug — but a self-scheduling `setTimeout`, re-armed in a `finally` after each pass completes, is the cleaner shape:

```ts
private scheduleActiveExpiry(intervalMs: number): void {
  const tick = async () => {
    if (this.closed) return
    try {
      await this.runActiveExpiry()
    } catch (err) {
      // route to a logger instead of an unhandled rejection
    } finally {
      if (!this.closed) {
        this.activeExpiryTimer = setTimeout(tick, intervalMs)
        this.activeExpiryTimer.unref?.()
      }
    }
  }
  this.activeExpiryTimer = setTimeout(tick, intervalMs)
  this.activeExpiryTimer.unref?.()
}
```

Why it is nicer here:
- The next delay is measured **from completion**, so there is no drift and no stream of wasted wakeups that just early-return while a slow pass (a busy `SerialTurnQueue`) is in flight.
- It removes the need for the `activeExpiryRunning` reentrancy flag entirely — overlap is impossible by construction.
- The `try/catch` in `tick` also closes the `void this.runActiveExpiry()` unhandled-rejection gap flagged on the next line, so both points collapse into one change.

Low priority — not blocking, but worth folding in while this area is being touched.

src/state/server-state.ts
fatal10110: Addressed in 8186907. Replica node states are now created with `activeExpiryIntervalMs: false`, so only masters run the active expiry sweep and replicas rely on the master’s replicated delete path. I also added a cluster regression test covering delayed replicas so a replica does not physically expire the key or emit its own expiry event on its clock.

src/state/server-state.ts
fatal10110: Addressed in 8186907. The active expiry tick now catches sweep failures before re-arming the next timer, so listener/notifier failures cannot surface as unhandled rejections. I added a regression test with a throwing keyspace notification listener to cover this path.

src/state/server-state.ts
fatal10110: Addressed in 8186907 with the replica sweep disabled at cluster state construction: replica `RedisServerState` instances now receive `activeExpiryIntervalMs: false`. The new regression test uses `replicaUpdateDelayMs` to verify the replica still has the replicated key after its TTL while the master-side expiry/delete propagation is delayed, preserving the lag semantics described here.

src/core/transports/resp2/server.ts
fatal10110: Addressed in 8186907. `Resp2Server.close()` now closes `RedisServerState` before handling the `server.close()` callback error, so the active expiry timer is cleared even when the TCP server was never listening. I added a regression test for closing an unstarted server.

src/state/server-state.ts
fatal10110: Folded into 8186907. The active expiry loop now uses a self-scheduling `setTimeout`, re-arms after each async tick completes, and catches tick failures before scheduling the next pass. That removed the interval overlap guard and closed the unhandled-rejection gap at the same time.

Issue #256, opened 2026-06-19
Feature: HGETEX


fatal10110:
Part of #207.

## Scope
`HGETEX key [EX seconds|PX ms|EXAT unix-s|PXAT unix-ms|PERSIST|KEEPTTL] FIELDS numfields field [field ...]` — get fields and optionally update their TTL.

## Depends on
Per-field TTL tracking added in the HEXPIRE/HPEXPIRE/HEXPIREAT/HPEXPIREAT sub-issue.

## Acceptance Criteria
- Returns field values like `HMGET`
- Optional TTL-modifying clause applies per the same semantics as `HEXPIRE`/`HPERSIST`
- Integration tests against real Redis 7.4+

Issue #255, opened 2026-06-19
Feature: HPERSIST, HTTL, HPTTL


fatal10110:
Part of #207.

## Scope
- `HPERSIST key FIELDS numfields field [field ...]` — remove field TTLs
- `HTTL key FIELDS numfields field [field ...]` — get field TTLs in seconds
- `HPTTL key FIELDS numfields field [field ...]` — get field TTLs in ms

## Depends on
Per-field TTL tracking added in the HEXPIRE/HPEXPIRE/HEXPIREAT/HPEXPIREAT sub-issue.

## Acceptance Criteria
- All three commands implemented, returning per-field result codes matching real Redis (`-2`=no field, `-1`=no TTL, else value)
- Integration tests against real Redis 7.4+

Issue #207, opened 2026-06-16
Feature: Hash-field TTL family (HEXPIRE, HPEXPIRE, HEXPIREAT, HPEXPIREAT, HPERSIST, HTTL, HPTTL, HGETDEL, HGETEX)


fatal10110:
## Description

Implement the Redis 7.4 per-field expiry commands for hashes:

- `HEXPIRE key seconds [NX|XX|GT|LT] FIELDS numfields field [field ...]` — set field TTL in seconds
- `HPEXPIRE key milliseconds [NX|XX|GT|LT] FIELDS numfields field [field ...]` — set field TTL in ms
- `HEXPIREAT key unix-time-seconds [NX|XX|GT|LT] FIELDS numfields field [field ...]` — set field expiry as Unix timestamp
- `HPEXPIREAT key unix-time-milliseconds [NX|XX|GT|LT] FIELDS numfields field [field ...]` — set field expiry as Unix timestamp in ms
- `HPERSIST key FIELDS numfields field [field ...]` — remove field TTLs
- `HTTL key FIELDS numfields field [field ...]` — get field TTLs in seconds
- `HPTTL key FIELDS numfields field [field ...]` — get field TTLs in ms
- `HGETDEL key FIELDS numfields field [field ...]` — get and delete hash fields atomically
- `HGETEX key [EX seconds|PX ms|EXAT unix-s|PXAT unix-ms|PERSIST|KEEPTTL] FIELDS numfields field [field ...]` — get fields and optionally update their TTL

## Notes

Requires per-field expiry tracking in the hash data structure, which is a non-trivial state change. The existing `RedisDataValue` hash type stores a plain `Map<string,Buffer>`; per-field TTLs require a richer entry type.

## Acceptance Criteria

- All nine commands implemented and registered
- Per-field lazy expiration integrated with `HGET`/`HGETALL`/`HMGET` etc.
- NX/XX/GT/LT condition flags on expiry-set commands
- Integration tests against real Redis 7.4+

fatal10110:
All sub-issues complete: #254 (HEXPIRE/HPEXPIRE/HEXPIREAT/HPEXPIREAT + per-field TTL infra), #255 (HPERSIST/HTTL/HPTTL), #257 (HGETDEL), #256 (HGETEX, PR #272 merged). Full hash-field TTL family implemented and covered by integration tests against real Redis 8.0.6.

PR #268, opened 2026-06-19
Add hash field expiry commands

### LINKED ISSUE #254: Feature: hash-field TTL infra + HEXPIRE/HPEXPIRE/HEXPIREAT/HPEXPIREAT
Part of #207.

## Scope
Foundational change: extend the hash `RedisDataValue` to track a per-field expiry timestamp (today it is a plain `Map<string,Buffer>`), plus lazy per-field eviction integrated into existing hash reads.

- `HEXPIRE key seconds [NX|XX|GT|LT] FIELDS numfields field [field ...]`
- `HPEXPIRE key milliseconds [NX|XX|GT|LT] FIELDS numfields field [field ...]`
- `HEXPIREAT key unix-time-seconds [NX|XX|GT|LT] FIELDS numfields field [field ...]`
- `HPEXPIREAT key unix-time-milliseconds [NX|XX|GT|LT] FIELDS numfields field [field ...]`

## Acceptance Criteria
- Hash data structure tracks per-field `expiresAt`
- Lazy per-field expiration wired into `HGET`/`HGETALL`/`HMGET`/`HKEYS`/`HVALS`/`HLEN`/etc. (a field past its TTL behaves as deleted)
- All four commands implemented with `NX`/`XX`/`GT`/`LT` condition flags
- Reply: array of per-field result codes (`2`=deleted immediately since past TTL, `1`=set, `0`=condition not met, `-2`=no such field)
- Integration tests against real Redis 7.4+

fatal10110:
## Summary
- Add per-field hash expiry metadata and lazy tracked hash cleanup.
- Implement HEXPIRE, HPEXPIRE, HEXPIREAT, and HPEXPIREAT with NX/XX/GT/LT conditions and Redis return codes.
- Cover mock and real Redis behavior for success paths, parser errors, lazy expiry visibility, and TTL clear/preserve semantics; update command/state docs.

## Root cause
Hash fields only stored field/value pairs, so Redis 7.4 hash-field TTL commands had no state model for per-field expiration and were absent from the command registry.

## Validation
- npm run build
- npm test
- TEST_BACKEND=mock node --enable-source-maps --import tsx --no-warnings --test ./tests-integration/ioredis/hash-integration.test.ts
- npm run clean:redis && TEST_BACKEND=real node --enable-source-maps --import tsx --no-warnings --test-concurrency 1 --test-timeout 30000 --test ./tests-integration/ioredis/hash-integration.test.ts
- npm run clean:redis && npm run test:integration:real
- npm run lint
- git push pre-push hook: npm test

Fixes #254

Part of #207. Leaves #255 (HPERSIST/HTTL/HPTTL) and #256 (HGETEX) open for follow-up work.

### DIFF
--- a/src/commands/hashes.ts
+++ b/src/commands/hashes.ts
@@ -5,7 +5,9 @@ import {
   t,
   type ParseContext,
 } from '../core/command-schema'
+import type { RedisExecutionContext } from '../core/redis-context'
 import {
+  ExpectedIntegerError,
   HashValueNotFloatError,
   HashValueNotIntegerError,
   RedisCommandError,
@@ -14,6 +16,7 @@ import {
 } from '../core/redis-error'
 import { RedisResult } from '../core/redis-result'
 import { RedisValue } from '../core/redis-value'
+import type { TrackedHashData } from '../state/tracked-values'
 import { bulk, integer, ok, array, parseIntegerToken } from './helpers'
 
 type FieldValuePair = { field: Buffer; value: Buffer }
@@ -26,8 +29,23 @@ type HgetdelArgs = {
   key: Buffer
   fields: Buffer[]
 }
+type HashExpireOption = 'NX' | 'XX' | 'GT' | 'LT'
+type HashExpireMode =
+  | 'seconds'
+  | 'milliseconds'
+  | 'unix-seconds'
+  | 'unix-milliseconds'
+type HashExpireArgs = {
+  key: Buffer
+  time: bigint
+  option: HashExpireOption | undefined
+  fields: Buffer[]
+}
 
 const LONG_MAX = 9223372036854775807n
+const LONG_MIN = -9223372036854775808n
+const MAX_SAFE_INTEGER_BIGINT = BigInt(Number.MAX_SAFE_INTEGER)
+const MIN_SAFE_INTEGER_BIGINT = BigInt(Number.MIN_SAFE_INTEGER)
 
 function createFieldValuePairsSchema() {
   return t.custom<FieldValuePair[]>(
@@ -123,6 +141,59 @@ function createHgetdelSchema() {
   )
 }
 
+function createHashExpireSchema() {
+  return t.custom<HashExpireArgs>(
+    (input: readonly Buffer[], index: number, ctx: ParseContext) => {
+      const key = input[index]
+      const timeToken = input[index + 1]
+      if (!key || !timeToken) {
+        throw new WrongNumberOfArgumentsError(ctx.commandName)
+      }
+
+      const time = parseHashExpireTime(timeToken)
+      let cursor = index + 2
+      if (cursor >= input.length) {
+        throw new WrongNumberOfArgumentsError(ctx.commandName)
+      }
+
+      let option: HashExpireOption | undefined
+      const maybeOption = input[cursor].toString().toUpperCase()
+      if (isHashExpireOption(maybeOption)) {
+        option = maybeOption
+        cursor += 1
+      }
+
+      const fieldsToken = input[cursor]
+      if (!fieldsToken) {
+        throw new WrongNumberOfArgumentsError(ctx.commandName)
+      }
+      if (fieldsToken.toString().toUpperCase() !== 'FIELDS') {
+        throw new RedisCommandError(
+          'Mandatory argument FIELDS is missing or not at the right position',
+        )
+      }
+
+      const fieldCountToken = input[cursor + 1]
+      if (!fieldCountToken) {
+        throw new WrongNumberOfArgumentsError(ctx.commandName)
+      }
+
+      const fieldCount = parseHashExpireFieldCount(fieldCountToken)
+      const fields = input.slice(cursor + 2)
+      if (fieldCount !== BigInt(fields.length)) {
+        throw new RedisCommandError(
+          'The `numfields` parameter must match the number of arguments',
+        )
+      }
+
+      return {
+        value: { key, time, option, fields },
+        nextIndex: input.length,
+      }
+    },
+  )
+}
+
 function parsePositiveFieldCount(token: Buffer): bigint {
   const raw = token.toString()
   if (!isIntegerToken(raw)) {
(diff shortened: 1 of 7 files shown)


src/commands/hashes.ts
```diff
@@ -137,6 +208,140 @@ function parsePositiveFieldCount(token: Buffer): bigint {
   return value
 }
 
+function parseHashExpireTime(token: Buffer): bigint {
```
fatal10110, line 217: **[Redis compat — high] Time validation missing: negative and out-of-range expire times are silently accepted instead of erroring.**

Ground-truthed against real Redis 8.0.6:
```
HEXPIRE h -1 FIELDS 1 f1            -> ERR invalid expire time, must be >= 0
HEXPIRE h 99999999999 FIELDS 1 f   -> ERR invalid expire time in 'hexpire' command\nHEXPIREAT h 9999999999999999 ...   -> ERR invalid expire time in 'hexpireat' command\nHEXPIRE h 9223372036854775807 ...  -> ERR invalid expire time in 'hexpire' command\n```\nReproduced against this mock (driving the command directly):\n```\nHEXPIRE  h -1                  -> [2]   (field + key silently deleted)\nHEXPIRE  h 9223372036854775807 -> [1]   (stored expiresAt = 9008981145513728, a bogus ~285y TTL)\n```\n\nTwo distinct bugs, both rooted here / in `hashExpireTimeToTimestamp` + `bigintToNumberSaturated` (hashes.ts:296,315):\n1. **Negative time -> silent data loss.** A negative value produces a past timestamp, so `expireHashFields` deletes the field (and the key if it was the last field) and returns `2` -- real Redis rejects it with `ERR invalid expire time, must be >= 0` and leaves the data intact. (Non-negative past times like `0`/`1` *are* valid and correctly return `2`.)\n2. **Overflow -> saturation.** `bigintToNumberSaturated` clamps to `Number.MAX_SAFE_INTEGER` and returns success `1` with a fabricated far-future TTL; real Redis returns an error.\n\nThe repo already has `InvalidExpireTimeError` -> `"invalid expire time in '<cmd>' command"` (redis-error.ts:194). Validate the *final* expire timestamp: reject `< 0` (with the `must be >= 0` wording) and reject values whose ms representation overflows the valid range, rather than saturating.

File: src/state/tracked-values.ts [hunk 2]
```diff
@@ -62,16 +78,55 @@ export class TrackedHashData {
...
+    }
+
+    if (entry.expiresAt !== expiresAt) {
+      entry.expiresAt = expiresAt
+      this.tracker.markChanged()
+    }
+    return true
+  }
+
   entries(): IterableIterator<RedisHashField> {
```
fatal10110, line 106: **[Redis compat / code design — high] HSCAN bypasses hash-field expiry -- expired fields leak.**

This PR makes `entries()`/`size`/`getField`/`hasField` expiry-aware, and correctly routes HGET/HMGET/HGETALL/HKEYS/HVALS/HLEN/HEXISTS/HSTRLEN/HRANDFIELD through them. But **HSCAN** ([src/commands/scan.ts:85](src/commands/scan.ts#L85)) still iterates the raw map directly:
```ts
const hash = ctx.db.getHash(args.key)
...
for (const { field, value } of hash.fields.values()) {  // not expiry-aware
```
so it surfaces fields whose TTL has elapsed but that have not been lazily collected yet.

Real Redis 8.0.6:
```
HSET hs keep 1 gone 2; HEXPIRE hs 1000 FIELDS 1 keep; HPEXPIRE hs 50 FIELDS 1 gone; sleep
HSCAN hs 0 -> 1) "0"  2) 1) "keep" 2) "1"      (gone is absent)
```
Reproduced against this mock -- HSCAN returns **both** `keep` and the expired `gone`.

Fix: route HSCAN through the expiry-aware accessor (`ctx.db.updateHash(key, h => h.entries())` or an equivalent read path) like the other hash readers. SSCAN/ZSCAN are unaffected (no field TTL).

src/commands/hashes.ts
```diff
@@ -137,6 +208,140 @@ function parsePositiveFieldCount(token: Buffer): bigint {
...
+function withExistingHash<TResult>(
+  ctx: RedisExecutionContext,
+  key: Buffer,
+  missing: () => TResult,
+  callback: (hash: TrackedHashData) => TResult,
+): TResult {
+  const hash = ctx.db.getHash(key)
+  if (!hash) return missing()
+
+  return ctx.db.updateHash(key, callback)
```
fatal10110, line 262: **[Redis compat — medium] WRONGTYPE error precedence.**

The key-type check happens here in `execute` (via `ctx.db.getHash`), which runs *after* the schema parser has already validated the time/numfields tokens at plan() time. So on a wrong-type key, malformed args win where real Redis returns WRONGTYPE first.

Real Redis 8.0.6 (`SET strk hello`):
```
HEXPIRE strk abc FIELDS 1 f1   -> WRONGTYPE Operation against a key holding the wrong kind of value
HEXPIRE strk 100 FIELDS 5 f1   -> WRONGTYPE Operation against a key holding the wrong kind of value
HEXPIRE strk 100 FIELDS 0 f1   -> WRONGTYPE Operation against a key holding the wrong kind of value
```
This mock returns `value is not an integer or out of range` / `The numfields parameter must match...` / `Parameter numFields should be greater than 0` respectively. (Arity is checked first by both, so the too-few-args case matches.)

Harder to fix cleanly given the layered parse-before-execute design, but worth a note: real Redis resolves key type early. Lower priority than #1/#2.

src/commands/hashes.ts
```diff
@@ -123,6 +141,59 @@ function createHgetdelSchema() {
...
+        cursor += 1
+      }
+
+      const fieldsToken = input[cursor]
+      if (!fieldsToken) {
+        throw new WrongNumberOfArgumentsError(ctx.commandName)
+      }
+      if (fieldsToken.toString().toUpperCase() !== 'FIELDS') {
+        throw new RedisCommandError(
+          'Mandatory argument FIELDS is missing or not at the right position',
```
fatal10110: **[Redis compat — medium] Missing FIELDS keyword yields the wrong error.**

When the FIELDS keyword is omitted but the token count is otherwise plausible, this throws `Mandatory argument FIELDS is missing or not at the right position`. Real Redis returns a wrong-number-of-arguments error instead.

Real Redis 8.0.6:
```
HEXPIRE h 100 1 f1   -> ERR wrong number of arguments for 'hexpire' command\n```\nHere: `cursor` lands on `1`, it is not an option, `fieldsToken=\"1\" != \"FIELDS\"`, so the `Mandatory argument FIELDS...` branch fires.\n\n(The `Mandatory argument FIELDS...` wording *is* correct when FIELDS is genuinely misplaced, e.g. `HEXPIRE h 100 1 FIELDS f1` -- verified against real Redis. Only the keyword-omitted case is wrong.)

File: src/commands/hashes.ts [hunk 5]
```diff
@@ -137,6 +208,140 @@ function parsePositiveFieldCount(token: Buffer): bigint {
...
+          const entry = hash.getField(field)
+          if (!entry) {
+            return RedisValue.integer(-2)
+          }
+
+          if (!shouldApplyHashExpire(entry.expiresAt, expiresAt, args.option)) {
+            return RedisValue.integer(0)
+          }
+
+          if (expiresAt <= now) {
```
fatal10110: **[Redis compat — medium, follow-up] Hash-field-expiry keyspace notifications.**

This is the branch where a past-time expiry removes a field (and may empty the key). Real Redis emits distinct keyspace events that a generic write/del mutation cannot reproduce.

Real Redis 8.0.6 with `notify-keyspace-events KEA`:
```
HEXPIRE hk 100 FIELDS 1 f1   -> __keyevent@0__:hexpire hk
HEXPIREAT hk 1 FIELDS 1 f1   -> __keyevent@0__:hdel hk          (field removed by past time -> hdel, NOT hexpire)
HEXPIREAT hk 1 FIELDS 1 f2   -> __keyevent@0__:hdel hk  then  __keyevent@0__:del hk   (last field -> key delete)
```
The notifier (src/state/keyspace-notifier.ts, not touched by this PR) cannot currently distinguish TTL-set (`hexpire`) from field-removal (`hdel`) from key-delete (`del`). Out of scope for the core command, but worth tracking as a follow-up so notification fidelity matches once the high-severity items land.

tests-integration/ioredis/hash-integration.test.ts
```diff
@@ -489,6 +493,395 @@ describe(`Hash Commands Integration (${testRunner.getBackendName()})`, () => {
     }
   })
 
+  test('HEXPIRE family sets and lazily expires hash fields', async () => {
```
fatal10110, line 496: **[Code quality — low] Coverage gaps.** Two behaviors the suite does not exercise (and both turned out to be real bugs -- see the high-severity comments):

1. **HSCAN after field expiry** -- no HSCAN test against a hash with an expired field, so the leak (HSCAN returning expired fields) is uncaught. Suggest: expire one field, assert HSCAN omits it.
2. **Whole-key removal when all fields expire** -- the HEXPIRE-family tests always leave at least one field alive, so the `isEmptyCollection` key-delete path for an HEXPIRE-emptied hash is never asserted end-to-end (only HDEL/HGETDEL cover empty-key removal). Suggest: expire *every* field, then assert `EXISTS == 0` / `TYPE == none`.

src/commands/hashes.ts
```diff
@@ -123,6 +144,62 @@ function createHgetdelSchema() {
...
+function createHashExpireSchema() {
+  return t.custom<HashExpireArgs>(
+    (input: readonly Buffer[], index: number, ctx: ParseContext) => {
+      const key = input[index]
+      if (!key || input.length - index < 5) {
+        throw new WrongNumberOfArgumentsError(ctx.commandName)
+      }
+
+      return {
+        value: { key, rawArgs: input.slice(index + 1) },
```
fatal10110, line 156: **[Re-review — low / informational] Deferred arg validation: wire-correct, but now the lone keyed command doing it.**

The schema now only checks arity and captures `rawArgs`; all content parsing (time / option / FIELDS / numfields) moved into `parseHashExpireArgs` at execute time. Two consequences, both verified:

- **Good news (ground-truthed against real Redis 8.0.6):** this actually matches real Redis MULTI/EXEC semantics. Bad arg-content inside a transaction is QUEUED and the error surfaces *inside* the EXEC array with the rest of the transaction still running -- only an arity error aborts at queue time with EXECABORT. Real Redis:
  ```
  MULTI; HSET h f v; HEXPIRE h abc FIELDS 1 f; HGET h f; EXEC
    -> 1) (integer) 1
       2) ERR value is not an integer or out of range
       3) "v"          (transaction NOT aborted)
  MULTI; HEXPIRE h 100; EXEC   -> ERR wrong number of arguments ... then EXECABORT
  ```
  HEXPIRE now reproduces this exactly.

- **Caveat (intra-repo consistency):** the sibling hash commands (e.g. HGETDEL/HSET/HMSET) still fully validate arg-content in their *schema*, so a bad-arg one of those inside MULTI goes EXECABORT at queue time -- which is the divergent-from-Redis behavior. So this PR is *more* wire-correct, not less; flagging only so the inconsistency is a deliberate decision and a follow-up can align the siblings rather than revert this.

Suggest: add a MULTI/EXEC test locking the inline-error (no-abort) behavior so a future refactor cannot silently flip it. (Also minor: `expireHashFields` inlines `getHash` + `updateHash` rather than reusing `withExistingHash` -- because the parse must run before the missing-vs-present branch; acceptable, just an altitude nit.)

fatal10110 (commented):
## Review — Hash field expiry commands (re-review @ e517c19)

**Verdict: ship** (posting as a comment — can't self-approve). Reviewed for code quality, design, and **real-Redis wire compatibility** (ground-truthed against Redis 8.0.6 and reproduced on the mock). The follow-up commit `e517c19` resolves every blocker from the first pass.

### Prior findings — all resolved
| # | Finding | Status |
|---|---------|--------|
| 1 | Time validation (negative → silent delete; overflow → saturated bogus TTL) | ✅ Fixed — negative → `ERR invalid expire time, must be >= 0` (field left intact); overflow → `ERR invalid expire time in '<cmd>' command`. The `(2^48-1)>>2` bound matches real Redis's exact cutoff for all four modes, relative `+now` included. |
| 2 | HSCAN returned expired fields | ✅ Fixed — routed through the expiry-aware `entries()`. |
| 3 | WRONGTYPE precedence vs arg-parse errors | ✅ Fixed — `getHash` (throws WRONGTYPE) runs before arg parsing. |
| 4 | Missing FIELDS keyword → wrong error | ✅ Fixed — arity gate handles the reported case; genuine-misplace wording verified correct. |
| 6 | Coverage gaps | ✅ Addressed — new tests cover HSCAN-omits-expired, whole-key removal (EXISTS=0/TYPE=none), WRONGTYPE×3, negative, and overflow. |

### One new, non-blocking note
The refactor moved arg-content validation from the schema (queue time) into `execute`. Ground-truthing confirmed this actually **matches** real Redis MULTI/EXEC semantics (bad arg-content → QUEUED + inline error in the EXEC array, no EXECABORT; only arity aborts). So HEXPIRE is now *more* wire-correct than its in-schema-validating siblings — not a regression. Left an inline comment suggesting a MULTI/EXEC test to lock the behavior, and a follow-up to align the sibling commands.

### Still open (pre-existing, out of scope)
Hash-field-expiry keyspace notifications (`hexpire` / `hdel` / `del` event types) — the notifier is untouched by this PR; tracked as a follow-up.

Verified clean: dead code fully removed, no OOB in the new parser, overflow math has no precision loss, HSCAN refactor does not over-dirty WATCH, tests are node:test/assert only. Suite green on the new head.

PR #285, opened 2026-06-23
feat(client-mocks): in-memory node-redis facade (createNodeRedisMock)

fatal10110:
## What

**Stream B** of the client-mocks plan ([docs/CLIENT-MOCKS-PLAN.md](docs/CLIENT-MOCKS-PLAN.md)), built on the `buildClusterNodes()` foundation (PR #283, now merged).

Adds `createNodeRedisMock()` (standalone **and** cluster). node-redis has **no** socket/connector hook (`#socketFactory` is private), so this is a **Tier 2 hand-written facade** replicating the `createClient`/`createCluster` surface, routing commands through a per-connection `ClientSession` on the shared in-memory pipeline.

> Tier 1.5 (`transformReply` reuse) was evaluated and skipped as unnecessary: node-redis leaves nearly all RESP2 replies untransformed, so the shared `decode()` (`RedisValue → native JS`) already yields node-redis-correct shapes. Curated methods are thin coercions, not a per-command reply table.

### New
- `src/client-mocks/node-redis-mock.ts` — `createNodeRedisMock()` + `NodeRedisMockClient` / `NodeRedisMockCluster` / `NodeRedisMockMulti`:
  - curated camelCase methods (`get/set/hSet/hGetAll/lRange/zAdd/zRange/...`) + generic `sendCommand()` escape hatch (documented fall-through),
  - `multi`/`exec`/`discard`, `watch`,
  - pub/sub (`subscribe`/`pSubscribe` + callbacks) via a dedicated push-draining session,
  - `duplicate()`, `quit`/`disconnect`/`destroy`, EventEmitter events,
  - cluster routes by slot (`calculateSlotForKeys`) reusing `buildClusterNodes`.

### Modified
- `src/index.ts` — export the node-redis mock symbols.
- `package.json` — `redis` as an **optional** peerDependency.

## Tests
- `tests/client-mocks/node-redis-mock.test.ts` — 19 tests (standalone 16 + cluster 3): common commands, `multi`, pub/sub, `sendCommand` fallback, slot routing.

Real Redis not needed (client-vs-mock).

## Notes / fidelity
- Facade reconstructs full on-the-wire error wording (`WRONGTYPE Operation against...`) since `decode()` drops the error code.
- Per-connection state (subscriptions, sessions) torn down in `quit()`/`disconnect()`; push-drain loop awaited so no async iteration dangles.

## Verification
tsc ✓ · lint ✓ · unit 258 ✓ · build + d.ts ✓.

## Coordinator note
Sibling PR #284 (ioredis, Stream A) is independent. Overlap only on `package.json` peerDeps (+ adjacent `src/index.ts` lines). Combined merge verified locally green (262 unit / 995 integration) — whichever lands second needs a trivial union resolve on `package.json`.

### DIFF
--- a/src/client-mocks/node-redis-mock.ts
+++ b/src/client-mocks/node-redis-mock.ts
@@ -1,0 +1,887 @@
+import { EventEmitter } from 'node:events'
+import { createRedisCommandExecutor } from '../commands'
+import { buildClusterNodes, type ClusterNodePipeline } from '../cluster'
+import { ClientSession } from '../core/client-session'
+import type { CommandExecutor } from '../core/command-executor'
+import { RedisCommandError } from '../core/redis-error'
+import { RedisResult } from '../core/redis-result'
+import type { RedisValue } from '../core/redis-value'
+import { isResponseStream, type ResponseStream } from '../core/response-stream'
+import { RedisServerState, RedisClusterTopology } from '../state'
+
+/**
+ * In-memory drop-in for the `node-redis` (`redis` package) client.
+ *
+ * node-redis exposes no socket/connector hook (its `#socketFactory` is private),
+ * so unlike the ioredis path we cannot drive the *real* client over a virtual
+ * socket. Instead this is a **hand-written facade** (the plan's "Tier 2") that
+ * replicates node-redis' public surface and routes every command through the
+ * shared in-memory command pipeline via a per-connection {@link ClientSession} —
+ * the exact same pipeline used by the networked server, MULTI/EXEC and Lua.
+ *
+ * Scope is deliberately honest: a curated set of camelCase methods carries
+ * node-redis-correct return types, and EVERYTHING ELSE falls through to the
+ * generic {@link NodeRedisMockClient.sendCommand} escape hatch, which decodes
+ * replies to native JS via the shared {@link decodeReply}. This avoids the
+ * `ioredis-mock` whack-a-mole trap of hand-shaping every command's reply.
+ */
+
+const DEFAULT_DATABASE_COUNT = 16
+
+/** A command argument node-redis accepts on the wire. */
+export type NodeRedisCommandArgument = string | Buffer
+
+/** Native JS value a reply decodes to (mirrors node-redis RESP2 defaults). */
+export type NodeRedisReply =
+  | string
+  | number
+  | bigint
+  | boolean
+  | Buffer
+  | null
+  | NodeRedisReply[]
+  | { [key: string]: NodeRedisReply }
+
+export type NodeRedisMockClusterOptions = {
+  masters: number
+  replicas?: number
+}
+
+export type CreateNodeRedisMockOptions =
+  | { cluster?: undefined; databaseCount?: number }
+  | { cluster: NodeRedisMockClusterOptions }
+
+/** A single `{ score, value }` member for ZADD-style methods. */
+export type NodeRedisZMember = { score: number; value: string }
+
+/** Listener invoked for each delivered pub/sub message: `(message, channel)`. */
+export type NodeRedisPubSubListener = (message: string, channel: string) => void
+
+/**
+ * Creates an in-memory node-redis-shaped client (standalone) or cluster client.
+ * Pass `{ cluster: { masters } }` for a {@link NodeRedisMockCluster}; otherwise
+ * a {@link NodeRedisMockClient} backed by a single in-memory pipeline.
+ */
+export async function createNodeRedisMock(
+  options: CreateNodeRedisMockOptions = {},
+): Promise<NodeRedisMockClient | NodeRedisMockCluster> {
+  if ('cluster' in options && options.cluster) {
+    return NodeRedisMockCluster.create(options.cluster)
+  }
+  const databaseCount =
+    ('databaseCount' in options ? options.databaseCount : undefined) ??
+    DEFAULT_DATABASE_COUNT
+  const state = new RedisServerState({ databaseCount })
+  const executor = createRedisCommandExecutor()
+  return new NodeRedisMockClient({ state, executor })
+}
+
+type FacadeBackend = {
+  state: RedisServerState
+  executor: CommandExecutor
+}
+
+/**
+ * Shared command-method surface implemented by both the standalone client and
+ * the cluster client. Standalone runs every command on its own session; the
+ * cluster routes by slot to the owning node's session — but the curated method
+ * bodies are identical, so they live in this base and dispatch through the
+ * abstract {@link CommandRunner.run}.
+ */
+abstract class CommandRunner {
+  /**
+   * Execute one already-tokenised command and return its decoded reply.
+   * Implementations pick the session (standalone: the only one; cluster: the
+   * slot owner for the command's keys).
+   */
+  protected abstract run(args: NodeRedisCommandArgument[]): Promise<RedisValue>
+
+  /** Generic escape hatch for any command, decoded to a native JS reply. */
+  async sendCommand(
+    args: NodeRedisCommandArgument[],
+  ): Promise<NodeRedisReply> {
+    return decodeReply(await this.run(args))
+  }
+
+  // --- strings -------------------------------------------------------------
+
+  async get(key: string): Promise<string | null> {
+    return asStringOrNull(await this.run(['GET', key]))
+  }
+
+  async set(
+    key: string,
+    value: string | number,
+    ...rest: NodeRedisCommandArgument[]
+  ): Promise<string | null> {
+    return asStringOrNull(await this.run(['SET', key, String(value), ...rest]))
+  }
+
+  async del(...keys: string[]): Promise<number> {
+    return asNumber(await this.run(['DEL', ...keys]))
+  }
+
+  async exists(...keys: string[]): Promise<number> {
+    return asNumber(await this.run(['EXISTS', ...keys]))
+  }
+
+  async incr(key: string): Promise<number> {
+    return asNumber(await this.run(['INCR', key]))
+  }
+
+  async expire(key: string, seconds: number): Promise<boolean> {
+    return asNumber(await this.run(['EXPIRE', key, String(seconds)])) === 1
+  }
+
+  async ttl(key: string): Promise<number> {
+    return asNumber(await this.run(['TTL', key]))
+  }
+
+  // --- hashes --------------------------------------------------------------
+
+  async hSet(key: string, field: string, value: string): Promise<number> {
+    return asNumber(await this.run(['HSET', key, field, value]))
+  }
+
+  async hGet(key: string, field: string): Promise<string | null> {
+    return asStringOrNull(await this.run(['HGET', key, field]))
+  }
+
+  async hGetAll(key: string): Promise<{ [field: string]: string }> {
+    const reply = decodeReply(await this.run(['HGETALL', key]))
+    // decode() turns a RESP2 map reply into a plain object already.
+    return (reply as { [field: string]: string }) ?? {}
+  }
+
+  // --- lists ---------------------------------------------------------------
+
+  async lPush(key: string, ...values: string[]): Promise<number> {
+    return asNumber(await this.run(['LPUSH', key, ...values]))
+  }
+
+  async rPush(key: string, ...values: string[]): Promise<number> {
+    return asNumber(await this.run(['RPUSH', key, ...values]))
+  }
+
+  async lRange(key: string, start: number, stop: number): Promise<string[]> {
+    return asStringArray(
+      await this.run(['LRANGE', key, String(start), String(stop)]),
+    )
+  }
+
+  // --- sets ----------------------------------------------------------------
+
+  async sAdd(key: string, ...members: string[]): Promise<number> {
+    return asNumber(await this.run(['SADD', key, ...members]))
+  }
+
+  async sMembers(key: string): Promise<string[]> {
+    return asStringArray(await this.run(['SMEMBERS', key]))
+  }
+
+  // --- sorted sets ---------------------------------------------------------
+
+  async zAdd(
+    key: string,
+    members: NodeRedisZMember | NodeRedisZMember[],
+  ): Promise<number> {
+    const list = Array.isArray(members) ? members : [members]
+    const args = ['ZADD', key]
+    for (const { score, value } of list) {
+      args.push(String(score), value)
+    }
+    return asNumber(await this.run(args))
+  }
+
+  async zRange(key: string, start: number, stop: number): Promise<string[]> {
+    return asStringArray(
+      await this.run(['ZRANGE', key, String(start), String(stop)]),
+    )
+  }
+}
+
+export type NodeRedisMockClientInit = FacadeBackend & {
+  database?: number
+}
+
+/**
+ * Standalone in-memory node-redis facade. Drives one {@link ClientSession}
+ * against a single keyspace; pub/sub uses a *dedicated* session that drains
+ * pushes off {@link ClientSession.readPushes}, mirroring node-redis' rule that a
+ * subscribed connection is reserved for pub/sub.
+ */
+export class NodeRedisMockClient extends CommandRunner {
+  private readonly emitter = new EventEmitter()
+  private readonly backend: FacadeBackend
+  private readonly database?: number
+  private readonly session: ClientSession
+  /** Dedicated session + push-reader loop, created lazily on first subscribe. */
+  private pubsub?: {
+    session: ClientSession
+    abort: AbortController
+    listeners: Map<string, Set<NodeRedisPubSubListener>>
+    patternListeners: Map<string, Set<NodeRedisPubSubListener>>
+    /** The push-draining loop; awaited on teardown so nothing dangles. */
+    drained: Promise<void>
+  }
+  private closed = false
+
+  constructor(init: NodeRedisMockClientInit) {
+    super()
+    this.backend = { state: init.state, executor: init.executor }
+    this.database = init.database
+    this.session = new ClientSession({
+      server: init.state,
+      executor: init.executor,
+      database: init.database,
+    })
+    // node-redis emits 'connect' then 'ready' once the handshake completes.
+    queueMicrotask(() => {
+      this.emitter.emit('connect')
+      this.emitter.emit('ready')
+    })
+  }
+
+  on(event: string, listener: (...args: unknown[]) => void): this {
+    this.emitter.on(event, listener)
+    return this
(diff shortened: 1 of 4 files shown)


src/client-mocks/node-redis-mock.ts
```diff
@@ -0,0 +1,887 @@
...
+    void this.teardown()
+    this.emitter.emit('end')
+  }
+
+  private async teardown(): Promise<void> {
+    if (this.closed) {
+      return
+    }
+    this.closed = true
+    this.session.close()
```
fatal10110, line 481: **[high · lifecycle] Standalone `teardown()` never calls `RedisServerState.close()` — leaks the active-expiry timer + entire server state per client.**

`createNodeRedisMock()` builds `new RedisServerState({ databaseCount })`, whose constructor arms a self-rescheduling active-expiry timer (`server-state.ts:84` `startActiveExpiry`). `teardown()` here closes only `this.session` (+ pubsub session) and never closes `this.backend.state`, so the timer re-arms forever — it stops only when `RedisServerState.close()` clears it (`server-state.ts:137`). Every mock client created in a test suite leaks one state graph + perpetual timer (the timer is `unref()`'d so it won't hang the process, but it's still a leak).

The sibling ioredis-mock does it right: `ioredis-mock.ts:189` `Promise.all([tracker.closeAll(), Promise.resolve(state.close())])`. Mirror that — `teardown()` must `this.backend.state.close()`.

src/client-mocks/node-redis-mock.ts
```diff
@@ -0,0 +1,887 @@
...
+    this.emitter.emit('end')
+  }
+
+  private teardown(): void {
+    if (this.closed) {
+      return
+    }
+    this.closed = true
+    for (const session of this.sessions.values()) {
+      session.close()
```
fatal10110, line 761: **[high · lifecycle] Cluster `teardown()` never calls `node.state.close()` for any master — same expiry-timer + state leak, multiplied per master.**

`NodeRedisMockCluster.create()` keeps only the `masters` and `replicationLinks`. `teardown()` closes cached sessions and replication links but never closes the masters' `RedisServerState` (each built by `buildClusterNodes` with active expiry on), so every master's expiry timer keeps running after `quit()`/`disconnect()`/`destroy()`.

Sibling ioredis-mock cluster teardown closes all node states: `ioredis-mock.ts:253` `node.state.close()`. Add the equivalent here (iterate masters → `node.state.close()`).

src/client-mocks/node-redis-mock.ts
```diff
@@ -0,0 +1,887 @@
...
+    this.assertOpen()
+    this.settled = true
+
+    await this.exec1(['MULTI'])
+    for (const args of this.queued) {
+      await this.exec1(args)
+    }
+    const result = await this.exec1(['EXEC'])
+
+    if (result.kind === 'null' || result.kind === 'null-array') {
```
fatal10110, line 631: **[high · client-fidelity] `multi().exec()` returns `null` on a dirty-watch abort — that's ioredis behavior. Real node-redis throws `WatchError`.** (The unit test at `tests/client-mocks/node-redis-mock.test.ts:133` encodes the wrong contract.)

Ground-truthed against `@redis/client` v6 source: `client/index.js:980-981` — `if (execResult === null) { throw new errors_1.WatchError(); }` (message `'One (or more) of the watched keys has been changed'`). node-redis NEVER returns null from `exec()` on abort; it throws. Holds across v4/v5/v6.

Real Redis RESP2 dirty-watch EXEC wire reply is `*-1` (null-array) — so the detection here is right, but the surfaced result is wrong. User code `try { ...exec() } catch (e) { if (e instanceof WatchError) retry() }` silently dead-codes against the mock. Throw a `WatchError`-shaped error instead of returning null, and flip the test.

src/client-mocks/node-redis-mock.ts
```diff
@@ -0,0 +1,887 @@
...
+  async exists(...keys: string[]): Promise<number> {
+    return asNumber(await this.run(['EXISTS', ...keys]))
+  }
+
+  async incr(key: string): Promise<number> {
+    return asNumber(await this.run(['INCR', key]))
+  }
+
+  async expire(key: string, seconds: number): Promise<boolean> {
+    return asNumber(await this.run(['EXPIRE', key, String(seconds)])) === 1
```
fatal10110: **[high · client-fidelity] `expire()` returns `boolean` (`=== 1`); node-redis returns the raw integer (0/1) as a `number`.**

`@redis/client` v6 `EXPIRE.js`: `transformReply: undefined` (raw integer pass-through, `NumberReply`) — no boolean transform, in any node-redis major. Real Redis wire reply is `:1`/`:0`.

User code `if ((await client.expire(k, 10)) === 1)` works against real node-redis but never against this mock (returns `true`); `=== true` is the inverse trap. Return the number, not the boolean.

src/client-mocks/node-redis-mock.ts
```diff
@@ -0,0 +1,887 @@
...
+
+function extractRoutingKeys(args: NodeRedisCommandArgument[]): Buffer[] {
+  // The facade only needs slot routing, and every command it routes places its
+  // key in the first argument slot. That covers the curated set + the common
+  // single-key sendCommand cases; multi-key cross-slot commands are out of the
+  // honest Tier-2 scope (documented).
+  if (args.length < 2) {
+    return []
+  }
+  return [toBuffer(args[1])]
```
fatal10110, line 898: **[medium · cluster correctness] Routing extracts only `args[1]` as the key — a cross-slot multi-key command silently runs against one node with no `CROSSSLOT`.**

`extractRoutingKeys` returns `[toBuffer(args[1])]`, so curated `cluster.del('a','b')` (or `exists`, or `MSET` via `sendCommand`) where keys hash to different slots picks the owner of `'a'` and runs the whole command there — key `'b'` on another master is neither found nor mutated, and the reply count is wrong, with **no error**. The real cluster path (`ClusterPolicy` + `calculateSlotForKeys` over all keys) throws `RedisCrossSlotError`.

"Documented out of scope" is acceptable for *raising* CROSSSLOT, but silent partial-operation returning a plausible success is worse than throwing. Either extract all keys for the curated/common commands, or throw on >1 distinct slot rather than routing by first key.

src/client-mocks/node-redis-mock.ts
```diff
@@ -0,0 +1,887 @@
...
+  switch (value.kind) {
+    case 'simple-string':
+      return value.value
+    case 'bulk-string':
+      return value.value === null ? null : value.value.toString('utf8')
+    case 'verbatim':
+      return value.value.toString('utf8')
+    case 'integer':
+      if (typeof value.value === 'bigint') {
+        return isSafeBigInt(value.value) ? Number(value.value) : value.value
```
fatal10110: **[medium · client-fidelity] `decodeReply` returns a `bigint` for `:` integers beyond `MAX_SAFE_INTEGER`; node-redis RESP2 always returns a JS `number` for the NUMBER (`:`) type.**

Ground-truth: `@redis/client` v6 `decoder.js` `#decodeUnsingedNumber` uses plain JS number arithmetic (`number * 10 + ...`) → always a `number`. Only the RESP3 `(` BIG_NUMBER type yields a bigint. The facade negotiates RESP2 (never sends `HELLO 3`), so a `:` reply must decode to `number`, never `bigint`.

Reachable via the `sendCommand` escape hatch (the documented primary path for uncurated commands): e.g. `INCR` past `2^53-1` returns `bigint` from the mock, `number` from real node-redis — breaking `typeof === 'number'` / `JSON.stringify` / arithmetic differently. (Masked for curated `incr()`/`del()` because `asNumber()` coerces.)

src/client-mocks/node-redis-mock.ts
```diff
@@ -0,0 +1,887 @@
...
+    const result = await this.exec1(['EXEC'])
+
+    if (result.kind === 'null' || result.kind === 'null-array') {
+      return null
+    }
+    if (result.kind !== 'array' && result.kind !== 'set') {
+      // Defensive: any non-array EXEC reply (shouldn't happen) → decode as-is.
+      return [decodeReply(result)]
+    }
+    return result.items.map(item => decodeReply(item))
```
fatal10110: **[medium · client-fidelity] An erroring command inside MULTI/EXEC throws on the first error item and discards all sibling replies; node-redis throws one `MultiErrorReply` carrying every reply + `errorIndexes`.**

`@redis/client` v6 `multi-command.js:34-43`: collects `errorIndexes`, builds `replies` (successes + errors), then `throw new MultiErrorReply(replies, errorIndexes)` — errors are aggregated, the full reply array is preserved on the thrown error.

Here `result.items.map(item => decodeReply(item))` throws a bare `RedisCommandError` on the first error item, so the user loses the other replies and the `MultiErrorReply` shape (`.replies` / `.errorIndexes` / `.errors()`). Build the aggregate instead of throwing per-item.

src/client-mocks/node-redis-mock.ts
```diff
@@ -0,0 +1,887 @@
...
+  }
+
+  async unwatch(): Promise<string> {
+    return asString(await this.run(['UNWATCH']))
+  }
+
+  /** Begin a MULTI transaction. Commands are queued, then replayed on exec(). */
+  multi(): NodeRedisMockMulti {
+    return new NodeRedisMockMulti(args =>
+      runOnSession(this.session, args, this.closed),
```
fatal10110: **[medium · concurrency] `multi().exec()` and curated commands share one `ClientSession`; concurrent (un-awaited) calls can interleave a normal command into the transaction.**

`exec()` issues `MULTI`, each queued cmd, then `EXEC` as separate `session.execute()` calls — the `SerialTurnQueue` serializes one `execute()`, not the span. Between them the session is in `mode === 'transaction'`. A `client.get('x')` fired concurrently (legal in JS, no await) lands while in transaction mode and gets `+QUEUED`'d into the wrong transaction (`TransactionPolicy.beforeExecute` queues any non-transaction-flagged command), corrupting both.

The networked server never hits this (a TCP connection delivers commands strictly serially); the facade exposes the shared session to arbitrary concurrent JS calls. Medium because it only bites on misuse — but consider a per-client lock around `exec()`, or a dedicated session for the transaction.

src/client-mocks/node-redis-mock.ts
```diff
@@ -0,0 +1,887 @@
...
+        decodeReply(val),
+      ])
+    case 'null':
+    case 'null-array':
+      return null
+    case 'error':
+      // Reconstruct the on-the-wire error text (`CODE message`) so the surfaced
+      // message matches what node-redis parses from `-WRONGTYPE …`, `-ERR …`,
+      // etc. The bare `value.message` would drop the code prefix.
+      throw new RedisCommandError(
```
fatal10110: **[low · client-fidelity] Errors surface as `RedisCommandError`, not node-redis's `ErrorReply`/`SimpleError` — message text matches, `instanceof` does not.**

The message reconstruction (`${code} ${message}`) is correct — real wording `WRONGTYPE Operation against a key holding the wrong kind of value` is preserved, and `err.message` matching (the test's `/WRONGTYPE/`) works. But node-redis throws `SimpleError extends ErrorReply` (`decoder.js:456-460`), so user code doing `catch (e) { if (e instanceof ErrorReply) }` (the node-redis idiom) never matches against the mock. Low-severity type-identity gap.

src/client-mocks/node-redis-mock.ts
```diff
@@ -0,0 +1,887 @@
...
+
+// --- reply decoding --------------------------------------------------------
+//
+// node-redis leaves most RESP2 replies untransformed, and the in-memory decode()
+// (RedisValue → native JS) already matches those defaults: bulk-string → utf8
+// string, integer → number, map → object, array/set → array. So the curated
+// methods are thin coercions over this shared decoder rather than per-command
+// reply tables — uncommon commands get the same native shapes via sendCommand.
+
+function decodeReply(value: RedisValue): NodeRedisReply {
```
fatal10110, line 936: **[low · reuse] `decodeReply` / `decodeKey` / `isSafeBigInt` are near-verbatim copies of `src/in-memory-client.ts` (`decode` / `decodeKey` / `isSafeBigInt`).**

Two independent `RedisValue → native` decoders will drift: a future RESP-decoding fix (verbatim handling, bigint policy, error-code prefixing) has to be applied in both. The only intentional delta worth keeping is this file's error branch (it prefixes the error code `CODE message`). Two call sites now justify extracting a shared `decodeRedisValue(value, { errorCodePrefix? })` helper both clients call. (Also relevant to the bigint finding above — fix it once.)

src/client-mocks/node-redis-mock.ts
```diff
@@ -0,0 +1,887 @@
...
+    await this.teardown()
+    this.emitter.emit('end')
+  }
+
+  /**
+   * node-redis exposes a synchronous destroy(); abort everything immediately.
+   * The push-drain loop settles on the next tick via the abort signal.
+   */
+  destroy(): void {
+    void this.teardown()
```
fatal10110: **[low · lifecycle] `destroy()` does `void this.teardown()` (standalone), discarding a promise that can reject; and cluster `quit()`/`disconnect()` are `async` but call a synchronous `teardown()`.**

Standalone `teardown()` is genuinely async (awaits `pubsub.drained`), so `void this.teardown()` in `destroy()` is a latent unhandled-rejection vector (safe today only because `drainPushes` swallows its own errors and `session.close()` is sync). Cluster `teardown()` is sync, yet `quit()`/`disconnect()` advertise `Promise<...>` while doing no async work — asymmetric with the standalone path. Worth a `void`-with-`.catch` on destroy and a comment on the cluster asymmetry.

src/client-mocks/node-redis-mock.ts
fatal10110: Fixed in fd39896. Standalone `teardown()` now calls `this.backend.state.close()` to clear the active-expiry timer. Guarded by a new `ownsState` flag (true only for the createNodeRedisMock-created client) so a `duplicate()` — which shares the same state — never closes it out from under its siblings.

src/client-mocks/node-redis-mock.ts
fatal10110: Fixed in fd39896. Cluster `teardown()` now iterates `this.masters` and calls `node.state.close()`. (Replicas are built with expiry disabled, so only masters arm a timer.)

src/client-mocks/node-redis-mock.ts
fatal10110: Fixed in fd39896. `exec()` now throws `WatchError` on a null/null-array EXEC instead of returning null. Using the real `redis` `WatchError` (resolved via lazy `import('redis')`, fallback to a shape-compatible local class), so `instanceof WatchError` matches. Test flipped to `assert.rejects(..., err => err instanceof WatchError)`.

src/client-mocks/node-redis-mock.ts
fatal10110: Fixed in fd39896. `expire()` returns the raw integer (`asNumber`, 0/1) — no boolean coercion. Test updated to assert `=== 1`.

src/client-mocks/node-redis-mock.ts
fatal10110: Fixed in fd39896. `extractRoutingKeys` now returns all keys for the common multi-key commands (DEL/EXISTS/UNLINK/TOUCH/MGET/WATCH/SINTER/SUNION/SDIFF/PFCOUNT); `calculateSlotForKeys` (generateMulti) returns -1 across slots, and the router now throws CROSSSLOT (surfaced as a node-redis ErrorReply carrying `CROSSSLOT Keys in request don't hash to the same slot`) rather than silently running against the first key. Uncommon multi-key commands via the generic sendCommand still route by first key (documented). Added a cross-slot test.

src/client-mocks/node-redis-mock.ts
fatal10110: Fixed in fd39896. `decodeReply` integer case now always returns a JS `number` for RESP2 `:` (`Number(value)` even past 2^53, matching node-redis' plain-number decoder); only the RESP3 `(` big-number type yields bigint. Removed the now-unused `isSafeBigInt`.

src/client-mocks/node-redis-mock.ts
fatal10110: Fixed in fd39896. `exec()` no longer throws on the first error item — it decodes every item, collecting an `ErrorReply` at each error position plus `errorIndexes`, then throws a single `MultiErrorReply(replies, errorIndexes)` (real redis class, so `.replies`/`.errorIndexes`/`.errors()` and instanceof all hold). Added an aggregation test.

src/client-mocks/node-redis-mock.ts
fatal10110: Fixed in fd39896. Added a per-client command lock (`runExclusive`): `run()` and the whole MULTI→queued→EXEC span both go through it, so a concurrent un-awaited call can't interleave into transaction mode — mirroring a real single connection's serial delivery. A dedicated transaction session was rejected because WATCH (issued earlier on the main session) must apply to the same session's MULTI/EXEC.

src/client-mocks/node-redis-mock.ts
fatal10110: Fixed in fd39896 (folded into the WatchError/aggregate work). Command errors now throw node-redis' real `ErrorReply` (resolved via lazy `import('redis')`) with the reconstructed `CODE message`, so `instanceof ErrorReply` matches the node-redis idiom; falls back to RedisCommandError only if the redis package is absent.

src/client-mocks/node-redis-mock.ts
fatal10110: Deferred to a follow-up PR, with reasoning: the bigint fix above intentionally diverges this decoder's integer policy (always number) from `in-memory-client`'s (keeps bigint for unsafe), and the error branches differ (code-prefixed ErrorReply vs bare RedisCommandError). A shared `decodeRedisValue(value, opts)` needs two option flags plus an in-memory-client regression pass; safer as its own focused PR than bundled into these correctness fixes.

src/client-mocks/node-redis-mock.ts
fatal10110: Fixed in fd39896. Standalone `destroy()` is now `void this.teardown().catch(err => this.emitter.emit('error', err))` (no dropped rejection). Cluster `teardown()` is genuinely synchronous; added a comment on `quit()`/`disconnect()` noting the async-signature/sync-work asymmetry vs the standalone path (which must await its pub/sub drain).

PR #286, opened 2026-06-23
feat(transports): socketless connection layer (case #3 — in-process MOVED following)

fatal10110:
## What

Extracts the **socketless connection layer** from the ioredis-mock work (PR #284) so it stands on its own, independent of any client mock. This is the infrastructure a browser/in-memory **cluster** target needs to follow `MOVED` redirects without a TCP socket (design "case #3").

### Pieces
- **`attachSession(transport, opts)`** — drives a fresh `ClientSession` over *any* `ConnectionTransport` + runs the RESP adapter loop, extracted out of `Resp2Server.handleConnection`. Net-free; `Resp2Server` now delegates to it. This is the run-loop that previously only existed inside the TCP server.
- **`VirtualClientSocket` / `createVirtualConnection()`** — a `net.Socket`-shaped `Duplex` bridged to a server-side session via `attachSession`. Lets a client lib "open a connection" with no real socket.
- **`InMemoryNodeRegistry`** — synthetic `host:port → pipeline` resolver: the lookup a client does when redirected by `-MOVED slot host:port` to reach the target node's state/executor in-process.

All exposed via `src/internal.ts`.

## Why (browser refactor)
Per [docs/BROWSER-REFACTOR-PLAN.md](docs/BROWSER-REFACTOR-PLAN.md), the browser drives the mock by feeding RESP bytes through a socketless transport — but the run-loop that consumes a transport was trapped inside the `net`-coupled `Resp2Server`. `attachSession` frees it (net-free, stream-free). The registry + virtual connection are the "fake connection" needed if a browser cluster client follows real `MOVED` redirects.

> `virtual-connection` imports `node:stream` `Duplex` — polyfillable in a browser bundle (same approach as the planned `buffer` polyfill). `attach-session` and `node-registry` are net- and stream-free.

## Scope / non-goals
- **No client mock**, no `ioredis`/`redis` dependency, no `package.json` change.
- Does **not** make `cluster.ts` browser-safe (the `Resp2Server`→`net` module split, plan blocker #5) — separate follow-up.
- Sibling PR (case #2, client-side slot routing) is independent; it sidesteps `MOVED` entirely and needs none of this.

## Tests
- `tests/core/virtual-connection.test.ts` — wire round-trip + teardown (no leaked sessions/sockets). 4 tests.
- `tests/client-mocks/node-registry.test.ts` — register/resolve/list/overwrite. 4 tests.

## Verification
tsc ✓ · lint ✓ · new tests 8/8 ✓ · build + d.ts ✓ (`InMemoryNodeRegistry`/`attachSession`/`createVirtualConnection` in `dist/`).

### DIFF
--- a/src/core/transports/virtual-connection.ts
+++ b/src/core/transports/virtual-connection.ts
@@ -3,6 +3,7 @@ import type { CommandExecutor } from '../command-executor'
 import type { RespEncodeOptions } from '../resp-encoder'
 import type { Logger } from '../../logger'
 import type { RedisClusterNodeRole, RedisServerState } from '../../state'
+import { formatSocketAddressParts } from '../network-address'
 import { attachSession } from './attach-session'
 import {
   type ConnectionTransport,
@@ -98,8 +99,12 @@ export class VirtualClientSocket extends Duplex {
     encoding: BufferEncoding,
     callback: (error?: Error | null) => void,
   ): void {
-    const buffer = Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk, encoding)
-    this.onClientWrite?.(Buffer.from(buffer))
+    // Defensive copy of caller-owned Buffer input; the string branch already
+    // allocates a fresh Buffer, so it needs no second copy.
+    const owned = Buffer.isBuffer(chunk)
+      ? Buffer.from(chunk)
+      : Buffer.from(chunk, encoding)
+    this.onClientWrite?.(owned)
     callback()
   }
 
@@ -166,6 +171,9 @@ class DuplexConnectionTransport implements ConnectionTransport {
 
   /** Feed a chunk the client wrote into the server-side read loop. */
   feed(chunk: Buffer): void {
+    // Deliberate divergence from InMemoryConnectionTransport.feed (a strict test
+    // harness that throws on a closed transport): a real socket silently drops
+    // writes after close, and this transport models a socket — so drop, not throw.
     if (this.inputEnded || this.closed) {
       return
     }
@@ -301,9 +309,10 @@ export function createVirtualConnection(
     nodeRole: opts.nodeRole,
     logger: opts.logger,
     encoder: opts.encoder,
-    clientAddress: `${opts.remoteAddress ?? '127.0.0.1'}:${
-      opts.remotePort ?? 6379
-    }`,
+    clientAddress: formatSocketAddressParts(
+      opts.remoteAddress ?? '127.0.0.1',
+      opts.remotePort ?? 6379,
+    ),
   })
 
   return {
--- a/src/core/transports/virtual-connection.ts
+++ b/src/core/transports/virtual-connection.ts
@@ -1,0 +1,314 @@
+import { Duplex } from 'node:stream'
+import type { CommandExecutor } from '../command-executor'
+import type { RespEncodeOptions } from '../resp-encoder'
+import type { Logger } from '../../logger'
+import type { RedisClusterNodeRole, RedisServerState } from '../../state'
+import { attachSession } from './attach-session'
+import {
+  type ConnectionTransport,
+  type ConnectionTransportEvent,
+  type ConnectionTransportListener,
+  type ConnectionTransportUnsubscribe,
+} from './connection-transport'
+
+export type CreateVirtualConnectionOptions = {
+  state: RedisServerState
+  executor: CommandExecutor
+  nodeRole?: RedisClusterNodeRole
+  logger?: Pick<Logger, 'error'>
+  encoder?: RespEncodeOptions
+  /** Synthetic remote address reported to the client lib. Default 127.0.0.1. */
+  remoteAddress?: string
+  /** Synthetic remote port reported to the client lib. Default 6379. */
+  remotePort?: number
+}
+
+export type VirtualConnection = {
+  /** The fake `net.Socket`-shaped stream handed to the client library. */
+  clientSocket: VirtualClientSocket
+  /** Resolves once the server-side adapter loop ends and the session is closed. */
+  done: Promise<void>
+  /** Tear down both ends of the wire and the server-side session. */
+  close(): void
+}
+
+/**
+ * A `net.Socket`-compatible {@link Duplex} handed to a client library (ioredis)
+ * as its connection stream. Only the surface the client touches is implemented:
+ * duplex read/write, the no-op socket tuning methods, the `remoteAddress` /
+ * `remotePort` getters, and the `'connect'` event (emitted on `nextTick`, like
+ * ioredis' `StandaloneConnector`). Bytes written by the client are forwarded to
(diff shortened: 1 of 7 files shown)


src/core/transports/virtual-connection.ts
```diff
@@ -0,0 +1,314 @@
...
+    onClientDestroy: () => transport.endInput(),
+  })
+
+  const attached = attachSession(transport, {
+    state: opts.state,
+    executor: opts.executor,
+    nodeRole: opts.nodeRole,
+    logger: opts.logger,
+    encoder: opts.encoder,
+    clientAddress: `${opts.remoteAddress ?? '127.0.0.1'}:${
```
fatal10110: **[redis-wire] medium** — `clientAddress` is built by raw string concat here, bypassing `formatSocketAddressParts()` that the socket path uses (`resp2/server.ts`). For a non-IPv4 synthetic host this diverges from real Redis: an IPv6 literal is not bracketed and an `::ffff:`-mapped address is not normalized.

Ground-truthed on redis-server 7.2.1:
- IPv6 client: real `addr=[::1]:56110`; this path emits `addr=::1:56110`
- mapped: real `addr=127.0.0.1:5000`; this path emits `addr=::ffff:127.0.0.1:5000`

Surfaces in `CLIENT INFO` / `CLIENT LIST` `addr=` and the MONITOR source field. Narrow today (only caller is the unit test, which hits the IPv4 default), but fix now so the two transports cannot drift: reuse `formatSocketAddressParts(opts.remoteAddress, opts.remotePort)` for the `clientAddress` passed to `attachSession`.

src/core/transports/virtual-connection.ts
```diff
@@ -0,0 +1,314 @@
...
+    private readonly clientSocket: VirtualClientSocket,
+    id?: string,
+  ) {
+    this.id = id ?? `virtual-${++DuplexConnectionTransport.nextId}`
+    this.signal = this.controller.signal
+  }
+
+  /** Feed a chunk the client wrote into the server-side read loop. */
+  feed(chunk: Buffer): void {
+    if (this.inputEnded || this.closed) {
```
fatal10110, line 177: **[architecture] low** — `feed` silently drops chunks after `endInput`/`close`, but the sibling `InMemoryConnectionTransport.feed` throws (`Cannot feed a closed transport`) on the same condition. The drop is correct here (matches real-socket behavior), yet the two transports now disagree on the same method contract. A one-line comment noting the deliberate difference (socket-drop vs test-harness-strict) would help future readers.

src/core/transports/virtual-connection.ts
```diff
@@ -0,0 +1,314 @@
...
+    // Backpressure is not modeled — the server pushes proactively via feedClient.
+  }
+
+  override _write(
+    chunk: Buffer | string,
+    encoding: BufferEncoding,
+    callback: (error?: Error | null) => void,
+  ): void {
+    const buffer = Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk, encoding)
+    this.onClientWrite?.(Buffer.from(buffer))
```
fatal10110: **[architecture] low (informational)** — double Buffer copy: when `chunk` is a string, `Buffer.from(chunk, encoding)` already allocates a fresh buffer, then `Buffer.from(buffer)` copies it again. Harmless (CLAUDE.md deprioritizes perf); only the Buffer-input branch needs the defensive copy.

tests/core/virtual-connection.test.ts
```diff
@@ -0,0 +1,110 @@
...
+  test('destroying the client socket tears down the server session', async () => {
+    const { state, executor } = freshPipeline()
+    const { clientSocket } = createVirtualConnection({ state, executor })
+    await once(clientSocket, 'connect')
+    assert.strictEqual(state.getConnectedClients().length, 1)
+
+    clientSocket.destroy()
+    await once(clientSocket, 'close')
+    // Give the server-side read loop a tick to observe the closed transport.
+    await new Promise(resolve => setImmediate(resolve))
```
fatal10110: **[architecture] low** — this destroy-teardown test leans on a single `setImmediate` tick instead of awaiting a settle handle (the sibling `close()` test awaits `done`). `createVirtualConnection` only destructures `clientSocket` here, so there is no `done` to await. Passes today because `session.close()` runs synchronously in the adapter finally; if the disposal chain ever grows an extra await it could go flaky. Prefer polling until `getConnectedClients().length === 0`, or expose and await `done`.

tests/core/virtual-connection.test.ts
```diff
@@ -0,0 +1,110 @@
+import { describe, test } from 'node:test'
```
fatal10110, line 1: **[architecture] low** — lifecycle coverage is happy-path only (connect / round-trip / idle close / idle destroy). The PR body claims no leaked sessions/sockets, but no test tears down while a ResponseStream / pubsub SUBSCRIBE / WATCH is active over the virtual wire — exactly the state the adapter finally is responsible for draining. A test that SUBSCRIBEs then destroys, asserting the push writer settles, would close the gap.

src/core/transports/virtual-connection.ts
fatal10110: Fixed in 57b9515. `clientAddress` now goes through `formatSocketAddressParts(opts.remoteAddress, opts.remotePort)` — same helper as the socket path — so IPv6 literals are bracketed and `::ffff:`-mapped addresses normalized. The `clientSocket.remoteAddress` getter still returns the raw address, matching a real `net.Socket`.

src/core/transports/virtual-connection.ts
fatal10110: Fixed in 57b9515. Added a comment on `feed` calling out the deliberate socket-style drop-on-closed and why it differs from the strict `InMemoryConnectionTransport.feed` (test harness throws).

src/core/transports/virtual-connection.ts
fatal10110: Fixed in 57b9515. `_write` now only defensively copies Buffer input; the string branch keeps the single `Buffer.from(chunk, encoding)` allocation. Comment added on the intent.

tests/core/virtual-connection.test.ts
fatal10110: Fixed in 57b9515. The destroy-teardown test now destructures and awaits `done` instead of a single `setImmediate` tick, so it stays correct if the disposal chain grows an await.

tests/core/virtual-connection.test.ts
fatal10110: Fixed in 57b9515. Added `tearing down with an active SUBSCRIBE settles cleanly`: SUBSCRIBEs over the virtual wire, reads the confirmation, then `close()` + `await done` — if the adapter finally failed to drain the active push stream, `done` would never settle and the test would time out. Asserts `getConnectedClients().length === 0` after.

PR #287, opened 2026-06-23
feat(client-mocks): in-memory cluster router (case #2 — client-side slot routing, no MOVED)

fatal10110:
## What

Extracts a **client-agnostic in-memory cluster router** from the node-redis cluster facade (PR #285) so it stands alone. This is the socket-free cluster path for the browser target (design "case #2").

- **`InMemoryCluster` / `createInMemoryCluster({ masters, replicasPerMaster? })`** — builds TCP-free node pipelines via `buildClusterNodes()` and routes each command to the slot owner's `ClientSession` **up front** (slot computed client-side). The node therefore never returns a `MOVED` that a caller would have to follow — the redirect is sidestepped entirely.
- `route(args) → ClientSession`, `execute(args) → Promise<RedisValue>` (raw value; callers decode), `close()` (tears down sessions + replication links). Exposed via `src/internal.ts`.

## Why (browser refactor)
Per [docs/BROWSER-REFACTOR-PLAN.md](docs/BROWSER-REFACTOR-PLAN.md), this is the recommended browser cluster approach: no fake socket, no `MOVED` round-trip, no `node:stream`. (The redirect-following alternative — fake connections — is the separate sibling PR #286.)

## Scope / non-goals
- **Client-agnostic**: in `(string|Buffer)[]` → out `RedisValue`. No `ioredis`/`redis` dep, no `package.json` change.
- **Routing heuristic**: single-key (first-arg). Covers single-key commands and the common case; genuine multi-key cross-slot commands misroute and surface the node's normal `-CROSSSLOT`/`-MOVED` error unchanged (documented in the class).
- **Not yet browser-safe**: it imports `buildClusterNodes` from `cluster.ts`, whose module graph pulls in `Resp2Server` (→`net`). Making it browser-importable needs the `cluster.ts` `net` split (plan blocker #5) — separate follow-up.
- No consumer wired yet — this is the staged primitive a browser cluster client will build on.

## Tests
- `tests/client-mocks/in-memory-cluster.test.ts` (5): round-trip through the slot owner, keys spread across >1 master, same-slot session caching, keyless command on first master, `execute` rejects after `close`.

## Verification
tsc ✓ · lint ✓ · new tests 5/5 ✓ · build + d.ts ✓.

### DIFF
--- a/src/client-mocks/in-memory-cluster.ts
+++ b/src/client-mocks/in-memory-cluster.ts
@@ -1,0 +1,153 @@
+import { buildClusterNodes, type ClusterNodePipeline } from '../cluster'
+import { ClientSession } from '../core/client-session'
+import type { RedisValue } from '../core/redis-value'
+import { isResponseStream, type ResponseStream } from '../core/response-stream'
+import { RedisClusterTopology } from '../state'
+
+export type InMemoryClusterOptions = {
+  masters: number
+  replicasPerMaster?: number
+}
+
+export type CommandArgument = string | Buffer
+
+/**
+ * Client-agnostic in-memory Redis cluster.
+ *
+ * Builds TCP-free node pipelines (via {@link buildClusterNodes}) and routes each
+ * command to the slot owner's {@link ClientSession} *up front* — the slot is
+ * computed client-side from the command's keys, so the node never has to answer
+ * with a `MOVED` redirect that a caller would then have to follow. That makes
+ * this the socket-free cluster path (the redirect-following alternative needs a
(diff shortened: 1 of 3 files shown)


fatal10110:
Closing: superseded by the virtual-network approach (#286). `InMemoryCluster` is a client-side slot router that pre-computes slots and never emits/follows MOVED — it duplicates ClusterPolicy's slot logic with a lossier single-key heuristic and is redundant once the cluster emits MOVED and InMemoryNodeRegistry translates the address. If a non-redirecting client ever needs cluster, it should follow MOVED over the registry rather than re-route client-side.

PR #316, opened 2026-06-27
Add Redis 7 function and admin command surface

### LINKED ISSUE #296: compat: implement Redis 7.0 command and behavior deltas
## Scope
Audit and implement the Redis 7.0 compatibility delta for the `redis-7.0` profile. Only add missing behavior; keep already-compatible commands unchanged.

## Commands / surface
- Root commands: `EXPIRETIME`, `PEXPIRETIME`, `LMPOP`, `BLMPOP`, `SINTERCARD`, `SSUBSCRIBE`, `SUNSUBSCRIBE`, `SPUBLISH`, `SORT_RO`, `ZINTERCARD`, `ZMPOP`, `BZMPOP`, `EVAL_RO`, `EVALSHA_RO`, `FUNCTION`, `FCALL`, `FCALL_RO`.
- Options/subcommands: `EXPIRE NX|XX|GT|LT`, `SET NX GET`, `BITPOS`/`BITCOUNT` bit indexes, `PUBSUB SHARDCHANNELS|SHARDNUMSUB`, `COMMAND DOCS|LIST|GETKEYSANDFLAGS`, `COMMAND INFO` arity variants, `SHUTDOWN NOW|FORCE|ABORT`, `CLUSTER SHARDS|LINKS|ADDSLOTSRANGE|DELSLOTSRANGE`, `CLIENT NO-EVICT`, `ACL DRYRUN`, `SLOWLOG GET -1`, stream metadata options, multi-section `INFO`.

## Behavior checks
- Profile gating exposes 7.0 commands at `redis-7.0+` and hides them in `redis-6.2`.
- Cover externally visible reply/error shifts: Lua `print` removal, script/function cluster behavior, ACL pubsub defaults, `X[AUTO]CLAIM` deleted-id handling, `LPOP`/`RPOP count` null-array shape, protected admin/debug/module commands, transaction rejection for unsafe admin commands.

## Acceptance criteria
- Unit coverage for registry/profile gates.
- Integration coverage for representative command availability and reply behavior.
- `npm test`, compatibility integration mock suite, build, and lint pass.

## References
- https://raw.githubusercontent.com/redis/redis/7.0.0/00-RELEASENOTES

## Profile-gating subtasks

- [x] #304
- [x] #305
- [x] #306
- [x] #307
- [x] #308

Note: #303 is a functionality slice discovered during this audit and intentionally references, but does not close, this parent issue.

### LINKED ISSUE #304: compat(7.0): gate root command availability by profile
Parent: #296

## Scope
Profile-gate the Redis 7.0 root command surface. This task is about command availability per compatibility profile, not implementing missing command behavior.

Commands to audit:
`EXPIRETIME`, `PEXPIRETIME`, `LMPOP`, `BLMPOP`, `SINTERCARD`, `SSUBSCRIBE`, `SUNSUBSCRIBE`, `SPUBLISH`, `SORT_RO`, `ZINTERCARD`, `ZMPOP`, `BZMPOP`, `EVAL_RO`, `EVALSHA_RO`, `FUNCTION`, `FCALL`, `FCALL_RO`.

## Acceptance
- `redis-6.2` hides every Redis 7.0 root command listed above.
- `redis-7.0`, `redis-7.2`, `redis-7.4`, `redis-8.0`, and supported Valkey profiles expose the applicable commands.
- Add/adjust registry unit coverage and representative mock integration coverage.

fatal10110:
## Summary
- add Redis 7 scripting/function roots: EVAL_RO, EVALSHA_RO, FUNCTION, FCALL, and FCALL_RO
- add a mock-local function registry backed by the existing Lua runtime
- add newly implemented admin roots: ACL WHOAMI/DRYRUN, empty SLOWLOG, and SHUTDOWN option handling
- merge the latest Redis 7 existing-behavior gates from main into the same validation matrix

## Scope notes
This completes the remaining root command availability slice from #304. The other Redis 7 profile-gating subtasks from #296 are already closed, including the existing-command behavior work merged in #315.

## Validation
- node --enable-source-maps --import tsx --no-warnings --test ./tests/compatibility/registry.test.ts ./tests/compatibility/behavior.test.ts ./tests/commands-scripts.test.ts
- REDIS_COMPAT loop for redis-6.2, redis-7.0, redis-7.2, redis-7.4, redis-8.0, valkey-8.0, valkey-9.0 with TEST_BACKEND=mock ./tests-integration/compatibility/profile-gates.test.ts
- npm run build
- npm test
- npm run test:integration:real
- npm run lint
- git push pre-push hook reran npm test

Fixes #304
Fixes #296


src/commands/connection.ts
```diff
@@ -776,6 +783,217 @@ export const lastsaveCommand = defineCommand({
...
+  }),
+  flags: ['admin', 'noscript'],
+  keys: () => [],
+  execute: (args, ctx) => {
+    const options = parseShutdownOptions(args.args, ctx)
+    if (options.abort) {
+      throw new RedisCommandError('No shutdown in progress.')
+    }
+
+    return RedisResult.create(RedisValue.simpleString('OK'), { close: true })
```
fatal10110: HIGH — wire-compat. SHUTDOWN on success sends `+OK\r\n` then closes; real Redis sends **no reply at all**, it just closes the socket.

Ground-truthed against real redis-server 7.2.1 (raw capture): `( printf 'PING\r\n'; sleep .5; printf 'SHUTDOWN NOSAVE\r\n'; sleep 1 ) | nc 127.0.0.1 6399 | xxd` returned only `+PONG\r\n` (7 bytes) — SHUTDOWN itself produced **zero** reply bytes, then the server closed.

A client/test that asserts "SHUTDOWN returns no reply / connection drops" sees a spurious `+OK` here. Fix: return a result that closes without emitting a reply.

src/commands/scripts.ts
```diff
@@ -266,3 +418,254 @@ async function runLuaScript(
...
+  ctx: RedisExecutionContext,
+  readOnly: boolean,
+): Promise<RedisResult> {
+  const fn = ctx.server.functionRegistry.findFunction(args.functionName)
+  if (!fn) {
+    throw new RedisCommandError('Function not found')
+  }
+
+  const { keys, argv } = splitFcallArgs(args)
+  return runLuaScript(fn.script, keys, argv, ctx, readOnly)
```
fatal10110, line 671: HIGH — wire-compat. FCALL_RO runs **any** function read-only and only fails at the first write call. Real Redis rejects a non-`no-writes` function **before execution**, keyed off the function's declared flags.

Real 7.2.1: `FCALL_RO wfunc 1 k1` (wfunc does SET) → `ERR Can not execute a script with write flag using *_ro command.` — returned pre-exec.

Root cause: `RedisFunctionRegistry`/`parseFunctionLibrary` never records per-function flags, so FCALL_RO cannot distinguish a `no-writes` function from a writing one. Needs flag tracking + a pre-exec check.

src/commands/scripts.ts
```diff
@@ -266,3 +418,254 @@ async function runLuaScript(
...
+}
+
+function functionReply(fn: { name: string }): RedisValue {
+  return RedisValue.array([
+    RedisValue.bulkString(Buffer.from('name')),
+    RedisValue.bulkString(Buffer.from(fn.name)),
+    RedisValue.bulkString(Buffer.from('description')),
+    RedisValue.bulkString(null),
+    RedisValue.bulkString(Buffer.from('flags')),
+    RedisValue.array([]),
```
fatal10110: HIGH — wire-compat. FUNCTION LIST always returns an empty `flags` array. Real Redis returns each function's actual flags.

Real 7.2.1, lib with `flags={'no-writes'}`: per-function reply has `flags 1) "no-writes"`. Here `functionReply()` hardcodes `flags []` for every function.

Same root cause as the FCALL_RO finding: the parser only handles the positional `register_function('name', function(keys,args)...end)` form and never extracts the `flags=` table form used to declare them.

src/commands/scripts.ts
```diff
@@ -266,3 +418,254 @@ async function runLuaScript(
...
+): RedisResult {
+  expectRestLength(args, 'function|stats', 0)
+  const libraries = ctx.server.functionRegistry.list()
+  const functionCount = libraries.reduce(
+    (count, library) => count + library.functions.length,
+    0,
+  )
+
+  return array([
+    RedisValue.bulkString(Buffer.from('running_script')),
```
fatal10110: HIGH — wire-compat. FUNCTION STATS builds a flat RESP2 array; real Redis encodes it as **RESP3 maps** (top-level map, `engines` map, per-engine map).

Real 7.2.1 `redis-cli -3 FUNCTION STATS`: `running_script => (nil)`, `engines => { LUA => { libraries_count => N, functions_count => M } }` — all `=>` maps. A RESP3 client (node-redis default) gets a flat array here and key-access fails.

Codebase already has `RedisValue.map(...)` (CONFIG GET, COMMAND, XINFO) — use it.

src/state/function-registry.ts
```diff
@@ -0,0 +1,134 @@
...
+  if (functions.length === 0) {
+    throw new Error('No functions registered')
+  }
+
+  return { name, code: Buffer.from(code), functions }
+}
+
+function* registeredFunctions(text: string): Iterable<string> {
+  const regex =
+    /redis\.register_function\s*\(\s*(["'])([^"']+)\1\s*,\s*function\s*\(\s*keys\s*,\s*args\s*\)\s*[\s\S]*?\s*end\s*\)/g
```
fatal10110: HIGH — correctness. The regex Lua parser truncates any function body containing an inner closure that ends in `end)`. The non-greedy `[\s\S]*?` stops at the **first** `end)`, which is the inner closure's, not the function's.

Reproduced: `register_function('h', function(keys, args)\n local ok = pcall(function() return 1 end)\n return ok\nend)` → extracted body `function(keys, args)\nlocal ok = pcall(function() return 1\nend` — cut mid-`pcall`, trailing `return ok end)` dangles → `buildFunctionScript` emits unbalanced Lua (fails at FCALL or mis-parses the next function).

Related limits, same parser: (1) only the positional `function(keys, args)` form with literal param names `keys, args` is matched — the documented table form `register_function{function_name=..., callback=..., flags=...}` and any other param names yield zero functions → spurious `No functions registered`; (2) `registeredFunctions()` and `buildFunctionScript()` keep two separately-compiled near-identical regexes (differ only by the body capture group) — drift risk, extract one shared constant. A real Lua tokenizer would avoid all three.

src/core/lua-runtime.ts
```diff
@@ -85,6 +88,14 @@ export class RedisLuaRuntime {
       return redisErrorToLuaReply(new ScriptNotAllowedCommandError())
     }
 
+    if (this.hostState.readOnly && plan.flags.includes('write')) {
+      return redisErrorToLuaReply(
+        new RedisCommandError(
+          'Write commands are not allowed from read-only scripts.',
```
fatal10110, line 100: MEDIUM — wire-compat. Read-only-script write error is missing the real trailing `script: <sha>, on @user_script:N.` suffix.

Real 7.2.1: `ERR Write commands are not allowed from read-only scripts. script: 2f78f69db1941d9e7ec570fa8bc819381f1a39e2, on @user_script:1.` Here only the first sentence is emitted. Tests/clients matching the full string diverge.

src/commands/scripts.ts
```diff
@@ -129,7 +141,146 @@ export const evalshaCommand = defineCommand<EvalShaArgs>({
...
+
+export const fcallCommand = defineCommand<FcallArgs>({
+  name: 'fcall',
+  since: { redis: '7.0.0', valkey: '7.2.0' },
+  schema: t.object({
+    functionName: t.string(),
+    numKeys: t.integer({ min: 0 }),
+    rest: t.variadic(t.bulk()),
+  }),
+  flags: ['write', 'movablekeys', 'noscript'],
```
fatal10110, line 307: MEDIUM — wire-compat. Static command flags diverge from real Redis (`COMMAND INFO`). Most impactful: EVAL/EVALSHA/FCALL carry a static `write` flag here; real Redis does **not** — write is determined dynamically.

Real 7.2.1 `COMMAND INFO` flag sets:
- `fcall`: noscript, stale, skip_monitor, no_mandatory_keys, movablekeys (no `write`)
- `fcall_ro`/`eval_ro`: readonly + the above
- `eval`/`evalsha`: noscript, stale, skip_monitor, no_mandatory_keys, movablekeys (no `write`)
- `slowlog`, `acl`, `function`: **empty** flag array (here: slowlog `[readonly,admin]`, acl/function `[admin,noscript]`)
- `shutdown`: admin, noscript, loading, stale, no_multi, allow_busy

Clients reading COMMAND INFO for write-detection / replica-safety / MULTI-eligibility get wrong answers. (Note these are the introspection/COMMAND-INFO flags; if the internal exec flags must stay for routing, the introspection output still needs to match real Redis.)

src/commands/scripts.ts
```diff
@@ -266,3 +418,254 @@ async function runLuaScript(
...
+  ctx: RedisExecutionContext,
+): RedisResult {
+  if (args.rest.length < 1 || args.rest.length > 2) {
+    throw new WrongNumberOfArgumentsError('function|restore')
+  }
+
+  const mode = args.rest[1]?.toString().toLowerCase() ?? 'append'
+  if (mode !== 'append' && mode !== 'flush' && mode !== 'replace') {
+    throw new RedisCommandError(
+      'FUNCTION RESTORE only supports FLUSH|APPEND|REPLACE',
```
fatal10110: MEDIUM — wire-compat. FUNCTION RESTORE bad-mode wording differs from real Redis.

Real 7.2.1: `FUNCTION RESTORE <payload> BADMODE` → `ERR Wrong restore policy given, value should be either FLUSH, APPEND or REPLACE.` Here: `FUNCTION RESTORE only supports FLUSH|APPEND|REPLACE`.

src/commands/connection.ts
```diff
@@ -776,6 +783,217 @@ export const lastsaveCommand = defineCommand({
...
+      if (args.args.length < 2) {
+        throw new WrongNumberOfArgumentsError('acl|dryrun')
+      }
+
+      const username = args.args[0].toString()
+      if (username !== DEFAULT_ACL_USER) {
+        throw new RedisCommandError(`User '${username}' not found`)
+      }
+
+      ctx.executor.plan(args.args[1], args.args.slice(2))
```
fatal10110: MEDIUM — wire-compat. ACL DRYRUN of an unknown command returns the wrong error.

Real 7.2.1: `ACL DRYRUN default nosuchcmd` → `ERR Command 'nosuchcmd' not found`. Here it routes through `executor.plan()`, whose UnknownRedisCommandError yields `ERR unknown command 'nosuchcmd', with args beginning with: ...`. Catch the unknown-command case and emit the DRYRUN-specific `Command '<c>' not found`.

src/commands/connection.ts
```diff
@@ -776,6 +783,217 @@ export const lastsaveCommand = defineCommand({
...
+        lines.push(
+          value('DRYRUN <username> <command> [<arg> ...]'),
+          value('    Check whether a user can run a command.'),
+        )
+      }
+      lines.push(value('HELP'), value('    Prints this help.'))
+      return array(lines)
+    }
+
+    throw unknownAclSubcommand(subcommand)
```
fatal10110: MEDIUM — wire-compat. ACL (and SLOWLOG) unknown-subcommand errors emit the **lowercased** subcommand; real Redis preserves the case the client typed.

Real 7.2.1: `ACL BOGUS` → `ERR unknown subcommand 'BOGUS'. Try ACL HELP.` Here `subcommand` is already `.toLowerCase()`d, so the client sees `'bogus'`. Same in `slowlogCommand`. (FUNCTION is already correct — it interpolates the original-case `args.subcommand`.) Use the original-case token in these two error strings.

src/state/function-registry.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  return Buffer.from(
+    `local __redis_functions = {}\n${body}\nreturn __redis_functions[${JSON.stringify(name)}](KEYS, ARGV)`,
+  )
+}
+
+function parseDump(payload: Buffer): RedisFunctionLibrary[] {
+  let entries: Array<{ code: string }>
+  try {
+    entries = JSON.parse(payload.toString()) as Array<{ code: string }>
```
fatal10110: MEDIUM — correctness. `parseDump` catches `JSON.parse` failures but not shape mismatches; the `as Array<{code}>` cast does no runtime check.

`FUNCTION RESTORE` with valid-JSON-but-wrong-shape (`{"code":"x"}`, `null`, `42`, `"hi"`) hits the unconditional `entries.map(...)` → `TypeError: entries.map is not a function` / `Cannot read properties of null`. `functionRestore` wraps it via `errorMessage(err)`, so the client receives `-ERR entries.map is not a function` — an internal JS message. Validate the parsed shape and throw the same clean `Invalid function payload`.

src/commands/connection.ts
```diff
@@ -776,6 +783,217 @@ export const lastsaveCommand = defineCommand({
...
+  keys: () => [],
+  execute: args => {
+    const subcommand = args.subcommand.toLowerCase()
+
+    if (subcommand === 'get') {
+      if (args.args.length > 1) {
+        throw new WrongNumberOfArgumentsError('slowlog|get')
+      }
+      if (args.args.length === 1) {
+        parseIntegerToken(args.args[0])
```
fatal10110: LOW — wire-compat. SLOWLOG GET with a non-integer count returns the wrong error.

Real 7.2.1: `SLOWLOG GET abc` → `ERR count should be greater than or equal to -1`. Here `parseIntegerToken` throws the generic `value is not an integer or out of range`. (`GET -1` / `GET 5` both fine.)

src/state/function-registry.ts
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  restore(payload: Buffer, mode: 'append' | 'flush' | 'replace'): void {
+    const libraries = parseDump(payload)
+    if (mode === 'flush') {
+      this.clear()
+    }
+
+    for (const library of libraries) {
+      this.load(library, mode === 'replace')
```
fatal10110: LOW — correctness. FUNCTION RESTORE in append/replace mode is non-atomic. If the Nth library `load()` throws (e.g. duplicate name in append mode), libraries 1..N-1 are already inserted and there is no rollback — the registry is left half-populated. Real Redis restores atomically. Validate/stage all libraries before mutating, or snapshot-and-rollback.

src/state/function-registry.ts
```diff
@@ -0,0 +1,134 @@
...
+      const fn = library.functions.find(entry => entry.name === name)
+      if (fn) {
+        return { ...fn, script: Buffer.from(fn.script) }
+      }
+    }
+
+    return null
+  }
+
+  dump(): Buffer {
```
fatal10110, line 58: LOW — wire-compat (acknowledged in PR). FUNCTION DUMP/RESTORE use a homegrown JSON+base64 payload, not the real RDB-style binary, so payloads are not interoperable with real Redis RESTORE. Minor extra divergence: real `FUNCTION DUMP` with no libraries returns a non-empty ~11-byte binary blob; here it returns `"[]"`. Self-consistent for the mock's own round-trip; flagging for awareness.

PR #374, opened 2026-09-21
refactor: remove dead encoder option from transport plumbing

### LINKED ISSUE #361: refactor: remove dead 'encoder' option threaded through server/session/transport
## Problem
`RespEncodeOptions` is `{ version?: RespVersion }` and is threaded through 5 layers as an `encoder?` option:
`Resp2Server` → `attachSession` → `Resp2SessionAdapter`, and `CreateVirtualConnectionOptions` → `attachSession`.

But `Resp2SessionAdapter.writeRedisResult` (`src/core/transports/resp2/session-adapter.ts:125-129`) always does `{ ...this.encoder, version: this.session.protocolVersion }`, so the option can never have any effect. Grep confirms no caller passes `encoder:` anywhere in `src/`, `tests/`, `tests-integration/`.

## Proposal
Remove the `encoder` option from `Resp2ServerOptions`, `AttachSessionOptions`, `Resp2SessionAdapterOptions`, `CreateVirtualConnectionOptions` and the private fields; encode with `{ version: this.session.protocolVersion }`.

## Compatibility
`Resp2ServerOptions`/`AttachSessionOptions` are reachable via `/core` — trivial type-level break for an option that was a no-op.

~-15 lines, purely mechanical.

fatal10110:
## What changed

`RespEncodeOptions` is `{ version?: RespVersion }` — a single optional field. It was threaded through the transport stack as an `encoder?` option across four option types plus two private fields:

- `Resp2ServerOptions` (`src/core/transports/resp2/server.ts`) — option, private field, forwarded to `attachSession`
- `AttachSessionOptions` (`src/core/transports/attach-session.ts`) — option, forwarded to the adapter
- `Resp2SessionAdapterOptions` (`src/core/transports/resp2/session-adapter.ts`) — option + private field
- `CreateVirtualConnectionOptions` (`src/core/transports/virtual-connection.ts`) — option, forwarded to `attachSession`

It could never have any effect. `Resp2SessionAdapter.writeRedisResult` spread the stored encoder and then immediately overrode its only field with the session's negotiated protocol version:

```ts
encodeRedisResult(result, {
  ...this.encoder,
  version: this.session.protocolVersion,
})
```

This PR removes the option and the private fields, and encodes with just `{ version: this.session.protocolVersion }`. It also drops the now-inaccurate `encoder` row from the `Resp2ServerOptions` table in `docs/API.md`, and — following review — adds the previously undocumented `nodeRole` row to that same table.

`RespEncodeOptions` itself is **kept** — it is still the options parameter of `encodeRedisValue` / `encodeRedisResult` and is still exported from `src/internal.ts`.

### Verification that the option was genuinely unused

`grep -rn "encoder" src tests tests-integration tests-package --include='*.ts'` on `origin/main` returned 22 hits. Every one was either (a) the plumbing removed here, or (b) an import/export of the `resp-encoder` module's real API (`RespEncodeOptions` type, `RespVersion`, `encodeRedisValue`, `encodeRedisResult`) — those stay. No call site anywhere passed `encoder:` other than the pass-through links in the chain itself.

## Diff

```
 docs/API.md                                  | 2 +-
 src/core/transports/attach-session.ts        | 3 ---
 src/core/transports/resp2/server.ts          | 5 -----
 src/core/transports/resp2/session-adapter.ts | 5 -----
 src/core/transports/virtual-connection.ts    | 3 ---
 5 files changed, 1 insertion(+), 17 deletions(-)
```

Pure deletion on the code side — no behavior change. The single insertion is the `nodeRole` documentation row added during review.

## Compatibility

`Resp2ServerOptions` and `AttachSessionOptions` are reachable through the published `js-redis-server/core` subpath (see `src/internal.ts`), so this is a **type-level breaking change** — but only for an option that was a no-op. Any external caller that was passing `encoder:` was getting nothing from it, and will now see a TypeScript excess-property error. Runtime behavior is identical either way: the encoder version has always come from the session's negotiated RESP version.

Review raised that a PR-body note is a fragile carrier for this, since the repo has no CHANGELOG and releases are cut manually. That process gap is tracked separately in #386 (make the `/core` guard detect export *removals*, and add a durable CHANGELOG) — it is not something this PR can fix in its own diff.

## Testing

| Suite | Result |
| :--- | :--- |
| `npm run lint` | pass |
| `npm test` (includes `tsc --noEmit`) | pass — 320 tests, 0 failures |
| `npm run build` | pass |
| `npm run test:integration:mock:ioredis` | pass — 475 tests, 0 failures |
| `npm run test:integration:mock:node-redis` | pass — 524 tests, 0 failures |
| `npm run test:integration:raw:mock` | pass — 105 tests, 0 failures |

The `:real` suites were not run (they need the docker-compose Redis cluster).

Closes #361


src/core/transports/resp2/server.ts
```diff
@@ -12,7 +11,6 @@ export type Resp2ServerOptions = {
   server: RedisServerState
   executor: CommandExecutor
   logger?: Pick<Logger, 'error'>
-  encoder?: RespEncodeOptions
   nodeRole?: RedisClusterNodeRole
```
fatal10110, line 14: **low — type-level breaking change (surfaced by both the Redis-compat and architecture lenses)**

Removing `encoder?` from `Resp2ServerOptions`, `AttachSessionOptions` and `CreateVirtualConnectionOptions` is source-breaking for TypeScript consumers of the published `js-redis-server/core` subpath. All three are public:

```
src/internal.ts:206  type AttachSessionOptions
src/internal.ts:212  type CreateVirtualConnectionOptions
src/internal.ts:226  type Resp2ServerOptions
```
and `package.json` maps `./core` -> `dist/core.*` built from `src/internal.ts` (tsup entry `core: "src/internal.ts"`, `dts: true`).

Failure scenario: a downstream consumer writing `new Resp2Server({ server, executor, encoder: { version: 3 } })` compiles today and fails with TS2353 (object literal may only specify known properties) after upgrading. Runtime behavior is unchanged — the field was provably discarded — so the break is compile-only, but it is silent until their build runs.

The PR body does disclose this in its Compatibility section, which is the right call. The gap is downstream of the PR: the repo has no `CHANGELOG.md` and no `.changeset/`, and `git log origin/main` shows releases are cut manually (`cd58b26 Release v0.2.0`), so the PR body is the only carrier of the note. Please make sure it reaches the release notes for the next publish. Mitigating: the package is `0.3.0`, so a 0.x minor bump legitimately carries a type-level removal, and the removed field was inert in every case.

Not a blocker — flagging so the disclosure is not lost at release time.

src/core/transports/resp2/session-adapter.ts
```diff
@@ -123,7 +119,6 @@ export class Resp2SessionAdapter {
       if (!result.options?.omitReply) {
         await this.transport.write(
           encodeRedisResult(result, {
-            ...this.encoder,
             version: this.session.protocolVersion,
```
fatal10110, line 122: **low / informational — pre-existing, NOT introduced by this PR. Do not block on it.**

Confirming the removal is a true no-op here, then recording one latent detail this line now owns alone.

Verification: `RespEncodeOptions` is byte-identical on `origin/main` and `pr-374` and has exactly one member —

```ts
export type RespEncodeOptions = {
  version?: RespVersion
}
```

— and `protocolVersion` is a non-optional getter (`src/core/client-session.ts:151`), so it can never be `undefined` and never fell back to the spread value. `{ ...this.encoder, version: this.session.protocolVersion }` had its only contributable key unconditionally overwritten in 100% of cases, not merely usually. Exhaustive scan of the PR tree finds two encode call sites total — this one (the single funnel for command replies via `handleFrame`/`writeExecutorResult`, streamed replies via `drainStream`, and session pushes via `writeSessionPushes`) and `client-session.ts:378` for pre-encoded MULTI/EXEC element bytes — and both still pass a correct version. The virtual-connection path shares `attachSession` and therefore this same funnel. Nothing lost a version.

The detail worth recording: `this.session.protocolVersion` is read inside the deferred `write` closure, i.e. at writeChain flush time, not when the reply value was produced. With the spread gone, this line is now the sole determinant of the wire version. A reply produced under RESP2 but flushed after a concurrent protocol switch would encode at the new version; real Redis serializes bytes into the output buffer at produce time and cannot exhibit this.

In practice the window looks unreachable today. The only in-session switches are HELLO and RESET, both ordinary frame-loop commands sharing this same writeChain, and real Redis closes the pubsub variant outright. Ground-truthed against redis-server 7.2.1:

```
sub (RESP2):  *3\r\n$9\r\nsubscribe\r\n$2\r\nch\r\n:1\r\n
message:      *3\r\n$7\r\nmessage\r\n$2\r\nch\r\n$2\r\nhi\r\n
HELLO 3:      -ERR Can't execute 'hello': only (P|S)SUBSCRIBE / (P|S)UNSUBSCRIBE / PING / QUIT / RESET are allowed in this context
```

The mock matches: `helloCommand` carries `flags: ["noscript"]` with no `"subscribed"`, so it is rejected in subscribed mode, while `resetCommand` carries `"subscribed"` and is allowed — the same allowlist. HELLO's own reply is also correct: real Redis encodes it at the *new* version (`%7\r\n...`, not `*14`), and the mock calls `ctx.session.setProtocolVersion(version)` before building the map (`src/commands/connection.ts:792`).

No action needed on this PR.

docs/API.md
```diff
@@ -297,7 +297,6 @@ new Resp2Server(options: Resp2ServerOptions)
 | `server`   | `RedisServerState`      | Yes      | The database server state containing database instances.       |
 | `executor` | `CommandExecutor`       | Yes      | The command executor handling pipeline and execution policies. |
 | `logger`   | `Pick<Logger, 'error'>` | No       | Optional logger for error logging.                             |
```
fatal10110, line 299: **low — optional, pre-existing gap in the table this PR is already editing**

Removing the `encoder` row is right: it advertised a "Custom RESP protocol encoding options" knob that silently did nothing, so the table is strictly more accurate now.

While you are in this table, though, it is still incomplete — `nodeRole` is undocumented:

```ts
// src/core/transports/resp2/server.ts (head)
export type Resp2ServerOptions = {
  server: RedisServerState
  executor: CommandExecutor
  logger?: Pick<Logger, 'error'>
  nodeRole?: RedisClusterNodeRole   // <- not in the table
}
```

`git show pr-374:docs/API.md | grep -n nodeRole` returns nothing. A reader constructing a cluster-mode `Resp2Server` by hand has no documented way to learn that `nodeRole` exists or that it drives replica-vs-master behavior in `ClusterPolicy`; they would have to read the source.

Pre-existing, so entirely your call whether to fold it in — but this PR is the one touching this exact table, which makes it the cheap moment. A one-line addition would leave the table complete rather than merely less wrong.

src/core/transports/resp2/server.ts
fatal10110: Agreed on all points, and I confirmed the `/core` reachability independently — `src/internal.ts` exports all three option types and `package.json` maps `./core` to the tsup `core` entry built from it with `dts: true`, so the removed `encoder?` key really was on the published type surface. TS2353 on a downstream `encoder: { version: 3 }` is the correct expectation.

Nothing to change in this diff, though: the gap you identified is downstream of the PR, at release time. That is now tracked in #386, which covers both halves — adding a CHANGELOG (or changesets) so a breaking-change note has somewhere durable to live between merge and publish, and making the `/core` guard detect *removals* so a deleted export fails CI instead of sailing through green. This PR is named there alongside #375 and #376 as one of the three that motivated it.

So: the disclosure stays in the PR body, and #386 is the durable carrier. Leaving the diff as-is.

src/core/transports/resp2/session-adapter.ts
fatal10110: Confirmed, and agreed it is informational only — no change made.

I re-verified the no-op claim from the other direction: `protocolVersion` is a plain non-optional getter returning `RespVersion` (`src/core/client-session.ts:151-153`, `return this.respVersion`), so it is never `undefined` and the spread's only contributable key was overwritten unconditionally, not merely usually. That matches your reading.

On the latent detail: agreed it is pre-existing and untouched by this diff. The `version: this.session.protocolVersion` read was already inside the deferred `write` closure on `origin/main` — the closure boundary is `this.writeChain = this.writeChain.catch(() => {}).then(write)` at line 134, and removing `...this.encoder` did not move the read relative to it. So the produce-time-vs-flush-time gap is exactly as reachable (or not) as it was before the PR; this line just no longer shares the determination with a key that never won anyway.

One nuance I would add rather than contradict: `writeSessionPushes` also funnels through `writeRedisResult` (line 143), so queued pushes serialize onto the same `writeChain` as command replies — the push path is not a second, independent ordering. That reinforces your conclusion rather than weakening it, since a push and a subsequent HELLO cannot interleave out of order on the chain. Your subscribed-mode argument then covers the pubsub pushes specifically: `helloCommand` lacks the `"subscribed"` flag, so it is rejected in that mode, matching the 7.2.1 ground truth you captured.

I have not tried to prove the window unreachable for every conceivable future push source, so I would treat it as a latent invariant worth preserving rather than a closed question — but nothing to act on here, and it is out of scope for a pure-deletion PR.

docs/API.md
fatal10110: Accepted and fixed in 4785d42.

Verified the gap first: `grep -rn "nodeRole\|RedisClusterNodeRole" docs/API.md` returns nothing across the whole file, so the option was undocumented everywhere, not just absent from this one table. `RedisClusterNodeRole` is `'master' | 'replica'` (`src/state/cluster-topology.ts:3`) and is re-exported from `/core` via `src/internal.ts:35`, so it is a public type a hand-rolling consumer can actually name.

Added row:

```
| `nodeRole` | `RedisClusterNodeRole`  | No       | Cluster role; a replica redirects keyed commands to master.    |
```

I kept the description to the observable consequence rather than naming `ClusterPolicy`, since the rest of this table describes what options do rather than which internal class consumes them — and the redirect behavior is the part a caller needs to predict.

You are right that this was the cheap moment. Doc-only, no code touched; re-ran the full suite anyway after the change (lint, prettier, `npm test` 320/320, build, and the three mock integration suites: ioredis 475/475, node-redis 524/524, raw-tcp 105/105 — all green).

PR #375, opened 2026-09-21
refactor: collapse RedisKeyspace into RedisDatabase

### LINKED ISSUE #362: refactor: merge RedisKeyspace into RedisDatabase and drop WrongRedisTypeError indirection
## Problem
`RedisDatabase` (`src/state/database.ts`) is largely a pass-through over `RedisKeyspace` (`src/state/keyspace.ts`): `get`, `getType`, `set`, `delete`, `expire`, `persist`, `getExpiration`, `flush`, `size`, `entriesSnapshot`, ... each just forwards one call. `RedisKeyspace` has no other users besides `database.ts`, `state/index.ts`, `internal.ts` and `tests/keyspace-update.test.ts`.

There is also an error dance: `keyspace.update` throws `WrongRedisTypeError` (a state-layer error) which `RedisDatabase.updateTyped` catches and re-throws as `WrongTypeRedisError` (core), and `getTyped` throws the core one directly. `state/` already imports core errors, so the intermediate class is redundant.

## Proposal
- Throw `WrongTypeRedisError` directly from the keyspace update path; delete `WrongRedisTypeError` and the try/catch in `updateTyped`.
- Merge `RedisKeyspace` into `RedisDatabase` (or, conversely, keep `RedisKeyspace` as the only data class and let `RedisDatabase` expose `keyspace` + `turnQueue` + `mutations` + typed `update*` helpers) — pick whichever keeps `keyspace-update.test.ts` easy to port.
- Remove now-unused re-exports from `state/index.ts` / `internal.ts`.

## Acceptance
- No behavior change (WATCH dirty semantics of `markChanged` vs `markCommitted`, empty-collection deletion, evict events all preserved).
- ~-80 lines, one fewer error class, one fewer layer to read when debugging a write.

fatal10110:
Closes #362

## Direction picked for (b): merge the keyspace **into** `RedisDatabase`

The issue offered two directions. I merged `RedisKeyspace` into `RedisDatabase` rather than keeping `RedisKeyspace` as the single data class, because the alternative is far more invasive:

- **Every command** reaches the keyspace through `ctx.db.get(...)` / `ctx.db.updateHash(...)`. Making `RedisKeyspace` the data class and `RedisDatabase` a thin owner means either rewriting all of those call sites to `ctx.db.keyspace.get(...)` (a huge diff across `src/commands/`), or keeping exactly the forwarding methods the issue asks us to delete. Merging deletes the forwarders without touching a single command.
- **The test ports cleanly.** `tests/keyspace-update.test.ts` keeps every assertion it had, now driven through the typed `update*` wrappers.

The merged class adds **no** public surface: `update()` is `private`, alongside `getTyped`, `updateTyped`, `getLiveEntry`, `evictIfExpired` and `emitWrite`. Its only caller anywhere in `src/` is the private `updateTyped`. (An earlier revision of this PR left it public so the test could port as a pure rename; review was right that a test needing access is not a reason to widen the storage core's API, especially as raw `update` hands the mutator the **untracked** value and so bypasses the hash-field TTL expiry that only `TrackedHashData` enforces. Fixed in 4ecbcde.)

### Why `src/state/keyspace.ts` still exists

It is now types-only: `KeyspaceEntry`, `SetOptions`, `ExpirationState`, `KeyspaceMutationTracker`. `src/state/tracked-values.ts` imports `KeyspaceMutationTracker`, and `database.ts` imports the `Tracked*` classes — putting the tracker type in `database.ts` would create a module cycle between the two for no benefit. The *layer* is gone; only the data-model shapes remain.

## (a) `WrongRedisTypeError` dropped

`keyspace.update()` threw a state-layer `WrongRedisTypeError`, which `RedisDatabase.updateTyped` caught and rethrew as the core `WrongTypeRedisError` — while `getTyped` threw the core error directly. `database.ts` already imported from `../core/redis-error`, so `update()` now throws `WrongTypeRedisError` itself and the `try`/`catch` is gone. One fewer error class, one fewer rethrow.

New unit test pins the client-visible outcome: `updateHash` against a string key throws `WrongTypeRedisError` with code `WRONGTYPE`, leaves the payload (not just the type discriminant) intact, and emits no mutation event.

## Behavior: unchanged

This is a transplant, not a rewrite — `update()` is byte-identical to `main` apart from `this.database` → `this.id` (same value) and the error swap:

- **WATCH dirty semantics are unchanged.** The `markChanged` / `markCommitted` / `!existing` carve-out is carried over verbatim: a `markCommitted`-only change dirties WATCH **only when it creates the key**. Reviewer ground-truthed this against redis-server 7.2.1 — 13/13 stream cases match real Redis exactly.
- **Empty-collection deletion is unchanged**, including the `string` and `stream` exclusions, and still emits a `delete` event only when the key previously existed.
- **Lazy expiry is unchanged.** `getLiveEntry` still evicts on read and emits an `evict` mutation event, so WATCH continues to see expiry as a delete.
- All mutation events keep their existing type/shape and the bus's clone-on-fanout behavior.

### Two sharp edges, now documented rather than glossed

An earlier revision of this PR claimed "ghost entries cannot happen" and added JSDoc saying nothing is persisted unless the mutator marks the change. **That is true only for a brand-new key**, and review caught it. For an *existing* key, `getLiveEntry` returns the stored entry rather than a copy, so the mutator writes straight through it; a mutator that throws part-way leaves a ghost empty hash that `getType` still reports as `hash`, with no event emitted and no WATCH dirtied. Reproduced on this branch:

```
CASE 1 unmarked delete on existing -> fields: [ '62' ]  | events: 0
CASE 2 throw on existing           -> getType: hash | fields: [] | events: 0
CASE 3 throw on new                -> getType: null | events: 0
```

This is **pre-existing and unchanged from `main`** — but since this PR is the one re-documenting the method, the JSDoc now states it narrowly instead of promising a rollback that never existed.

Second: `markCommitted` suppresses the mutation event **outright**, and that same bus also drives keyspace notifications. The carve-out is WATCH-faithful but not notification-faithful — real Redis keeps `signalModifiedKey` and `notifyKeyspaceEvent` independent, and the single-bus design here cannot express "notify but do not dirty". Noted in both the `update()` JSDoc and the `KeyspaceMutationTracker` docstring, with a pointer to #379, which tracks the split.

## Diff

Against `origin/main`:

```
 AGENTS.md                                          |   2 +-
 docs/ARCHITECTURE.md                               |  24 +-
 src/internal.ts                                    |   2 -
 src/state/database.ts                              | 284 +++++++++++++++++---
 src/state/index.ts                                 |   1 -
 src/state/keyspace.ts                              | 291 +--------------------
 src/state/tracked-values.ts                        |   5 +-
 .../ioredis/key/keyspace-empty-cleanup.test.ts     |   2 +-
 tests/keyspace-update.test.ts                      | 265 ++++++++-----------
 9 files changed, 392 insertions(+), 484 deletions(-)
```

**`src/` only: 268 insertions, 315 deletions — net -47 lines** (was -71 before the added JSDoc caveats). Overall net -92.

Stale re-exports removed: `RedisKeyspace` and `WrongRedisTypeError` from both `src/state/index.ts` and `src/internal.ts`. Docs updated: both mermaid diagrams plus the top-level pipeline diagram in `docs/ARCHITECTURE.md`, the layer table, the state/data-model prose, the `AGENTS.md` state-layer bullet, a `tracked-values.ts` comment, and the `keyspace-empty-cleanup.test.ts` header. No `RedisKeyspace` / `WrongRedisTypeError` / `keyspace.update` reference remains anywhere in the tree.

## Breaking change for the `/core` subpath

`RedisKeyspace` and `WrongRedisTypeError` were both re-exported from `src/internal.ts`, which is the published `./core` entry point (`dist/core.js`). Removing them is a **breaking change for consumers of `js-redis-server/core`** — anyone importing either symbol from that subpath will break. Neither is reachable from the main `.` entry point, and neither is used anywhere in this repo outside the files touched here. Worth a minor/major bump depending on how strictly `/core` is treated as public API.

Review correctly notes that `tests-package/dual-format.test.ts`'s `assertCore` only spot-checks six names, so dropping two `/core` exports leaves CI green. That gap is real but pre-existing, and the right fix — pinning the full `/core` export list — is its own change: a partial assertion bolted on here would give false confidence. Filed as follow-up rather than widened into this PR.

## Tests run

Re-run after the review fixes:

| Suite | Result |
| :--- | :--- |
| `npm run lint` | pass (clean) |
| `npm test` (incl. `tsc --noEmit`) | pass — 321/321 |
| `npm run build` | pass |
| `npm run test:integration:mock:node-redis` | pass — 524/524 |
| `npm run test:integration:compatibility:mock` | pass — 8/8 on each of redis-6.2, 7.0, 7.2, 7.4, 8.0, valkey-8.0, valkey-9.0 |

The `:real` suites were not run (they need a docker-compose Redis cluster).


src/state/database.ts
```diff
@@ -121,6 +179,75 @@ export class RedisDatabase {
     return this.getTyped<RedisStreamData>(key, 'stream')
   }
 
+  /**
+   * Read-modify-write a single key under its expected type. The mutator is
+   * handed the live value plus a tracker; nothing is persisted (and no event is
```
fatal10110: **HIGH — this JSDoc (and the PR body's "Ghost entries still cannot happen") is false for an _existing_ key.**

"nothing is persisted (and no event is emitted) unless the mutator marks the change" holds only on the **new-key** path. For an existing key, `const entry: KeyspaceEntry = existing ?? {...}` binds `entry` to the **live stored object** — `getLiveEntry` returns the entry itself, not a clone (unlike `get()`, which clones) — so the mutator writes straight into the keyspace. The early `if (!dirty && !committed) return result` reverts nothing.

Ran it against this PR's code:

```
seeded fields: [ '61', '62' ]
CASE 1 unmarked delete on existing key -> fields now: [ '62' ] | events emitted: 0
CASE 2 throwing mutator on existing key -> fields now: []     | events emitted: 0 | getType: hash
CASE 3 throwing mutator on NEW key      -> getType: null      | events emitted: 0
```

CASE 2 is the one that matters: a mutator that throws part-way through an existing hash leaves a **ghost empty hash** — `getType` still reports `hash` with zero fields, and no `write`/`delete` event fires, so WATCH is never dirtied and no keyspace notification is emitted.

Real Redis cannot represent that state (ground-truthed on redis-server 7.2.1 locally):

```
HDEL last field  -> after-empty EXISTS h  -> :0   TYPE -> +none
LPOP last elem   -> after-empty EXISTS l  -> :0   TYPE -> +none
SREM last member -> after-empty EXISTS st -> :0   TYPE -> +none
ZREM last member -> after-empty EXISTS z  -> :0   TYPE -> +none
```

To be clear about scope: **this behavior is unchanged from `main`** — `update()` is a verbatim transplant, and I verified that method-by-method. What this PR adds is (a) new JSDoc that actively mis-states the contract and (b) making the method public. Both make the sharp edge easier to cut yourself on.

Suggest narrowing the JSDoc to the new-key case — the in-body comment just below it ("For a new key, mutate a not-yet-committed entry...") already says it correctly and narrowly — or making the rollback real for existing keys.

src/state/database.ts
```diff
@@ -121,6 +179,75 @@ export class RedisDatabase {
...
+   * emitted) unless the mutator marks the change. Prefer the typed
+   * `updateHash`/`updateList`/... wrappers — they hand the mutator a tracked
+   * value instead of the raw one.
+   */
+  update<TValue extends RedisDataValue, TResult>(
```
fatal10110: **MEDIUM — `update()` is new public surface on the storage core with zero production callers.**

The only caller anywhere in `src/` is the **private** `updateTyped`:

```
src/state/database.ts:333:    return this.update(key, expectedType, createValue, (value, tracker) =>
```

Every other member of this class that is internal-by-intent is marked `private` — `getTyped`, `updateTyped`, `getLiveEntry`, `evictIfExpired`, `emitWrite`. `update()` is the sole exception, and the PR body's stated reason is that keeping it public lets `tests/keyspace-update.test.ts` port as a rename.

A test needing access isn't a reason to widen production API on the storage core — and here it isn't even load-bearing. Raw `update()` hands the mutator the **untracked** value, bypassing the `Tracked*` layer, which is the only place hash-field TTL expiry is enforced (`TrackedHashData` in `src/state/tracked-values.ts`). A command author who reaches for `db.update(key, 'hash', ...)` instead of `db.updateHash(...)` silently sees logically-expired hash fields as live, and additionally inherits the manual `markChanged`/`markCommitted` discipline described in my comment on the JSDoc above. Nothing — no lint rule, no `@internal` tag, no naming convention — stops that.

Suggest `private update(...)` (or at minimum an `@internal` tag), with the hash/list/set/zset/stream test cases ported to the `Tracked*` wrappers. The new WRONGTYPE test reproduces identically through `db.updateHash(...)`, since `updateTyped` delegates to `update` and throws the same error. The only case not reachable through a wrapper is `RedisStringData` — but there is no `updateString` wrapper and nothing in `src/` calls `update` with `'string'`, so that test is pinning a branch production cannot reach.

src/state/database.ts
```diff
@@ -121,6 +179,75 @@ export class RedisDatabase {
...
+    key: Buffer,
+    expectedType: TValue['type'],
+    createValue: () => TValue,
+    mutator: (value: TValue, tracker: KeyspaceMutationTracker) => TResult,
```
fatal10110, line 213: **MEDIUM — `KeyspaceMutationTracker` appears in a public signature but is exported from no public entry point.**

`RedisDatabase` is re-exported from `src/internal.ts`, i.e. the published `./core` subpath, and `update()` is now public — so this parameter type is part of `/core`'s public API. But `KeyspaceMutationTracker` is exported from none of `src/internal.ts`, `src/index.ts`, or `src/state/index.ts`; those export only `ExpirationState`, `KeyspaceEntry`, and `SetOptions` from `./keyspace`.

Consequence: an inline arrow mutator still type-checks via contextual inference, so this is an ergonomics break rather than a compile error. But a consumer who wants a named mutator —

```ts
function bump(v: RedisHashData, tracker: KeyspaceMutationTracker) { /* ... */ }
db.update(k, 'hash', createHashData, bump)
```

— has no way to import the type, and must hand-redeclare it or fall back to `any`.

Either add `KeyspaceMutationTracker` to the export lists, or make this moot by making `update()` private per the comment above (my preference — it's the smaller change and it removes the question entirely).

src/internal.ts
```diff
@@ -114,13 +114,11 @@ export {
   RedisClusterTopology,
   RedisDatabase,
   RedisServerState,
-  RedisKeyspace,
```
fatal10110, line 118: **MEDIUM — the `/core` break is correctly flagged in the PR body, but nothing in CI would have caught it.**

Credit where due: the PR body calls this out explicitly, names both removed symbols, notes neither is reachable from the main `.` entry point, and recommends a version bump. That's the right handling, and at `0.3.0` pre-1.0 semantics make the removal defensible.

The gap is test coverage of the published surface. `tests-package/dual-format.test.ts`'s `assertCore` only asserts a handful of names (`defineCommand`, `CommandRegistry`, `t`, `Resp2Server`, `RedisServerState`, `createRedisCommandExecutor`); it never enumerated `RedisKeyspace` or `WrongRedisTypeError`, so dropping two exported symbols from `./core` leaves CI fully green.

Worth either pinning the `/core` export list in that test or adding an explicit CHANGELOG/release note, so the next `/core` removal isn't silent. (Verified there are no remaining in-tree consumers of either symbol.)

tests/keyspace-update.test.ts
```diff
@@ -242,6 +229,39 @@ describe('RedisKeyspace.update — ghost entries and empty-collection cleanup (#
...
+        )
+      },
+      (err: unknown) => {
+        assert.ok(err instanceof WrongTypeRedisError)
+        assert.strictEqual(err.code, 'WRONGTYPE')
+        return true
+      },
+    )
+
+    assert.strictEqual(db.getType(key), 'string')
```
fatal10110, line 184: **LOW — this test asserts the key's _type_ survives, but not its _value_.**

The PR body says the new test pins that WRONGTYPE "leaves the existing value intact", but the assertions only cover `getType(key) === 'string'` and `events.length === 0`. A regression that threw WRONGTYPE *after* clobbering the payload — while preserving the type discriminant — would still pass.

One line closes the gap:

```ts
assert.deepStrictEqual(db.getString(key), Buffer.from('v'))
```

Otherwise this is a good test: it pins the error class, the `WRONGTYPE` code, and the no-mutation-event invariant, and it correctly clears `events` *after* the `setString` setup so the assertion isn't vacuous.

For the record, I diffed the whole test port assertion-by-assertion — nothing was weakened or dropped, and `db.subscribe` is wired in `setup()` before any mutation, so no test was made vacuous by late subscription. The "byte-for-byte the same" claim checks out.

docs/ARCHITECTURE.md
```diff
@@ -140,7 +138,7 @@ graph TD
...
-| **State**     | In-memory keyspace, mutation events, cluster topology, script cache, connected clients, pub/sub, monitor feed      | [`RedisServerState`](../src/state/server-state.ts), [`RedisDatabase`](../src/state/database.ts), [`RedisKeyspace`](../src/state/keyspace.ts)                                                                                                                                                                                       |
+| **State**     | In-memory keyspace, mutation events, cluster topology, script cache, connected clients, pub/sub, monitor feed      | [`RedisServerState`](../src/state/server-state.ts), [`RedisDatabase`](../src/state/database.ts)                                                                                                                                                                                       |
```
fatal10110: **LOW — three `RedisKeyspace` references survived the sweep.**

The docs update is genuinely thorough — both mermaid diagrams in this file, this layer table, the state/data-model prose, and the `AGENTS.md` bullet are all correct and accurate to the new structure. Three stragglers:

1. `docs/ARCHITECTURE.md:37` (outside this diff) — the top-level pipeline diagram still reads `J[(RedisServerState<br/>→ Database → Keyspace)]`.
2. `src/state/tracked-values.ts:706` — comment still says "...while letting `keyspace.update` decide the dirty signal". This is the only stale reference left in `src/`.
3. `tests-integration/ioredis/key/keyspace-empty-cleanup.test.ts:11` — "...rather than in `RedisKeyspace.update()`". That file's header comment is the primary prose explanation of the empty-collection invariant, so it's worth keeping accurate.

src/state/database.ts
```diff
@@ -121,6 +179,75 @@ export class RedisDatabase {
...
+      }
+      return result
+    }
+
+    this.entries.set(id, entry)
+    // A markChanged write always dirties WATCH. A markCommitted-only change
+    // dirties only when it creates the key (`!existing`): real Redis treats
+    // bringing a watched key into existence as a write, but leaves a WATCH
+    // intact for in-place metadata changes to an already-existing key.
+    if (dirty || !existing) {
```
fatal10110, line 265: **Verified against real Redis — this carve-out is correct. No change requested.**

This is the highest-risk line in the PR, so I ground-truthed it rather than trusting the description. redis-server 7.2.1 locally: `WATCH k` on connection 1, mutation on connection 2, then `MULTI`/`PING`/`EXEC` on connection 1. `EXEC -> *-1` means the WATCH was dirtied and the transaction aborted; an array means it survived.

```
DIRTIED  | XADD (data write) on watched stream           EXEC: *-1
DIRTIED  | XDEL (data write) on watched stream           EXEC: *-1
INTACT   | XGROUP CREATE on existing watched stream      EXEC: *1 +PONG
INTACT   | XGROUP CREATECONSUMER on watched stream       EXEC: *1 +PONG
INTACT   | XREADGROUP (fills PEL) on watched stream      EXEC: *1 +PONG
INTACT   | XACK (drains PEL) on watched stream           EXEC: *1 +PONG
INTACT   | XCLAIM on watched stream                      EXEC: *1 +PONG
INTACT   | XGROUP SETID on watched stream                EXEC: *1 +PONG
INTACT   | XGROUP DELCONSUMER on watched stream          EXEC: *1 +PONG
```

And the `!existing` half of the condition:

```
DIRTIED  | SET on watched NONEXISTENT key                     EXEC: *-1
DIRTIED  | LPUSH on watched NONEXISTENT key                   EXEC: *-1
DIRTIED  | XGROUP CREATE ... MKSTREAM on watched NONEXISTENT  EXEC: *-1
```

So `if (dirty || !existing)` matches real Redis exactly, and the `markCommitted` call sites in `tracked-values.ts` (`addGroup`, and the no-mark `deleteGroup`/`addConsumer`/`deleteConsumer`) line up with the measured behavior.

The other two semantics carried over here also check out:

- **Lazy expiry** — a watched key that expires dirties the WATCH both passively and after a lazy-expiry read by another connection (`EXEC: *-1` both ways), which is what the `evict` event reproduces.
- **Empty-collection deletion and its exclusions** — `SET k "" -> EXISTS :1, TYPE +string, STRLEN :0` and `XADD` then `XDEL` of the only entry -> `EXISTS :1, TYPE +stream, XLEN :0`, so `isEmptyCollection` returning `false` for `string` and `stream` is right.

fatal10110:
### Late wire-compat findings (ground-truthed vs real Redis 7.2.1)

A second reviewer pass finished after the inline comments above were posted. **All five of this PR's claims verified TRUE**, including the non-obvious `markCommitted` / `!existing` carve-out — 13/13 stream cases reproduce real Redis exactly, and probe output from a `main`-built mock and a PR-built mock diffs clean, so **this PR introduces no behavioral change**.

One caveat worth adding to this PR, and some follow-up work that is **pre-existing and not caused by this diff**.

#### The caveat (worth addressing here)

The comment at `src/state/database.ts:241-244` and the `KeyspaceMutationTracker` docstring at `src/state/keyspace.ts:26-31` justify `markCommitted` with "real Redis does not treat [these] as touching a WATCH on the stream key." That is true and now *measured*. But the implementation achieves it by **suppressing the mutation event outright**, and that same event stream also drives keyspace notifications.

Real Redis keeps the two signals independent (`signalModifiedKey` vs `notifyKeyspaceEvent`). The single-bus design here cannot express "notify but do not dirty". Since this PR is the one consolidating and re-documenting that logic, it reads as a clean bill of health when it is only *WATCH*-faithful. A short caveat noting the notification gap would keep the comment honest.

#### Pre-existing divergences found while probing (carried over verbatim; follow-ups, not blockers)

| Sev | Behavior | Real Redis | Mock |
|---|---|---|---|
| high | Stream consumer-group cmds (`XGROUP CREATE`/`CREATECONSUMER`/`DELCONSUMER`/`SETID`/`DESTROY`, `XREADGROUP`, `XCLAIM`) | emits keyspace events | emits **nothing** |
| high | Last element removed from a collection | `hdel` **then** `del` | only `del` |
| medium | `SET..EX` / `SETEX` / `PSETEX` | `set` **and** `expire` | only `set` |
| medium | `XGROUP CREATE ... MKSTREAM` | `xgroup-create` | `xgroup` |

The second one is pinned by `tests/keyspace-update.test.ts:86,121,154` ("emits a single delete event"), so the divergence is currently enshrined in tests.

Also spotted outside this diff: `GETEX key EX n` emits `getex` where real Redis emits `expire`; `SETRANGE` does not enforce `proto-max-bulk-len`.

Splitting `signalModifiedKey` from `notifyKeyspaceEvent` would fix the first, second and fourth rows together — worth its own issue.

src/state/database.ts
fatal10110: Confirmed before changing anything — reproduced on this branch and got your CASE 1/2/3 output exactly. `getLiveEntry` returns the stored entry, not a copy (unlike `get()`), so the early `if (!dirty && !committed) return` reverts nothing for an existing key.

Narrowed the JSDoc to the new-key case in 4ecbcde, and named the ghost-empty-hash outcome explicitly rather than leaving it implied. Also corrected the PR body, which had a bolded "Ghost entries still cannot happen" — that was the line that overstated it. Behavior itself untouched, since as you say it is unchanged from `main`.

src/state/database.ts
fatal10110: Agreed, and the untracked-value argument is the part that convinced me: `Tracked*` is the only place hash-field TTL expiry is enforced, so a command author reaching for `db.update(key, 'hash', ...)` would silently see logically-expired fields as live. Verified your caller claim independently — exactly one caller in all of `src/`, the private `updateTyped`.

`private update(...)` in 4ecbcde, with the hash/list/set/zset/stream cases ported to the `Tracked*` wrappers. A test needing access was not a good enough reason; you were right.

src/state/database.ts
fatal10110: Took your preferred option — making `update()` private in 4ecbcde removes the question entirely, so `KeyspaceMutationTracker` no longer appears in any public signature and no export list needs widening.

src/internal.ts
fatal10110: Agreed the gap is real, and deliberately not fixing it here: the right fix is snapshotting the full `/core` export list, which is its own change: a partial assertion bolted onto this PR would give false confidence without catching the next removal.

That is now tracked in #386, which proposes exactly that (snapshot the exported symbol list, fail CI when one disappears, plus CHANGELOG notes for breaking changes). Leaving `assertCore` untouched here; the `/core` break stays documented in the PR body.

tests/keyspace-update.test.ts
fatal10110: Good catch — the PR body claimed more than the test proved. Added your line in 4ecbcde:

```ts
assert.deepStrictEqual(db.getString(key), Buffer.from('v'))
```

The test also now drives `db.updateHash(...)` rather than raw `update`, since that went private.

Related, and worth putting on the record here rather than leaving it buried in a commit message: making `update()` private removed the only way to reach it with `'string'`, so the "empty string value is never auto-deleted" test now goes through `setString` instead. That is the production-reachable path — nothing in `src/` calls `update` with `'string'`, as you noted — and the test picked up a value assertion in the move. **The explicit trade: `isEmptyCollection`'s `case 'string'` arm loses direct coverage.** It is unreachable exhaustiveness now, so I judged that better than `db['update'](...)` bracket-access, which would have preserved the test's letter while evading the point of your comment above. Happy to be overruled.

docs/ARCHITECTURE.md
fatal10110: All three fixed in 4ecbcde — the top-level pipeline diagram now reads `RedisServerState → RedisDatabase`, the `tracked-values.ts` comment says `RedisDatabase.update`, and the `keyspace-empty-cleanup.test.ts` header reads `RedisDatabase.update()`.

Re-swept afterwards: zero occurrences of `RedisKeyspace`, `WrongRedisTypeError`, or `keyspace.update` anywhere in the tree.

src/state/database.ts
fatal10110: Thanks for ground-truthing this rather than taking the description on trust — the 13/13 stream table is exactly the evidence this carve-out needed. Untouched, as you asked.

On your PR-level note: added the caveat in 4ecbcde, to both the `update()` JSDoc and the `KeyspaceMutationTracker` docstring — `markCommitted` suppresses the mutation event outright, and since the same bus drives keyspace notifications, the carve-out is WATCH-faithful but loses a notification real Redis still fires. Both point at #379 for the `signalModifiedKey` / `notifyKeyspaceEvent` split.

tests/keyspace-update.test.ts
```diff
@@ -1,247 +1,202 @@
...
-      (str, tracker) => {
-        str.value = Buffer.alloc(0)
-        tracker.markChanged()
-      },
-    )
+    // Strings never go through `update` in production (there is no
+    // `updateString` wrapper) — they are written whole via `set`/`setString`,
+    // which has no empty-collection rule at all. `isEmptyCollection` returning
+    // false for 'string' keeps the two consistent.
+    db.setString(key, Buffer.alloc(0))
```
fatal10110: **MEDIUM — the coverage trade is worse than "loses direct coverage": this test no longer detects a change to the arm it was written for, and nothing else does either.**

You flagged this yourself and invited an overrule, so: I think the call to avoid `db['update'](...)` bracket-access was right, but the replacement should not have kept the old test's name and place. As written, test 8 is a `setString`/`getString` round-trip sitting under a `describe('RedisDatabase.update — ...')` block, exercising a method (`set`) that, as your own comment says, "has no empty-collection rule at all". It cannot fail for the reason it is named.

Mutation-tested both versions against this branch:

```
# mutant: isEmptyCollection  case 'string': return true   (from false)

4ecbcde tests/keyspace-update.test.ts   -> 10 pass,  0 fail   (mutant SURVIVES)
                    full unit suite     -> 321 pass, 0 fail   (mutant SURVIVES)

743ca49 tests/keyspace-update.test.ts   -> 9 pass,   1 fail   (mutant KILLED)
  not ok 1 - RedisDatabase.update — ghost entries and empty-collection cleanup (#124)
```

So the old test was the only thing in the repo pinning that arm, and after the port the mutant is invisible to all 321 unit tests. I also checked for a wire-level backstop and there isn't one — no integration test asserts `SET k "" -> EXISTS 1`:

```
$ grep -rn "\.set([^)]*''" tests-integration/   # no matches
```

The invariant itself is real — ground-truthed again on redis-server 7.2.1 — it just has no guard any more:

```
SET k ""      -> +OK
EXISTS k      -> :1
TYPE k        -> +string
STRLEN k      -> :0
```

**That said, I do not think this blocks the PR**, because the mutant is currently unreachable rather than latent-dangerous. With `update` private, `updateTyped` is its only caller and the five wrappers pass only `hash`/`list`/`set`/`zset`/`stream` — `case 'string'` is pure TypeScript exhaustiveness (the switch has no `default`, so it cannot simply be deleted). The risk is a future `updateString` wrapper, or someone re-widening `update`, silently turning `SET k ""` into a delete.

Two cheap options, either is fine:

1. Say so in the code. One line at `src/state/database.ts:458` (and/or in the block comment at L445-448, which currently reads as a live, tested rule) — `// 'string' is unreachable today: update() is private and no updateString wrapper exists; kept for exhaustiveness.` — converts the untested arm from "looks like a tested rule" into "documented exhaustiveness". This is the honest form and matches how you handled the other sharp edges in this commit.
2. Rename/relocate the test to match what it now does (e.g. `'SET with an empty value keeps the key'`, moved next to the other string tests), so the `update` describe block stops claiming coverage it does not have.

Option 1 alone is enough for me.

For the record, the rest of the port holds up. I mapped all ten tests old→new and mutation-tested the rewritten mutators; nothing else was weakened, and four tests got stronger:

| # | change |
| :-- | :-- |
| 3 | + `deleted === 0` |
| 8 | + value assertion, **but subject changed — this comment** |
| 9 | + `entries.length === 0`, `events.length === 1`, `type === 'write'` |
| 10 | + `deepStrictEqual(db.getString(key), Buffer.from('v'))` (as requested) |

Tests 2 and 3 really are still on the unmarked path — `TrackedHashData.deleteField` calls `getField`, gets `undefined` and returns `false` *before* any `markChanged()` (`tracked-values.ts:93-105`), so they are not passing for a new reason. `list.trim(1, 0)` is exactly equivalent to the old `values.length = 0` + `markChanged()` (`start > stop` branch, `tracked-values.ts:265-280`). And test 9's move from `markChanged` to `addGroup`/`markCommitted` is a genuine upgrade: mutating `if (dirty || !existing)` to `if (dirty)` now fails test 9, which the old version did not catch.

src/state/database.ts
```diff
@@ -121,6 +179,92 @@ export class RedisDatabase {
...
   }
 
+  /**
+   * Read-modify-write a single key under its expected type, shared by the typed
+   * `updateHash`/`updateList`/... wrappers. Private on purpose: it hands the
+   * mutator the *untracked* value, which skips the `Tracked*` layer and with it
+   * hash-field TTL expiry. Go through a wrapper.
+   *
+   * The mutator gets the value plus a tracker, and must mark what it did:
+   * `markChanged` for a WATCH-dirtying write, `markCommitted` to persist
```
fatal10110: **LOW — new in 4ecbcde: the rewritten JSDoc drops the `!existing` qualifier that the other three copies of this explanation keep.**

The narrowing on the ghost-entry half is correct and I re-verified it (see the verdict comment). But two new lines in this same block state the `markCommitted` contract unqualified:

- L189-190: "`markCommitted` to persist **without dirtying**"
- L200-201: "`markCommitted` suppresses the mutation event ***outright***"

Both are false when `markCommitted` creates the key, which is precisely the case the carve-out exists for. Probed on this branch (4ecbcde), through the public surface:

```
D1 addGroup on NEW stream   -> getType=stream events=write
D2 addGroup on EXISTING     -> getType=stream events=(none)
```

`if (dirty || !existing)` at L262 is what makes D1 emit. So on a brand-new key `markCommitted` neither persists "without dirtying" nor suppresses the event "outright" — it does exactly the opposite.

This is the same failure mode as the finding this commit fixes (a JSDoc promising more than the code does), just inverted: the old text over-promised a rollback, the new text over-promises suppression. Worth flagging because the other three places all get it right and are more precise than this one:

- `src/state/keyspace.ts:29-31` — "A brand-new key still dirties — coming into existence is itself a write — so `RedisDatabase.update` only suppresses the dirty signal for in-place changes to an already-existing key."
- `src/state/tracked-values.ts:703-708` — "…`XGROUP CREATE ... MKSTREAM` can create a brand-new key that still must be persisted (and, as a key creation, does dirty the WATCH)."
- the in-body comment at L258-261 here: "A markCommitted-only change dirties only when it creates the key (`!existing`)".

The repo's own wire-level test says it plainest, and it runs against **real** Redis in CI (`raw-real`, green on this head):

```
tests-integration/raw-tcp/cluster-stream-watch.test.ts:186-202
  // Creating the watched key itself is a write, even via XGROUP CREATE MKSTREAM:
  // real Redis dirties a WATCH on a key that comes into existence.
  test('XGROUP CREATE MKSTREAM creating the watched key dirties it', ...)
    WATCH k; XGROUP CREATE k g 0 MKSTREAM; MULTI; SET k after; EXEC  ->  *-1
```

Two words fix it: "`markCommitted` to persist without dirtying **an existing key**", and "suppresses the mutation event outright **for an in-place change to an existing key**".

Narrow impact — `markCommitted` has exactly one call site in the tree (`TrackedStreamData.addGroup`, `tracked-values.ts:711`), and a maintainer reaching this JSDoc will also read the correct comment ~70 lines down at the code site. Not a blocker.

src/state/keyspace.ts
```diff
@@ -1,9 +1,8 @@
-import {
-  cloneRedisDataValue,
-  type RedisDataValue,
-  type RedisDataTypeName,
-} from './data-types'
-import { RedisMutationBus } from './mutation-events'
+// Data model for a database's keyspace. The storage itself lives on
+// `RedisDatabase` (src/state/database.ts) — these are the shapes it stores and
+// the contract its `update` mutators are handed.
```
fatal10110: **LOW — the symbol sweep is clean; three bits of *conceptual* prose still describe the old structure.**

Verified your sweep claim exactly as stated — zero occurrences tree-wide:

```
$ git grep -n "RedisKeyspace"       4ecbcde -- .   # (no output)
$ git grep -n "WrongRedisTypeError" 4ecbcde -- .   # (no output)
$ git grep -n "keyspace\.update"    4ecbcde -- .   # (no output)
```

Three survivors say "keyspace" as if it were still a thing that owns behavior, which is the exact impression this PR exists to remove:

1. **This file's header, L2-3** — "the contract its `update` mutators are handed". After this commit no mutator on any reachable path is handed a tracker: the five typed wrappers hand the mutator a `Tracked*` object, and the tracker is an internal detail of the now-private `update`. A contributor reading this goes looking for a public `update` to pass a tracker into and finds a private method — the same confusion the JSDoc rewrite was meant to end.

2. **`docs/ARCHITECTURE.md:356`** — "In-place collection updates run through a **keyspace-owned** mutation tracker". Nothing is keyspace-owned any more; the tracker is constructed inside `RedisDatabase.update`. (`docs/ARCHITECTURE.md:337`, "the keyspace removes the hash key when the last live field disappears", reads fine to me — that paragraph opens with "Each database _is_ its keyspace", so it's the concept, not the class.)

3. **`AGENTS.md:69`** — "drives `WATCH` today, keyspace notifications **later**". This one is now self-contradictory with the caveats you added in this very commit, which say the same bus drives keyspace notifications *today*. It does:

```
src/state/server-state.ts:84-88
  // mutations surface as Redis keyspace/keyevent notifications. The bridge is
  // always wired; it is a cheap no-op while `notifyKeyspaceEvents` is empty.
  const notifier = new KeyspaceNotifier(this.pubsubBroker)
  for (const database of this.databases) { database.subscribe(...) }
```

So the new caveat is correct and `AGENTS.md` is the stale one. Cheap to fix while you're in here, and #3 in particular is worth it because the caveat comments now point readers at that bus.

Non-blocking. Separately and even more trivially: removing `→ Keyspace` from the `docs/ARCHITECTURE.md:141` table cell left the row's padding wider than the table needs, so `prettier --check` on `docs/ARCHITECTURE.md` now warns where it didn't on `main`. CI only checks `**/*.{js,ts,json}` and `npm run format` only covers `js|ts|json`, so nothing breaks — mentioning it only so it isn't a surprise later.

fatal10110:
## Re-review of `4ecbcde` — **Ship.** 🟢

All seven prior findings are resolved or properly deferred, and I can now prove the claim the whole PR rests on: **the fixes changed no behavior whatsoever.** Transpiling both revisions of all three touched `src/` files with the repo's own TypeScript and `removeComments: true` gives byte-identical JavaScript —

```
database.ts        transpiled, comments stripped, 743ca49 vs 4ecbcde  ->  IDENTICAL
keyspace.ts        transpiled, comments stripped, 743ca49 vs 4ecbcde  ->  IDENTICAL
tracked-values.ts  transpiled, comments stripped, 743ca49 vs 4ecbcde  ->  IDENTICAL
```

which is what you'd expect, since the entire non-comment `src/` delta between the two commits is one token:

```
-  update<TValue extends RedisDataValue, TResult>(
+  private update<TValue extends RedisDataValue, TResult>(
```

### Prior findings

| :-- | :-- | :-- | :-- |
| 1 | high | JSDoc falsely claimed ghost entries can't happen | ✅ **Fixed** — verified by probe, not by reading the prose |
| 2 | med | `update()` public with zero production callers | ✅ **Fixed** — `private`; emitted `/core` d.ts shows `private update;` |
| 3 | med | `KeyspaceMutationTracker` in a public signature | ✅ **Moot** — gone from every exported signature |
| 4 | med | `/core` export coverage gap | ✅ **Correctly deferred** to #386 (open); `tests-package/` untouched |
| 5 | low | WRONGTYPE test asserted type but not value | ✅ **Fixed** — `deepStrictEqual(db.getString(key), Buffer.from('v'))` |
| 6 | low | Three stale `RedisKeyspace` references | ✅ **Fixed** — sweep verified clean; 3 bits of conceptual prose remain (inline) |
| 7 | — | WATCH carve-out (no change requested) | ✅ **Untouched**; caveat added in both places as promised |

**On #1 — I wrote the probe rather than trusting the text.** Driven through the public surface only, since `update` is private now:

```
A1  new-key + throw        -> getType=null   get=null      events=0
A2  existing-key + throw   -> getType=hash   fields=0      events=0   <- ghost, as documented
B   existing-key unmarked  -> getType=hash   fields=0      events=0   size()=1
```

A2 and B are exactly the "throws **or forgets to mark**" and "ghost empty hash that `getType` still reports as `hash`" wording. The narrowed doc now matches reality, and the PR body header is corrected.

**On #2 — I checked the port for weakened assertions rather than taking "every assertion preserved" on trust.** Mapped all ten tests old→new and then mutation-tested the production code to confirm each one still kills what it claims to:

| mutant | result |
| :-- | :-- |
| `case 'hash'` / `'list'` / `'set'\|'zset'` → `false` | tests 4 / 5 / 6 **fail** ✅ |
| `case 'stream'` → `true` | test 9 **fails** ✅ |
| `if (dirty \|\| !existing)` → `if (dirty)` | test 9 **fails** ✅ (the old version did *not* catch this) |
| drop the WRONGTYPE check | test 10 **fails** ✅ |
| commit the fresh entry before the mutator | tests 1, 2 **fail** ✅ |
| **`case 'string'` → `true`** | **321/321 still pass ❌** |

Nine of ten hold. Four tests were strengthened (3, 8, 9, 10), none weakened by assertion count, and tests 2/3 really are still on the unmarked path — `deleteField` returns `false` before any `markChanged()`. `list.trim(1, 0)` is exactly equivalent to the old `values.length = 0` + `markChanged()`.

**On #6 — your deliberate trade.** I think you called the `db['update'](...)` question right, but the trade is a bit worse than "loses direct coverage": the old test was the *only* thing in the repo killing that mutant, and there's no wire-level backstop either. Details and a two-line remedy inline. **Not a blocker** — with `update` private and no `updateString`, the arm is unreachable exhaustiveness, so the mutant has no runtime effect today.

### Previously-verified invariants — all re-confirmed, none regressed

- **WATCH carve-out** — re-ground-truthed the two cases the changed test touches on redis-server 7.2.1: `XGROUP CREATE ... MKSTREAM` on a missing key creates it with `XLEN 0` and **dirties** the WATCH (`EXEC -> nil`); `XGROUP CREATE` on an existing stream does **not** (`EXEC -> ["PONG"]`). Mock matches both. The full 13-case table is unchanged and still pinned at wire level by `tests-integration/raw-tcp/cluster-stream-watch.test.ts`, green in CI against **real** Redis.
- **Lazy expiry** — `evict` still emitted on read.
- **Empty-collection deletion + exclusions** — emptied hash/list/zset → single `delete`; `string` and `stream` still excluded.
- **Byte-identical transplant** — every non-comment added line in `database.ts` traces to `main`'s `keyspace.ts` apart from the three documented deltas (`this.database`→`this.id`, `WrongRedisTypeError`→`WrongTypeRedisError`, `private`).

### Tests — every number in your table reproduced independently

`tsc --noEmit` clean · `eslint` clean · unit **321/321** · ioredis mock **475/475** · node-redis mock **524/524** · raw mock **105/105** · full mock integration **1112/1112** · compatibility profiles **8/8** on each spot-checked profile · `tsup` build clean. CI is green on all 14 checks including the four `:real` suites you couldn't run locally.

### New findings (3, all non-blocking)

| Sev | Location | Issue |
| :-- | :-- | :-- |
| medium | `tests/keyspace-update.test.ts:149` | Test 8's reroute leaves `isEmptyCollection`'s `'string'` arm killed by nothing; mutation-verified both ways |
| low | `src/state/database.ts:189` | Rewritten JSDoc states the `markCommitted` contract unqualified — false when it creates the key |
| low | `src/state/keyspace.ts:3` | Symbol sweep is clean, but 3 bits of conceptual prose still describe a keyspace that owns behavior; `AGENTS.md:69` now contradicts your own new caveat |

### Two notes, neither a finding

- The new caveat correctly names `markCommitted` as a notification-loss path, but it's not the only one: the `isEmptyCollection` early-return also collapses the operation event into a bare `del` (real Redis fires `hdel` **then** `del`). That's pre-existing and identical on `main`, and #379's title already covers it — just don't let the JSDoc's framing narrow the eventual fix.
- `TrackedHashData` & co. are still declared-but-unexported types reachable through `updateHash`'s callback in the published `/core` d.ts. Same shape as finding #3, but unchanged from `main` and not something this PR introduced.

Nothing here blocks merge. The high finding is genuinely fixed, the medium ones are fixed or deferred with an issue, and the remaining items are doc precision and test-guard strength.

src/state/database.ts
fatal10110: You are right, and it is the more embarrassing kind of right — I repeated the exact failure mode I was fixing, inverted. Probed it on this branch before rewording and got your D1/D2 result:

```
D1 addGroup on NEW stream      -> getType=stream events=write
D2 addGroup on EXISTING stream -> getType=stream events=(none)
```

Fixed in 7eda4e2. Both sentences now carry the `!existing` qualifier: "`markCommitted` to persist an in-place change to an **already-existing** key without dirtying — creating the key dirties either way", and the caveat is scoped to "where `markCommitted` does suppress". Tightened the `KeyspaceMutationTracker` caveat in `keyspace.ts` the same way, since it had the same unqualified "suppresses outright" phrasing right after the sentence that gets it right.

tests/keyspace-update.test.ts
fatal10110: Thank you for mutation-testing it — that is a sharper characterisation than mine. I said "loses direct coverage"; the accurate statement is that the mutant is now invisible to all 321 unit tests with no wire-level backstop, where the pre-port test killed it. Fair correction.

Took both of your options in 7eda4e2, since you were right that the placement was the worse half:

1. Annotated the arm at `src/state/database.ts` — that it is unreachable today (`update` private, `updateTyped` its only caller, five wrappers none of which passes `'string'`), that it is kept because the switch has no `default` so deleting it breaks the build, and the failure mode if it goes live: a future `updateString` or a re-widened `update` turning `SET k ""` into a delete.
2. Moved the test into its own `describe('RedisDatabase.set — empty values')` block. You were right that leaving a `setString`/`getString` round-trip under a `describe('RedisDatabase.update — ...')` heading claimed coverage it does not have. Its comment now also says explicitly that it does **not** cover the `case 'string'` arm.

Also noted for the record: your check that tests 2 and 3 are still on the unmarked path via `deleteField`'s early `false`, and that `list.trim(1, 0)` hits the `start > stop` branch, were the two things I most wanted a second pair of eyes on in the port.

src/state/keyspace.ts
fatal10110: All three fixed in 7eda4e2.

1. `keyspace.ts` header now says the behavior lives on `RedisDatabase` and that `KeyspaceMutationTracker` is internal to its private `update`, with a pointer to the `Tracked*` wrappers — so nobody goes hunting for a public `update` to pass a tracker into.
2. `docs/ARCHITECTURE.md` now reads "a mutation tracker owned by `RedisDatabase.update`".
3. `AGENTS.md` was indeed the stale one, and you are right that it contradicted the caveat I added in the same commit. It now states that the bus drives both `WATCH` and keyspace notifications via `KeyspaceNotifier` today, and names the one-bus-for-two-signals divergence with a pointer to #379 — so it reinforces the caveat comments instead of undercutting them. Agreed on leaving L337 alone; that paragraph opens with "Each database _is_ its keyspace", so it reads as the concept.

Good catch on the table padding too — `prettier --write docs/ARCHITECTURE.md` fixed it, a one-line change confined to the row I had edited. `prettier --check` is clean on both md files now, even though CI does not look at them.

PR #376, opened 2026-09-21
refactor: collapse pub/sub session methods into table-driven pair

### LINKED ISSUE #358: refactor: collapse six ClientSession pub/sub methods into a table-driven pair
## Problem
`ClientSession` has six near-identical methods for pub/sub bookkeeping (`src/core/client-session.ts:448-678`):
`subscribePubSubChannels`, `unsubscribePubSubChannels`, `subscribePubSubShardChannels`, `unsubscribePubSubShardChannels`, `subscribePubSubPatterns`, `unsubscribePubSubPatterns`.

They differ only in: which `Map` they use, which broker function they call (`subscribe` / `ssubscribe` / `psubscribe`), the push frame name (`message` / `smessage` / `pmessage`), and which counter goes in the reply. ~230 lines for what is a 3-row table. The same 6 methods are mirrored on the `RedisClientSession` interface (`src/core/redis-context.ts:57-62`), and each new one needs a new field, new getter and new reset branch.

## Proposal
Table-drive it:
```ts
const KINDS = {
  channel: { subscribe: 'subscribe', frame: 'message',  broker: 'subscribe',  count: s => s.pubsubRegularSubscriptionCount },
  shard:   { subscribe: 'ssubscribe', frame: 'smessage', broker: 'ssubscribe', count: s => s.pubsubShardChannelCount },
  pattern: { subscribe: 'psubscribe', frame: 'pmessage', broker: 'psubscribe', count: s => s.pubsubRegularSubscriptionCount },
}
subscribe(kind, targets): RedisResult[]
unsubscribe(kind, targets): RedisResult[]
```
- `RedisClientSession` shrinks from 6 pub/sub methods to 2; `resetPubSub` loops over the three maps.
- Update callers in `src/commands/pubsub.ts`.

## Acceptance
- No behavior change: existing pub/sub unit + integration tests (ioredis/node-redis, mock + real) pass unmodified.
- Net ~-150 lines.

## Notes
Frame ordering in the "unsubscribe all" path (`.reverse()` of insertion order) and the `pubsubRegularSubscriptionCount` vs shard count differences must be preserved.

fatal10110:
## What changed

`ClientSession` carried six near-identical pub/sub methods (`subscribePubSubChannels`, `unsubscribePubSubChannels`, and the shard / pattern variants of each). They were the same bookkeeping loop written three times, differing only in the `Map` field, the broker entry point (`subscribe` / `ssubscribe` / `psubscribe`), the push frame name (`message` / `smessage` / `pmessage`), the confirmation frame name, and which counter went into the reply.

- Added a `PUBSUB_KINDS` const table with one row per kind and a `PubSubKind` type (`'channel' | 'shard' | 'pattern'`, exported from `src/core/redis-context.ts` and re-exported from `src/internal.ts`).
- Collapsed the six methods into `pubsubSubscribe(kind, targets)` / `pubsubUnsubscribe(kind, targets)`.
- Replaced the three private maps with a single `Record<PubSubKind, Map<string, PubSubRegistration>>`, so `resetPubSub()` is now a loop over its values and the `pubsub*Count` getters read from it.
- Shrunk the `RedisClientSession` interface from six pub/sub methods to two.
- Updated the callers in `src/commands/pubsub.ts` to pass the kind.

## Breaking change to the published `js-redis-server/core` surface

`RedisClientSession` is exported from `src/internal.ts`. External implementors (test doubles) and callers (custom commands) need this mapping:

```
subscribePubSubChannels(ch)        -> pubsubSubscribe('channel', ch)
unsubscribePubSubChannels(ch)      -> pubsubUnsubscribe('channel', ch)
subscribePubSubShardChannels(ch)   -> pubsubSubscribe('shard', ch)
unsubscribePubSubShardChannels(ch) -> pubsubUnsubscribe('shard', ch)
subscribePubSubPatterns(p)         -> pubsubSubscribe('pattern', p)
unsubscribePubSubPatterns(p)       -> pubsubUnsubscribe('pattern', p)
```

Also removed from `ClientSession` (not on the interface, but public via `/core`):

```
pubsubRegularSubscriptionCount  -> no replacement; it was channels + patterns.
                                   Use pubsubChannelCount + pubsubPatternCount.
```

The package is at `0.3.0` with no `CHANGELOG.md`; this section is the release note. `PubSubKind` is exported because it appears in the signature of the published interface — declaration emit requires it.

## Behavior

Unchanged — wire output is byte-identical. Deliberately preserved:

- `pmessage` carries three items (pattern, channel, message); `message` / `smessage` carry two.
- Channel and pattern confirmations report the combined regular count; shard confirmations report the shard count. The two counters are **not** unified.
- The no-argument unsubscribe path returns a single frame with a `null` bulk-string name plus the current count when nothing is subscribed (including the cross-kind count: 2 patterns + 0 channels replies `:2`).
- `refreshPubSubMode()` is still called per target inside the loop, before the frame is constructed.
- Buffers are still defensively copied on their way into a frame (now via the `bulk()` helper).

### Emission order of a bare UNSUBSCRIBE

`.reverse()` is carried over unchanged from `main`. It is **not** a Redis guarantee — Redis iterates the client's subscription dict.

The mechanism, verified locally against Redis 7.2.1: the order is **stable within one `redis-server` process and different after every restart**, because the dict hash seed is randomized at startup. Three probes per process, three processes:

| server start | run 1 | run 2 | run 3 |
| --- | --- | --- | --- |
| #1 | `k6 k3 k4 k1 k2 k5` | identical | identical |
| #2 | `k3 k2 k1 k4 k5 k6` | identical | identical |
| #3 | `k3 k1 k2 k6 k5 k4` | identical | identical |

Reverse insertion order would be `k6 k5 k4 k3 k2 k1` in every one. Reverse insertion order remains a fine deterministic choice for a mock, and the code is unchanged; the doc comment now documents the mechanism (rather than quoting one unreproducible sequence) and warns against pinning the order in a test.

## Line-count delta

`git diff --stat origin/main`:

```
 src/commands/pubsub.ts                            |  12 +-
 src/core/client-session.ts                        | 408 ++++++++++------------
 src/core/redis-context.ts                         |  15 +-
 src/internal.ts                                   |   1 +
 tests-integration/raw-tcp/pubsub-protocol.test.ts |  25 ++
 tests/commands-foundation.test.ts                 |   4 +-
 6 files changed, 235 insertions(+), 230 deletions(-)
```

Net **+5 lines** (−66 before review; two rounds of review fixes added doc comments, 25 lines of test coverage, and the per-kind frame-name types). The issue estimated ~−150 — the difference is almost entirely JSDoc explaining the per-kind differences and the wire-compat invariants, in the style of the rest of the file.

## Review fixes

- **Corrected a false Redis claim.** The "reverse insertion order" doc comment asserted a contract Redis does not offer. Ground-truthed and rewritten (see above). Code unchanged.
- **Pinned the frame names per kind.** They were bare `string`, then literal unions — but a union still admitted a *crossed* row (`shard` taking `'psubscribe'` compiled clean). They are now derived from a prefix map (`'' | 's' | 'p'`) with each row bound to its own kind, so a crossed row is `Type '"psubscribe"' is not assignable to type '"ssubscribe"'`. Verified both directions with the repo's `tsc`. The generic's default keeps `pubsubCount` and `PUBSUB_KIND_ENTRIES` compiling untouched — no other file changed.
- **Made the `counter` column actually authoritative.** `pubsubRegularSubscriptionCount` hardcoded `channel + pattern` outside the table, so a fourth kind would compile clean and silently under-count. `pubsubCount()` now sums every kind whose row carries that counter, and `pubsubSubscriptionCount` sums all kinds. Behavior identical.
- **`listen` is a `readonly` property**, not method shorthand — restores the strict variance check the shorthand form opts out of, and matches its four siblings.
- **Extended the RESET test** (`tests-integration/raw-tcp/pubsub-protocol.test.ts`) to subscribe all three kinds and assert `PUBLISH` / `SPUBLISH` all return `0` afterwards. Verified it fails when `resetPubSub()`'s loop is narrowed back to one kind. Existing assertions untouched.
- **Fixed the `bulk()` comment** to cover both roles (broker payloads *and* caller-owned command args).

### Second round

- **Renamed to `pubsubSubscribe` / `pubsubUnsubscribe`.** I declined this initially on the belief that the unprefixed names were the repo owner's spec in #358; that premise was wrong — the issue wording was assistant-drafted, so there was no accepted proposal to defer to. The reviewer's own argument stands on its own anyway: this PR already breaks `RedisClientSession` on `/core` and already ships a migration table, so the rename is free now and would cost a second breaking release later.
- **Deleted `pubsubRegularSubscriptionCount`.** Deriving `pubsubCount()` from the table orphaned it — 7 call sites on `main`, 0 after. The review offered deletion or rewiring `pubsubSubscriptionCount` through it; rewiring would reintroduce a hand-enumerated aggregation outside the table, which is exactly what the previous fix removed, so deletion it is. Added to the migration mapping above.
- **Documented the ordering mechanism** instead of quoting one sequence (see above).

### CodeQL

The alert on `pattern.replace('*', 'x')` in the RESET test was a false positive: `pattern` is built as `` `raw-reset-pattern:${randomKey()}:*` `` and `randomKey()` is base-36, so there is exactly one `*` and replacing the first occurrence replaces the only one.

Rather than dismiss the alert, the test now builds both strings from a shared prefix, so there is no rewriting left to analyze. That resolves it legitimately instead of suppressing it — no one has to re-triage the same false positive on the next scan — and it reads better than deriving one string from the other by substitution.

## Test suites run

All pass locally:

| Suite | Result |
| --- | --- |
| `npm run lint` | pass |
| `npm test` (includes `tsc --noEmit`) | pass — 320 tests |
| `npm run build` | pass |
| `npm run test:integration:mock:ioredis` | pass — 475 tests |
| `npm run test:integration:mock:node-redis` | pass — 524 tests |
| `npm run test:integration:raw:mock` | pass — 105 tests |

The `:real` suites were not run locally (they need a live Redis via docker-compose) — leaving those to CI.

No pub/sub assertion was weakened. The only test edits are the two call sites in `tests/commands-foundation.test.ts` that named the removed methods, and the added coverage in the RESET test.

Closes #358


src/core/client-session.ts
```diff
@@ -882,6 +795,18 @@ export class ClientSession implements RedisClientSession {
     }
   }
 
+  /**
+   * Resolves the counter a confirmation frame reports. Channels and patterns
+   * share one count; shard channels are counted on their own.
+   */
+  private pubsubCount(counter: PubSubKindSpec['counter']): number {
```
fatal10110, line 842: **medium — the `counter` column is only half-authoritative**

The table presents `counter` as the single source of truth for which count a kind reports, but `pubsubCount()` dispatches on it and then delegates to a hand-enumerated getter:

```ts
get pubsubRegularSubscriptionCount(): number {
  return this.pubsubChannelCount + this.pubsubPatternCount
}
```

So the aggregation is still hardcoded to channel+pattern, outside the table.

**Failure scenario:** add a fourth kind with `counter: 'regular'`. `Record<PubSubKind, PubSubKindSpec>` and the `Record<PubSubKind, Map<…>>` initializer both force you to add a row, so the compiler reports the change as complete — but `pubsubRegularSubscriptionCount` still omits the new kind's subscriptions from every confirmation frame, and `pubsubSubscriptionCount` under-counts, so `refreshPubSubMode()` never enters subscribed mode. The exhaustive `Record` gives false confidence precisely because this one spot escaped the table.

Deriving it closes the gap: sum `this.pubsubSubscriptions[kind].size` over `Object.entries(PUBSUB_KINDS)` where `spec.counter === counter`, and define `pubsubRegularSubscriptionCount` in terms of that.

src/core/client-session.ts
```diff
@@ -54,6 +56,64 @@ type PubSubRegistration = {
...
+   * every push frame it delivers — `pmessage` carries the matched pattern in
+   * front of the channel, the other two do not.
+   */
+  listen(
+    broker: RedisPubSubBroker,
+    target: Buffer,
+    emit: (items: RedisValue[]) => void,
+  ): Unsubscribe
+  /** Push frame name used to deliver a message to a subscriber. */
+  readonly message: string
```
fatal10110: **medium — frame names typed as bare `string`, so a crossed row is invisible to `tsc`**

`message`, `subscribed` and `unsubscribed` are `readonly …: string`, while `counter` right below them *is* a literal union (`'regular' | 'shard'`). That asymmetry matters here.

Before this PR these frame names were inline string literals at their single use site, where crossing them was physically impossible. Now they are three interchangeable `string` slots in three adjacent, near-identical rows — the classic failure mode a lookup table introduces.

**Failure scenario:** someone copy-pastes the `channel` row to add a kind and forgets to change `subscribed` to `ssubscribe`, or types `'punsubscibe'`. `tsc` is silent; the only thing that catches it is the raw-tcp integration test.

Tightening costs nothing at runtime:

```ts
readonly message: 'message' | 'smessage' | 'pmessage'
readonly subscribed: 'subscribe' | 'ssubscribe' | 'psubscribe'
readonly unsubscribed: 'unsubscribe' | 'sunsubscribe' | 'punsubscribe'
```

src/core/redis-context.ts
```diff
@@ -54,12 +61,8 @@ export interface RedisClientSession {
   readonly pubsubShardChannelCount: number
   readonly pubsubPatternCount: number
   readonly pubsubSubscriptionCount: number
-  subscribePubSubChannels(channels: readonly Buffer[]): RedisResult[]
-  unsubscribePubSubChannels(channels: readonly Buffer[]): RedisResult[]
-  subscribePubSubShardChannels(channels: readonly Buffer[]): RedisResult[]
-  unsubscribePubSubShardChannels(channels: readonly Buffer[]): RedisResult[]
-  subscribePubSubPatterns(patterns: readonly Buffer[]): RedisResult[]
-  unsubscribePubSubPatterns(patterns: readonly Buffer[]): RedisResult[]
+  subscribe(kind: PubSubKind, targets: readonly Buffer[]): RedisResult[]
```
fatal10110: **medium — `subscribe` / `unsubscribe` are the least specific names on this interface**

Every other pub/sub member of `RedisClientSession` is prefixed: `pubsubChannelCount`, `pubsubShardChannelCount`, `pubsubPatternCount`, `pubsubSubscriptionCount`, `resetPubSub()`. These two new methods are the only pub/sub members not marked as such — and they sit two lines below `watch(keys)` / `unwatch()`, a *second* subscription concept on the same interface.

It also collides conceptually with the repo's `Unsubscribe` type from `state/mutation-events`, used for `WatchRegistration.unsubscribe` and `PubSubRegistration.unsubscribe` in `client-session.ts` itself. At a call site, `session.unsubscribe(...)` reads as "detach a listener" rather than "run the UNSUBSCRIBE command".

No compiler collision today, but `pubsubSubscribe(kind, targets)` / `pubsubUnsubscribe(kind, targets)` keeps the entire collapse (still −66 lines) and keeps the surface coherent.

src/core/client-session.ts
```diff
@@ -445,252 +511,99 @@ export class ClientSession implements RedisClientSession {
...
-            this.pubsubShardChannels.values(),
-            entry => entry.value,
-          ).reverse()
-
-    if (targets.length === 0) {
+  /**
+   * UNSUBSCRIBE / SUNSUBSCRIBE / PUNSUBSCRIBE bookkeeping for one kind.
+   *
+   * With no targets Redis drops every current subscription of that kind, in
+   * reverse insertion order, and replies with a single nil-named frame when
```
fatal10110: **low, but it is the one thing this PR newly asserts that is wrong — flagged by both the wire-compat and architecture lenses**

> With no targets Redis drops every current subscription of that kind, in reverse insertion order

Real Redis makes no such guarantee. It iterates the client's `pubsub_channels` dict, so emission order is hash order. Ground-truthed against **Redis 7.2.1**: subscribed 6 of each kind in order, then sent bare `UNSUBSCRIBE` / `PUNSUBSCRIBE` / `SUNSUBSCRIBE`:

```
subscribed: k1 k2 k3 k4 k5 k6 | q1* q2* q3* q4* q5* q6* | t1 t2 t3 t4 t5 t6
emitted:    k2 k3 k1 k4 k5 k6 | q2* q1* q5* q3* q4* q6* | t4 t1 t6 t2 t3 t5
```

Reverse insertion order would be `k6 k5 k4 k3 k2 k1`. A separate 3-channel run gave `c2 c3 c1`.

The `.reverse()` on the line below is unchanged from `origin/main` and is a perfectly defensible deterministic choice for a mock — the problem is only that this new comment promotes an acknowledged approximation into a documented Redis contract. In a repo whose whole point is wire fidelity, that is how a future contributor ends up writing a test that pins "reverse insertion order" as correct Redis behavior.

Suggest: "…in reverse insertion order (Redis leaves this order unspecified; we pick a deterministic one)."

src/core/redis-context.ts
```diff
@@ -54,12 +61,8 @@ export interface RedisClientSession {
...
+  unsubscribe(kind: PubSubKind, targets: readonly Buffer[]): RedisResult[]
```
fatal10110: **low — breaking change to a published type, with no release note**

`RedisClientSession` is exported from `src/internal.ts`, published as the `js-redis-server/core` subpath under the header "Deep internals — building blocks for power users". Dropping six methods from it breaks any external implementor (a test double) or caller (a custom command doing `ctx.session.subscribePubSubChannels(args.channels)`) at compile time.

The package is at `0.3.0` and there is no `CHANGELOG.md`, so at minimum the PR description / release notes should carry the full mapping:

```
subscribePubSubChannels(ch)        -> subscribe('channel', ch)
unsubscribePubSubChannels(ch)      -> unsubscribe('channel', ch)
subscribePubSubShardChannels(ch)   -> subscribe('shard', ch)
unsubscribePubSubShardChannels(ch) -> unsubscribe('shard', ch)
subscribePubSubPatterns(p)         -> subscribe('pattern', p)
unsubscribePubSubPatterns(p)       -> unsubscribe('pattern', p)
```

src/core/client-session.ts
```diff
@@ -445,252 +511,99 @@ export class ClientSession implements RedisClientSession {
...
-
-    const frames: RedisResult[] = []
-    for (const pattern of targets) {
-      const key = pubsubId(pattern)
-      const existing = this.pubsubPatterns.get(key)
-      if (existing) {
-        existing.unsubscribe()
-        this.pubsubPatterns.delete(key)
+  resetPubSub(): void {
+    for (const registrations of Object.values(this.pubsubSubscriptions)) {
```
fatal10110, line 637: **low — this rewrite is correct, but shard/pattern teardown is unpinned by any test**

`Object.values(this.pubsubSubscriptions)` does correctly cover all three kinds (verified: it yields `Map<string, PubSubRegistration>[]` over the three-key `Record`). The gap is coverage.

`resetPubSub()` is the only pub/sub teardown path — reached from `RESET` (`src/commands/connection.ts:871`) and from `close()` — and the only test exercising it is **`tests-integration/raw-tcp/pubsub-protocol.test.ts:249`** ("RESET exits subscribed mode and removes all subscriptions"), which issues only `SUBSCRIBE channel` → `RESET` → `PUBLISH`. No RESET test anywhere issues `PSUBSCRIBE` or `SSUBSCRIBE`.

**Failure scenario:** a future edit narrows this loop back to one kind (`this.pubsubSubscriptions.channel`). Shard/pattern broker listeners then survive RESET and connection close — the broker's subscriber map retains the closed session forever (leak), and a later `PUBLISH`/`SPUBLISH` enqueues pushes onto a dead session. The whole suite still passes.

Adding `PSUBSCRIBE` + `SSUBSCRIBE` to that same test and asserting `PUBLISH`/`SPUBLISH` both return `0` afterwards is ~10 lines and pins this line.

For the record, the rest of the refactor *is* well covered by that file: the counter split at `:136`, the empty-unsubscribe nil frames for all three kinds at `:398`, per-kind frame names throughout.

src/core/client-session.ts
```diff
@@ -54,6 +56,64 @@ type PubSubRegistration = {
...
+ * `SUBSCRIBE`/`UNSUBSCRIBE` families are otherwise identical bookkeeping, so
+ * they are driven from this table instead of being written out per kind.
+ */
+type PubSubKindSpec = {
+  /**
+   * Registers the session's listener with the broker and builds the items of
+   * every push frame it delivers — `pmessage` carries the matched pattern in
+   * front of the channel, the other two do not.
+   */
+  listen(
```
fatal10110: **low — `listen` is a method shorthand while its four siblings are `readonly` properties**

Beyond the inconsistency, method shorthand opts into **bivariant** parameter checking, which silently discards the contravariance check `"strict": true` otherwise gives you (`strictFunctionTypes` does not apply to method-shorthand declarations). A `listen` whose `emit` parameter were narrowed incompatibly would type-check.

Cosmetic today since the table is the only implementor, but the property form matches the siblings and restores strict variance:

```ts
readonly listen: (
  broker: RedisPubSubBroker,
  target: Buffer,
  emit: (items: RedisValue[]) => void,
) => Unsubscribe
```

src/core/client-session.ts
```diff
@@ -936,6 +861,11 @@ function pubsubId(value: Buffer): string {
   return value.toString('hex')
 }
 
+/** Defensive copy — broker payloads must never alias into a reply frame. */
```
fatal10110: **low — the comment describes one of `bulk()`'s two roles**

`bulk()` is called from the three `listen` closures in `PUBSUB_KINDS` (broker payloads — what this comment describes), but *also* on caller-owned `SUBSCRIBE`/`UNSUBSCRIBE` targets in `subscribe()` and `unsubscribe()`, where the invariant is different: the parsed command args are owned by the caller and may alias the decoder buffer.

A maintainer who later reasons "subscribe targets aren't broker payloads, so `bulk` is overkill here" would silently drop a defensive copy.

Suggest: "Defensive copy — neither broker payloads nor caller-owned command args may alias into an emitted frame."

src/core/client-session.ts
```diff
@@ -165,23 +241,28 @@ export class ClientSession implements RedisClientSession {
...
-    return this.pubsubShardChannels.size
+    return this.pubsubSubscriptions.shard.size
   }
 
   get pubsubPatternCount(): number {
-    return this.pubsubPatterns.size
+    return this.pubsubSubscriptions.pattern.size
   }
 
   get pubsubRegularSubscriptionCount(): number {
```
fatal10110: **low — new in this commit: the `pubsubCount()` rework orphaned this getter**

Surfaced independently by both review lenses. The fix for the `counter` finding is correct, but it removed this getter's last consumer and left the getter behind.

On `origin/main` `pubsubRegularSubscriptionCount` had 7 call sites (`client-session.ts:184, 477, 499, 517, 632, 654, 672`). After this commit:

```
$ grep -rn pubsubRegularSubscriptionCount src tests tests-integration tests-package examples docs
src/core/client-session.ts:255:  get pubsubRegularSubscriptionCount(): number {
```

Definition only. `pubsubCount('regular')` now derives the total from the table directly, and `pubsubSubscriptionCount` sums `Object.values(this.pubsubSubscriptions)` rather than going through here.

It is also **not** on the `RedisClientSession` interface — `redis-context.ts:60-63` lists only `pubsubChannelCount` / `pubsubShardChannelCount` / `pubsubPatternCount` / `pubsubSubscriptionCount` — but it *is* public API, because `ClientSession` itself is exported from `src/internal.ts:80`. So this is unreferenced, untested surface on the published `js-redis-server/core` subpath.

Delete it, or have `pubsubSubscriptionCount` consume it (`pubsubRegularSubscriptionCount + pubsubShardChannelCount`) so it stays on a live path. The other three per-kind getters are fine as-is — they genuinely feed `sub=` / `psub=` in `src/commands/connection.ts:258-259`.

src/core/client-session.ts
```diff
@@ -445,252 +526,105 @@ export class ClientSession implements RedisClientSession {
...
+   *
+   * With no targets Redis drops every current subscription of that kind and
+   * replies with a single nil-named frame when there was nothing to drop.
+   *
+   * Real Redis emits those frames in the iteration order of the client's
+   * subscription dict — effectively hash order, and *not* a documented
+   * contract: subscribing `k1..k6` to Redis 7.2.1 and sending a bare
+   * UNSUBSCRIBE came back as `k4 k2 k1 k5 k6 k3`. We emit in reverse insertion
+   * order instead, purely because a mock should be deterministic. Do not pin
+   * this order in a test as though it were Redis behavior.
```
fatal10110: **low — the correction is right; the evidence it cites is the one part that will not survive contact with a reader**

The claim is now accurate and the "do not pin this order" guidance is good — that part is resolved. Flagging only the framing, and both lenses landed on the same spot independently.

**Third independent run, Redis 7.2.1** (subscribe `k1..k6` in order, then bare unsubscribe):

```
channel  =>  k2 k4 k3 k6 k5 k1        counts: 5 4 3 2 1 0
pattern  =>  q3* q2* q1* q6* q5* q4*  counts: 5 4 3 2 1 0
shard    =>  t3 t2 t6 t1 t4 t5        counts: 5 4 3 2 1 0
```

So three runs, three different orders — yours `k4 k2 k1 k5 k6 k3`, the first review's `k2 k3 k1 k4 k5 k6`, this one `k2 k4 k3 k6 k5 k1`. Your reading is confirmed.

But the run also turned up the *mechanism*, which is the thing actually worth writing down: **repeating the probe against the same `redis-server` process gave byte-identical output every time, while a restart gave a different order.** That is the dict hash seed being randomized at process start. It holds for all three kinds, consistent with 7.x where `pubsub_channels`, `pubsub_patterns` (list→dict in 7.0) and `pubsub_shard_channels` are all dicts.

The comment currently spends its evidence budget on a single observed sequence. Nobody can reproduce `k4 k2 k1 k5 k6 k3` — the next person who tries will get a fourth order and may conclude the comment is simply wrong. Citing the mechanism is both shorter and checkable:

> Real Redis emits those frames in the iteration order of the client's subscription dict. That order is stable within one `redis-server` process and changes across restarts, because the dict hash seed is randomized at startup — three separate runs of this experiment produced three different orders. It is not a documented contract. We emit in reverse insertion order instead, purely because a mock should be deterministic. Do not pin this order in a test as though it were Redis behavior.

For the record, the warning is currently honored by the suite: no test pins multi-target bare-unsubscribe ordering. `tests-integration/raw-tcp/pubsub-protocol.test.ts:201-213` has exactly one subscription left per registry at that point, and `:341-349` sorts the names before asserting.

src/core/client-session.ts
```diff
@@ -54,6 +56,74 @@ type PubSubRegistration = {
...
+   */
+  readonly listen: (
+    broker: RedisPubSubBroker,
+    target: Buffer,
+    emit: (items: RedisValue[]) => void,
+  ) => Unsubscribe
+  /** Push frame name used to deliver a message to a subscriber. */
+  readonly message: 'message' | 'smessage' | 'pmessage'
+  /** Confirmation frame name echoed back by SUBSCRIBE/SSUBSCRIBE/PSUBSCRIBE. */
+  readonly subscribed: 'subscribe' | 'ssubscribe' | 'psubscribe'
```
fatal10110: **low, non-blocking — answering the caveat you volunteered: the crossed row is closable in four lines, and the ripple you would expect does not happen**

You flagged that the unions catch typos but still admit a crossed row, and judged that pinning per kind would mean restating the table in the type system. First half confirmed, second half turns out not to be true — so here is the measurement rather than an opinion.

**The crossed row is real.** Took your table, changed the `shard` row's `subscribed` to `'psubscribe'`, ran the repo's `tsc --noEmit`:

```
$ tsc --noEmit
$ echo $?
0
```

Clean. Silent.

**A prefix map closes it without restating anything.** The three families *are* a prefix scheme — `''` / `'s'` / `'p'` — so the nine literals derive from three:

```ts
type PubSubFramePrefix = { channel: ''; shard: 's'; pattern: 'p' }

type PubSubKindSpec<K extends PubSubKind = PubSubKind> = {
  readonly listen: (...) => Unsubscribe
  readonly message: `${PubSubFramePrefix[K]}message`
  readonly subscribed: `${PubSubFramePrefix[K]}subscribe`
  readonly unsubscribed: `${PubSubFramePrefix[K]}unsubscribe`
  readonly counter: 'regular' | 'shard'
}

const PUBSUB_KINDS: { readonly [K in PubSubKind]: PubSubKindSpec<K> } = { /* table body unchanged */ }
```

Applied to the actual file, table body untouched:

```
$ tsc --noEmit          # correct table
$ echo $?
0

$ # shard row -> subscribed: 'psubscribe'
$ tsc --noEmit
src/core/client-session.ts(103,5): error TS2322: Type '"psubscribe"' is not assignable to type '"ssubscribe"'.
```

**The ripple is what I expected to kill it, and it did not materialize.** The `= PubSubKind` default keeps `PubSubKindSpec` usable bare, so `pubsubCount(counter: PubSubKindSpec['counter'])` at `:825` compiles untouched, as does `PUBSUB_KIND_ENTRIES` at `:122`. Whole-repo `tsc --noEmit` passes with **zero** other edits.

Net: +1 type alias, 3 changed lines, one generic parameter with a default. It also subsumes the typo case you already fixed (`'punsubscibe'` → `Type '"punsubscibe"' is not assignable to type '"punsubscribe"'`).

Genuinely your call — the raw-tcp suite does cover the crossed case, so this is belt-and-braces and I would not hold the PR for it. But "restating the table" was the stated reason for declining, and on the evidence it does not.

fatal10110:
## Re-review of `d6c0b9f` — verdict: **ship**

Re-ran the full review against the new head: both specialist lenses (Redis wire-compat, Node architecture) plus independent verification of the two claims that carried the most weight. The fixes hold. Three new low-severity comments posted, none blocking.

### Prior findings

| # | Finding | Status |
|---|---|---|
| 1 | False Redis claim on unsubscribe order | **Fixed** — claim is now correct; one framing nit posted |
| 2 | Frame names typed as bare `string` | **Fixed** — residual crossed-row risk is real but covered; see note |
| 3 | `counter` only half-authoritative | **Fixed** — verified live, not by inspection |
| 4 | `listen` method shorthand vs `readonly` property | **Fixed** |
| 5 | RESET teardown unpinned for shard/pattern | **Fixed** — independently confirmed as a real pin, below |
| 6 | `bulk()` comment covered one of two roles | **Fixed**, with a third role still unnamed — see note |
| 7 | Breaking change to published surface, no release note | **Fixed** — mapping is in the PR body |
| 8 | `pubsubSubscribe` / `pubsubUnsubscribe` rename | **Declined** — my view below |

### The RESET pin — independently confirmed red

Not taking the summary's word for it. Narrowed `resetPubSub()`'s loop and ran `tests-integration/raw-tcp/pubsub-protocol.test.ts` on the PR head:

```
resetPubSub narrowed to channel-only : 8 pass / 1 fail   <- "RESET exits subscribed mode and removes all subscriptions"
resetPubSub narrowed to shard-only   : 8 pass / 1 fail
resetPubSub narrowed to pattern-only : 8 pass / 1 fail
```

Then the sharper question — is each kind *individually* pinned, or is the pre-existing channel assertion doing all the work? Dropped one kind at a time instead:

```
resetPubSub drops only pattern : 8 pass / 1 fail
resetPubSub drops only shard   : 8 pass / 1 fail
resetPubSub drops only channel : 8 pass / 1 fail
```

All three independently caught. And the control that makes it conclusive — the **pre-fix** test file against the same channel-only narrowing:

```
9 pass / 0 fail
```

The old test sailed straight through the exact leak it was supposed to catch. The 25 added lines are load-bearing, not decoration.

**Every test edit in this PR audited for weakened assertions**, given the stated concern. `tests-integration/raw-tcp/pubsub-protocol.test.ts` is **+25 −0** against `main` — a single purely-additive hunk, zero deleted lines. All four pre-existing assertions survive verbatim (`['subscribe', channel, 1]`, `'RESET'`, `PUBLISH → 0`, `GET still-normal → null`), and the new ones are `deepStrictEqual` on exact values. The only other test change in the PR is `tests/commands-foundation.test.ts` **+2 −2**, a mechanical rename of two call sites with the assertions untouched. Nothing was loosened.

### The counter rework — probed hardest, clean

This was the most likely place for an accidental behavior change. Drove the mock and a real Redis 7.2.1 through byte-identical probes; the full 8-case counter comparison `diff`s **identical**. The cases that matter:

```
psubscribe after subscribe      -> :2    (patterns continue the regular counter)
ssubscribe alongside 2 regular  -> :1    (shard counter independent)
2 patterns + 0 channels, bare UNSUBSCRIBE   -> ['unsubscribe',  nil, 2]   (not :0)
2 channels + 0 patterns, bare PUNSUBSCRIBE  -> ['punsubscribe', nil, 2]
2 regular + 0 shards,    bare SUNSUBSCRIBE  -> ['sunsubscribe', nil, 0]
interleaved: sub 1, ssub 1, psub 2, sub 3, ssub 2, psub 4
teardown:    unsub 3, sunsub 1, punsub 2
```

The cross-kind empty case is exactly the one the rework could have broken, and `pubsubCount('regular')` summing table rows reproduces it. No regression in anything the first review ground-truthed: all six confirmation frame names, `pmessage` as `*4` vs `message`/`smessage` as `*3`, RESP2/RESP3 push encoding (`>3`/`>4`), `refreshPubSubMode()` in all three positions, every defensive `Buffer.from`, `resetPubSub()` wired to both `RESET` and `close()`, zero stale references to the six removed methods, and `PubSubKind`'s export still required for declaration emit.

Suites on the PR head: unit 320/320, raw-tcp 105/105, ioredis 475/475, node-redis 524/524, `tsc --noEmit` and prettier clean.

### On the naming (`subscribe` → `pubsubSubscribe`)

You declined on the grounds that #358 specifies these names. Worth saying plainly: **that issue text was assistant-drafted, not an owner decision** — there is no accepted proposal to defer to here, so the argument reduces to the merits. Judging it on those:

I'd rename, and the reason is narrower than the style argument the first review made. **This PR is already a breaking change to `RedisClientSession` on the published `js-redis-server/core` subpath.** You're paying that break exactly once, right now, and the PR body already carries the migration table. Renaming later means paying it a second time for a cosmetic gain nobody will think is worth it — so this is the only cheap moment the rename will ever have. The cost today is one interface line, one class line, six call sites in `src/commands/pubsub.ts`, two in `tests/commands-foundation.test.ts` — all files this PR already touches.

The supporting case still stands: `watch(keys)` / `unwatch()` sit two lines above on the same interface, every other pub/sub member carries the `pubsub` marker, and the collision is concrete rather than theoretical — at `client-session.ts:603`, `existing.unsubscribe()` (a `state/mutation-events` `Unsubscribe` closure) is invoked *inside a method also named `unsubscribe`*.

Strength: **worth doing, not blocking.** If you'd rather keep the names, that's a defensible call and I wouldn't hold the PR — but "#358 says so" shouldn't be what decides it.

### New, from this round (all low, all posted inline)

1. **`pubsubRegularSubscriptionCount` is now dead code.** The `pubsubCount()` rework removed its last consumer; it had 7 call sites on `main` and has 0 now. Both lenses flagged it independently. It's the one genuine over-correction — a two-line cleanup.
2. **The unsubscribe-order comment quotes an unreproducible measurement.** The conclusion is right; a third run gave a *third* order (`k2 k4 k3 k6 k5 k1`). That run also pinned down the mechanism worth documenting instead: the order is stable within one `redis-server` process and changes across restarts, because the dict hash seed is randomized at startup.
3. **The crossed-row risk you volunteered is closable in ~4 lines**, and the ripple into `pubsubCount` that would have made it not worth it doesn't actually occur (a default type parameter keeps `PubSubKindSpec` usable bare). Verified both directions with the repo's own `tsc`. Still your call — the raw-tcp suite does cover the case.

Two more, not worth their own threads: the `bulk()` comment now names broker payloads and caller-owned args but not the third role — the session's own stored registration buffers, which reach `bulk(target)` via the bare-UNSUBSCRIBE path at `:586` → `:610` and are the longest-lived aliasing risk of the three. And the `PUBSUB_KIND_ENTRIES` cast at `:122` is sound but leaves two different iteration mechanisms over the same table (`Object.entries` in `pubsubCount`, `Object.values(this.pubsubSubscriptions)` in the other three loops); co-locating each spec with its registration map would remove the cast and unify them. Neither is worth churn on its own.

The CodeQL hit on `pattern.replace('*', 'x')` is a **false positive**: `randomKey()` is `Math.random().toString(36).substring(2, 10)`, base-36 only, so the constructed pattern contains exactly one `*`. The substitution is also deliberately correct — `...:x` *does* match the pattern, so the `→ 0` assertion proves the pattern subscription was dropped rather than passing vacuously.

PR #378, opened 2026-09-21
fix: route node-redis mock cluster commands by real command keys

### LINKED ISSUE #363: fix/refactor: node-redis mock mis-routes multi-key commands; dedupe reply decoders
## Problem
`src/client-mocks/node-redis-mock.ts` (1042 lines) and `src/in-memory-client.ts` (405 lines) duplicate logic and, in one place, re-implement something the pipeline already provides.

1. **Cluster routing heuristic**: `extractRoutingKeys` + `MULTI_KEY_COMMANDS` (`node-redis-mock.ts:~870-900`) guess routing keys by command name (first arg unless the command is in a hard-coded set). Commands like `MSET`, `EVAL`, `ZUNIONSTORE`, `SINTERSTORE`, `RENAME`, `BLPOP`... are routed by first argument only — wrong slot / missing CROSSSLOT detection. The executor already knows the truth: `executor.plan(name, args).keys` (the same call `ClusterPolicy` uses). Replace the heuristic with it (this is a latent **bug fix**, not just cleanup).
2. **Duplicated reply decoding**: `decodeReply`/`decodeKey`/`toBuffer` in `node-redis-mock.ts:936-1000` vs `InMemoryRedisClient.decode`/`decodeKey`/`toBuffer` in `in-memory-client.ts:212-270,380-400`. Differences are small (push shape, bigint policy, error constructor). Extract one `decodeRedisValue(value, opts)` with those three knobs.
3. **Emitter boilerplate**: `on/once/off` wrappers around a private `EventEmitter` appear on both `NodeRedisMockClient` and `NodeRedisMockCluster`. Extend `EventEmitter` instead.

## Acceptance
- Add a node-redis-mock cluster test with a cross-slot `MSET` / `ZUNIONSTORE` that currently mis-routes, expecting `CROSSSLOT` (red before, green after).
- Existing `tests-integration/node-redis` mock suite passes.
- ~-130 lines.

fatal10110:
## Part 1 — the routing bug

`NodeRedisMockCluster` picked the node for every command with a hand-rolled
heuristic: the whole argument list for a nine-entry hard-coded
`MULTI_KEY_COMMANDS` set, and otherwise **the single argument right after the
command name**. That guess is wrong for every command whose first argument is
not a key, and incomplete for every multi-key command outside the set.

Concrete mis-routings, all through the facade's cluster client:

| invocation | keys the heuristic used | keys the command actually has | consequence |
| --- | --- | --- | --- |
| `EVAL "<script>" 1 mykey` | the **script text** | `mykey` | routed to whichever node owns the script text's slot → `-MOVED`, the script never runs |
| `ZDIFF 2 a b`, `LMPOP 2 a b LEFT`, `SINTERCARD 2 a b` | `"2"` | `a`, `b` | routed by the numkeys literal |
| `BITOP AND dst src` | `"AND"` | `dst`, `src` | routed by the operation name |
| `MSET k1 v1 k2 v2` | `k1` | `k1`, `k2` | routed by one key while spanning several |
| `RENAME a b`, `SINTERSTORE dst a b`, `COPY src dst` | first key only | both | same |
| `ZUNIONSTORE dest 2 a b` | `dest` | `dest`, `a`, `b` | sources outside the destination's slot never noticed client-side |
| `GEORADIUS key … STORE dst` | `key` | `key`, `dst` | STORE target ignored for routing |

The executor already computes this exactly, and it is the same extraction
`ClusterPolicy` routes on — so ask it: `executor.plan(name, args).keys`.

Preserved behaviour:

- **CROSSSLOT**: a multi-slot key set still surfaces node-redis' `ErrorReply`
  with `CROSSSLOT Keys in request don't hash to the same slot` — now raised by
  `ClusterPolicy` on the node the command reaches, rather than reconstructed
  client-side (see the SORT fix below).
- **Keyless commands** (slot `null`) still run on the first master.
- **Unplannable commands** (unknown name, bad arity) have no keys to route on,
  so they go to the first master and the normal pipeline produces the canonical
  error reply, exactly as the standalone client does. No second error path, no
  internal crash out of `sendCommand`.

### The review caught a regression — fixed

Routing by `plan().keys` was right; pairing it with the facade's **own**
client-side `slot === -1` refusal was not. `SORT`'s routing keys deliberately
carry its BY/GET **patterns** (`sortRoutingKeys`, `src/commands/keys.ts`) so
that `ClusterPolicy.getSortClusterPatternError` can check them. Those patterns
are not keys, so they usually hash elsewhere — and the facade's CROSSSLOT branch
fired before the command reached any node, swallowing the SORT-specific error.

Reproduced against `origin/main`, this branch, and a local cluster-enabled
Redis 7.2.1:

```
main    SORT mylist BY w_*  ->  ERR BY option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.
branch  SORT mylist BY w_*  ->  CROSSSLOT Keys in request don't hash to the same slot     <- regression
real    SORT mylist BY w_*  ->  ERR BY option of SORT denied in Cluster mode.
```

```
$ redis-cli COMMAND GETKEYS SORT k BY 'w_*' GET 'p_*' GET '#'
k
```

The fix is subtraction, not a SORT special case: the `slot === -1` branch is
gone. Now that routing keys match what `ClusterPolicy` computes, the facade's
own CROSSSLOT construction is redundant — slot `-1` owns no node, falls through
to the first master, and that node's policy recomputes the same keys and raises
the identical error, which `decodeReply` surfaces as a byte-identical
`ErrorReply`. Maintaining a second copy of rules the policy already owns is the
very per-command special-casing this PR set out to delete. After the fix all
five probes match `origin/main` byte-for-byte.

### Red-then-green evidence

**The earlier evidence was overstated and the review was right to say so.** The
two cross-slot tests I added did *not* guard the routing fix: whichever node a
bad router picks rejects the span anyway, so `/CROSSSLOT/` matched either way.
Only a **wrong-node** assertion can discriminate. They are replaced.

Five routing cases, each a command whose first argument is not a key, each run
against one key per master. With 3 masters the slot space splits into
`[0,5460] / [5461,10921] / [10922,16383]`, and `alpha` (865), `abc` (7638) and
`k1` (12706) sit one per range — so whatever slot a command's first argument
hashes to, it covers at most one of the three, and a first-argument router
mis-routes at least two of every three runs.

Against the restored `origin/main` heuristic:

```
not ok 4 - ZDIFF routes by its keys, not its first argument
      error: 'MOVED 865 127.0.0.1:0'
not ok 5 - LMPOP routes by its keys, not its first argument
not ok 6 - SINTERCARD routes by its keys, not its first argument
not ok 7 - BITOP routes by its keys, not its first argument
not ok 8 - EVAL routes by its declared KEYS, not by the script text
# pass 25
# fail 5
```

Previously only the `EVAL` case failed there. With the fix: **32/32 pass**.

The SORT test fails against the state this PR's second commit fixes:

```
not ok 9 - SORT BY/GET patterns keep the cluster-mode SORT error
      + actual   "CROSSSLOT Keys in request don't hash to the same slot"
      - expected 'ERR BY option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.'
```

The CROSSSLOT tests remain as **error-wording** guards — tightened from
`/CROSSSLOT/` to the exact message, and commented in the file as not being
routing guards.

### Why the red test is not in `tests-integration/`

That suite drives the **real** node-redis client over TCP against
`createRedisCluster`, so `NodeRedisMockCluster` is never instantiated and
`routingKeys()` is not in the loop — it cannot see this bug. The facade's tests
live in `tests/client-mocks/`, whose own header explains why ("the facade IS the
client surface"). The two cross-slot integration tests are kept only to pin the
per-command CROSSSLOT wording against real Redis under `TEST_BACKEND=real`, and
now say so in a comment. **They should be verified against the real backend in
CI**; the `:real` suites were not run here (docker-compose cluster).

## Part 2 — the dedupe

New `src/core/decode-redis-value.ts` holds `decodeRedisValue`,
`decodeRedisKey`, `toRedisArgument` and `redisErrorText` once. The three real
differences survive as options, not flattened:

- `pushShape`: `'items'` (node-redis routes a push by its type tag and hands
  listeners only the payload) vs `'tagged'` (`[name, ...items]`).
- `narrowBigInt`: `'always'` vs `'when-safe'`.
- `error`: node-redis' `ErrorReply` vs `RedisCommandError`.

Plus `returnBuffers`, used only by the in-memory client.

`CommandRunner` now extends `EventEmitter` — which is what a real node-redis
client is — replacing the duplicated `on`/`once`/`off` wrappers on both clients.

### The divergence test did not do what its header claimed

Re-review caught that `tests/core/decode-redis-value.test.ts` promised a
tripwire it did not install: it re-declared the two option objects as **local
literals**, so it described what `decodeRedisValue` does with two values and
said nothing about which values the clients pick. Three of five divergences were
unguarded, and the two that did fail were already covered by older tests.

Both objects are exported now — `NODE_REDIS_DECODE_OPTIONS` from the facade, and
a new `IN_MEMORY_DECODE_OPTIONS` hoisted out of the `InMemoryRedisClient`
constructor — and the test imports them rather than re-declaring them. Each
divergence asserts the option value *and* the reply it produces.

Verified by flipping each option one at a time and running the full `tests/**`
suite (`340/340` restored):

| flip | result |
| --- | --- |
| facade `pushShape` `'items'` → `'tagged'` | **339/340** |
| facade `narrowBigInt` `'always'` → `'when-safe'` | **339/340** |
| in-memory `narrowBigInt` `'when-safe'` → `'always'` | **339/340** |
| in-memory `pushShape` `'tagged'` → `'items'` | **338/340** |
| in-memory `error` → plain `Error` | **338/340** |
| facade `error` → `RedisCommandError` | **334/340** |

Previously the first three of those were 340/340 green.

One wrinkle worth recording: node-redis' error classes are resolved lazily when
a client is first created, so the suite creates and quits one in `before`.
Without it the facade's `error` option falls back to `RedisCommandError` and the
`instanceof ErrorReply` assertion is vacuous.

## Other review items

- **`toBuffer` was only half as strict as its comment.** Rejecting `number` was
  right, but `Buffer.from` happily converts arrays and TypedArrays that
  `@redis/client`'s encoder refuses — so `sendCommand(['SET','k',[104,105]])`
  sent `hi` on the mock and threw against the real client. Now an explicit
  `string | Buffer` check whose message mirrors the encoder's, argument index
  included; the test asserts the message, not just the class.
- **`teardown()`'s leak fix was incomplete.** `this.session.close()` sat one
  line *above* the new `try`, so a throw there still skipped `state.close()` and
  stranded the active-expiry timer. Moved inside.
- **The event test pinned a divergence in the forgiving direction.** Its second
  `quit()` relied on the facade emitting `'end'` again where real node-redis
  rejects with `ClientClosedError`. It emits `'end'` directly instead, which
  shows `once` not re-firing without asserting anything about a closed client.
  The underlying close-path divergence is tracked in **#389**.
- **Raw `SUBSCRIBE` through the cluster facade** now lands on `masters[0]`
  (no keys → first master) and wedges the cached session that also serves every
  keyless command and the first slot range. `main` wedged a different node, so
  this is deterministic rather than new. Documented on `sessionForCommand`
  rather than fixed: the real fix is a dedicated pub/sub session plus a
  push-drain loop — what `NodeRedisMockClient` already has — which is a feature,
  not a routing change. Use the standalone client for pub/sub. Deferral
  confirmed in review.
- **`'when-safe'` comment was backwards** — no real client widens past 2^53. It
  now says plainly that this is a deliberate lossless-over-faithful choice.
- **`flat-pairs`** notes that real node-redis negotiates RESP3 and returns
  `[field, value]` tuples, referencing **#385**. Behaviour unchanged here.
- **`routingKeys` doc** names the two invariants it leans on (one profile and
  registry per cluster; a `plan()` error meaning "no routable keys") and stops
  implying `plan()` is free.

### Out of scope

- **`mSet` / `zUnionStore` signature narrowness.** Real fault, but widening two
  methods by hand does not fix it — the whole curated surface can drift from
  node-redis with nothing in CI noticing. Note for whoever picks this up: the
  obvious form of the fix **does not compile**. `const _: RedisClientType =
  mockClient` is impossible, because `RedisClient` has a `#private` field plus
  four `private` members and TypeScript checks those nominally. A structural
  `Pick<RedisClientType, 'mSet' | 'zUnionStore' | …>` pins the same surface and
  does compile.
- **#387** — `ClusterPolicy` rejects `SORT … BY nosort` and non-glob `BY`
  patterns that both Redis 7.2.1 and 8.0.6 allow, and the SORT cluster rules are
  not version-gated. Worth reading next to this PR's argument — "delete the
  facade's copy and let `ClusterPolicy` own the SORT rules" — since it means the
  owner being delegated to gets one neighbouring rule wrong. `sortRoutingKeys`
  already exempts `nosort`; `sortPatternMayCrossSlot` does not.
- **#388** — `CONFIG <unknown-subcommand>` wording, with MONITOR timestamp
  precision.
- **#389** — the `quit()`/close-path lifecycle divergence described above.

All three are strictly pre-existing: `origin/main` is byte-identical on every
case.

### Confirmed correct, for the record

The SORT error wording this PR pins is **right**. Redis 7.2.1 uses a shorter
`ERR BY option of SORT denied in Cluster mode.`, but this repo's default profile
is `redis-8.0` (`DEFAULT_SPEC`, `src/core/compatibility/profile.ts`), and 8.0.6
emits the long wording verbatim.

## `git diff --stat origin/main`

```
 docs/TESTING.md                                  |   6 +-
 src/client-mocks/node-redis-mock.ts              | 402 ++++++++++++-----------
 src/core/decode-redis-value.ts                   | 157 +++++++++
 src/in-memory-client.ts                          | 136 ++------
 tests-integration/node-redis/string/core.test.ts |  15 +
 tests-integration/node-redis/zset/setops.test.ts |  15 +
 tests/client-mocks/node-redis-mock.test.ts       | 204 +++++++++++-
 tests/core/decode-redis-value.test.ts            | 187 +++++++++++
 8 files changed, 815 insertions(+), 307 deletions(-)
```

## Test suites run

| suite | result |
| --- | --- |
| `npm run lint` | pass |
| `npm test` (incl. `tsc --noEmit`) | 340/340 pass |
| `npm run build` | pass |
| `npm run test:package` | 4/4 pass |
| `npm run test:integration:mock:node-redis` | 526/526 pass |
| `npm run test:integration:mock:ioredis` | 475/475 pass |

The `:real` suites were not run — they need the docker-compose Redis cluster.

<sub>Correction to an earlier note on this PR: one run of
`test:integration:mock:node-redis` reported `525 pass / 1 fail`. I did not
capture which test failed and it has not recurred in eleven subsequent full
runs. My earlier claim that it was a known pre-existing flake does not hold up —
it is simply **unexplained**. Neither file this PR changes in that path is
reachable from that suite, but that is an argument, not an identification.</sub>

Closes #363


src/client-mocks/node-redis-mock.ts
```diff
@@ -776,9 +779,7 @@ export class NodeRedisMockCluster extends CommandRunner {
    * command's keys. Keyless commands run on the first master.
    */
   private sessionForCommand(args: NodeRedisCommandArgument[]): ClientSession {
-    const keys = extractRoutingKeys(args)
-    const slot =
-      keys.length > 0 ? this.topology.calculateSlotForKeys(keys) : null
+    const slot = this.topology.calculateSlotForKeys(this.routingKeys(args))
```
fatal10110, line 810: **high — regression: client-side CROSSSLOT now pre-empts the server's SORT cluster guard**

Routing by `plan().keys` is right, but `plan().keys` for `SORT` deliberately contains more than the routing keys. `sortRoutingKeys` (`src/commands/keys.ts:825`) pushes the **BY and GET patterns** into the key list precisely so `ClusterPolicy.getSortClusterPatternError` (`src/core/execution-policies/cluster-policy.ts:74`) can run its dedicated check:

```ts
function sortRoutingKeys(args: SortArgs): Buffer[] {
  const keys = [args.key]
  if (args.by && !isNoSortPattern(args.by)) keys.push(args.by)   // <- a pattern, not a key
  for (const pattern of args.get) if (!isSelfSortPattern(pattern)) keys.push(pattern)
  ...
```

Feeding that list to `calculateSlotForKeys` on this line makes `SORT mylist BY w_*` look cross-slot (`mylist`=5282, `w_*`=1585) → `-1` → the facade throws CROSSSLOT **before the command ever reaches a session**, so the policy's SORT-specific error never runs.

Ground-truthed against real Redis 7.2.1 (cluster-enabled, port 6398) driven by real node-redis v6:

```
SORT mylist BY w_*   ->  ErrorReply: "ERR BY option of SORT denied in Cluster mode."
SORT mylist GET p_*  ->  ErrorReply: "ERR GET option of SORT denied in Cluster mode."
```

Mock on `origin/main`:

```
SORT BY pattern   ->  ErrorReply: ERR BY option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.
SORT GET pattern  ->  ErrorReply: ERR GET option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.
```

Mock on this branch:

```
SORT BY pattern   ->  ErrorReply: CROSSSLOT Keys in request don't hash to the same slot
SORT GET pattern  ->  ErrorReply: CROSSSLOT Keys in request don't hash to the same slot
```

`COMMAND GETKEYS` confirms the patterns are not keys: `redis-cli -p 6399 COMMAND GETKEYS SORT k BY w_* GET p_* GET #` → `k`.

**The clean fix is probably to delete the `slot === -1` branch entirely** (lines 784-793), not to special-case SORT. With `routingKeys()` now producing the same key set `ClusterPolicy` computes, the facade's own CROSSSLOT construction is redundant: `getSlotOwner(-1)` finds no owner, falls back to `?? this.masters[0]`, and that node's `ClusterPolicy` recomputes the same keys and raises `RedisCrossSlotError` — which `decodeReply` surfaces as a byte-identical `ErrorReply("CROSSSLOT Keys in request don't hash to the same slot")`. Verified by deleting the block and re-running the facade suite: **25/25 pass**, and SORT BY/GET then returns the correct policy error again.

That also removes the second, drifting construction of a client-visible error that `ClusterPolicy` already owns — the same per-command special-casing in the facade that this PR's fix is otherwise removing.

tests/client-mocks/node-redis-mock.test.ts
```diff
@@ -279,4 +279,87 @@ describe('createNodeRedisMock (cluster)', () => {
...
+          keys: [key],
+          arguments: [`${key}-value`],
+        }),
+        'OK',
+      )
+      assert.strictEqual(await cluster.get(key), `${key}-value`)
+    }
+  })
+
+  test('mSet across slots is refused with CROSSSLOT', async () => {
```
fatal10110: **medium — these two CROSSSLOT tests are green *before* the fix, so they don't guard the bug**

Empirically checked: I extracted the PR into a scratch dir, restored the old `MULTI_KEY_COMMANDS` + first-arg heuristic inside `routingKeys()`, and re-ran this file:

```
# pass 24
# fail 1
```

The only failure is test 5, `routes by the command's real keys, not its first argument`. Both `mSet across slots is refused with CROSSSLOT` (this test) and `zUnionStore across slots is refused with CROSSSLOT` (line 325) **pass with the buggy router**.

The reason is the same one that makes the facade's `slot === -1` branch redundant: with a first-arg heuristic the cross-slot key set still reaches *some* master, and that node's `ClusterPolicy` computes the real keys and throws the identical `RedisCrossSlotError`. So `/CROSSSLOT/` matches either way.

They are fine as shape/regression guards for the error wording, but the PR description presents them as coverage for the routing fix, and a future revert to a first-arg heuristic would sail past them. The genuinely red-then-green `eval` test at line 285 is the one doing the work — worth saying so in the comments here so nobody mistakes these for the guard.

src/core/decode-redis-value.ts
```diff
@@ -0,0 +1,152 @@
...
+      return ''
+  }
+}
+
+/** Coerce a client-supplied command argument to its wire `Buffer`. */
+export function toRedisArgument(arg: string | number | Buffer): Buffer {
+  if (Buffer.isBuffer(arg)) {
+    return arg
+  }
+  return Buffer.from(typeof arg === 'number' ? String(arg) : arg)
```
fatal10110, line 149: **medium — `toRedisArgument` accepting `number` silently widens the node-redis facade**

`node-redis-mock.ts` previously had its own `toBuffer(arg: string | Buffer)`. Both of its call sites (`runOnSession`, and now `routingKeys`) were switched to this shared `toRedisArgument(arg: string | number | Buffer)`, which stringifies numbers. `NodeRedisCommandArgument` is still `string | Buffer`, so TypeScript still guards typed call sites — but anything arriving through `any`, plain JS, or a runtime value now gets accepted where it used to throw.

Real node-redis v6:

```
SET n <number>  ->  TypeError: "arguments[2]" must be of type "string | Buffer", got number instead.
```

Mock on `origin/main`:

```
SET n 5(number)          ->  TypeError: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received type number (5)
cluster SET n 5(number)  ->  TypeError: ...
```

Mock on this branch:

```
SET n 5(number)          ->  OK
GET n                    ->  "5"
cluster SET n 5(number)  ->  OK
```

That is the wrong direction for a facade whose whole job is to fail the same way the real client fails: a test suite that passes numbers now green-lights code that throws against real node-redis.

The `number` branch is legitimately needed by `InMemoryRedisClient` (`RedisCommandArgument` includes `number`), so the shared helper is right to have it — but the node-redis facade should keep a narrower local coercion, or reject `number` explicitly.

src/client-mocks/node-redis-mock.ts
```diff
@@ -469,8 +488,8 @@ export class NodeRedisMockClient extends CommandRunner {
   destroy(): void {
     // teardown() is async (it awaits the push-drain loop); surface a rejection
     // as an 'error' event rather than dropping it as an unhandled rejection.
-    void this.teardown().catch(err => this.emitter.emit('error', err))
-    this.emitter.emit('end')
+    void this.teardown().catch(err => this.emit('error', err))
```
fatal10110, line 493: **medium — this comment is now false, and the chain it describes leaks the expiry timer**

The comment directly above says the `.catch(...)` surfaces the rejection "as an 'error' event rather than dropping it as an unhandled rejection". With `CommandRunner` now extending `EventEmitter`, `this` *is* the emitter — and `EventEmitter#emit('error', err)` **throws** `ERR_UNHANDLED_ERROR` when no `'error'` listener is registered. So the catch handler itself throws, the void-discarded promise rejects anyway, and under Node's default `--unhandled-rejections=throw` the process crashes.

Full chain, when `drainPushes()` catches a non-abort error:

1. `this.emit('error', err)` (line 556) throws `ERR_UNHANDLED_ERROR` — no listener.
2. `pubsub.drained` rejects → `teardown()` rejects at `await pubsub.drained` (line 508).
3. There is no `try/finally`, so `if (this.ownsState) { this.backend.state.close() }` (lines 512-517) **never runs** — the self-rescheduling active-expiry timer keeps the process alive.
4. `quit()` / `disconnect()` reject with a confusing `ERR_UNHANDLED_ERROR` instead of the original error; `destroy()` crashes.

The semantics are unchanged from `origin/main` (the private `new EventEmitter()` threw on an unlistened `'error'` too), so this is not a regression — but the line is in the diff, the comment asserting the opposite is now visibly wrong, and moving `state.close()` into a `finally` is a small fix while you're here.

(The rest of the `extends EventEmitter` refactor checks out: zero name collisions between `EventEmitter.prototype`/statics and any `CommandRunner` / `NodeRedisMockClient` / `NodeRedisMockCluster` member — `off`, `emit`, `once` and the static `create` included — and `super()` is called in both constructors.)

src/core/decode-redis-value.ts
```diff
@@ -0,0 +1,152 @@
...
+    case 'map-pairs': {
+      const out: { [key: string]: NativeRedisReply } = {}
+      for (const [key, val] of value.entries) {
+        out[decodeRedisKey(key)] = decode(val)
+      }
+      return out
+    }
+    case 'flat-pairs':
+      // Flat on the wire in RESP2; keep the flat array shape here too.
+      return value.entries.flatMap(([key, val]) => [decode(key), decode(val)])
```
fatal10110, line 115: **medium — `flat-pairs` is flattened for both clients, but real node-redis returns tuples**

Pre-existing (byte-identical to `origin/main`) — flagging it here because this PR makes this module the single home for the decision and already models per-client divergence via `pushShape` / `narrowBigInt`, so a `flatPairsShape: 'flat' | 'pairs'` option belongs right next to them.

The comment on line 109 ("Flat on the wire in RESP2") is the justification, but the node-redis client this decoder is impersonating negotiates **RESP3**, where these come back as an array of `[field, value]` pairs.

Real node-redis v6 → Redis 7.2.1, `sendCommand` (connection reports `resp=3`):

```
HRANDFIELD h 2 WITHVALUES  ->  [["f1","v1"],["f2","v2"]]
ZRANGE z 0 -1 WITHSCORES   ->  [["a",1],["b",2]]
ZRANDMEMBER z 1 WITHSCORES ->  [["b",2]]
```

```
$ redis-cli -3 -p 6397 --no-raw HRANDFIELD hz 2 WITHVALUES
1) 1) "f1"
   2) "v1"
2) 1) "f2"
   2) "v2"
```

This mock, same commands:

```
HRANDFIELD h 2 WITHVALUES  ->  ["f1","v1","f2","v2"]
ZRANGE z 0 -1 WITHSCORES   ->  ["a",1,"b",2]
ZRANDMEMBER z 1 WITHSCORES ->  ["b",2]
```

So `const [[f, v]] = await client.sendCommand(['HRANDFIELD','h','2','WITHVALUES'])` passes against real node-redis and destructures garbage against the mock. (`map`/`map-pairs` *is* correct — `CONFIG GET maxmemory` gives `{"maxmemory":"0"}` in both.)

Fine to leave for a follow-up, but worth an issue rather than letting the "RESP2 wire shape" comment stand as settled.

tests-integration/node-redis/string/core.test.ts
```diff
@@ -81,6 +81,19 @@ describe(`String Commands Integration (node-redis, ${testRunner.getBackendName()
     assert.strictEqual(get3, 'value3')
   })
 
+  test('MSET cross-slot error', async () => {
```
fatal10110, line 84: **low — this integration test can't observe the change, and duplicates existing coverage**

The PR description's reasoning for putting the red-then-green test in the facade suite is **correct** — I verified it: `tests-integration/test-config.ts` builds the fixture with `createRedisCluster(...)` plus `createCluster` from the real `redis` package over TCP, so `NodeRedisMockCluster` is never instantiated and `routingKeys()` is not in the loop. The integration suite genuinely cannot see this bug.

But that same fact makes these two added tests (`MSET cross-slot error` here, and `ZUNIONSTORE rejects sources outside the destination slot` at `tests-integration/node-redis/zset/setops.test.ts:156`) green before **and** after the change — they assert only `ClusterPolicy` behaviour, and this PR touches zero files under `src/core/execution-policies/` or `src/commands/`.

There is already equivalent coverage:
- `tests-integration/node-redis/commands.test.ts:44` — `cross slot error` (mGet)
- `tests-integration/ioredis/string/core.test.ts:55` — MGET cross-slot

So they cost real cluster-startup runtime in both the `mock` and `real` backends for coverage that exists. Worth dropping, or keeping only if you specifically want the per-command CROSSSLOT wording pinned against real Redis in CI — in which case say that in the comment, because right now they read as coverage for the routing fix.

src/client-mocks/node-redis-mock.ts
```diff
@@ -192,6 +188,16 @@ abstract class CommandRunner {
     return asNumber(await this.run(['DEL', ...keys]))
   }
 
+  async mSet(
```
fatal10110, line 190: **low — `mSet` / `zUnionStore` accept a strict subset of the real node-redis signatures**

The facade is explicitly a "curated" surface, so a subset is defensible — but these two are narrower in ways that break the *most idiomatic* node-redis call, which is the failure mode this facade exists to prevent.

From the installed `redis@6.0.0`:

```ts
// node_modules/@redis/client/dist/lib/commands/MSET.d.ts
export type MSetArguments =
  | Array<[RedisArgument, RedisArgument]>
  | Array<RedisArgument>
  | Record<string, RedisArgument>
// returns SimpleStringReply<'OK'>  (mock returns Promise<string>)
```

```ts
// ZUNIONSTORE.d.ts
parseCommand(parser, destination: RedisArgument, keys: ZKeys, options?: ZUnionOptions)
// ZKeys = ZVariadicKeys<RedisArgument> | ZVariadicKeys<ZKeyAndWeight>   (generic-transformers.d.ts:173)
```

So against the mock these are type errors and against real node-redis they are fine:

```ts
await client.mSet({ a: '1', b: '2' })        // Record form — the common one
await client.mSet(['a', '1', 'b', '2'])      // flat form
await client.zUnionStore('d', 'src')         // single key, not an array
await client.zUnionStore('d', ['a','b'], { WEIGHTS: [1,2] })
```

A user swapping `createClient()` for `createNodeRedisMock()` gets a compile break unrelated to their own code.

Nothing catches this: `tests-package/` has only `dual-format.test.ts`, and nothing type-checks the facade against `RedisClientType`, so the drift is invisible to CI. A `const _: RedisClientType = mockClient` style assertion somewhere in `tests-package/` would pin the whole curated surface at once — probably worth more than widening these two by hand.

(The three tested shapes — `mSet([[k,v],...])`, `zUnionStore(dest, keys)`, `eval(script, {keys, arguments})` — were verified to work identically against real node-redis v6 and a real cluster, including the exact CROSSSLOT messages the new integration tests assert. `eval` is also not double-decoding: `run()` returns a raw `RedisValue`, so exactly one `decodeReply` is applied.)

src/client-mocks/node-redis-mock.ts
```diff
@@ -799,6 +800,40 @@ export class NodeRedisMockCluster extends CommandRunner {
...
+   * does, rather than inventing a second error path here.
+   */
+  private routingKeys(args: NodeRedisCommandArgument[]): readonly Buffer[] {
+    if (args.length === 0) {
+      return []
+    }
+
+    const [name, ...rest] = args
+    try {
+      return this.masters[0].executor.plan(
```
fatal10110, line 852: **low — `plan()` for its side data: cost, an over-broad catch, and an unguarded cross-node invariant**

Three small things about this call, none blocking:

**1. It is not free, and the comment presents it as free.** `createPlan` (`src/core/command-executor.ts:331-352`) runs `parseCommandArgs(...)`, then `definition.keys(args)`, then defensively copies every argument: `rawCommand: Buffer.from(rawCommand), rawArgs: rawArgs.map(arg => Buffer.from(arg))`. The same `plan()` is recomputed immediately afterwards inside `session.execute` → `executeRawWithPlan` (line 119). So every cluster command now pays **two** schema parses and two full argument copies, and all but `.keys` is discarded. CLAUDE.md deprioritises performance so this is fine to ship — but a `CommandExecutor.routingKeysFor(name, args)` that stops after `definition.keys(args)` would be cheaper *and* a cleaner seam than a client mock reaching into `plan()` for side data. With one call site the abstraction may not be earned yet; if so, say that in the comment rather than implying the call is cheap.

**2. The catch is broader than the doc comment's justification.** It swallows every `RedisCommandError`, while the comment only argues for "unknown name, bad arity". I checked and it's safe today — no `defineCommand` declares `t.key()` in its schema while returning `keys: () => []`, and `RedisCrossSlotError` / `RedisMovedError` / `RedisClusterDownError` are policy errors that `plan()` never throws (policies don't run there). But commands this repo doesn't register (OBJECT, MEMORY, RESTORE, LATENCY, DEBUG under the default profile) do have real keys in Redis —

```
$ redis-cli -p 6399 COMMAND GETKEYS OBJECT ENCODING k   ->  k
$ redis-cli -p 6399 COMMAND GETKEYS MEMORY USAGE k      ->  k
$ redis-cli -p 6399 COMMAND GETKEYS RESTORE k 0 payload ->  k
```

— and today land on `masters[0]` with `keys: []`, which is observationally fine only because the node answers the same unknown-command error wherever it lands. The moment one is implemented, a keyed command could silently run on the wrong node instead of producing MOVED. Narrowing to `UnknownRedisCommandError` + arity errors, or just a comment noting the coupling, would close that.

**3. `masters[0]`'s executor is used to route a command that runs elsewhere.** The comment asserts "any master's executor answers identically" with nothing enforcing it. True today because `src/cluster.ts:69,120-140` hoists one `resolveCompatibilityProfile(...)` and hands every node the same registry+profile (only the `CLUSTER` definitions are node-specific). If per-node profiles or command sets ever land, routing would use master-0's schema for a command executed on another node, with no error. Worth naming as an invariant rather than an observation.

(`calculateSlotForKeys([])` → `null`, confirmed at `src/state/cluster-topology.ts:25-31`, so dropping the old `keys.length > 0` guard is safe. `masters[0]` is always defined — `validateOptions` rejects `masters < 1` at `src/cluster.ts:167`. `args.length === 0` → `[]` → `null` → `masters[0]` → `runOnSession` throws `'command requires at least a name'`. All correct.)

src/core/decode-redis-value.ts
```diff
@@ -0,0 +1,152 @@
...
+ * flat pairs, nulls — decodes identically, so it lives in
+ * {@link decodeRedisValue} once.
+ */
+export type DecodeRedisValueOptions = {
+  /**
+   * How an integer reply that arrived as a `bigint` is narrowed.
+   *  - `'always'`: node-redis parses a RESP2 `:` with plain JS number
+   *    arithmetic, so the reply is a `number` — precision loss past 2^53
+   *    included. Only RESP3's `(` BIG_NUMBER yields a bigint there.
+   *  - `'when-safe'`: keep a plain `number` like a real client, and widen to
```
fatal10110: **low — "like a real client" is backwards here**

No real client widens to `bigint` — they all lose precision past 2^53. `'when-safe'` is the option that *diverges* from real-client behaviour; `'always'` is the faithful one.

Real node-redis v6 → Redis 7.2.1:

```
sendCommand(['EVAL', 'return 9007199254740993', '0'])  ->  number 9007199254740992
```

This PR's mocks, same command:

```
in-memory client :  BIGINT(9007199254740992)
node-redis mock  :  9007199254740992          <- 'always' is correct
```

So a consumer of the in-memory client doing arithmetic or `===` on a large integer reply gets `TypeError: Cannot mix BigInt and other types` — something neither ioredis nor node-redis ever produces.

The behaviour is unchanged from `origin/main`; only the justification is new, and it claims the opposite of what it does. Either drop the "like a real client" clause or state plainly that this is a deliberate lossless-over-faithful choice for the in-memory client.

src/core/decode-redis-value.ts
```diff
@@ -0,0 +1,152 @@
...
+   */
+  narrowBigInt: 'always' | 'when-safe'
+  /**
+   * Shape of a `push` reply.
+   *  - `'items'`: just the payload items; the type tag is consumed by the
+   *    client's own push routing (node-redis).
+   *  - `'tagged'`: `[name, ...items]` — the RESP2 wire shape, so a raw
+   *    push-mode consumer sees what a real client would read off the socket.
+   */
+  pushShape: 'items' | 'tagged'
```
fatal10110, line 43: **low — nothing pins the two clients' option sets apart, so a later edit can silently re-converge them**

The dedupe itself is behaviour-preserving — I ran an identical 24-command decode probe (`map`, `map-pairs`, `set`, `double`, `boolean`, `big-number`, `verbatim`, `flat-pairs`, `null` vs `null-array`, `push`, bulk-string-null, `error`) through both clients on `origin/main` and on this branch and diffed the output: the only difference was `HRANDFIELD`'s random ordering. Both option sets map case-for-case onto the old bodies. Good refactor.

The gap is that the *divergences* are untested, so the safety net for the next person is missing:

- Flipping node-redis to `pushShape: 'tagged'`, or the in-memory client to `narrowBigInt: 'always'`, fails **no test**.
- `pushShape: 'items'` is effectively uncovered: the facade's pub/sub path (`dispatchPush`, `node-redis-mock.ts:572`) reads `value.name` / `value.items` off the `RedisValue` directly and decodes push *items* individually — it never decodes the push frame. Only `sendCommand(['SUBSCRIBE', ...])` would exercise this branch. (That's also why the refactor is safe: `drainSubscribeAck` returns a raw `RedisValue`, so neither `pushShape` value can disturb the existing pub/sub flow.)
- The `'when-safe'` widening branch (value past 2^53 → `bigint`) has no test either. `tests/in-memory-client.test.ts:20` covers "returns integers as numbers, not bigint" and `:55` covers `returnBuffers`, but not the widening.

Two small tests — one asserting the in-memory client tags a push `[name, ...items]` while the node-redis mock does not, one asserting the bigint widening threshold — would lock in exactly what this options object exists to express.

Same story for `extends EventEmitter`: `grep` across `tests/`, `examples/`, `docs/` for `.on('connect'|'ready'|'end'|'error')` / `.once(` / `.off(` returns only `process.off('unhandledRejection', ...)`. The whole event surface of both public clients was just rewritten with zero coverage on either side of the change. Also worth noting none of `quit()` / `disconnect()` / `destroy()` / `teardown()` calls `removeAllListeners()`, so a quit client keeps user listeners alive — same as the old private emitter, but now on a public object.

tests/client-mocks/node-redis-mock.test.ts
```diff
@@ -279,4 +279,87 @@ describe('createNodeRedisMock (cluster)', () => {
...
+
+    await assert.rejects(
+      () =>
+        cluster!.mSet([
+          ['{x}:a', '1'],
+          ['{y}:b', '2'],
+        ]),
+      (err: unknown) => {
+        assert.ok(err instanceof Error)
+        assert.match(err.message, /CROSSSLOT/)
```
fatal10110: **low — `/CROSSSLOT/` is looser than the helper this repo already has**

A mangled message — bare `CROSSSLOT`, or the wrong slot wording — still passes this assertion. The integration suite pins the full string:

```ts
// tests-integration/node-redis/string/core.test.ts:93
errorWithMessage("CROSSSLOT Keys in request don't hash to the same slot")
```

Worth tightening here and at line 334 (line 277 is pre-existing, same pattern).

For reference, the exact wording is right — real node-redis v6 against a real cluster surfaces `message === "CROSSSLOT Keys in request don't hash to the same slot"`, byte-identical to what `RedisCrossSlotError` (`src/core/redis-error.ts:332`) + `redisErrorText()` reconstruct. (Real node-redis actually throws the subclass `SimpleError`, but `instanceof ErrorReply` holds for both, so the `assert.ok(err instanceof ErrorReply)` assertions are sound.)

tests/core/decode-redis-value.test.ts
```diff
@@ -0,0 +1,167 @@
...
+  redisErrorText,
+  toRedisArgument,
+  type DecodeRedisValueOptions,
+} from '../../src/core/decode-redis-value'
+import { RedisCommandError } from '../../src/core/redis-error'
+import type { RedisValue } from '../../src/core/redis-value'
+
+// The node-redis facade and InMemoryRedisClient share this decoder and differ
+// only through its options. Those differences are deliberate, so they are
+// asserted here: flipping either client onto the other's option value must
```
fatal10110: **medium (new) — this file does not install the tripwire this comment promises**

> flipping either client onto the other's option value must fail a test rather than pass silently

It doesn't. `NODE_REDIS` (line 18) and `IN_MEMORY` (line 24) are **locally re-declared literals**, not the objects the clients actually pass. The real ones are module-private:

- `NODE_REDIS_DECODE_OPTIONS` — `src/client-mocks/node-redis-mock.ts:962` (not exported)
- the in-memory options — built inline in the constructor, `src/in-memory-client.ts:82-92` (not extractable)

So the tests assert `decodeRedisValue`'s behaviour for two option *values*, which was never in doubt, and say nothing about which values the clients choose.

Measured on this head (`0baa5b2`), one flip at a time, full `tests/**` suite each run:

| flip | result |
|---|---|
| facade `pushShape: 'items'` → `'tagged'` | **340/340 pass** |
| facade `narrowBigInt: 'always'` → `'when-safe'` | **340/340 pass** |
| in-memory `narrowBigInt: 'when-safe'` → `'always'` | **340/340 pass** |
| in-memory `pushShape: 'tagged'` → `'items'` | 339/340 — `pub/sub: a subscriber connection sees another connection PUBLISH` |
| in-memory `error` → plain `Error` | 339/340 — `throws a RedisCommandError on an error reply` |

Three of the five divergences are unguarded, and the two that *are* caught were already caught before this PR by pre-existing tests — so the new file adds no tripwire at all. `pushShape` on the facade is structurally unreachable from a test, too: `dispatchPush` (`node-redis-mock.ts:572`) reads `value.name`/`value.items` off the `RedisValue` directly and never routes a push frame through `decodeReply`.

This is the previous round's *"nothing pins the two clients' option sets apart"* finding, and it's still open — the comment now claims it's closed.

Smallest fix that makes the claim true: `export const NODE_REDIS_DECODE_OPTIONS`, hoist the in-memory literal to an exported `IN_MEMORY_DECODE_OPTIONS`, and import both here instead of re-declaring them. Otherwise, drop the three-line claim.

(The assertions themselves are all correct and worth keeping — verified `narrowBigInt: 'always'` → `9007199254740992` against real node-redis v6, and `DEFAULT_RESP = 3` at `@redis/client/dist/lib/RESP/types.js:5` backs the new `flat-pairs`/#385 comment.)

tests/client-mocks/node-redis-mock.test.ts
```diff
@@ -209,6 +210,36 @@ describe('createNodeRedisMock (standalone)', () => {
...
+    assert.strictEqual(
+      client.on('end', () => seen.push('on')),
+      client,
+      'on() returns the client for chaining',
+    )
+    client.once('end', () => seen.push('once'))
+    client.on('end', removed).off('end', removed)
+
+    await client.quit()
+    await client.quit() // 'once' must not fire a second time
```
fatal10110: **medium (new) — this line pins a divergence from real node-redis rather than a behaviour worth keeping**

The second `quit()` is what makes `seen` end in `'on'` instead of stopping at `['on', 'once']` — the facade's `teardown()` early-returns on `this.closed` and then `quit()` emits `'end'` again. Real node-redis **rejects** the second `quit()` and emits `'end'` exactly once:

```js
// node_modules/@redis/client/dist/lib/client/socket.js:323
async quit(fn) {
    if (!this.#isOpen) {
        throw new errors_1.ClientClosedError();   // 'The client is closed'
    }
    this.#isOpen = false;
    const reply = await fn();
    this.destroySocket();                        // <- 'end' is emitted here, once
    return reply;
}
```

and `client.quit()` goes straight there — `index.js:1104: return this._self.#socket.quit(async () => {...})`, aliased at `:1112: quit = this.QUIT`.

So against real node-redis this test's line 228 throws and the assertion on line 231 would be `['on', 'once']`. The facade is the more forgiving of the two, which is the wrong direction for a fake whose job is to fail where the real client fails: an `afterEach` that calls `quit()` on top of an explicit `quit()` is green on the mock and throws in production-shaped code.

Two ways out, either fine:
- keep the comment's intent but drop the double-quit — `client.emit('end')` directly exercises `once` not re-firing, without asserting anything about a closed client; or
- make `quit()`/`disconnect()` on a closed facade reject with node-redis' `ClientClosedError` and assert *that* here, which is strictly more faithful.

The rest of the test is good — `on()` returning `this` for chaining is real (`EventEmitter#on` returns the emitter), and it's the only coverage the `extends EventEmitter` change has.

src/client-mocks/node-redis-mock.ts
```diff
@@ -774,23 +784,19 @@ export class NodeRedisMockCluster extends CommandRunner {
   /**
    * Resolve (and cache) a session on the master that owns the slot for the
    * command's keys. Keyless commands run on the first master.
+   *
+   * There is deliberately no client-side rejection here. A key set spanning
+   * slots gives `-1`, which owns no node and falls through to the first master;
```
fatal10110, line 789: **low (new) — routing keyless commands to `masters[0]` makes a raw `SUBSCRIBE` wedge the cluster facade's keyless path permanently**

Not about the CROSSSLOT deletion — this falls out of `plan().keys` returning `[]` for `SUBSCRIBE`, so it now lands on `masters[0]`, which is also where every other keyless command goes. `sessionFor()` caches one session per node, and that session is now stuck in subscriber mode for the life of the facade.

Same script both trees, `cluster: { masters: 3 }`, everything through `sendCommand`:

```
                   origin/main                    pr-378 (0baa5b2)
PING before    ->  OK "PONG"                      OK "PONG"
SUBSCRIBE ch   ->  OK ["ch",1]                    OK ["ch",1]
FLUSHALL       ->  OK "OK"                        ErrorReply: ERR Can't execute 'flushall': only
                                                  (P|S)SUBSCRIBE / (P|S)UNSUBSCRIBE / PING / QUIT /
                                                  RESET are allowed in this context
PING after     ->  OK "PONG"                      OK [""]
DBSIZE         ->  OK 0                           ErrorReply: ERR Can't execute 'dbsize': ...
GET alpha      ->  OK null                        ErrorReply: ERR Can't execute 'get': ...
```

`GET alpha` is collateral: `alpha` is slot 865, inside master-0's `[0,5460]`, so every key in the first master's range is affected too, not just keyless commands.

On `main` the first-argument heuristic sent `SUBSCRIBE` to `slot('ch')`'s node, so the wedge existed there as well — it just didn't land on the node that serves keyless commands. Strictly, `main` was accidentally better here; the PR makes the collision deterministic.

Low because `NodeRedisMockCluster` exposes no pub/sub API, so `sendCommand` is the only way in, and the error text itself is faithful to real RESP2 subscriber mode. But the standalone client already solves this — it keeps a **separate** `pubsub.session` (`node-redis-mock.ts:503-512`) precisely so the command session never enters subscriber mode. Either do the same here, or add a sentence to the `sessionForCommand` doc comment recording that a raw `SUBSCRIBE` through the cluster facade is not supported and why.

src/client-mocks/node-redis-mock.ts
```diff
@@ -868,34 +917,15 @@ async function drainSubscribeAck(stream: ResponseStream): Promise<RedisValue> {
...
-  // Single-key heuristic: every other routed command keys off its first arg.
-  // Uncommon multi-key commands sent via the generic sendCommand fall through
-  // here and route by first key (documented honest scope).
-  return [toBuffer(args[1])]
+/**
+ * Deliberately narrower than the shared `toRedisArgument`, which also accepts
+ * `number` for {@link InMemoryRedisClient}. Real node-redis rejects a number
+ * argument with a TypeError rather than stringifying it, and a facade that
+ * quietly accepted one would green-light test code that throws against the real
+ * client. `Buffer.from` raises that TypeError for us.
```
fatal10110: **low (new) — the `number` claim is right; "rejects like node-redis" in general is not**

`Buffer.from` genuinely does throw on a number, so the JSDoc's core point and the new test at `tests/client-mocks/node-redis-mock.test.ts:233` both hold, and re-localising `toBuffer` was the right call:

```
$ node -e "Buffer.from(5)"
TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an
instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received type number (5)
```

The gap is everything else `Buffer.from` accepts that node-redis' encoder refuses:

```
// node_modules/@redis/client/dist/lib/RESP/encoder.js:17
throw new TypeError(`"arguments[${i}]" must be of type "string | Buffer", got ${typeof arg} instead.`);
```

```
Buffer.from acceptance probe:
  5                 -> TypeError            (agrees with node-redis)
  true / null / {}  -> TypeError            (agrees)
  [104, 105]        -> OK, bytes "hi"       <- node-redis throws
  new Uint8Array()  -> OK, "hi"             <- node-redis throws
```

So `client.sendCommand(['SET', 'k', [104,105] as any])` silently sends `hi` on the mock and throws against the real client — the exact failure mode this function exists to prevent, one type away from the one it does prevent.

Also, the test asserts only the class, not the message, so the two TypeErrors' very different wording isn't pinned either.

Cheapest fix that matches the doc comment:

```ts
function toBuffer(arg: NodeRedisCommandArgument): Buffer {
  if (Buffer.isBuffer(arg)) return arg
  if (typeof arg !== 'string') {
    throw new TypeError(`"arguments" must be of type "string | Buffer", got ${typeof arg} instead.`)
  }
  return Buffer.from(arg)
}
```

Or keep `Buffer.from` and narrow the comment to "rejects a number, which is the case `InMemoryRedisClient` needs and the facade must not inherit" — array-likes still slip through.

(Flagged independently by both the wire-compat and architecture passes.)

src/client-mocks/node-redis-mock.ts
```diff
@@ -479,22 +500,27 @@ export class NodeRedisMockClient extends CommandRunner {
...
-      // Wait for the drain loop to observe the abort and finish, so no async
-      // iteration dangles past teardown.
-      await pubsub.drained
-      pubsub.listeners.clear()
-      pubsub.patternListeners.clear()
-    }
-    if (this.ownsState) {
-      // Only the creating client owns the state graph; closing it clears the
-      // self-rescheduling active-expiry timer. Duplicates share this state and
+    try {
```
fatal10110, line 502: **low (new) — `this.session.close()` sits one line above the `try`, so the invariant this `finally` was added for is only half-enforced**

The comment in the `finally` states the rule exactly right:

> a push-drain loop that rejects must not strand the state: closing it is what clears the self-rescheduling active-expiry timer, and a live timer keeps the process alive

But line 502 — `this.session.close()` — runs *outside* the `try`, and it can throw for the same class of reason. If it ever does, `state.close()` is skipped, the active-expiry timer stays armed, and because `this.closed` was already set to `true` on line 501 a retried `quit()` returns early and can never re-close it. The process then hangs at exit with no way to recover from user code.

Moving line 502 to the first statement inside the `try` costs nothing and closes the gap:

```ts
this.closed = true
try {
  this.session.close()
  if (this.pubsub) { ... }
} finally {
  if (this.ownsState) this.backend.state.close()
}
```

The rest of the change is correct and the leak it was written for is genuinely fixed. For the record on the two things it deliberately leaves: `pubsub.listeners`/`patternListeners` not being cleared when `await pubsub.drained` rejects is **harmless** — `this.pubsub` is already `undefined` and the record is only reachable from the local `const` and the finished drain closure, so it's collectable; and `teardown()` is idempotent, since `this.closed` is set synchronously before the first `await`.

tests/client-mocks/node-redis-mock.test.ts
```diff
@@ -263,18 +294,160 @@ describe('createNodeRedisMock (cluster)', () => {
...
+      ['BY', 'w_*'],
+      ['GET', 'p_*'],
+    ]) {
+      await assert.rejects(
+        () => cluster!.sendCommand(['SORT', 'mylist', option, pattern]),
+        (err: unknown) => {
+          assert.ok(err instanceof ErrorReply)
+          assert.strictEqual(
+            err.message,
+            `ERR ${option} option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.`,
```
fatal10110, line 398: **medium (new, pre-existing — not introduced here, not blocking)** — this assertion is **correct**, and checking it turned up a neighbouring rule `ClusterPolicy` gets wrong

First, the good news, because I went looking for the opposite. The long wording here is right for this repo's default profile. `DEFAULT_SPEC = 'redis-8.0'` (`src/core/compatibility/profile.ts:63`), and Redis 8.0.6, cluster-enabled, all slots owned:

```
SORT mylist BY 'w_*'        -> ERR BY option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.
SORT mylist GET 'p_*'       -> ERR GET option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.
SORT mylist GET '#'         -> 1 2 3                      (allowed — matches isSelfSortPattern)
SORT {x}list BY '{x}w_*'    -> 1 2 3                      (allowed — matches the hash-tag relaxation)
SORT {x}list BY '{y}w_*'    -> ERR BY option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.
```

Byte-identical to what this test pins, and `sortPatternMayCrossSlot`'s hash-tag logic matches too. (Redis **7.2.1** uses a shorter `ERR BY option of SORT denied in Cluster mode.` and denies `GET '#'` — see the second point below.)

**The bug this surfaced.** `ClusterPolicy` rejects a `BY` pattern that contains no `*`, including the extremely common `BY nosort`:

```
                        Redis 8.0.6      Redis 7.2.1      mock (pr-378 AND origin/main)
SORT mylist BY nosort   3 1 2            3 1 2            ErrorReply: ERR BY option of SORT denied in Cluster mode when...
SORT mylist BY w_foo    3 1 2            3 1 2            ErrorReply: ERR BY option of SORT denied in Cluster mode when...
```

Real Redis sets `dontsort` for any `BY` pattern without a `*` and skips the cluster check entirely. `sortRoutingKeys` already knows this — `isNoSortPattern` (`src/commands/keys.ts:817`) exempts `nosort` from the routing keys — but `sortPatternMayCrossSlot` (`src/core/execution-policies/cluster-policy.ts:147-153`) doesn't:

```ts
if (!pattern.includes(0x2a)) {
  return topology.calculateSlotForKeys([source, pattern]) === -1   // true for almost any pair
}
```

One-line fix: `if (!pattern.includes(0x2a)) return false`.

**Second, smaller:** these SORT cluster rules aren't version-gated. `getSortClusterPatternError` has no `profile.has(...)` check, unlike the `cluster.multi-db` gate ~50 lines above it, so `profile: 'redis-7.2'` still emits the 8.0 wording, still allows `GET '#'`, and still allows the hash-tag relaxation — all three of which 7.2.1 does differently. Wants a feature gate keyed at 7.4.

**Why I'm raising it on your PR at all:** both are strictly pre-existing (`origin/main` is byte-identical on every case above) and neither blocks this commit. But this commit's entire argument is *"delete the facade's copy and let `ClusterPolicy` own the SORT rules"* — which I agree with, and which this verifies — so it's worth knowing that the owner you're delegating to gets one of those rules wrong. Both deserve an issue rather than scope creep here.

fatal10110:
## Re-review of `0baa5b2` — **ship** 🚢

The high finding is properly fixed, and fixed by subtraction as suggested rather than by a SORT special case. Nothing in the six new comments below blocks: one medium is a comment that overstates what a new test guarantees, one medium is a new test pinning a small client divergence, the rest are low, and the `SORT BY nosort` item is pre-existing on `main`.

I re-ran every claim rather than reading the summary. Details below.

### Prior findings

| # | Finding | Status |
|---|---|---|
| **high** | facade's `slot === -1` pre-empts the cluster SORT guard | ✅ **fixed** — branch deleted, verified byte-for-byte (below) |
| med | the two CROSSSLOT tests don't guard the routing fix | ✅ **fixed** — replaced with wrong-node tests, verified red (below) |
| med | `toRedisArgument` accepting `number` widens the facade | ✅ **fixed** — local `toBuffer` + test; one gap remains, see the `toBuffer` comment |
| med | `destroy()` comment false; chain leaks the expiry timer | ✅ **fixed** — `finally` added, comment corrected; one line still outside the `try`, see that comment |
| med | `flat-pairs` flattened, real node-redis returns tuples | ✅ **fixed as agreed** — comment + #385, which accurately describes it |
| low | `'when-safe'` comment claims the opposite of what it does | ✅ **fixed** — now explicitly the less-faithful choice. (`src/in-memory-client.ts:82-83` still calls it the faithful one; worth harmonising next time that file is touched) |
| low | integration tests can't observe the change | ✅ **resolved as offered** — kept, with comments that now say exactly what they pin |
| low | `plan()` cost / broad catch / cross-node invariant | ✅ **addressed** — doc names both invariants and admits the double plan. The `catch` is still broader than its justification; fine at this size |
| low | `/CROSSSLOT/` looser than the repo's own helper | ✅ **fixed** — exact-string + `instanceof ErrorReply` |
| low | nothing pins the two clients' option sets apart | ⚠️ **still open** — see the `decode-redis-value.test.ts` comment; the new file asserts against its own literals |
| low | `mSet`/`zUnionStore` narrower than the real signatures | ⏸️ **deferred — agreed, but the proposed follow-up won't compile** (below) |

### Verifying the two load-bearing claims

**1. "Matches `origin/main` byte-for-byte."** Confirmed, and wider than the five probes claimed: a **19-case** probe through `NodeRedisMockCluster` — SORT BY / GET / plain / `BY nosort` / `BY` non-glob / `GET '#'` / `STORE` cross-slot, MSET, DEL, MGET, ZUNIONSTORE, SINTERCARD, EVAL cross-slot, GEORADIUS STORE, keyless PING, unknown command, arity error, empty args, plain GET — diffed `origin/main` vs `0baa5b2`: **identical, every byte**, always an `ErrorReply`. CROSSSLOT never degrades to MOVED, because `validateClusterSlot` checks `slot === -1` before ownership — the same precedence real Redis uses. An independent 27-case pass reached the same result. The only behavioural difference found anywhere between the two trees is the raw-`SUBSCRIBE` routing noted in its own comment.

**2. "Five failures where there was one."** Confirmed by restoring the old heuristic on top of this head and running the file:

```
not ok 4 - ZDIFF routes by its keys, not its first argument        error: 'MOVED 865 127.0.0.1:0'
not ok 5 - LMPOP routes by its keys, not its first argument        error: 'MOVED 865 127.0.0.1:0'
not ok 6 - SINTERCARD routes by its keys, not its first argument   error: 'MOVED 865 127.0.0.1:0'
not ok 7 - BITOP routes by its keys, not its first argument        error: 'MOVED 7638 127.0.0.1:0'
not ok 8 - EVAL routes by its declared KEYS, not by the script text error: 'MOVED 7638 127.0.0.1:0'
# tests 32   # pass 27   # fail 5
```

and against the pre-fix facade (`1ff169a`), the new SORT test goes red exactly as claimed:

```
not ok 9 - SORT BY/GET patterns keep the cluster-mode SORT error
  expected: 'ERR BY option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.'
  actual:   "CROSSSLOT Keys in request don't hash to the same slot"
not ok 18 - rejects a non string/Buffer argument the way node-redis does
```

The insight behind the rewrite is right, and the reasoning in the test comments checks out independently: `computeSlotRange` (`src/cluster.ts:146-160`) really does give `[0,5460] / [5461,10921] / [10922,16383]`, and `alpha`=865, `abc`=7638, `k1`=12706 really do land one per master. All four table commands really do have a non-key first argument (`zdiff`/`lmpop`/`sintercard` behind `numkeys`, `bitop` behind the operation token). The hard-coded replies aren't load-bearing, incidentally — a mis-route rejects with `-MOVED` before the value is ever compared.

**No weakened assertions.** Every replaced test got *tighter* (`assert.match(/CROSSSLOT/)` → `instanceof ErrorReply` + `assert.strictEqual` on the full wire text); the EVAL case survives verbatim under a clearer name; the two `tests-integration/` edits are comment-only. Net coverage is a superset.

### Suite state on this head

`tests/**` 340/340 · `test:integration:mock:node-redis` 526/526 on **three** consecutive runs · `test:integration:mock:ioredis` 475/475 · `tsc --noEmit`, `eslint .`, prettier all clean. Leaving the earlier 525/1 as unexplained is the right call — six runs here plus nine there, no recurrence, nothing captured.

### On deferring the `mSet`/`zUnionStore` widening

**Agreed, and it's the right call — but the follow-up needs rewording before someone picks it up.** `const _: RedisClientType = mockClient` cannot compile, no matter how complete the surface is: `RedisClient` declares `#private` plus `private _self`, `private _commandOptions`, `private reAuthenticate` and `private _commandOptionsProxy` (`@redis/client/dist/lib/client/index.d.ts:204,213-214,248,250`), and TypeScript checks private and `#private` members **nominally** — no class outside that inheritance chain is ever assignable. Independently, `RedisClientType` defaults to `RESP=3` and types replies through `TYPE_MAPPING` (`BlobStringReply | NullReply`, not `string | null`), so even the methods that do exist would mismatch this deliberately RESP2-shaped facade.

What *does* work is structural: `Pick<>` drops private members from `keyof`, so

```ts
const _: Pick<RedisClientType, 'mSet' | 'zUnionStore' | 'get' | 'set'> = client
```

compiles and pins exactly the curated surface, one entry per method. File the follow-up as **"add a `Pick<RedisClientType, …>` conformance assertion in `tests-package/`"**, not as assigning the mock to `RedisClientType` — otherwise the next person burns a session discovering it can't work.

### New findings

| sev | location | one-liner | lens |
|---|---|---|---|
| med | `tests/core/decode-redis-value.test.ts:15` | header promises a tripwire the file doesn't install — 3 of 5 option flips leave 340/340 green | arch |
| med | `tests/client-mocks/node-redis-mock.test.ts:228` | the double-`quit()` pins a divergence: real node-redis throws `ClientClosedError` and emits `'end'` once | redis |
| med | `tests/client-mocks/node-redis-mock.test.ts:381` | **pre-existing, non-blocking** — assertion is correct for the `redis-8.0` default, but `ClusterPolicy` wrongly rejects `BY nosort` / any non-glob `BY`, and these rules aren't version-gated | redis |
| low | `src/client-mocks/node-redis-mock.ts:789` | a raw `SUBSCRIBE` now wedges `masters[0]`'s cached session, taking every keyless command and all of master-0's key range with it | redis |
| low | `src/client-mocks/node-redis-mock.ts:925` | `toBuffer`'s `number` rejection is real; arrays and TypedArrays still slip through where node-redis throws | both |
| low | `src/client-mocks/node-redis-mock.ts:503` | `this.session.close()` sits above the new `try`, so the timer-leak invariant is only half-enforced | arch |

<sub>Ground-truthed against `redis-server` 7.2.1 (cluster-enabled, all slots owned) and, for the `redis-8.0` default profile, a disposable `redis:8.0` container (`Redis server v=8.0.6`, no published ports, removed after). Mock behaviour was exercised from read-only exports of `origin/main` and `0baa5b2`; the shared working tree was never modified.</sub>

PR #391, opened 2026-09-21
fix(demo): resolve vite peer conflict breaking demo deploy

fatal10110:
## What's broken

`Deploy browser demo` has failed on **every push to `main` since 2026-07-21** — 7 consecutive runs, ~2 months. The GitHub Pages site has been frozen at the last green deploy, `c6ad045` (2026-07-02).

| date | commit | result |
| --- | --- | --- |
| 2026-07-02 | `c6ad045` | ✅ last success |
| 2026-07-21 | `7b128bd` | ❌ first failure |
| 2026-09-09 | `35ac0bf` | ❌ |
| 2026-09-21 | `268b57c`, `e9b6a15`, `f2dabe7`, `cad826e`, `822c833` | ❌ |

It fails at `npm ci` in `examples/browser-demo`, before the build ever runs:

```
npm error code ERESOLVE
npm error While resolving: vite-plugin-node-polyfills@0.23.1
npm error Found: vite@8.1.5
npm error Could not resolve dependency:
npm error peer vite@"^2.0.0 || ^3.0.0 || ^4.0.0 || ^5.0.0 || ^6.0.0 || ^7.0.0" from vite-plugin-node-polyfills@0.23.1
npm error Conflicting peer dependency: vite@7.3.6
```

## Root cause

`7b128bd` — "chore(deps): bump esbuild (#352)", a Dependabot security bump — jumped `vite` two majors to get off the vulnerable `esbuild`:

```diff
-    "vite": "^6.0.0",
-    "vite-plugin-node-polyfills": "^0.23.0"
+    "vite": "^8.1.5",
+    "vite-plugin-node-polyfills": "^0.23.1"
```

`vite-plugin-node-polyfills@0.23.1` declares peer support only up to `vite ^7`. Dependabot wrote a lockfile pinning `vite@8.1.5` anyway (it regenerates locks without enforcing peer ranges), so the committed lockfile has been un-installable by a plain `npm ci` from the moment it landed. The `vite@7.3.6` in the error message is npm's own suggestion for what would satisfy the plugin's peer range — it is not what the lockfile holds. The bump itself was never wrong; it was just never reconciled with the plugin.

Worth noting: `0.23.1` is published to npm as **deprecated — "this version is broken"**. It has been the pinned version for two months.

## The fix

Bump `vite-plugin-node-polyfills` to `^0.28.0`. `0.26.0` was the first release to add `^8.0.0` to its peer range; `0.28.0` (2026-05-18) is the current release and the plugin is actively maintained.

`vite` stays on `8.1.5` — the resolved tree is otherwise byte-identical. The whole lockfile diff is the one plugin entry:

```diff
-      "version": "0.23.1",
-      "deprecated": "this version is broken",
+      "version": "0.28.0",
-        "vite": "... || ^6.0.0 || ^7.0.0"
+        "vite": "... || ^6.0.0 || ^7.0.0 || ^8.0.0"
```

No `--legacy-peer-deps`, no `--force`, no workflow changes.

### Alternatives rejected

- **Pin `vite` back to `^7`.** Would undo a security bump and walk the demo backwards onto an older major, just to accommodate a deprecated plugin version. Strictly worse when a supported plugin release exists.
- **Drop the plugin entirely.** Not viable. The demo imports `../../src`, and `src/` uses `Buffer` in 87 files plus `node:crypto`, `node:events` and `node:stream`. Vite 8 does not polyfill any of those for the browser. `dist/assets/index-*.js` contains 65 `allocUnsafe` / 88 `isBuffer` / 76 `nextTick` occurrences — the polyfills are load-bearing, not vestigial.
- **`--legacy-peer-deps` / `--force` in the workflow.** Would turn the check green while leaving the demo on a plugin its own author marked broken, and would silently accept whatever peer mismatch lands next. The point of this PR is a trustworthy signal, not a green square.

The plugin's `shims/{buffer,global,process}/dist/index.js` paths that `vite.config.ts` aliases are unchanged in `0.28.0`, so the config needed no edits.

## Verification

Everything below was run on this branch, from a clean `node_modules`.

**Install and build**

- `npm ci` in `examples/browser-demo`: clean, 165 packages, no `ERESOLVE`. Verified with both npm 10.9.0 and **npm 11** (what the workflow's `node-version: 24` ships).
- `npm run build`: ✅ 368 modules, built in 547ms.
- Output is real, not an empty bundle:
  ```
  dist/index.html                     4.12 kB │ gzip:   1.38 kB
  dist/assets/index-kHJ-D0s7.css      2.85 kB │ gzip:   0.77 kB
  dist/assets/index-ClX6bmqE.js   1,325.07 kB │ gzip: 351.73 kB
  ```
  `dist/index.html` references both assets under the correct `/js-redis-server/` base.

**Runtime — the demo actually runs**

A bundle that builds can still throw the moment it touches `Buffer` or `process`, so the built `dist/` was served with `vite preview` and driven headlessly in a real browser.

Single-instance mode:

| command | reply |
| --- | --- |
| `SET hello world` | `"OK"` |
| `GET hello` | `"world"` |
| `HSET h a 1 b 2` | `(integer) 2` |
| `HGETALL h` | `1) "a" 2) "1" 3) "b" 4) "2"` |
| `LPUSH q job` / `LRANGE q 0 -1` | `(integer) 1` / `1) "job"` |
| `INCR counter` | `(integer) 1` |
| `TYPE h` | `"hash"` |
| `APPEND bin éé` | `(integer) 4` |
| `STRLEN bin` | `(integer) 4` |
| `EVAL "return redis.call('GET', KEYS[1])" 1 hello` | `"world"` |
| `EVAL "return {1,2,'three'}" 0` | `1) (integer) 1  2) (integer) 2  3) "three"` |

`APPEND bin éé` → `4` is the specific Buffer check: `éé` is 2 UTF-16 characters but 4 UTF-8 bytes, and the polyfilled `Buffer` counted bytes, not characters. If the buffer shim had degraded to a string-ish stand-in this would have returned `2`. The `EVAL` cases additionally prove `node:crypto` works (SHA1 script cache) and that the `lua-redis-wasm` WASM + glue still load from jsDelivr at the version `vite.config.ts` injects.

Cluster mode (`+ Cluster`) also works — 3 masters at `0-5460` / `5461-10921` / `10922-16383`:

- `SET foo bar` → `-MOVED 12182 master-2`, followed, `"OK"`
- `GET foo` → `"bar"`
- `MSET k1 a k2 b` → `(error) CROSSSLOT Keys in request don't hash to the same slot`

**Browser console: zero messages of any kind across the whole session — no errors, no warnings.**

**Repo-level, since the lockfile change could touch the workspace**

- root `npm ci`: clean
- `npm run lint`: ✅ clean
- `npm test`: ✅ **342 passed, 0 failed** (55 suites)
- `examples/browser-demo/format.test.ts`: ✅ 12 passed

## Notes / things left alone

- `npm audit` in the demo reports 9 advisories (`elliptic`, `qs`, `nanoid`, `postcss`). All of them are **pre-existing and unchanged by this PR** — I diffed the old and new lockfiles and every one of those packages resolves to the identical version before and after. They belong in their own Dependabot PR, not bundled into a deploy fix.
- `examples/browser-demo/format.test.ts` is not run by any workflow. Unrelated to this fix, but it is a test nobody is executing.
- The build emits two pre-existing warnings — a direct `eval` in `vm-browserify` and a >500 kB chunk. Both predate this change and neither affects the demo.


examples/browser-demo/package-lock.json
```diff
@@ -2764,21 +2764,20 @@
       }
     },
     "node_modules/vite-plugin-node-polyfills": {
-      "version": "0.23.1",
-      "resolved": "https://registry.npmjs.org/vite-plugin-node-polyfills/-/vite-plugin-node-polyfills-0.23.1.tgz",
-      "integrity": "sha512-hR2YlaMqVo2f/0w6V9YgaeKsIy8z5gEsMFHX8MQlh8ATEAbfI4EWjE6qnCvkZ9lyaXKoJD2rAwL/up8/Rzsjkg==",
-      "deprecated": "this version is broken",
+      "version": "0.28.0",
```
fatal10110, line 2767: **Supply chain: verified clean.** Post-merge review, so this is a record rather than a gate — but a dependency bump merged without review is exactly where a bad package lands, so I checked the registry rather than taking the bump on faith.

| check | result |
| --- | --- |
| `0.28.0` is genuinely current | ✅ `dist-tags.latest = 0.28.0`, published **2026-05-18**, 4 months before this PR — not a freshly-cut version |
| maintainer continuity | ✅ same single publisher for both: `_npmUser` on `0.23.1` **and** `0.28.0` is `davidmyersdev <davidrmyersii@gmail.com>`. No ownership transfer, no new maintainer |
| release cadence | ✅ unbroken `0.23.1 → 0.24.0 → 0.25.0 → 0.26.0 → 0.27.0 → 0.28.0`, same author since 2022. No suspicious jump into the bumped version |
| integrity pin | ✅ lockfile `sha512-NXct/ci2ef…Xu7+w==` matches the registry's `dist.integrity` char-for-char |
| registry signature | ✅ present on both versions |
| new transitive deps | ✅ **none** — deps are the same two packages (`@rollup/plugin-inject`, `node-stdlib-browser`); only the `node-stdlib-browser` *range* moved |
| payload size | ✅ `unpackedSize` **shrank** 750,486 → 541,256 bytes (27 → 33 files). No blob added |

No provenance attestation on either version — worth knowing, but not a regression introduced here.

I also reproduced the build independently from a clean `npm ci` in an isolated export of this commit, and it is **byte-identical** to what the description reports: `index-ClX6bmqE.js` at 1,325,070 bytes and `index-kHJ-D0s7.css` at 2,853 bytes — same content hashes, same sizes. Reproducible build.

Nothing here warrants a revert.

examples/browser-demo/package.json
```diff
@@ -14,6 +14,6 @@
   },
   "devDependencies": {
     "vite": "^8.1.5",
-    "vite-plugin-node-polyfills": "^0.23.1"
+    "vite-plugin-node-polyfills": "^0.28.0"
```
fatal10110, line 17: **The `npm audit` advisories are not "dev-only" in the sense that matters.** (medium — follow-up issue, not a revert)

Leaving the 9 advisories out of this PR was the right call, and the equivalence claim checks out exactly — I diffed both lockfiles:

```
package                before       after        dev    same
elliptic               6.6.1        6.6.1        True   True
qs                     6.15.2       6.15.2       True   True
nanoid                 3.3.16       3.3.16       True   True
postcss                8.5.21       8.5.21       True   True
```

`npm audit` on the merged state: `{low: 6, moderate: 2, high: 1, total: 9}` — unchanged by this PR.

The part worth flagging is the **justification**, not the decision. These are `dev: true` because the plugin is a devDependency — but `vite-plugin-node-polyfills` is a *bundler* plugin: it inlines its dependency graph into the production artifact. `elliptic` really is in the file served to every visitor of the Pages site. From the built bundle:

```
$ grep -oc elliptic dist/assets/index-ClX6bmqE.js   →  9
$ grep -oE "(secp256k1|p256|p384|curve25519)" … | sort | uniq -c
      1 curve25519
      5 p256
      6 p384
      6 secp256k1
```

Those are real curve tables, not an incidental string match — `brorand` (5 hits) is in there too. `vite.config.ts` calls `nodePolyfills()` with no `include`, which means *every* builtin gets polyfilled; `src/` imports `node:crypto` (script-cache SHA1, hyperloglog), so the whole `crypto-browserify → browserify-sign/create-ecdh → elliptic` chain gets pulled in.

So the advisory chain is client-side here, and `npm audit`'s `dev` label will keep understating it. Note the chain now also flags the plugin itself:

```
low  vite-plugin-node-polyfills  via=['node-stdlib-browser']
low  node-stdlib-browser         via=['crypto-browserify']
low  elliptic                    via=['Elliptic Uses a Cryptographic Primitive with a Risky Implementation']
```

Not exploitable today — the demo only ever calls `createHash` (sha1/sha256), never ECDSA verify — so this is not urgent. But the fix is cheap and also sheds ~100 kB from a bundle that already trips the 500 kB warning: `nodePolyfills({ include: ['buffer', 'crypto', 'stream'] })`, then re-check what lands.

Worth an issue. Also: the comment at the top of `vite.config.ts` says it polyfills "the one Node builtin … (buffer)", which is not what the code does and not true of `src/` either.

fatal10110, line 17: **The real finding is that nothing stops this recurring.** (medium — one follow-up PR covers all three)

This PR fixes the breakage correctly. It does not close the hole that let a broken lockfile sit on `main` for two months and seven red deploys, and there are three separate contributing gaps:

**1. No workflow builds the demo on `pull_request`.** `ci.yml` has zero references to `browser-demo` (`git show pr-391:.github/workflows/ci.yml | grep -c browser-demo` → `0`); every job runs only root `npm ci` and root scripts. `deploy-demo.yml` is the only thing that builds the demo, and it is `on: push: branches: [main]`. So the failure mode is structural: a PR touching `examples/browser-demo/package-lock.json` goes green on all of CI, merges, and *then* turns the deploy red — post-merge, on the protected branch, with no PR to revert cleanly. That is exactly the `7b128bd` story. This is also why #391 had to be merged unreviewed to test it, so fixing it pays for itself twice.

**2. No `.github/dependabot.yml` exists.** `git ls-tree -r --name-only pr-391 .github/` returns only the three workflow files. The root cause here was Dependabot regenerating this lockfile without honouring peer ranges — and `7b128bd` is authored by `dependabot[bot]` with a 1350-line diff to `examples/browser-demo/package-lock.json`, a directory it was never explicitly configured for. This will recur at vite 9: `0.28.0` peers top out at `^8.0.0`, and `^0.28.0` on a `0.x` version only admits `0.28.x` patches, so there is no automatic escape hatch. Grouping `vite` + `vite-plugin-node-polyfills` so they move together is the durable fix.

**3. `deploy-demo.yml`'s `paths:` filter omits the root manifests.** The filter is `['examples/browser-demo/**', 'src/**', '.github/workflows/deploy-demo.yml']`, but the job then runs root `npm ci` — by its own comment, because the demo imports `../../src`. A root-only dependency bump therefore changes what the demo bundles without triggering a redeploy, and the *next* unrelated `src/` commit picks it up, so a breakage gets attributed to an innocent commit. Worth noting for `lua-redis-wasm` specifically: `vite.config.ts` reads the CDN version from the **demo's** `node_modules` while `main.ts`'s `../../src` graph resolves the loader from the **root** tree. Both are `1.5.0` today, but nothing enforces that they agree.

Smallest fix for all three: a `demo` job in `ci.yml` (`npm ci` at root + in the demo + `npm run build`), gated on `paths` so it doesn't slow every PR; a `dependabot.yml` with a `/examples/browser-demo` npm entry and a vite group; and `package.json`/`package-lock.json` added to the deploy `paths:` filter.

examples/browser-demo/package-lock.json
```diff
@@ -2764,21 +2764,20 @@
...
+      "resolved": "https://registry.npmjs.org/vite-plugin-node-polyfills/-/vite-plugin-node-polyfills-0.28.0.tgz",
+      "integrity": "sha512-NXct/ci2ef4fRyCfTb8fk2HmR80Rv7icLd+cRH41TnUugDzdKMFKqFPpZYCFUInZMMem9bkLv5pkq02+7Xu7+w==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
         "@rollup/plugin-inject": "^5.0.5",
-        "node-stdlib-browser": "^1.2.0"
+        "node-stdlib-browser": "^1.3.1"
```
fatal10110, line 2774: **Lockfile diff verified — with one correction and one near-miss worth recording.** (low)

I parsed both lockfiles and compared the `packages` maps key-by-key rather than reading the rendered diff:

```
lockfileVersion: 3 -> 3
pkg count:       196 -> 196
ADDED:           none
REMOVED:         none
CHANGED entries: 2 of 196
```

**Correction:** it is **two** entries, not "exactly one" — `packages[""]` (the root `devDependencies` pointer) and `packages["node_modules/vite-plugin-node-polyfills"]`. Zero behavioural impact; I only flag it because the "exactly one entry" claim was offered as part of the reason review could be skipped, and a future reader auditing that decision should find the evidence matching. The substance of the claim — nothing added, nothing removed, no transitive resolution moved — is correct.

**The near-miss:** this line moves the `node-stdlib-browser` range `^1.2.0` → `^1.3.1`, and because no other lockfile entry changed, the pin did **not** move with it. That is precisely the shape of a latent `npm ci` failure — a pin outside its own declared range. It happens to be fine:

```
installed node-stdlib-browser: 1.3.1   satisfies ^1.3.1 -> true
```

The pin was *already* at `1.3.1`, so the widened range is satisfied exactly. Worth stating explicitly since it was luck rather than design, and it is the one thing in this diff that could have silently broken `npm ci` again.

**npm-version sensitivity.** The npm 11 decision was correct and I verified the stated reason rather than taking it: I copied the committed manifests into a scratch dir and ran `npm install --package-lock-only` under npm 10.9.0 — all 10 `libc` fields were stripped, and those 10 entries were the *only* difference (no version drift). The other direction is safe too: `npm ci` under npm 10.9.0 against the committed lockfile exits 0, installs 165 packages, no `ERESOLVE`, and leaves the lockfile byte-identical. So CI is fine on either npm.

The residual snag is contributor-side: root `engines` is `node: >=22`, and Node 22 ships npm 10. Anyone on the declared minimum who runs `npm install` here produces a spurious 10-entry diff that the next person on Node 24 flips back. A root `.nvmrc` with `24` (matching both workflows' `node-version: 24`) would settle it.

examples/browser-demo/package.json
fatal10110, line 17: **Runtime evidence reproduced independently — and `format.test.ts` is dead in three ways, not one.** (low)

First, the runtime claims. I exported this commit to an isolated tree, ran a clean `npm ci` + `npm run build`, served `dist/` and drove it in a real browser. The two checks that actually prove the polyfills survived the bump:

```
redis> APPEND bin éé
(integer) 4

redis> EVAL "redis.call('SET',KEYS[1],'world') return {1,2,redis.call('GET',KEYS[1])}" 1 hello
1) (integer) 1
2) (integer) 2
3) "world"
```

Both match real Redis 7.2.1, ground-truthed on a local server: `APPEND bin éé` → `:4\r\n`, and the Lua table → `*3\r\n:1\r\n:2\r\n$5\r\n…`. I confirmed the characters the terminal actually received were NFC `U+00E9 U+00E9` (2 UTF-16 units, 4 UTF-8 bytes) — so `4` is the byte count and a string stand-in would have returned `2`. That single `EVAL` also exercises the `node:crypto` SHA1 script cache, the jsDelivr WASM load, and the `redis.call` bridge in both write and read directions. **Zero console messages** across the session, matching the description.

So the polyfills are intact and load-bearing — `Buffer` ×1228, `isBuffer` ×88, `allocUnsafe` ×65, `nextTick` ×76 in the built bundle (the last three match the description exactly).

Second, `format.test.ts`. Confirmed it passes and confirmed nothing runs it — but it is worse than "no workflow runs it":

- **Never executed**: root `test` script globs `"./tests/**/*.test.ts"`, rooted at `./tests`, so `examples/` is unreachable.
- **Never typechecked**: root `tsconfig.json` has `"include": ["src/**/*.ts"]`, so `tsc --noEmit` never sees it.
- **Never linted**: `eslint.config.mjs:17` ignores `'examples/**'`.

Run manually it is green — `# tests 12 / # pass 12 / # fail 0`, and it is on-convention (`node:test` + `node:assert`, no Jest/Chai). But `format.ts` is the console's reply formatter: edit it, break the contract, and the failure ships straight to Pages with nothing objecting. A type error there is equally invisible, since vite/rolldown transpiles without typechecking.

Cheap fix, and it rides along free with the `demo` CI job proposed in the other comment: `node --import tsx --test "examples/browser-demo/**/*.test.ts"`.

PR #392, opened 2026-09-21
chore: fail CI when a /core export disappears

### LINKED ISSUE #386: assertCore does not catch /core export removals; no CHANGELOG for breaking-change notes
Process gap noticed while reviewing PRs #374, #375 and #376, each of which removes or changes symbols on the published `js-redis-server/core` subpath.

## Problem
`assertCore` (the guard intended to protect the `/core` surface) does not catch **removals** of exports from `src/internal.ts`. PR #375 deletes the `RedisKeyspace` and `WrongRedisTypeError` re-exports and CI stays green; #374 removes an option from exported option types; #376 changes six interface methods to two on `RedisClientSession`.

Each PR body *does* flag the break — but that relies on an author remembering. The repo has no CHANGELOG and releases are manual, so a correctly-written PR note can still evaporate before it reaches release notes.

## Proposal
Two independent pieces, either useful alone:

1. **Make the guard detect removals.** Snapshot the `/core` (and root) export surface — e.g. generate `dist/core.d.ts` and diff the exported symbol list against a committed baseline, failing CI when a symbol disappears without the baseline being updated in the same commit. `tests-package/` is the natural home.
2. **Add a CHANGELOG** (or changesets), so a breaking-change note has somewhere durable to live between merge and publish.

## Acceptance
A PR that deletes an export from `src/internal.ts` without updating the baseline fails CI with a message naming the removed symbol.

fatal10110:
Closes #386

Two independent pieces, as the issue proposes.

> **Updated after review** (`0b5f96f`). The union gap is closed rather than disclosed, the diff logic is now unit-tested, and a degraded extraction now throws instead of returning a plausible-looking thin surface. Two claims in the original description were wrong and are corrected below: the scoreboard is **two** of four, not three, and the "no class statics" gap did not exist. Details in the review threads.

## 1. A guard that detects removals

`assertCore` (`tests-package/dual-format.test.ts:55`) spot-checks seven names — six that must exist and one that must not. Anything outside that list can disappear with CI green. That is how #374, #375, #376 and #377 each shipped a `js-redis-server/core` break on a green build.

New: `tests-package/export-surface.test.ts` snapshots the **full** export surface of both published entry points into `tests-package/export-surface.json` and diffs it on every `npm run test:package`. `assertCore` is untouched — its behavioral assertions (the facade is *not* on `/core`, the executor is *not* on the root, `SET`/`GET` actually work through the resolved package) are things a symbol list cannot express.

### Declarations, not runtime

The surface is read out of the emitted `dist/index.d.ts` / `dist/core.d.ts` with the TypeScript compiler API, not by enumerating the built JS.

`Object.keys(require('js-redis-server/core'))` sees only **value** exports. Most of `/core` is type-only — every `*Options` type, `RedisClientSession`, `CommandPlan`, `CommandFlag` — and that is where two of the four breaks lived. The declaration bundle is the only artifact carrying both. The runtime is still checked, but in the other direction: two tests assert that every symbol the *current declarations* mark `value` is actually present on the built CJS and ESM modules, so a declaration that promises a binding the bundler dropped is caught on the same run that introduces it.

### What the baseline records

Per exported symbol:

| field | what it pins |
| --- | --- |
| `kind` | `value` (has a runtime binding) or `type`. A `value` → `type` downgrade is breaking: it still type checks at the import site, but the binding is gone. |
| `members` | Own properties and methods of an exported interface, class, object type alias or `const` namespace. Statics are prefixed `static:`. |
| `variants` | Literal constituents of a union, so dropping `'noscript'` from `CommandFlag` or a key from `FeatureId` reads as a removal. |

Inherited members are deliberately not walked — otherwise 120-odd `*Error` subclasses each snapshot `message`/`stack`/`cause` out of `lib.es5.d.ts` and the baseline churns on every TypeScript bump. `RedisCommandError` still pins its own `code`.

`const` members resolve through the checker, gated on the type being anonymous or mapped. That is what makes `FEATURE_GATES: Record<FeatureId, VersionGate>` pin its 20 gate keys, while keeping `getCommand: CommandDefinition<…>` and the command arrays from dragging in interface members that are snapshotted under their own name, or `Array.prototype` noise.

### Why below symbol level

**Two** of the four breaks were invisible to a symbol-list-only snapshot. (The original description said three; #377's merge commit also removed `RedisTurnQueue` and `createNoopParkHandler` as top-level exports, so a plain list catches it.)

| PR | removed from `/core` | symbol list | this guard |
| --- | --- | --- | --- |
| #375 | `RedisKeyspace`, `WrongRedisTypeError` | caught | caught |
| #377 | `RedisTurnQueue`, `createNoopParkHandler` | caught | caught |
| #377 | `CommandRegistry.override/has/getNames`, `CommandPlan.flags`, `ClientSessionOptions.turnQueue` | **missed** | caught |
| #376 | six `RedisClientSession` pub/sub methods → two | **missed** | caught |
| #374 | `encoder` from four transport option types | **missed** | caught |

#374 and #376 carry the argument on their own, and member level also catches *more* of #377 than a symbol list would.

### Additions

**Additions fail the build**, in a separate test with different wording: "Additions are NOT a breaking change; this is a bookkeeping failure."

Explicit bump rather than silent self-update, for three reasons:

1. A baseline that does not grow is a hole. A symbol added in PR A never lands in the baseline, so deleting it in PR B is invisible — the exact gap this exists to close.
2. Silent self-update only works locally. In CI the write is discarded, so the failure mode becomes "author forgot to commit the regenerated file" and nobody finds out until the removal it was supposed to catch.
3. The baseline diff is the review signal. "This PR widens the published API" belongs in the diff, not only in the PR body.

Cost is one command (`npm run export-baseline`) on a deliberate, rare action.

### Failing loudly when extraction degrades

A half-resolved program is the dangerous failure: it returns a plausible-looking symbol list with members and type-only kinds quietly missing, and `skipLibCheck` — which all-declaration inputs need — suppresses the `TS2307` that would reveal it. Reproduced: copying `core.d.ts` away from its shared chunk gave 267 symbols with 49 of 67 member-sets and 51 type-only kinds silently gone.

`readExportSurface` now enforces two structural, diagnostic-independent invariants: zero exports for an entry point, and any exported symbol whose declaration did not resolve. On that same degraded input:

```
--- healthy ---
OK symbols 266 withMembers 71 kind=type 65
--- degraded (chunk missing) ---
THREW: …/core.d.ts: 180 exported symbol(s) have no resolvable declaration, so the
extracted surface would be wrong. Rebuild with `npm run build`. First few:
AtLeastOneInputKeyError, BitOffsetError, BitOpNotSingleKeyError, …
```

An empty entry throws `resolved to zero exports — the build is broken`. This matters most for `npm run export-baseline`, which would otherwise happily commit the gutted surface.

### The guard's own diff is pinned

`findRemovals` / `findAdditions` live in `export-surface.ts` and are unit-tested in `export-surface-diff.test.ts` against synthetic surfaces — 13 cases covering symbol removal, member removal, a static distinguished from its instance twin, variant removal, `value`→`type` downgrade, an entry losing its whole `members` list, a rename producing both results, and the empty-baseline path.

Every assertion in the suite-level test is `deepStrictEqual(diff, [])`, which passes just as happily when the producer is broken. Two neuterings, both on the current tree:

**Invert the `missing` predicate** → 16 of 27 fail, 12 of them in the diff unit test.

**Delete the member-removal loop, keep the symbol loop** → 3 fail, *all three in the diff unit test*:

```
    not ok 3 - names a deleted member of a surviving symbol
    not ok 4 - distinguishes a deleted static from the instance member
    not ok 7 - treats a member losing its whole list as removal, not as unchanged
not ok 3 - findRemovals
# tests 27
# pass 24
# fail 3
```

Both surface-level tests stayed green — nothing is removed on this tree, so they had nothing to notice. That is the vacuous-guard failure, and the unit test is the only thing that sees it.

### What it does **not** cover

- **Type shape.** Members and variants are pinned by *name only*. Narrowing `logger?: Logger` to `logger: Logger`, widening a parameter, changing a return type, or reordering an overload set all pass. A full `.d.ts` text diff would catch it but would churn on every JSDoc edit and every tsup/TypeScript bump. Name-level is where signal stayed well above noise.
- **Generic parameters.** Adding a required type parameter to an exported generic is breaking and invisible here.
- **Union arms, individually.** A union records the *union* of its constituents' members, so moving a field from one arm to another is invisible. Deleting it outright is not.
- **Behavior.** A method that keeps its name and drops its effect is `assertCore`'s and the integration suites' problem.
- **`dist/cli.js`.** Not an importable surface.

~~Class statics~~ — the original description listed statics as uncovered. That was wrong: `ts.isClassDeclaration` walks `declaration.members`, which includes them. They were already captured, just flattened into the instance namespace, which let a `static create()` mask an instance `create()`. They are now prefixed `static:`, so both are pinned.

`.d.mts` is read alongside `.d.ts` and compared as a **full surface**, not just symbol names — the baseline itself is only ever read from `.d.ts`, so this is the sole thing pinning the ESM declarations at member and kind level.

### Red-run evidence

A guard nobody has seen fail is not a guard. Three removals from `main`'s source, rebuilt:

```
- export type { RedisTurnHandle } from './core/turn-queue'   # src/internal.ts:59
- export { SerialTurnQueue } from './core/turn-queue'        # src/internal.ts:111
- readonly pubsubShardChannelCount: number                   # RedisClientSession
```

plus a fourth edit exercising the union and annotated-const paths — dropping the `cluster.multi-db` feature gate (its `FeatureId` arm, its `FEATURE_GATES` entry, and its one call site in `cluster-policy.ts`).

> The original description used `RedisClientSession.pubsubSubscribe` for the member case. That is not reproducible as written: it has three internal callers, so the **build** fails before the guard runs. `pubsubShardChannelCount` has no caller through the interface, so the build succeeds and the guard is what fails. Corrected.

`npm run test:package`:

```
        not ok 1 - no exported symbol, member or variant was removed
            BREAKING: 1 export(s) disappeared from "js-redis-server".
              - FeatureId — union variant 'cluster.multi-db' gone

        not ok 1 - no exported symbol, member or variant was removed
            BREAKING: 5 export(s) disappeared from "js-redis-server/core".
              - FEATURE_GATES.cluster.multi-db — member gone
              - FeatureId — union variant 'cluster.multi-db' gone
              - RedisClientSession.pubsubShardChannelCount — member gone
              - RedisTurnHandle (type) — gone
              - SerialTurnQueue (value) — gone
```

Every category in one run: a value symbol, a type symbol, an interface member, a union variant, an annotated-`const` key — and the **root** entry point firing too, not just `/core`.

**All four of `dual-format.test.ts`'s tests stayed green throughout** — that is the gap, reproduced.

Additions, verified separately by appending one export to `src/internal.ts`:

```
    not ok 2 - baseline records every current export
      error: |-
        The export baseline is stale for "js-redis-server/core" — 1 addition(s).
          - RedisKeyspaceProbe (value)
        Additions are NOT a breaking change; this is a bookkeeping failure. Run
        `npm run export-baseline` and commit tests-package/export-surface.json so
        the next PR that deletes one of these is caught.
```

All source edits were reverted; nothing from any red run is on this branch.

## 2. CHANGELOG

`CHANGELOG.md`, [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) format, plain Markdown.

**Why not changesets.** Releases here are a hand-cut `v<version>` tag that `release.yml` asserts against `package.json`, and the repo publishes two names from one tree. Changesets wants to own the version bump, the tag and the changelog generation; adopting it means a new dependency, a CI job and a different release procedure, to solve a problem that is currently "there is nowhere to write it down". A file costs nothing and fits the process that already exists. If releases ever get automated, changesets can consume this file's history.

The `Unreleased` section covers everything since `v0.3.0` and is written from the PR bodies — #376's six-method migration block and #377's replacement table quoted as-is. Two corrections from review:

- #377's "No wire-visible change" was not unconditional. `Resp2SessionAdapter.writeError` passes a `RedisCommandError` (and a `Resp2ParseError`) straight through, and `ClientSession` now uses `signal.throwIfAborted()`, which rethrows `signal.reason` verbatim — so `controller.abort(new WrongTypeRedisError(...))` puts `-WRONGTYPE …` on the wire where it previously produced `-ERR internal server error`. The entry now says which reasons are masked and which are not.
- `pmessage` is a **4**-element push frame (`pmessage`, pattern, channel, payload), not three. The "three" came from the internal `items` array before `pubsubFrame` prepends the name. Reworded so a contributor writing a conformance test from the CHANGELOG does not encode a regression.

`CHANGELOG.md` is added to `package.json` `files`. `CONTRIBUTING.md` documents the three recorded fields and the release-time `Unreleased` rename.

## Prettier and Markdown

`npm run format` globs `js|ts|json` and CI runs `prettier --check "**/*.{js,ts,json}"`, so Markdown is checked by neither. `.prettierignore`'s last line is `.md` — which matches a file *named* `.md`, not `*.md`, so it is almost certainly a typo and currently does nothing.

On `main`, `prettier --check "**/*.md"` fails on 6 files. On this branch it fails on exactly the same 6 — `CHANGELOG.md` is prettier-clean. I deliberately did **not** run `prettier --write` over `CONTRIBUTING.md`: it would have reformatted pre-existing prose and code blocks I did not touch, trading a clean diff for a cosmetic win. Widening the `format` glob or fixing the `.prettierignore` typo reformats those files and is a separate call for the owner.

## Testing

| Suite | Result |
| --- | --- |
| `npm run lint` | pass |
| `npm test` (incl. `tsc --noEmit`) | pass — 342/342 |
| `tsc -p tsconfig.tests-package.json` | pass (new) |
| `npm run build` | pass |
| `npm run test:package` | pass — 27/27 (was 4 on main) |
| `npm run test:package` as `js-valkey-server` | pass — 27/27 |
| `npm pack --dry-run --ignore-scripts` | pass — `CHANGELOG.md` included |
| `npm run test:integration:mock:node-redis` | pass — 526/526 |

The `js-valkey-server` leg matters because the CI `packages` matrix renames the package before running `test:package`; the guard reads `dist/*.d.ts` (name-independent) and resolves the runtime through `package.json`'s own `name`.

The `:real` suites were not run locally (they need the docker-compose Redis cluster) — CI ran them green on the previous push.

`npm run export-baseline` is idempotent and formats its output through the repo's own Prettier config, so the committed JSON passes `prettier --check`.

## Files

```
 CHANGELOG.md                            | new
 CONTRIBUTING.md                         | +38 -2
 package.json                            | +4
 tsconfig.tests-package.json             | new  (type-checks tests-package/)
 tests-package/export-surface.ts         | new  (extractor + diff)
 tests-package/export-surface.test.ts    | new  (the guard)
 tests-package/export-surface-diff.test.ts | new (pins the diff itself)
 tests-package/export-surface.json       | new  (baseline: 91 root + 266 /core)
 tests-package/update-export-baseline.ts | new  (regenerator)
```


tests-package/export-surface.ts
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  if (ts.isIntersectionTypeNode(node)) {
+    // `NodePipeline & { host: string; port: number }` contributes the literal's
+    // own members; the referenced half is snapshotted under its own name.
+    return node.types.flatMap(membersOfTypeNode)
+  }
+
+  // Unions, type references, mapped types, keyof, … have no own members to pin.
+  return []
```
fatal10110: **HIGH — union type nodes silently void member coverage, including root-entry public option types.**

`membersOfTypeNode` handles `isIntersectionTypeNode` by flatMapping constituents (line 83), but the structurally identical union case falls through to this `return []`. The PR frames this as a narrow "no union members" gap about string-literal unions. It is much wider: **any union-typed export loses 100% of its member coverage.**

The committed baseline shows the asymmetry directly:

```
CreateIoredisMockOptions {"kind":"type"}                      <-- union, NO members
CreateRedisMockOptions   {"kind":"type","members":["basePort","cluster",
                          "compatibility","databaseCount","logger","port"]}
```

`CreateIoredisMockOptions` is a **root entry point** public option type on `createIoredisMock()` (src/client-mocks/ioredis-mock.ts:21):

```ts
export type CreateIoredisMockOptions =
  | { cluster?: false; databaseCount?: number; seed?: readonly SeedEntry[] }
  | { cluster: { masters: number; ... }; seed?: readonly SeedEntry[] }
```

Delete `seed` and every consumer passing `{ seed: [...] }` breaks at compile time — and this guard stays green, because there is no `members` array to compare. Unlike `CreateInMemoryClientOptions` (an intersection of separately-exported refs), these constituents are anonymous literals exported nowhere else, so there is no indirect coverage.

Fully uncovered today — object-literal unions: `CompatibilitySpec`, `CreateIoredisMockOptions`, `CreateNodeRedisMockOptions`, `SeedEntry`, `ExpirationState`, `RedisMutationEvent`. String-literal unions: `CommandFlag`(12), `FeatureId`(20), `RedisDataTypeName`(6), `ClientSessionMode`(3), `PubSubKind`(3), `RespVersion`, `RedisFlavor`.

`CommandFlag` and `RespVersion` are the protocol-relevant ones: `CommandFlag` drives MULTI/EXEC eligibility, Lua exclusion, replica safety and MONITOR visibility, so dropping `'noscript'` or `'admin'` ships green with no CHANGELOG entry prompted.

This is the #374 class of break the header comment cites as the reason to build this at all, so I would not defer it. The fix is symmetric with the intersection branch already above, ~4 lines:

```ts
if (ts.isUnionTypeNode(node)) return node.types.flatMap(membersOfTypeNode)
```

plus collecting `literal.text` for `isLiteralTypeNode` + `isStringLiteral` constituents — which closes the string-union gap too.

_Surfaced from both the architecture and wire-compat lenses._

tests-package/export-surface.test.ts
```diff
@@ -0,0 +1,227 @@
...
+
+before(() => {
+  for (const entry of entryNames) {
+    actual.set(entry, readExportSurface(declarationFileFor(entry)))
+  }
+})
+
+type Change = { readonly symbol: string; readonly detail: string }
+
+function removals(
```
fatal10110: **MEDIUM — the guard's own diffing logic has no red-before-green test.**

`removals` (here) and `additions` (line 108) are pure `(ExportSurface, ExportSurface) => ...` functions with no I/O — ideal unit-test targets — but they are unexported locals in a `.test.ts`, so they are structurally untestable. All five suite tests assert against the real, currently-consistent baseline, so they pass by construction today.

Failure mode: a future refactor inverts line 96 to `if ((current.members ?? []).includes(member))`, or drops the member loop while keeping the symbol loop. `gone` is now always `[]`, the suite goes permanently green, and the next `/core` member deletion ships exactly like #374/#376/#377 did. Nothing notices, because every assertion is `deepStrictEqual(x, [])` — which passes when the producer is broken.

That is the classic vacuous-guard failure, and it is the highest-value thing to test in a PR whose entire deliverable *is* the guard. The HIGH union finding above is direct evidence that gaps in this logic do not surface on their own.

I did reproduce the red run manually (removed `SerialTurnQueue`, `CommandPlan`, and `RedisClientSession.pubsubShardChannelCount`, rebuilt → 11 pass / 3 fail, each named correctly), so the logic is right *today*. The point is that nothing keeps it right.

Suggest moving both functions into `export-surface.ts` and adding a unit test with synthetic before/after surfaces covering: symbol gone, member gone, value→type downgrade, pure addition, and a rename producing both failures. ~40 lines.

tests-package/export-surface.ts
```diff
@@ -0,0 +1,222 @@
...
+ * Parse one emitted declaration bundle and return its full export surface.
+ *
+ * `entryDeclarationFile` is an absolute path to a `dist/*.d.ts` (or `.d.mts`).
+ * The bundle's shared chunk is resolved by the compiler, so symbols that tsup
+ * hoisted out of the entry file are still seen.
+ */
+export function readExportSurface(entryDeclarationFile: string): ExportSurface {
+  const program = ts.createProgram([entryDeclarationFile], {
+    noEmit: true,
+    skipLibCheck: true,
```
fatal10110, line 244: **MEDIUM — a degraded extraction is indistinguishable from a successful one, and `skipLibCheck: true` hides it.**

`readExportSurface` never inspects program diagnostics and has no sanity floor. I reproduced the degradation: copy `dist/core.d.ts` to a directory **without** its shared chunk (`redis-error-*.d.ts`) and re-run the extractor.

```
healthy: symbols 267  withMembers 67  kind=type 65
broken : symbols 267  withMembers 18  kind=type 14
```

49 of 67 member-sets silently vanish, and 51 type-only symbols silently reclassify as `value`:

```
healthy  RedisClientSession -> {"kind":"type","members":[...34 members...]}
broken   RedisClientSession -> {"kind":"value"}          // no error raised
```

No exception, no diagnostic — it returns a plausible-looking 267-symbol surface.

Critically, **adding a `getSemanticDiagnostics()` check would not catch this** while `skipLibCheck: true` is set on this line. I verified both directions:

```
skipLibCheck=true  | total diags 0   | TS2307 0
skipLibCheck=false | total diags 65  | TS2307 2
                     "Cannot find module './redis-error-cHS6xBHh.js' or its
                      corresponding type declarations."
```

All inputs here are declaration files, so `skipLibCheck` suppresses exactly the diagnostic that matters.

Today the `kind` reclassification makes this partly self-detecting (a degraded baseline trips the value→type check, a degraded read trips the runtime-reachability test), but that is luck, not design — and it does not hold for the 199 `/core` symbols that record no members.

Two structural invariants cover it without depending on `skipLibCheck`: throw when `(resolved.declarations?.length ?? 0) === 0` (naming the symbol), and throw when `getExportsOfModule` returns zero exports for an entry point. Both hold on the current build (`noDecl: 0`, exports 91/266), so they cost nothing and convert a silent degradation into a named failure.

tests-package/export-surface.test.ts
```diff
@@ -0,0 +1,227 @@
...
+            `Additions are NOT a breaking change; this is a bookkeeping failure. ` +
+            `Run \`${REFRESH}\` and commit tests-package/export-surface.json so the ` +
+            `next PR that deletes one of these is caught.`,
+        )
+      })
+
+      test('ESM and CJS declarations expose the same symbols', () => {
+        const cjs = Object.keys(actual.get(entry) ?? {})
+        const esm = Object.keys(
+          readExportSurface(declarationFileFor(entry, '.d.mts')),
```
fatal10110: **MEDIUM — the ESM/CJS parity test computes both full surfaces, then throws away everything but the names.**

This line already calls `readExportSurface(...)`, which returns the complete `ExportSurface` with kinds and members — and then `Object.keys` discards all of it. Combined with the fact that the baseline only ever reads `.d.ts` (the `before` hook at line 68 uses the default `'.d.ts'`), **the `.d.mts` surface is not pinned at member or kind level anywhere.**

Failure scenario: tsup's dual emit drifts and `dist/core.d.mts` loses `CommandRegistry.registerAll`, or downgrades a class to a type-only re-export, while `dist/core.d.ts` keeps it. Symbol names still match, so this test passes and the baseline never looked at `.d.mts` — the drift ships green. That is precisely what this test's own message ("the dual build has drifted") claims to catch.

The two surfaces are currently deep-equal for both entry points (I checked: 0 differences in kinds or members), so

```ts
assert.deepStrictEqual(esmSurface, cjsSurface, ...)
```

is a free, strictly stronger assertion that passes today.

tests-package/export-surface.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    if (ts.isTypeAliasDeclaration(declaration)) {
+      for (const name of membersOfTypeNode(declaration.type)) {
+        members.add(name)
+      }
+      continue
+    }
+
+    if (ts.isVariableDeclaration(declaration)) {
+      // `const` object namespaces — `t`, `RedisValue`, `RedisResult`. In a
```
fatal10110: **MEDIUM — this branch only works for consts with an *inferred* type literal; the comment below is false for most of them.**

"In a declaration file these carry an explicit type literal" does not hold: 57 of 266 `/core` consts are annotated with a named type, and for every one of them `membersOfTypeNode` hits a type reference and returns `[]`.

The one that matters is `FEATURE_GATES`, declared `Record<FeatureId, VersionGate>` (src/core/compatibility/feature-gates.ts:3). Delete a feature-gate key and **both** the const's key and the corresponding `FeatureId` union member vanish with CI green — this gap compounds with the union gap, since `FeatureId` is a 20-member string union that also records nothing. A consumer doing `FEATURE_GATES['<removed-feature>']` breaks silently.

(Most of the other 56 are `CommandDefinition<...>` values, which legitimately have nothing to pin — the `CommandDefinition` interface is exported and snapshotted separately. So the practical loss is `FEATURE_GATES` plus any future annotated const namespace.)

Asking the checker instead of the syntax resolves `Record`, `Omit`, `Pick` and mapped types uniformly:

```ts
checker.getTypeOfSymbolAtLocation(symbol, decl).getProperties().map(p => p.getName())
```

For a `const` this does **not** drag in the `lib.es5.d.ts` inheritance the header comment (lines 24-29) rightly wants to avoid. At minimum, the comment should stop claiming the annotated case works.

CHANGELOG.md
```diff
@@ -0,0 +1,126 @@
...
+  | `CommandRegistry.override(d)`    | `register(d, { override: true })`                  |
+  | `CommandRegistry.has(n)`         | `get(n) !== undefined`                             |
+  | `CommandRegistry.getNames()`     | `getAll().map(d => d.name)`                        |
+  | `CommandPlan.flags`              | `plan.definition.flags`                            |
+
+  Nothing changed on the root barrel. One residue worth knowing about for
+  `/core` consumers who pass their own `signal` to `ClientSession`: aborts now
+  surface the caller's `signal.reason` (a `DOMException`, or whatever was passed
+  to `abort()`) instead of a fixed `Error`. No wire-visible change — the adapter
+  maps any non-`RedisCommandError` to `-ERR internal server error` either way.
```
fatal10110: **MEDIUM — "no wire-visible change" is not unconditional, and the exception is reachable from `/core`.**

`Resp2SessionAdapter.writeError` does not map *any* non-`RedisCommandError` to `-ERR internal server error` — it has an earlier pass-through branch (src/core/transports/resp2/session-adapter.ts:153-172):

```ts
if (err instanceof RedisCommandError) {
  await this.writeRedisResult(RedisResult.error(err.message, err.code))
  return
}
if (err instanceof Resp2ParseError) {
  await this.writeRedisResult(RedisResult.error(err.message, 'ERR'))
  return
}
this.logger?.error(err)
await this.writeRedisResult(RedisResult.error('internal server error', 'ERR'))
```

Combined with the #377 change this entry describes — `ClientSession` now uses `signal.throwIfAborted()`, which rethrows `signal.reason` **verbatim** (src/core/redis-context.ts:168-171) — a `/core` consumer who passes their own `signal` and calls `controller.abort(new WrongTypeRedisError(...))` now gets `-WRONGTYPE ...` on the wire, where pre-#377 they got `-ERR internal server error`. That is a wire-visible change, and it is exactly the consumer group this entry is addressed to.

The guarantee holds only for reasons that are neither `RedisCommandError` nor `Resp2ParseError`. Worth narrowing the sentence to say so, since the whole point of this CHANGELOG is that `/core` consumers can trust it.

_Wire-compat lens._

tests-package/export-surface.ts
```diff
@@ -0,0 +1,222 @@
...
+//
+// Two levels are captured per entry point:
+//
+//   1. the exported symbol names, each tagged `value` (has a runtime binding:
+//      class, function, const) or `type` (interface / type alias only), and
+//   2. the *own* members declared on each exported interface, class, object
+//      type alias, or `const` object namespace.
+//
+// Level 2 is what makes this guard worth having. Three of the four `/core`
+// breaks this was written in response to were member-level, not symbol-level:
```
fatal10110: **LOW — the stated rationale is factually wrong: it is two of four, not three.**

This comment (and the same claim in the test header and the PR body) says a symbol-list-only snapshot would have stayed green for #374, #376 **and** #377. I checked #377's actual merge commit (`822c833`) rather than its description, and it removed two **top-level exported symbols** from `/core`:

```
pre-#377  src/internal.ts:59  export type { RedisTurnHandle, RedisTurnQueue } from './core/turn-queue'
pre-#377  src/internal.ts:98    createNoopParkHandler,

post-#377                       export type { RedisTurnHandle } from './core/turn-queue'
post-#377                       (createNoopParkHandler gone)
```

So a plain symbol list would have caught #377. The accurate scoreboard is **two of four invisible** (#374 `encoder` on four option types; #376 the six `RedisClientSession` methods) and **two of four caught** (#375 `RedisKeyspace`/`WrongRedisTypeError`; #377 `RedisTurnQueue`/`createNoopParkHandler`).

This does not weaken the case for member-level recording — #374 and #376 alone justify it, and member-level also catches *more* of #377 than a symbol list would (`CommandRegistry.override/has/getNames`, `CommandPlan.flags`, `ClientSessionOptions.turnQueue`). But this comment is the documented design rationale, so it should be right. Worth fixing in all three places.

For what it is worth, I verified the guard would in fact catch all four: every break site has member coverage in the committed baseline — `Resp2ServerOptions`(4), `AttachSessionOptions`(5), `Resp2SessionAdapterOptions`(3), `CreateVirtualConnectionOptions`(6), `RedisClientSession`(34), `CommandRegistry`(4), `ClientSessionOptions`(9), `CommandPlan`(5).

tests-package/export-surface.ts
```diff
@@ -0,0 +1,222 @@
...
+  return []
+}
+
+function ownMembers(symbol: ts.Symbol): string[] {
+  const members = new Set<string>()
+
+  for (const declaration of symbol.declarations ?? []) {
+    if (
+      ts.isInterfaceDeclaration(declaration) ||
+      ts.isClassDeclaration(declaration)
```
fatal10110, line 190: **LOW — the "no class statics" gap is inverted: statics *are* captured, but flattened.**

`ts.isClassDeclaration(declaration)` walks `declaration.members`, which includes static members, so they land in the baseline. From the committed file:

```
RedisResult {"kind":"value","members":["create","encoded","error","nil",
                                       "ok","options","preEncoded","value"]}
```

`create` / `nil` / `ok` / `error` / `preEncoded` are the statics (src/core/redis-result.ts:17-38), merged into the same list as the instance fields `value` / `options` / `encoded`. I also confirmed removal is detected: dropping a static from a probe class removed it from the extracted surface.

Two consequences:
1. The gap list in the PR body and CONTRIBUTING will send a future maintainer off to implement coverage that already exists.
2. Because both namespaces share one `Set`, a class with **both** `static create(...)` and an instance `create(...)` records `'create'` once — deleting either leaves the baseline unchanged and the removal goes undetected.

Either correct the claim, or prefix statics (`static:create`) to keep the namespaces distinct.

CHANGELOG.md
```diff
@@ -0,0 +1,126 @@
...
+
+  Also removed from `ClientSession` (not on the interface, but public via
+  `/core`):
+
+  ```
+  pubsubRegularSubscriptionCount  -> no replacement; it was channels + patterns.
+                                     Use pubsubChannelCount + pubsubPatternCount.
+  ```
+
+  Wire output is unchanged — `pmessage` still carries three items, channel and
```
fatal10110: **LOW — "`pmessage` still carries three items" miscounts the wire frame.**

The behavior is correct; the prose is not. On the wire a `pmessage` push is a **four**-element array. Ground-truthed against real Redis 7.2.1, raw socket capture after `PSUBSCRIBE pat*` + `PUBLISH pat123 hello`:

```
*4\r\n$8\r\npmessage\r\n$4\r\npat*\r\n$6\r\npat123\r\n$5\r\nhello\r\n
```

The mock, driven through the identical script, emits the identical byte string.

The "three" comes from the internal `items` array (src/core/client-session.ts:124 emits `[bulk(pattern), bulk(channel), bulk(message)]`); `pubsubFrame` then prepends the frame name via `RedisValue.push(name, items)`. But the sentence is framed as a statement about *wire output*, so a contributor writing a conformance test from this CHANGELOG would assert a 3-element array and encode a real regression.

Suggested: "`pmessage` still carries the pattern ahead of the channel (a 4-element push frame)".

The rest of this entry checks out byte-for-byte: channel and pattern confirmations do report the combined count (`ch1`→1, `pat*`→2, `ch2`→3, `other*`→4), shard confirmations report a separate count (`sh1`→1, `sh2`→2), and the counters are not unified.

_Wire-compat lens._

tests-package/export-surface.test.ts
```diff
@@ -0,0 +1,227 @@
...
+  }
+})
+
+describe('published export surface reaches the runtime', () => {
+  // The declaration bundle is a claim about the JS. These two tests check it is
+  // true for the half of the surface that has a runtime binding: a symbol the
+  // types promise but the bundler dropped would otherwise ship broken.
+  for (const entry of entryNames) {
+    const expectedValues = () =>
+      Object.entries(baseline.entries[entry] ?? {})
```
fatal10110: **LOW — the runtime reachability tests check last-refresh's symbol list, not the current one.**

`expectedValues()` derives from the committed `baseline`, but `actual` (line 64) already holds the current `.d.ts` surface and is what every other test compares against.

Consequence: a PR adds a new value export that the bundler drops from the emitted JS. The "baseline records every current export" test fails (stale baseline), but these two runtime tests pass — they are validating a symbol list that predates the new export. The genuinely broken export only surfaces on the *second* run, after someone refreshes the baseline.

Switching `expectedValues()` to read `actual.get(entry)` makes the check track reality and costs nothing.

tests-package/export-surface.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  for (const name of Object.keys(surface).sort()) {
+    sorted[name] = surface[name]
+  }
+
+  return sorted
+}
+
+/** The two published entry points, keyed by the name used in the baseline. */
+export const PUBLISHED_ENTRIES = {
```
fatal10110, line 406: **LOW — `PUBLISHED_ENTRIES` is an identity map whose values are never read.**

`declarationFileFor` (line 209) interpolates `entry` — the key — not `PUBLISHED_ENTRIES[entry]`. `readAllEntrySurfaces` (line 217) and the test (line 50) both only call `Object.keys(...)`. So the values are dead, and the root-vs-`/core` mapping this constant looks like it owns is actually re-encoded by hand in the test:

```ts
return entry === 'index' ? packageName : `${packageName}/core`   // test line 53
```

Adding a third entry point — or the planned phase-4 `/core` removal — means editing two places that must agree, with nothing linking them.

Either narrow it to `readonly ['index', 'core']`, or give the values real content (the import-path suffix: `{ index: '', core: '/core' }`) so `importPath()` derives from it instead of duplicating it.

package.json
```diff
@@ -75,6 +76,7 @@
     "test:integration:raw:mock": "tsc --noEmit && TEST_BACKEND=mock node --enable-source-maps --import tsx --no-warnings --test \"./tests-integration/raw-tcp/**/*.test.ts\"",
     "test:integration:raw:real": "tsc --noEmit && TEST_BACKEND=real node --enable-source-maps --import tsx --no-warnings --test-concurrency 1 --test-timeout 30000 --test \"./tests-integration/raw-tcp/**/*.test.ts\"",
     "test:package": "npm run build && node --enable-source-maps --import tsx --no-warnings --test \"./tests-package/**/*.test.ts\"",
+    "export-baseline": "npm run build && node --enable-source-maps --import tsx --no-warnings ./tests-package/update-export-baseline.ts",
```
fatal10110: **LOW — this adds ~490 lines of compiler-API TypeScript to a directory `tsc --noEmit` never sees.**

`tsconfig.json` has `"include": ["src/**/*.ts"]`, so `tests-package/` is outside the program entirely. Every other test script (`test`, `test:integration:*`) is prefixed with `tsc --noEmit`; `test:package` and this new `export-baseline` script are not.

So a type error in `export-surface.ts` is stripped by tsx at runtime and never reported — for example the `{} as Record<PublishedEntry, ExportSurface>` cast at line 215 going stale after a third entry point is added, or the unvalidated `as ExportBaseline` at test line 46. The extractor can rot with no signal.

This is pre-existing for `dual-format.test.ts`, but this PR multiplies the unchecked surface by ~6x and adds the first non-test **library** module (`export-surface.ts`) there. Worth adding `tests-package` to a type-check pass now, while it is one line.

tests-package/export-surface.ts
```diff
@@ -0,0 +1,452 @@
...
+export function findAdditions(
+  baseline: ExportSurface,
+  current: ExportSurface,
+): readonly string[] {
+  const added: string[] = []
+
+  for (const [name, entry] of Object.entries(current)) {
+    const before = baseline[name]
+
+    if (!before) {
```
fatal10110, line 381: **MEDIUM (new) — `findAdditions` is blind to a `type` → `value` kind upgrade, which permanently disarms the `value` → `type` downgrade check.**

`findRemovals` handles one direction (line 350: `entry.kind === 'value' && now.kind === 'type'`). Nothing handles the other. For a symbol present in both surfaces, `findAdditions` never inspects `kind` at all — this `if (!before)` is the only kind-sensitive branch, and it only fires for a brand-new name.

Run against the PR's own exported functions:

```
findRemovals ({Foo:{kind:'type'}}, {Foo:{kind:'value'}})  -> []
findAdditions({Foo:{kind:'type'}}, {Foo:{kind:'value'}})  -> []
```

So a type-only export that gains a runtime binding (an interface replaced by a class, or a type alias joined by a same-named const) produces **no** failure — the "baseline records every current export" test stays green, nobody runs `npm run export-baseline`, and the baseline keeps `kind: "type"` indefinitely.

Now the real break lands. The binding is removed again, and:

```
findRemovals ({Foo:{kind:'type'}}, {Foo:{kind:'type'}})   -> []
```

baseline `type` vs current `type` — nothing to report. A consumer's `import { Foo } from 'js-redis-server/core'` breaks at runtime with CI green. That is the #375 shape, and it is the one failure this `kind` field was added to catch.

It is also the one mutation the new 13-test suite doesn't cover: no case in `export-surface-diff.test.ts` passes a `type` → `value` pair. Two lines here (push `${name} — is now a value export (was type-only)`) plus a test case closes it.

_Architecture lens; reproduced on the PR head._

tests-package/export-surface.ts
```diff
@@ -0,0 +1,452 @@
...
+      // typically a missing shared chunk. Everything derived from it (kind,
+      // members) would be silently wrong.
+      unresolved.push(symbol.getName())
+      continue
+    }
+
+    surface[symbol.getName()] = describeSymbol(checker, resolved)
+  }
+
+  if (unresolved.length > 0) {
```
fatal10110, line 294: **MEDIUM (new) — the two new invariants are themselves untested, and the PR's own vacuity argument applies to them verbatim.**

This throw and the zero-exports throw at line 268 are the fix for the silent-degradation finding, and they work — I reproduced the orphaned-chunk case and got the message you quote in the PR body (`180 exported symbol(s) have no resolvable declaration`).

But nothing keeps them working. I deleted **both** `throw` blocks on the current tree and re-ran the suite:

```
# tests 27
# pass 27
# fail 0
```

Every test stays green, because a healthy build never reaches either branch. `export-surface-diff.test.ts` never calls `readExportSurface` (`grep -c` → 0), and `export-surface.test.ts` only ever calls it on the real `dist/`.

This is precisely the argument the PR makes about `findRemovals` — "a guard whose diff quietly stops diffing passes every assertion it has" — applied to the guard-against-a-degraded-guard. Compare the contrast you already drew: deleting the member loop fails 3 tests *because* the diff functions are unit-tested against synthetic input. These throws have no such backstop, so the next refactor that drops them takes the suite back to reporting a plausible 267-symbol surface on a missing chunk.

Both fixtures are trivial and need no build — an orphaned copy of `dist/core.d.ts` for one, a file containing only `export {}` for the other. They belong next to the `findRemovals`/`findAdditions` cases in `export-surface-diff.test.ts`, in the same synthetic-input style.

_Architecture lens; reproduced on the PR head._

tests-package/export-surface.ts
```diff
@@ -0,0 +1,452 @@
...
+    for (const constituent of node.types) {
+      collectFromTypeNode(constituent, into)
+    }
+    return
+  }
+
+  if (ts.isLiteralTypeNode(node)) {
+    const literal = node.literal
+
+    if (ts.isStringLiteral(literal) || ts.isNumericLiteral(literal)) {
```
fatal10110, line 128: **MEDIUM (new) — `variants` records a numeric and a string literal identically, and silently drops four other constituent kinds — so some unions are pinned completely and others only partly, with nothing in the baseline to tell them apart.**

**Collision.** `into.variants.add(literal.text)` uses `.text` for both branches, and `.text` strips the quotes. So:

```
type RespVersion = 2 | 3      -> {"kind":"type","variants":["2","3"]}
type X           = '2' | '3'  -> {"kind":"type","variants":["2","3"]}   // byte-identical
```

`export-surface.json:902` is the first of those. Normalising RESP version handling to strings — `2 | 3` → `'2' | '3'` — breaks every consumer passing `{ version: 2 }`, and both `findRemovals` and `findAdditions` return `[]`. The guard is silent on a real break in the one union whose variants it does record.

**Silent partial coverage.** Only `isStringLiteral` / `isNumericLiteral` are collected; four other constituent shapes fall through the comment at line 144 and contribute nothing *while their siblings are recorded*:

```
type Neg   = -1 | 0 | 1          -> {"variants":["0","1"]}   // PrefixUnaryExpression
type Big   = 1n | 2n             -> {}                       // BigIntLiteral
type Undef = 'a' | undefined     -> {"variants":["a"]}       // KeywordTypeNode
type Tmpl  = `a${string}` | 'lit'-> {"variants":["lit"]}     // TemplateLiteralType
```

`-1` is the canonical Redis sentinel (TTL, `LPOS` rank, `XAUTOCLAIM`), so `-1 | 0 | 1` is a plausible future union here. Deleting `-1` from it would ship green — and the entry would still *look* pinned, which is worse than the pre-fix behaviour where a union recorded nothing at all and was visibly untrusted. Partial-but-plausible is the same failure shape as the half-resolved program you just added invariants for.

Neither shape exists on either entry point today, so the committed baseline is correct — this is about the next union someone writes. `literal.getText()` (which yields `2` vs `'2'`) or a `num:`/`str:` tag, symmetric with `static:`, fixes the collision in one line; `PrefixUnaryExpression` + `BigIntLiteral` is ~4 more. At minimum the comment at 144-145 should name the kinds that are *not* captured, so the next reader doesn't take "literal constituents of a union" at face value.

_Both lenses._

tests-package/export-surface.ts
```diff
@@ -0,0 +1,452 @@
...
+): string[] {
+  const type = checker.getTypeOfSymbolAtLocation(symbol, declaration)
+
+  if (!(type.flags & ts.TypeFlags.Object)) {
+    return []
+  }
+
+  const objectFlags = (type as ts.ObjectType).objectFlags
+
+  if (!(objectFlags & (ts.ObjectFlags.Anonymous | ts.ObjectFlags.Mapped))) {
```
fatal10110, line 171: **LOW (new) — the gate does its job, but its discriminator is `type` vs `interface`, which is not a distinction consumers can see — and a `type` → `interface` refactor therefore reads as BREAKING.**

First, the gate earns its keep, and I measured it rather than taking the rationale on trust. On the real `/core` build, 55 of 59 exported consts are blocked; ungated they would add **742** member entries, including exactly the `Array.prototype` noise you predicted:

```
BLOCKED connectionCommands [Reference] wouldHavePulled=41  e.g. length,toString,toLocaleString,pop,push,concat
BLOCKED aclCommand         [Reference] wouldHavePulled=9   e.g. name,since,schema,flags,capabilities,monitor
PASS    FEATURE_GATES      [Mapped|Instantiated] members=20
PASS    RedisValue         [Anonymous] members=16
```

That is why the baseline only grew 929 → 1076 lines. No argument with the decision.

The issue is the predicate. `ObjectFlags.Anonymous` does not mean "written inline" — a type alias creates no distinct type, so a const annotated with a **named object type alias** is anonymous too, while one annotated with an **interface** is not. Same authorial intent, opposite outcome:

```ts
export type Opts = { alpha: string; beta: number }
export declare const DEFAULTS: Opts
// -> DEFAULTS {"kind":"value","members":["alpha","beta"]}   // duplicated; Opts pins these already

export interface Iface { delta: string }
export declare const ANNOTATED: Iface
// -> ANNOTATED {"kind":"value"}                             // blocked, as the docstring intends
```

So converting `type Opts = {…}` to `interface Opts {…}` — invisible to every consumer — is reported as a breaking change:

```
before: {"DEFAULTS":{"kind":"value","members":["alpha","beta"]}, "Opts":{...}}
after : {"DEFAULTS":{"kind":"value"},                            "Opts":{...}}
REMOVALS (reported as BREAKING): [ 'DEFAULTS.alpha — member gone', 'DEFAULTS.beta — member gone' ]
```

and the contributor is told to write a CHANGELOG breaking-change entry for a no-op — or, more likely, to regenerate the baseline and hope nothing real was hidden in the same run.

Second, narrower: the docstring says the checker pass makes `Record<…>`, `Omit<…>` "and other utility types answer with their real keys". Intersections don't — `type.flags & ts.TypeFlags.Object` is false for an intersection, so the early return at line 165 fires first:

```ts
export declare const wired: Opts & Named   // -> {"kind":"value"}, no members
export type          Wired = Opts & Named  // -> members ["alpha","beta","gamma"]
```

The const form records nothing while the structurally identical alias form records everything — and the intersection is the exact case the comment at 109-111 calls out as handled.

Neither shape exists on either entry point today (I checked all 59 consts), so nothing in the committed baseline is wrong. Gating on the *annotation node* — absent, or not a `TypeReferenceNode` — would express the stated intent ("pin what is only written here") without depending on how the annotation happened to be spelled.

_Both lenses._

tests-package/export-surface.ts
```diff
@@ -0,0 +1,452 @@
...
+    .map(property => property.getName())
+}
+
+function describeSymbol(
+  checker: ts.TypeChecker,
+  symbol: ts.Symbol,
+): ExportEntry {
+  const collected: Collected = { members: new Set(), variants: new Set() }
+
+  for (const declaration of symbol.declarations ?? []) {
```
fatal10110, line 187: **LOW (new) — this loop flattens a declaration-merged symbol's *type* half into its *value* half, which re-introduces one level up the masking that `static:` was just added to fix.**

`RedisValue` is the only export on either entry point declared twice — `export type RedisValue = …` (a 16-arm discriminated union, `src/core/redis-value.ts:1`) and `export const RedisValue = { … }` (the factory namespace, `:19`). I enumerated both entry points for symbols carrying a type *and* a value declaration; it is the sole hit.

This loop walks every declaration into one `collected.members` Set. Now that `collectFromTypeNode` descends unions, the type half contributes its arms' field names for the first time:

```
before this commit  16: array, bigNumber, boolean, bulkString, double, error, flatPairs,
                        integer, map, mapPairs, null, nullArray, push, set, simpleString, verbatim
after               24: + code, entries, format, items, kind, message, name, value
```

Those eight are fields of the *type*, not properties of the *value* — `RedisValue.kind` is not something a consumer can call. Two consequences:

1. **The masking you just fixed, one level up.** `static:` stops `static create()` hiding an instance `create()`; nothing stops the union arms' `value` / `name` / `items` / `message` hiding a future factory of the same name. The two halves happen not to overlap today (16 + 8 = 24, no collisions), so this is latent rather than live — but deleting a factory named `name` would go unreported, and `RedisValue.push()` already collides in spirit with the `push` arm.
2. **The entry stops being reviewable.** Someone reading those 24 names in the baseline cannot tell which are callable.

A `type:` prefix, symmetric with `static:`, settles it.

Separately and in the same loop: there is no branch for `EnumDeclaration` or `ModuleDeclaration`, so `export enum E { A, B }` baselines as `{"kind":"value"}` and removing `E.A` ships green. No enums are exported today — raising it only because completeness is this guard's whole premise, and the "what it does not cover" list in the PR body is otherwise scrupulous.

_Both lenses._

tests-package/export-surface.json
```diff
@@ -0,0 +1,1076 @@
...
+      "DiscardWithoutMultiError": { "kind": "value" },
+      "ExecCommandAbortError": { "kind": "value" },
+      "ExecWithoutMultiError": { "kind": "value" },
+      "ExecutorResult": { "kind": "type" },
+      "ExpectedFloatError": { "kind": "value" },
+      "ExpectedIntegerError": { "kind": "value" },
+      "ExpirationState": { "kind": "type", "members": ["expiresAt", "kind"] },
+      "ExpireGtLtConflictError": { "kind": "value" },
+      "ExpireNxXxGtLtConflictError": { "kind": "value" },
+      "FEATURE_GATES": {
```
fatal10110, line 455: **LOW (new) — baseline redundancy: these 20 members are byte-for-byte `FeatureId.variants`, which is itself recorded under both entry points. Adding one feature gate now costs four baseline edits.**

`FEATURE_GATES` is `Record<FeatureId, VersionGate>`, so its key list is fully derivable from `FeatureId`. The baseline records it three more times anyway:

```
line 455  core.FEATURE_GATES.members   (20 strings)
line 480  core.FeatureId.variants      (the same 20)
line  69  index.FeatureId.variants     (the same 20)
```

So a routine compatibility PR adding one feature id produces a ~60-line baseline diff carrying one bit of information. That is the churn that turns a reviewable baseline into a rubber-stamped one — and a rubber-stamped baseline is where a real removal hides, which is the failure this PR exists to prevent.

To be clear about the overall picture, because I went looking for churn and mostly did not find it: the baseline grew 929 → 1076 lines, regeneration is byte-identical across three consecutive runs, and the delta touches only the 21 symbols the fix intended (I diffed old vs new entry by entry). No `lib.es5` inheritance leaked in, no tsup chunk hash is recorded, and nothing in it is sensitive to a TypeScript bump. The "own members only" rule is doing its job. This one duplication is the exception.

Two options: skip `members` for a mapped-type const whose key type is itself an exported union, or state in the file header that the duplication is deliberate belt-and-braces. Either is fine; silently paying 60 lines per feature id is the thing to avoid.

**One genuine dependency-churn vector worth a line in the header while you are here.** The header says inherited members are not walked so the baseline won't "churn on every TypeScript bump" — true for the error subclasses. But a re-exported builtin *class* would snapshot its own declared members straight out of `@types/node`. I confirmed the extractor resolves them when the entry file sits in `dist/`:

```
export type { Socket } from 'net'
  -> Socket {"kind":"value","members":[ ...35 members: addListener, address, bufferSize,
                                        bytesRead, connect, ..., static:from, static:toWeb ]}
```

Nothing re-exports a builtin today (`dist/core.d.ts` only *imports* `Socket` and `Duplex` for use in signatures, so they never enter `getExportsOfModule`) — but the day one is re-exported, every `@types/node` bump churns 35 lines and the removal guard starts reporting Node's API changes as breaks in this package.

_Architecture lens, with the churn measurement from the main review._

fatal10110:
## Re-review of `0b5f96f` — verdict: **ship**

All twelve prior findings are fixed. I re-ran each one as a red run against a pristine build of the head rather than reading the diff for intent, because "fixed" and "reworded" look the same in a commit message. Both lenses completed — architecture and wire-compat.

### Prior findings

| # | Finding | Status | Red-run evidence on `0b5f96f` |
|---|---|---|---|
| 1 | **HIGH** union nodes returned `[]`, voiding member coverage for every union-typed export | **Fixed** | Removed `'blocking'` from `CommandFlag`, rebuilt → 1 failure: `CommandFlag — union variant 'blocking' gone`. `CreateIoredisMockOptions` now pins `["cluster","databaseCount","seed"]` |
| 2 | **MED** diff logic structurally untestable / vacuous guard | **Fixed** | Deleted the member loop → **exactly 3** failures, all in `findRemovals`, both surface tests green. Inverted `missing()` → 16/27. Both figures match the PR body exactly |
| 3 | **MED** silent degradation, `skipLibCheck` hides TS2307 | **Fixed** | Orphaned `core.d.ts` → `THREW: … 180 exported symbol(s) have no resolvable declaration`. (The throws themselves are now untested — see the inline note) |
| 4 | **MED** ESM/CJS parity compared only `Object.keys` | **Fixed** | Dropped `registerAll` from `redis-error-*.d.mts` only → `"js-redis-server/core" describes a different surface … the dual build has drifted` |
| 5 | **MED** annotated consts recorded nothing; `FEATURE_GATES` empty | **Fixed** | Dropped the `cluster.multi-db` gate (union arm + gate entry + call site) → 2 failures: `FEATURE_GATES.cluster.multi-db — member gone` **and** `FeatureId — union variant 'cluster.multi-db' gone` |
| 6 | **MED** #377 CHANGELOG entry claimed an unconditional guarantee | **Fixed** | End-to-end repro on the built `dist/core.js`: `abort(WrongTypeRedisError)` → `-WRONGTYPE Operation against a key holding the wrong kind of value\r\n`; plain `Error` / default / string reason → `-ERR internal server error\r\n`. Real Redis 7.2.1 wording matches byte-for-byte. There is no RESP3 adapter, so "only reasons that are neither of those two classes are masked" is exhaustive |
| 7 | **LOW** "three of four" scoreboard wrong | **Fixed** | "Two" in the module header (`:9`, `:21-27`) and PR body; no "three" survives. The test header dropped the claim rather than restating it, which is fine |
| 8 | **LOW** statics captured but flattened | **Fixed** | `RedisResult: ["encoded","options","static:create","static:error","static:nil","static:ok","static:preEncoded","value"]` |
| 9 | **LOW** `pmessage` miscounted as 3 items | **Fixed** | Raw socket, real Redis 7.2.1: `*4\r\n$8\r\npmessage\r\n$4\r\npat*\r\n$4\r\npatX\r\n$10\r\nviapattern\r\n` vs `*3` for `message`/`smessage`. Pattern does precede channel. Mock byte-identical |
| 10 | **LOW** runtime reachability read the baseline | **Fixed** | `expectedValues()` now reads `surfaceOf(entry)` |
| 11 | **LOW** `PUBLISHED_ENTRIES` identity map | **Fixed** | Carries `{dist, subpath}`; both consumed |
| 12 | **LOW** `tests-package/` outside every `tsc` pass | **Fixed** | Injected a type error into `export-surface.ts` → `error TS2322`. Clean run exits 0 in 0.76s; no chicken-and-egg, the extractor imports only `typescript` and `node:url` |

### Baseline churn — the risk I was most worried about, and it came back clean

This was the thing that could have sunk the round: union constituents, `variants` and checker-resolved const members all landed at once, and a baseline nobody can review is a baseline people rubber-stamp.

- **929 → 1076 lines (+16%).** Regeneration is **byte-identical across three consecutive runs**.
- The delta touches **only the 21 symbols the fix intended**. I diffed old against new entry by entry: 13 unions gained `variants`, 7 types gained `members`, `FEATURE_GATES` gained its 20 keys, `RedisResult`/`NodeRedisMockCluster` re-spelled statics. Nothing else moved.
- **The anonymous/mapped gate holds.** 55 of 59 consts blocked; ungated they would have added **742** member entries, including `length,toString,pop,push,concat` off `Array.prototype` for the command arrays. The stated rationale is accurate.
- No `lib.es5` inheritance leaked in, no tsup chunk hash is recorded, nothing is sensitive to a TypeScript bump. The `on`/`emit`/`push`/`name` entries I chased down are all genuine own members of this repo's interfaces.

The one real redundancy (`FEATURE_GATES` ≡ `FeatureId.variants`, ×2 entry points → ~60 lines per feature id) and one latent dependency-churn vector are in the inline note on `export-surface.json`.

### `variants` direction is right

Confirmed both ways end-to-end, not just in the unit test: adding `'newflag'` to `CommandFlag` fails **"baseline records every current export"** as a bookkeeping addition while the removals test stays green; removing `'blocking'` fails **"no exported symbol, member or variant was removed"** as BREAKING. Removal is the breaking direction and that is how it reads.

### New findings — six, none blocking

| Sev | Location | Issue |
|---|---|---|
| MED | `export-surface.ts:381` | `findAdditions` blind to a `type`→`value` upgrade, which permanently disarms the `value`→`type` downgrade check |
| MED | `export-surface.ts:294` | Both new structural invariants are untested — deleting them leaves 27/27 green |
| MED | `export-surface.ts:128` | `variants` can't distinguish `2` from `'2'`; negative/bigint/`undefined`/template constituents silently dropped |
| LOW | `export-surface.ts:171` | The const gate discriminates on `type` vs `interface`; a `type`→`interface` refactor reads as BREAKING |
| LOW | `export-surface.ts:187` | Declaration-merge flattening on `RedisValue` re-introduces the masking `static:` just fixed |
| LOW | `export-surface.json:455` | `FEATURE_GATES` members duplicate `FeatureId.variants` across three places |

Every one is latent — none is wrong in the committed baseline today, and all six are about the next symbol someone adds. The first two are the ones I'd actually spend time on; neither needs to block this merge.

### Smaller notes, no inline comment

- **The `prettier --write CONTRIBUTING.md` pushback is correct — drop the request.** Reproduced: `origin/main:CONTRIBUTING.md` and the PR head are prettier-dirty by the *identical* deviations, none on lines this PR touches. Markdown is unreachable from `.lintstagedrc` (`*.ts` only) and from `npm run format` (`**/*.+(js|ts|json)`). It would be pure unrelated churn.
- `export-surface-diff.test.ts:70` — "`CommandRegistry` stays, three of its methods do not" describes #377's real shape, but the synthetic case below it removes **two** (`getAll`, `has`). Worth a word change, given the round we just had about documented rationale being right.
- CHANGELOG cosmetics, all optional: `### Removed` precedes `### Added`/`### Fixed`, against Keep a Changelog's prescribed order (arguably deliberate, since the breaking removals are the point); `## [0.3.0] and earlier` isn't a parseable KAC release heading; "4-element push frame" uses RESP3 vocabulary for what is a plain `*` array in RESP2 (the arity and field order — the load-bearing part — are right in both). All 7 link references resolve, none unused.
- Everything else in the CHANGELOG ground-truths correct: the subscribe/psubscribe/ssubscribe counter claims (`ch1`→1, `pat*`→2, `ch2`→3, `pat2*`→4, with `sh1`→1 restarting separately), the #375 empty-collection and lazy-expiry carry-over, and the #378 routing entry including the `SORT … BY` cluster carve-out — all byte-identical between real Redis 7.2.1 and the mock.

### Merge

CI is **14/14 green** on `0b5f96f`, including both `package` matrix legs (`js-redis-server` and `js-valkey-server`) where this guard, the new diff unit test and the new `tests-package` type-check actually run. `mergeStateStatus: CLEAN`.

One caveat for whoever merges: the branch carries a `main` merge commit (`f305f61`) with the fix commit rebased on top. **Squash** folds it harmlessly; a **rebase** merge would want a look first.

The union gap was the one thing holding this back last round, and it is closed — verified by breaking a real union and watching the build go red for the right reason. Ship it.

<sub>Reviewed read-only against an exported copy of the head; the shared working tree was never checked out or modified.</sub>

PR #394, opened 2026-09-21
chore(ci): build demo on PRs and add Dependabot config

fatal10110:
Follow-up to #391. That PR fixed the `vite` / `vite-plugin-node-polyfills` peer conflict correctly; this one closes the gaps that let it sit on `main` for two months and seven red deploys.

Each gap below was re-verified against `main` before changing anything. Rebased onto `main` after #392.

---

## Gap 1 — no PR-time demo build

**Verified:** `grep -rn browser-demo .github/` matches only `deploy-demo.yml`, and that workflow was `on: push: branches: [main]`. `ci.yml` never installs or builds the demo. So the only thing that built the demo ran *after* merge, on the protected branch.

**Change:** `deploy-demo.yml` now also triggers on `pull_request`, and the existing `build` job runs unchanged on both events. The two Pages steps (`configure-pages`, `upload-pages-artifact`) and the whole `deploy` job are gated on `github.event_name != 'pull_request'`.

Deliberately **one job in one workflow** rather than a copy of the build steps in `ci.yml`: a duplicated build can drift from the deploy build and stop catching the thing it exists to catch, which is the same class of bug as this incident. The PR build and the deploy build are now literally the same steps.

`concurrency` was `group: pages` for the whole workflow. PR builds would have queued behind Pages deploys, so it is now `pages-${{ github.event_name == 'pull_request' && github.ref || 'deploy' }}` with `cancel-in-progress` on for PRs only — deploys stay serialized exactly as before, superseded PR builds get cancelled.

### `permissions` are now scoped per job

Bringing `build` into PR context while leaving `permissions` workflow-scoped was a privilege widening introduced by this diff: the job runs `npm ci` twice over an author-controlled lockfile, so install lifecycle scripts execute while holding `pages: write` + `id-token: write`, and `checkout@v4` leaves the token in `.git/config`. Fixed:

```yaml
permissions:
  contents: read          # workflow-level

jobs:
  build:
    permissions: { contents: read, pages: write }
  deploy:
    permissions: { contents: read, pages: write, id-token: write }
```

[...]

PR #397, opened 2026-09-21
chore: pin the export guard's own invariants and close the kind-upgrade hole

fatal10110:
Follow-up to #392, addressing the six latent findings from its re-review. None of them affected the committed baseline — every fix here is about the next thing someone writes. The guard's behaviour on the current surface is unchanged except where it gained coverage.

`npm run test:package`: **27 → 42 tests.**

## 1. The invariants had no test — the vacuity argument, one layer up

#392 added two `throw`s to `readExportSurface` so a half-resolved program fails loudly instead of returning a plausible 267-symbol surface with members and type-only kinds quietly missing. They work. Nothing kept them working: **deleting both left the suite at 27/27**, because a healthy `dist/` never reaches either branch.

That is exactly the argument #392 makes about `findRemovals` — "a guard whose diff quietly stops diffing passes every assertion it has" — applied to the guard-against-a-degraded-guard. Fixed the same way: `tests-package/export-surface-extractor.test.ts` drives `readExportSurface` over synthetic `.d.ts` fixtures written at run time.

**Red-run, both throws deleted on this branch:**

```
    not ok 1 - throws when the entry resolves to zero exports
    not ok 2 - throws, naming symbols, when a re-export target does not resolve
# tests 42
# pass 40
# fail 2
```

Exactly the two new invariant tests, and nothing else — the surface-level tests stayed green, as they did before, because the real build is healthy.

The same file pins the extraction rules that no current export exercises, which are precisely the ones that would degrade silently when first written: a `-1 | 0 | 1` union, a bigint union, a class with colliding static and instance names, a declaration-merged symbol, an enum, a mapped-type const.

> Fixtures live under `dist/` (gitignored) rather than the OS temp dir: a `.d.ts` in `/tmp` cannot resolve anything from `node_modules`, which would make the dependency case below fail for the wrong reason. They are removed in `after`.

## 2. `findAdditions` was blind to a `type` → `value` upgrade

`findRemovals` reports `value` → `type`. Nothing reported the other direction, so a type-only export that gained a runtime binding produced no failure at all — no refresh, baseline keeps `kind: "type"` indefinitely. When the binding is later removed, the downgrade check compares baseline `type` against current `type` and says nothing, while the consumer's `import { Foo }` breaks at runtime.

That is the #375 shape, with the field that exists to catch it silently disarmed by an ordinary, non-breaking change. Now reported as an addition (not a removal — it is not breaking in itself):

```
RedisTurnQueue — is now a value export (was type-only)
```

Two new diff cases cover it, including the end-to-end sequence: upgrade reported → baseline refreshed → later removal caught as BREAKING.

## 3. `variants` was ambiguous and silently partial

Two problems, both worse than they look:

**Collision.** `literal.text` strips quotes, so `2 | 3` and `'2' | '3'` were byte-identical in the baseline. Normalising `RespVersion` from numbers to strings — which breaks every consumer passing `{ version: 2 }` — diffed to nothing, in the one union whose variants were recorded.

**Partial coverage.** Only string and numeric literals were collected. `-1` (a `PrefixUnaryExpression`, and the canonical Redis sentinel for TTL / `LPOS` rank / `XAUTOCLAIM`), `1n`, `undefined` and template-literal arms fell through *while their siblings were recorded*. An entry that looks pinned but is not is worse than the pre-#392 behaviour, where a union recorded nothing and was visibly untrusted.

Union arms are now recorded by **source text**, all of them; object-shaped arms still contribute `members`. `getText()` handles `-1`, `1n`, `undefined` and template literals for free and keeps `2` apart from `'2'`.

Side effect: five unions gain coverage they never had.

| symbol | was | now |
| --- | --- | --- |
| `RedisDataValue` | *(nothing)* | 6 arms |
| `ExecutorResult` | *(nothing)* | 2 arms |
| `CommandExecutionResult` | *(nothing)* | 3 arms |
| `RedisCommandArgument` | *(nothing)* | 3 arms |
| `NodeRedisCommandArgument` | *(nothing)* | 2 arms |

## 4. The const gate discriminated on `type` vs `interface`

The gate blocked on `ObjectFlags.Anonymous`. A type alias creates no distinct type, so a const annotated with a named **alias** was "anonymous" and recorded its members; one annotated with an **interface** was not. Same authorial intent, opposite outcome — and converting `type Opts = {…}` to `interface Opts {…}`, invisible to every consumer, was reported as BREAKING, sending the author off to write a changelog entry for a no-op.

It now also blocks when the type's `aliasSymbol` names something this surface already exports, so both spellings give the same answer and the members stay pinned under `Opts` where they belong. `FEATURE_GATES` is unaffected: its `aliasSymbol` is lib's `Record`, which this package does not export.

The docstring also claimed the checker pass made intersections answer with their keys. It does not — `type.flags & Object` is false for an intersection. The asymmetry with the alias form is left in place (for a const the members are already pinned under the referenced names, which is what the new gate is for) and now documented instead of misdescribed.

## 5. Declaration merging, enums, and a dependency-churn guard

- **`type:` prefix.** `RedisValue` is the only declaration-merged export on either entry point — a 16-arm union plus a factory const. Since #392 taught the extractor to descend unions, the type half contributed 8 field names (`kind`, `items`, `value`, …) into the same list as the callable factories. `RedisValue.kind` is not something a consumer calls, and nothing stopped a future factory named `name` or `items` from being deleted unreported. The type half is now prefixed `type:`, symmetric with `static:`.
- **Enums.** No branch existed, so `export enum E { A, B }` baselined as `{"kind":"value"}` and removing `E.A` would ship green. None are exported today; added because completeness is this guard's premise.
- **`node_modules`.** A symbol declared in a dependency is pinned by name only. Re-exporting a dependency's type would otherwise snapshot its declared members, turning every version bump into baseline churn and its API changes into "breaks" here.

> On the builtin case specifically: the re-review's residue was that `export type { Socket } from 'net'` would snapshot 35 members out of `@types/node`. I could not reproduce that under this extractor's compiler options — a re-exported builtin does not resolve at all and trips the unresolvable-declaration invariant first. So the guard is real but the builtin is not what exercises it; the test uses `ioredis` (a declared peer dep, which does resolve) and the header says so rather than repeating a claim I could not verify.

**Red-run for the guard**, neutered on this branch:

```
    not ok 10 - does not pull members out of node_modules
# tests 42
# pass 41
# fail 1
```

## 6. `FEATURE_GATES` / `FeatureId` redundancy — left as-is, with the measurement

The review estimated that adding one feature gate produces a "~60-line baseline diff", making the file rubber-stampable. I measured it on merged `main`: add one `FeatureId` arm plus its `FEATURE_GATES` entry, regenerate, and the baseline diff is

```
3	0	tests-package/export-surface.json
```

**three added lines**, one per recording (`core.FEATURE_GATES.members`, `core.FeatureId.variants`, `index.FeatureId.variants`). Not the churn the finding describes.

And the redundancy is not pure. TypeScript keeps the two in sync only because the const is annotated `Record<FeatureId, VersionGate>`; loosen that to `Partial<Record<…>>` and a dropped key breaks consumers reading the object at runtime while the type union still advertises it. `FEATURE_GATES` is a value export — the object's keys are surface in their own right.

So no de-duplication. Both the measurement and the reason are now in the file header, which is what the finding offered as its alternative.

## What still isn't covered

Unchanged from #392, plus one correction: **class statics and enum members are covered** (statics were already, contrary to #392's own gap list). Still out of scope:

- **Type shape.** Members and variants are pinned by name/text only. Narrowing `logger?: Logger` to `logger: Logger`, changing a return type, reordering overloads — all pass.
- **Generic parameters.** Adding a required type parameter to an exported generic is breaking and invisible.
- **Union arms individually.** A union records the union of its arms' members, so moving a field between arms is invisible; deleting it outright is not.
- **Behaviour**, and **`dist/cli.js`** (not an importable surface).

## Testing

| Suite | Result |
| --- | --- |
| `npm run lint` | pass |
| `npm test` (incl. `tsc --noEmit`) | pass — 342/342 |
| `tsc -p tsconfig.tests-package.json` | pass |
| `npm run build` | pass |
| `npm run test:package` | pass — **42/42** (was 27) |
| `npm run test:package` as `js-valkey-server` | pass — 42/42 |
| `npm run test:integration:mock:node-redis` | pass — 526/526 |

The `:real` suites were not run (docker-compose Redis). `npm run export-baseline` is idempotent — re-running on a clean tree produces no diff — and the fixture scratch directory is removed after the run, leaving `git status` clean.

## Not touched

`.prettierignore`'s `.md` → `*.md` typo is PR #394's. `CONTRIBUTING.md` is edited for content only, not reformatted — `origin/main` and this branch are prettier-dirty by identical deviations, and Markdown is unreachable from both `.lintstagedrc` and `npm run format`.

---

# Review round 2 (`bc41487`)

Nine findings addressed. `test:package`: **42 → 52 tests**, every new behaviour with its own red run.

## The HIGH — `type:` is no longer conditional

The prefix was applied only when the symbol had a value half. Both directions of the finding reproduce, and both are fixed by the same change: **the prefix follows the declaration a member was written on, and nothing else.** `type:` for interfaces *and* type aliases, `static:` for class statics, bare for the value side.

Unconditional is what makes it stable. Adding a declaration cannot rename an existing member, and because `interface` and `type` agree, converting one to the other stays invisible.

**Red run — direction 1**, conditional prefix reintroduced (`isMerged`):

```
    not ok 12 - adding a value half is purely additive, not a removal
      error: |-
        + [
        +   'Foo.a — member gone'
        + ]
        - []
# tests 52  # pass 41  # fail 10
```

A strictly additive change reported as BREAKING — exactly what the `aliasSymbol` gate was added to eliminate, one level up.

**Red run — direction 2**, interface members flattened into the value namespace:

```
    not ok 11 - namespaces the interface spelling of a merge identically
    not ok 13 - converting an interface to a type alias is invisible
    not ok 18 - the const gate fires through a renamed re-export
# tests 52  # pass 44  # fail 7
```

Baseline churn is large and entirely mechanical. I diffed every entry with the prefixes normalised away: **0 substantive changes**, same 91 + 266 symbols, no members lost or gained.

## Quote and paren leakage

Raw `getText()` copied the source's quote characters into the baseline. String literals are now re-quoted in one canonical style and parentheses unwrapped; `2` stays `2`, so the collision fix holds. The two message templates lost their own quotes, so the headline case reads `CommandFlag — union variant 'noscript' gone` again.

**Red runs:** canonicalisation removed → `not ok 5 - canonicalises quoting so the baseline does not track source style` (51/52). Paren unwrap removed → `not ok 6 - ignores redundant parentheses around a union arm` (49/51 at the time).

## Empty union arms

An arm the extractor descends into but which yields nothing — an intersection of two named types, an index-signature-only object — vanished on both paths. It now falls back to its source text:

```
type U  = 'lit' | (Foo & Bar)          -> variants ["'lit'", "Foo & Bar"]
type U2 = 'lit' | { [k: string]: number } -> variants ["'lit'", "{ [k: string]: number }"]
```

The fallback measures what the arm *itself* contributed rather than watching the shared set grow. My first attempt used set growth and was wrong in a way the real build caught immediately: both arms of `CreateIoredisMockOptions` declare `seed` and `cluster`, so the second arm looked like it contributed nothing and fell back to source text. Fixed before the baseline was committed; the current baseline has no such entry.

**Red run:** fallback removed → `not ok 8 - records a union arm that contributes no nameable members` (51/52).

## `/node_modules/` path substring → `isSourceFileFromExternalLibrary`

The path test described where the *file* sat, not where the dependency came from. A `dist/` that itself lives under a `node_modules/` directory lost every member and variant while still looking plausible — and neither invariant catches it, because every declaration resolves. New test reads the identical source from both locations and asserts the surfaces are equal.

**Red run:** substring restored → `not ok 2 - a surface read from inside node_modules keeps its members` (51/52).

## Const gate by symbol identity

**The finding's first example needed a correction, and it changed the test.** Comparing `aliasSymbol` by *name* against the set I was building did not reproduce the renamed-re-export failure, because I build the set from **resolved** symbols — which carry the original name `Opts`, not the export name `Options`. The bug is real, but it lives in the comparison against *export* names, which is what the pre-#397 code did. So the neutering had to restore that, not just swap identity for name:

**Red run:** `aliasSymbol` matched against export names →

```
    not ok 18 - the const gate fires through a renamed re-export
    not ok 19 - the const gate does not fire on a mere name collision
# tests 52  # pass 50  # fail 2
```

Both directions of the finding are now covered — the second (a non-exported alias colliding with an unrelated export, dropping members nothing else pins) is the one only identity comparison fixes.

## Namespaces and enum values

`declare namespace fn { … }` merged onto a function now records its members. Enum members record their values, since a `const enum` value is inlined into consumers' builds — the same class of break as `2 | 3` becoming `'2' | '3'`, which the union handling goes out of its way to surface.

**Red runs:** namespace branch removed → `not ok 15 - pins the members of a merged namespace` (51/52). Enum initializers dropped → `not ok 14 - pins enum members with their values` (51/52).

## Header corrections

- **Builtins.** Your version is the one in the header now, because it is the one that reproduces: the program is created with **no `types` and no `typeRoots`**, so `@types/node` never enters it; add `types: ['node']` and `Socket` resolves with one declaration. Not a `/tmp` artifact, not a property of `moduleResolution: Bundler` or of declaration-only inputs — one config key away. The invariant's message now names that cause too, instead of only telling the reader to rebuild.
- **`FEATURE_GATES`.** You are right and I was wrong: the declaration file emits the *annotation*, so `Partial<Record<…>>` records the full key set exactly as `Record<…>` does, and a dropped key is invisible on both. Re-pointed at the case that does hold — the annotation being dropped entirely, so the inferred object type is emitted and the keys become independent surface. The 3-line measurement stands.
- **"Non-object constituent"** did not describe the code: a named object type used as an arm *is* a variant, and `RedisDataValue` — the PR's own headline example — is six of them. Reworded to "every union arm the extractor does not descend into", here and in `CONTRIBUTING.md`.
- The type-reference-arm hazard (renaming a non-exported internal type that appears as an arm reads as BREAKING) is written down rather than fixed. Not reachable today — every reference-shaped variant is an exported type or a global.

## Test-file fixes

Stale rationale corrected: the scratch directory is under `dist/` because the `ioredis` case needs the repo's `node_modules` to resolve, not because of a builtin case that the file explicitly does not contain. `after()` now guards `scratch` so a `before()` failure does not bury the real error under an `ERR_INVALID_ARG_TYPE`.

## Testing

| Suite | Result |
| --- | --- |
| `npm run lint` | pass |
| `npm test` (incl. `tsc --noEmit`) | pass — 342/342 |
| `tsc -p tsconfig.tests-package.json` | pass |
| `npm run build` | pass |
| `npm run test:package` | pass — **52/52** |
| `npm run test:package` as `js-valkey-server` | pass — 52/52 |
| `npm run test:integration:mock:node-redis` | pass — 526/526 |

`:real` suites not run. Baseline regeneration idempotent; fixture scratch removed, `git status` clean.


tests-package/export-surface.ts
```diff
@@ -172,19 +251,56 @@ function checkerMembersOfConst(
...
+
+  // A declaration-merged symbol — `RedisValue` is the only one on either entry
+  // point — has a type half and a value half whose names mean different things.
+  // `RedisValue.kind` is a field of the union; `RedisValue.push()` is a factory
+  // on the const. Flattening both into one list re-introduces, one level up,
+  // exactly the masking `static:` was added to stop: a future factory named
+  // `name` or `items` could be deleted unreported because a union arm already
+  // contributed that string. It also makes the entry unreviewable — you cannot
+  // tell which of the 24 names are callable.
+  const isMerged =
```
fatal10110: **HIGH — `isMerged` is wrong in both directions: it makes `type:` conditional on a value half existing (so an *additive* change reports BREAKING), and it misses the `interface` spelling entirely (so the masking it exists to stop still happens).**

The comment above is right about the hazard. The predicate is the problem: `type:` is a property of *which declaration a member came from*, not of whether some other declaration happens to exist.

**Direction 1 — false BREAKING on a purely additive change.** I hit this while verifying the `type`→`value` fix, on that fix's own happy path. Adding a factory const beside an existing exported type alias:

```ts
export type Foo = { a: string }                                  // -> {"kind":"type","members":["a"]}
export type Foo = { a: string }
export declare const Foo: { make(): void }                       // -> {"kind":"value","members":["make","type:a"]}

findAdditions -> ["Foo — is now a value export (was type-only)", "Foo.make", "Foo.type:a"]
findRemovals  -> ["Foo.a — member gone"]        <-- BREAKING, for a strictly additive change
```

That is exactly the outcome the `aliasSymbol` gate was added to eliminate ("sent the author off to write a changelog entry for a no-op"), reintroduced one level up — and it fires on the same commit that taught `findAdditions` to report the upgrade, so the intended "upgrade reported, refresh, move on" flow instead hands the author a breaking-change failure. The reverse is just as bad: deleting the const half of `RedisValue` is genuinely breaking, but the one true removal arrives buried under 8 spurious `type:*` removals.

**Direction 2 — the interface spelling is not covered, so it still masks.** `interface Foo {…}` + `declare const Foo` is the other, more idiomatic way to spell this merge; it takes the `isInterfaceDeclaration` branch at line 304 and flattens unprefixed:

```
interface+const : {"kind":"value","members":["kind","make","value"]}
typealias+const : {"kind":"value","members":["make","type:kind","type:value"]}
```

and the masking itself:

```ts
export interface Coll { name: string }
export declare const Coll: { name(): void; other(): void }   // -> ["name","other"]
export interface Coll { name: string }
export declare const Coll: { other(): void }                 // -> ["name","other"]

findRemovals(before, after) -> []
```

The factory `Coll.name()` was deleted, unreported, because the interface half still contributes `name` — verbatim the scenario this comment describes.

Both directions go away if the prefix stops being conditional and is applied by *source declaration* instead: a member collected from a type alias or interface is always `type:`, a static is always `static:`, regardless of what else the symbol declares. That costs a one-time baseline churn, and it is what actually makes the two namespaces disjoint — which is the stated goal. If the conditional form is kept instead, it at least needs `ts.isInterfaceDeclaration` in the predicate and the prefix applied in the interface branch too.

Not live on today's build — I enumerated every multi-declaration export on both entry points: `RedisValue` (`TypeAliasDeclaration+VariableDeclaration`) is the only merge, and `createRedisServer` is two overloads, correctly flattened.

_Architecture lens, both reviewers independently._

tests-package/export-surface.ts
```diff
@@ -172,19 +251,56 @@ function checkerMembersOfConst(
...
 }
 
+/**
+ * Re-exporting a dependency's type would snapshot that dependency's own
+ * declared members, so every version bump of it would churn the baseline and
+ * the removal check would start reporting its API changes as breaks in this
+ * package. Nothing does so today; this keeps it that way by construction. The
+ * symbol itself is still pinned, just without members.
+ */
+function isThirdPartyDeclaration(declaration: ts.Declaration): boolean {
```
fatal10110: **MEDIUM — a substring match on the absolute path means any checkout living under a `/node_modules/` path silently loses every member and variant, and the new invariants do not catch it.**

`fileName.includes('/node_modules/')` tests where the *file* is, not where the *dependency* is. Identical source, identical content, two locations:

```
<scratch>/norm/x.d.ts
  -> {"Bar":{"kind":"value","members":["go"]},"Foo":{"kind":"type","members":["alpha","beta"]}}

<scratch>/node_modules/js-redis-server/dist/x.d.ts
  -> {"Bar":{"kind":"value"},"Foo":{"kind":"type"}}
```

Levels 2 and 3 — the two the module header calls "what make this guard worth having" — evaporate, and the surface still looks entirely plausible. Reachable from a linked or vendored checkout, a monorepo that installs the package into a workspace, or CI that runs against an installed copy.

The dangerous direction is `npm run export-baseline`: run it from such a tree and the emptied baseline gets committed, after which every real member removal passes green forever. That is precisely the silent-degradation failure this PR added two invariants against — and neither invariant covers it, because every declaration resolves fine; only the members are dropped.

The compiler already answers this exactly, and `program` is in scope in `readExportSurface`:

```ts
program.isSourceFileFromExternalLibrary(declaration.getSourceFile())
```

(or compare against the entry file's own root directory). Worth threading `program` through `describeSymbol` for it.

_Architecture lens._

tests-package/export-surface.ts
```diff
@@ -105,60 +129,115 @@ function collectFromTypeNode(
...
+        continue
+      }
 
-    if (
-      literal.kind === ts.SyntaxKind.TrueKeyword ||
-      literal.kind === ts.SyntaxKind.FalseKeyword ||
-      literal.kind === ts.SyntaxKind.NullKeyword
-    ) {
-      into.variants.add(literal.getText())
+      into.variants.add(sourceTextOf(constituent))
```
fatal10110, line 226: **MEDIUM — "every constituent contributes" is not true: an arm that is an intersection of references, or an object with only index signatures, contributes *nothing* — the partial-but-plausible shape this very comment argues against.**

`isMemberBearing` routes intersections, type literals and nested unions to `collectFromTypeNode` instead of `variants`. When such an arm has no syntactically nameable own members, it is dropped on both paths:

```ts
export type Foo = { f: string }
export type Bar = { b: string }
export type U = 'lit' | (Foo & Bar)      // -> {"kind":"type","variants":["'lit'"]}

// delete the whole (Foo & Bar) arm — a breaking narrowing for any consumer passing that shape
findRemovals -> []
```

Same for an index-signature-only arm:

```ts
export type U2 = 'lit' | { [k: string]: number }   // -> {"kind":"type","variants":["'lit'"]}
```

This is the failure shape argued against four lines below — "the entry would look pinned while … arms vanished silently next to their captured siblings" — reached through a different door. The union is recorded, one arm is visibly pinned, and the other does not exist as far as the baseline is concerned.

Keeping the intent, the fix is a fallback rather than a new branch: descend a member-bearing arm, and if it contributed no members, record its source text as a variant instead.

```ts
for (const constituent of node.types) {
  if (isMemberBearing(constituent)) {
    const before = into.members.size
    collectFromTypeNode(constituent, into)
    if (into.members.size > before) continue
  }
  into.variants.add(sourceTextOf(unwrapParens(constituent)))
}
```

Related wording, same root confusion: "non-object constituent" (line 18 here, and the identical sentence in `CONTRIBUTING.md`) does not describe what the code does. Object *types referenced by name* are recorded as variants — the PR's own headline example is one:

```
"RedisDataValue": { "variants": ["RedisHashData","RedisListData","RedisSetData", …] }
```

Six object types, all in `variants`, cited in the docs as the example of a non-object constituent. The accurate rule is "every arm the extractor does not descend into — anything that is not an inline object literal, intersection or nested union — by source text".

_Architecture lens._

tests-package/export-surface.ts
```diff
@@ -105,60 +129,115 @@ function collectFromTypeNode(
...
+/**
+ * A union arm's source text, whitespace-collapsed.
+ *
+ * Source text rather than `literal.text`, because `.text` strips the quotes and
+ * makes `2 | 3` and `'2' | '3'` byte-identical in the baseline — so normalising
+ * `RespVersion` from numbers to strings, which breaks every consumer passing
+ * `{ version: 2 }`, would diff to nothing. `getText()` keeps them apart and
+ * handles `-1`, `1n`, `undefined` and template-literal arms for free.
+ */
+function sourceTextOf(node: ts.Node): string {
```
fatal10110, line 278: **MEDIUM — source text is the right call, but raw `getText()` leaks the quote characters into the baseline: the guard's own headline message now renders `''noscript''`, redundant parens read as BREAKING, and the baseline is coupled to `.prettierrc`.**

No argument with the direction — I confirmed the collision this closes (`2 | 3` vs `'2' | '3'` are now distinct) and that all four previously-vanishing arm kinds are captured. Three consequences of using the raw text, one of them live today.

**Live — the message is mis-rendered.** `findRemovals` wraps the variant in its own quotes (`union variant '${variant}' gone`) and the variant now already carries them. The canonical scenario this file's header cites, dropping `'noscript'` from `CommandFlag`, prints:

```
CommandFlag — union variant ''noscript'' gone
```

87 of the 107 variants in the committed baseline are quoted string literals, so this is what nearly every real variant removal looks like. Before this PR `.text` stripped the quotes and it read cleanly.

**Live — redundant parens read as a removal.** `unwrapParens` is applied when deciding whether to descend, but not when recording the text:

```ts
export type P = 'a' | 'b'      // variants: ["'a'","'b'"]
export type P = ('a') | 'b'    // variants: ["'b'","('a')"]
findRemovals -> ["P — union variant ''a'' gone"]
```

A no-op reformat reported as BREAKING.

**Latent — the baseline depends on source quote style.** tsup's DTS emit normalises whitespace and arm order (I checked both; neither can move the baseline, so reordering and reformatting a union are correctly invisible) but passes literal quote characters through verbatim. Flipping two arms in `src/core/command-definition.ts` to double quotes and rebuilding:

```
dist: type CommandFlag = "readonly" | "write" | 'denyoom' | 'admin' | …
findRemovals -> CommandFlag — union variant ''readonly'' gone
                CommandFlag — union variant ''write'' gone
```

CI's `prettier --check` plus `singleQuote: true` stops anyone landing that by hand, so the reachable path is a `.prettierrc` change or a prettier default change — at which point a pure reformat reports 87 breaking removals across `CommandFlag`, `FeatureId`, `RedisDataTypeName`, `ClientSessionMode`, `PubSubKind`, `RedisFlavor` and `CompatibilityProfile`.

One change addresses all three without giving up the collision fix — canonicalise string literals, unwrap parens, keep `getText()` for everything else:

```ts
function sourceTextOf(node: ts.Node): string {
  const inner = ts.isTypeNode(node as ts.TypeNode) ? unwrapParens(node as ts.TypeNode) : node
  if (ts.isLiteralTypeNode(inner) && ts.isStringLiteral(inner.literal)) {
    return JSON.stringify(inner.literal.text)   // '2' -> "2", still distinct from 2
  }
  return inner.getText().replace(/\s+/g, ' ').trim()
}
```

`2` stays `2`, `-1` / `1n` / `undefined` / template arms are untouched, and dropping the outer quotes from the two message templates makes the output readable again.

Same root cause, worth a header line rather than code: a union arm that is a *type reference* is pinned by its spelling, so renaming a non-exported internal type that appears as an arm reports BREAKING for a consumer-invisible refactor. Not live — every reference-shaped variant in the baseline is either an exported type or a global (`Buffer`, `Promise<RedisResult>`).

_Architecture lens._

tests-package/export-surface.ts
```diff
@@ -31,8 +33,27 @@
...
+//
+// A symbol whose declaration lives in `node_modules` is pinned by name only.
+// Nothing re-exports a dependency's type today, but doing so would otherwise
+// snapshot that dependency's declared members, turning every version bump of it
+// into baseline churn and its API changes into "breaks" in this package.
+// (A re-exported Node *builtin* never reaches that guard: under this program's
+// options `export type { Socket } from 'net'` does not resolve at all and trips
+// the unresolvable-declaration invariant first.)
+//
+// `FEATURE_GATES.members` deliberately repeats `FeatureId.variants`. TypeScript
```
fatal10110: **MEDIUM — the measurement is exactly right; the `Partial<Record<…>>` reason underneath it does not reproduce.**

I re-ran the churn measurement rather than taking it on trust. One `FeatureId` arm plus its `FEATURE_GATES` entry, regenerated:

```
3	0	tests-package/export-surface.json

+          "'probe.new-gate'",      # index.FeatureId.variants
+          "probe.new-gate",        # core.FEATURE_GATES.members
+          "'probe.new-gate'",      # core.FeatureId.variants
```

Three lines, exactly as claimed. The "~60-line diff" estimate was wrong and the decision not to de-duplicate is well supported. No objection to the outcome.

The justification is the problem. The claim is that the redundancy earns its keep because loosening the annotation to `Partial<Record<…>>` would let a dropped key break runtime consumers while `FeatureId` still advertises it. It would not — the declaration file emits the *annotation*, not the object literal, so `getProperties()` answers with the full key set either way:

```
export declare const G: Record<Fid, Gate>           -> members ["a","b","c"]
export declare const G: Partial<Record<Fid, Gate>>  -> members ["a","b","c"]

findRemovals(strict, loosened-and-key-dropped) -> []
```

Under the exact scenario the header describes, `FEATURE_GATES.members` still mirrors `FeatureId.variants` and the dropped key is still invisible. So a maintainer who loosens that annotation would be doing it believing this guard has them covered — the opposite of what the paragraph intends.

The redundancy *does* pay off, just for a different reason: when the annotation is dropped entirely and the inferred literal type is emitted, the keys become independent surface and a drop is caught.

```
export declare const G: { a:{r:string}; b:{r:string}; c:{r:string} }   # key c removed
findRemovals -> ["G.c — member gone"]
```

Worth re-pointing the paragraph at that case, since this header is now the documented answer to a review finding and will be read as settled. Same shape as #392's "stated rationale is factually wrong" finding you fixed in this module header.

_Architecture lens._

tests-package/export-surface.ts
```diff
@@ -172,19 +251,56 @@ function checkerMembersOfConst(
     return []
   }
 
+  const alias = type.aliasSymbol?.getName()
```
fatal10110: **MEDIUM — the new gate compares `aliasSymbol` by *name*, so it misses a renamed re-export (the `type`/`interface` flip-flop is still reachable) and can silently drop members on a name collision.**

`exportedNames` is a set of strings, so the gate asks "is there an export called `Opts`?" rather than "is this the exported symbol `Opts`?".

**Misses the gate.** Export the type under a different name and the gate never fires — `aliasSymbol.getName()` is `Opts`, `exportedNames` holds `Options`:

```ts
type Opts = { alpha: string; beta: number }
declare const DEFAULTS: Opts
export { type Opts as Options, DEFAULTS }

-> { "DEFAULTS": {"kind":"value","members":["alpha","beta"]},
     "Options":  {"kind":"type","members":["alpha","beta"]} }
```

Members duplicated, and converting `Opts` from `type` to `interface` — invisible to every consumer — flips `DEFAULTS` between having members and not, which is reported as BREAKING. That is the exact bug this gate was added to fix, still reachable through the rename.

**Silently drops members.** The other direction is worse because nothing else pins them: a const annotated with a *non-exported* alias whose name happens to collide with an unrelated export has its members dropped, and no other entry records them.

Comparing resolved symbols by identity instead of name fixes both — build the set from the aliased symbols (`checker.getAliasedSymbol` where `SymbolFlags.Alias` is set, which `readExportSurface` already computes at line 423) and test `type.aliasSymbol` against it.

Not triggered by today's build: `FEATURE_GATES` is the only const reaching the gate and its alias is lib's `Record`, and no `/core` export resolves to a declaration whose name differs from the export name.

_Architecture lens._

tests-package/export-surface.ts
```diff
@@ -198,13 +314,41 @@ function describeSymbol(
       continue
     }
 
+    if (ts.isEnumDeclaration(declaration)) {
```
fatal10110: **LOW — enums landed; the `ModuleDeclaration` half of that finding did not, and enum *values* are not pinned.**

Two gaps around this branch, both on the same "completeness is this guard's premise" grounds the branch was added on.

**Namespace merging records nothing.** #392's finding named "no branch for `EnumDeclaration` *or* `ModuleDeclaration`". The `ModuleDeclaration` falls through the loop unmatched:

```ts
export declare function fn(): void
export declare namespace fn { const helper: number; function sub(): void }

-> fn {"kind":"value"}
```

`fn.helper` and `fn.sub` are callable published surface and deleting either ships green.

**Enum values are not surface.** Members are pinned by name only, so a `const enum` value change — breaking, and inlined into consumers' builds — diffs to nothing:

```ts
export declare const enum CE { A = 0 }   // -> members ["A"]
export declare const enum CE { A = 5 }   // -> members ["A"]
findRemovals -> []   findAdditions -> []
```

This one is worth weighing against the `sourceTextOf` rationale added in the same PR: the argument there is that `2 | 3` → `'2' | '3'` must be visible *because* `{ version: 2 }` breaks at runtime. An enum value change is the same class of break and is invisible. Recording `A=0` rather than `A` would be symmetric.

Neither shape is exported today, same as enums before this PR. Either add the branches, or add both to the "what still isn't covered" list — that list is careful enough elsewhere that a reader will treat an omission as coverage.

_Architecture lens._

tests-package/export-surface.ts
```diff
@@ -31,8 +33,27 @@
...
+// would otherwise collide in one list: `static:` for class statics (so
+// `static create()` cannot mask an instance `create()`), and `type:` for the
+// type half of a declaration-merged symbol (`RedisValue` is the only one — its
+// union arms' field names are not callable factories).
+//
+// A symbol whose declaration lives in `node_modules` is pinned by name only.
+// Nothing re-exports a dependency's type today, but doing so would otherwise
+// snapshot that dependency's declared members, turning every version bump of it
+// into baseline churn and its API changes into "breaks" in this package.
+// (A re-exported Node *builtin* never reaches that guard: under this program's
```
fatal10110: **LOW — the builtin claim is correct; the mechanism is narrower than stated, and the invariant's error text sends the contributor somewhere useless.**

Worth pinning down, since this has now been stated three different ways across #392 and here. **Your version is the right one** — I reproduced it with the fixture inside `dist/` so `node_modules` is reachable (`@types/node@22.20.0` installed), and every spelling trips the invariant rather than reaching the members code:

```
export type { Socket } from 'net'          -> THROWS: 1 exported symbol(s) have no resolvable declaration … First few: Socket
export { Socket } from 'net'               -> THROWS (same)
export type { Socket } from 'node:net'     -> THROWS (same)
export type { Buffer } from 'node:buffer'  -> THROWS (same)
```

So the 35-member `@types/node` snapshot from #392's inline note does not happen, and substituting `ioredis` is right.

The *reason* is narrower than "under this program's options", which is what lines 46-48 say. The program is built with no `types` and no `typeRoots`, so `@types/node` never enters it:

```
resolveModuleName('net', …)             -> (UNRESOLVED)
program files under @types/node         -> 0
same program + types: ['node']          -> exports: ['Socket'], declarations: [1]
```

It resolves the moment `@types/node` is in the program. That is not a property of `moduleResolution: Bundler` or of declaration-only inputs — it is one config key away, so the parenthetical reads as more permanent than it is.

The actionable half: if anyone does re-export a builtin from `src/`, `npm run test:package` fails on a perfectly healthy build with

> `… have no resolvable declaration, so the extracted surface would be wrong. Rebuild with \`npm run build\`.`

which cannot fix it and points at the build. Since this invariant exists so degradation "fails loudly with a name", worth having the message admit the other cause — "…or an export this program cannot resolve, such as a Node builtin re-export". One clause, saves the next person a long detour.

_Architecture lens._

tests-package/export-surface-extractor.test.ts
```diff
@@ -0,0 +1,256 @@
...
+
+let scratch: string
+
+before(() => {
+  const distDir = fileURLToPath(new URL('../dist/', import.meta.url))
+  mkdirSync(distDir, { recursive: true })
+  scratch = mkdtempSync(join(distDir, 'export-surface-fixtures-'))
+})
+
+after(() => {
```
fatal10110, line 46: **LOW — this header justifies the scratch location by a case the file deliberately does not contain, and `after()` can throw over the real failure.**

**Stale rationale.** Lines 22-24 say the scratch lives under `dist/` because "a fixture in `/tmp` cannot resolve `@types/node`, and would make the builtin case below fail for the wrong reason". There is no builtin case below — line 244 explains at length why a builtin *cannot* be used and substitutes `ioredis`. A reader who later wants to move these fixtures to the OS temp dir will check the stated reason, find no builtin case, conclude the constraint is obsolete, and break `does not pull members out of node_modules`, which genuinely does need the repo's `node_modules` to resolve `ioredis`. The real constraint is "the fixture must sit inside a tree that resolves the repo's `node_modules`, for the `ioredis` case".

**Unguarded cleanup.** `after()` dereferences `scratch` unconditionally. If `before()` throws — `mkdtempSync` on a full or read-only CI disk — the cleanup hook throws `ERR_INVALID_ARG_TYPE: The "path" argument must be of type string … Received undefined` on top of it, and that is what lands at the top of the log instead of the real failure. `if (scratch) rmSync(…)`.

For what it's worth the lifecycle is otherwise sound: I confirmed the scratch dir is removed and `git status` is clean after a full `test:package` run, and `node --test` gives each `tests-package` file its own process, so the shared `scratch` binding is not raced.

_Architecture lens._

fatal10110:
## Review — ship, after the `isMerged` finding

Every claim in the description reproduces. The two red runs are exact, the baseline is idempotent and reviewable, and both push-backs hold up on their merits. One finding is worth fixing before merge; the rest are latent or documentation-level.

**No wire-compat lens on this one.** The PR touches no `src/` file — it is `tests-package/` tooling plus the regenerated baseline. There is no protocol surface, no RESP encoding, no error wording and no command flag in the diff, so there was nothing to ground-truth and I did not start a Redis. Saying that plainly rather than manufacturing protocol findings. The weight is entirely on the architecture lens.

### Claims I reproduced

| Claim | Result |
|---|---|
| Both invariant `throw`s deleted → `40/42`, failing exactly the two new tests | **Confirmed, exactly** — `not ok 1 - throws when the entry resolves to zero exports`, `not ok 2 - throws, naming symbols, when a re-export target does not resolve`, nothing else red |
| `node_modules` guard neutered → `41/42` | **Confirmed, exactly** — `not ok 10 - does not pull members out of node_modules` |
| `type` → `value` upgrade reported, refreshed, later removal caught as BREAKING | **Confirmed end to end** on real `.d.ts` fixtures, not just synthetic surfaces (below) |
| Five unions gain incidental coverage | **Confirmed** — `RedisDataValue` 6 arms, `ExecutorResult` 2, `CommandExecutionResult` 3, `RedisCommandArgument` 3, `NodeRedisCommandArgument` 2 |
| Regeneration idempotent, tree clean after the run | **Confirmed** — two consecutive `export-baseline` runs produce no diff; no fixture dir survives in `dist/`, `git status` clean |
| `test:package` still fast | **Confirmed** — tsc 0.8s + tests 2.6s + build ≈ 8s total. Fine for the normal loop |

The upgrade sequence, driven through `readExportSurface` on real declaration files:

```
S1 type-only        {"Foo":{"kind":"type","members":["a"]}}
S2 gains a binding  {"Foo":{"kind":"value","members":["make","type:a"]}}
   findAdditions -> ["Foo — is now a value export (was type-only)", …]
S3 binding removed  findRemovals(S2,S3) -> ["Foo — was a value export, is now type-only (no runtime binding)"]
   old hole:        findRemovals(S1,S3) -> []      <-- what this closes
```

### The two push-backs

**`FEATURE_GATES` churn — your measurement is right, the reason is not.** I re-measured on this branch: adding one feature gate is exactly `3 0 tests-package/export-surface.json`. The ~60-line estimate was wrong and keeping the redundancy is well supported. But the `Partial<Record<…>>` argument now in the header does not reproduce — the declaration file emits the annotation, not the literal, so `Partial<Record<Fid,Gate>>` and `Record<Fid,Gate>` both record all keys and a dropped key is still invisible. The redundancy pays off for a different reason (annotation dropped entirely → inferred literal → drop caught). Details inline.

**Builtin re-export — you are right, and here is the actual mechanism.** `export type { Socket } from 'net'` genuinely does not resolve and trips the unresolvable-declaration invariant, reproduced with the fixture inside `dist/` and `@types/node` installed. It is not a `/tmp` artifact and not a property of `moduleResolution: Bundler`: the program is built with no `types`/`typeRoots`, so `@types/node` never enters it — add `types: ['node']` and `Socket` resolves with one declaration. Substituting `ioredis` is the right call.

### Findings

| Sev | Location | Issue |
|---|---|---|
| **HIGH** | `export-surface.ts:293` | `isMerged` makes `type:` conditional on a value half existing → an additive change (adding a factory const beside an exported type) reports BREAKING; and the `interface`+`const` spelling is not covered, so the masking it exists to stop still happens |
| MED | `export-surface.ts:273` | `/node_modules/` path substring → any checkout under such a path silently loses all members and variants; `export-baseline` there commits a disarmed baseline |
| MED | `export-surface.ts:162` | Intersection-of-refs and index-signature-only union arms contribute nothing to either list — the partial-but-plausible shape the comment argues against |
| MED | `export-surface.ts:203` | Raw `getText()` leaks quotes: `union variant ''noscript'' gone` today, redundant parens read as BREAKING, and the baseline is coupled to `.prettierrc` |
| MED | `export-surface.ts:50` | The `Partial<Record<…>>` rationale does not reproduce (measurement is right) |
| MED | `export-surface.ts:254` | `aliasSymbol` compared by name, not identity → renamed re-export misses the gate; name collision silently drops members |
| LOW | `export-surface.ts:317` | `ModuleDeclaration` half of the #392 finding not done; enum *values* not pinned |
| LOW | `export-surface.ts:46` | Builtin determination is right but stated too broadly, and the invariant's "Rebuild" advice cannot fix that case |
| LOW | `export-surface-extractor.test.ts:41` | Header justifies the scratch location by a case the file deliberately omits; `after()` can throw over a `before()` failure |

Only the HIGH one is worth blocking on, and it is a small change. Everything below it is latent — I enumerated the built surface for each and none fires on today's `dist/`.

### Baseline churn — clean again

1076 → 1107 lines (+31). Eighteen symbol entries changed, every one explained by the stated intent: thirteen unions re-spelled with quotes or newly populated, `RedisValue` gaining its eight `type:` members. Normalising away quotes and the prefix, **nothing** was lost or unexpectedly gained. Still reviewable.

The one thing that would change my read is the HIGH finding's second direction — the `interface`+`const` masking is a live hole in the guard's own coverage, not just a false positive.

tests-package/export-surface.ts
```diff
@@ -172,39 +343,147 @@ function checkerMembersOfConst(
...
+ * it still looks entirely plausible. Neither invariant catches that, because
+ * every declaration resolves fine. Run `export-baseline` from such a tree and
+ * the emptied baseline gets committed, after which every real member removal
+ * passes green forever.
+ */
+function isExternalDeclaration(
+  program: ts.Program,
+  declaration: ts.Declaration,
+): boolean {
+  return program.isSourceFileFromExternalLibrary(declaration.getSourceFile())
```
fatal10110, line 377: **HIGH — the swap from `/node_modules/` to `isSourceFileFromExternalLibrary` does not fix the failure the comment above says it fixes. A `dist/` that lives under `node_modules/` still loses most of its surface, because the *shared chunk* is still classified external.**

The path test was wrong for the reason the comment gives. The replacement is wrong for a neighbouring reason: TypeScript flags a file as an external-library import based on where it was *resolved from*, and that includes a **relative** import resolved from inside a `node_modules` path. `dist/` is a two-file bundle — the entry plus `redis-error-*.d.ts`, where nearly every symbol actually declares — so the entry escapes (program roots are never flagged) and the chunk does not.

Instrumenting the real `dist/` copied to `<scratch>/node_modules/js-redis-server/dist/`:

```
--- normal
  isSourceFileFromExternalLibrary = false | redis-error-cHS6xBHh.d.ts
  isSourceFileFromExternalLibrary = false | core.d.ts
--- vendored
  isSourceFileFromExternalLibrary = true  | redis-error-cHS6xBHh.d.ts     <-- the chunk
  isSourceFileFromExternalLibrary = false | core.d.ts
```

and the surface that falls out of it:

```
core : normal [symbols,members,variants]=[266,454,72]   vendored=[266,143,3]
         findRemovals(normal, vendored) = 380 entries
index: normal [symbols,members,variants]=[91,122,35]    vendored=[91,92,5]
         findRemovals(normal, vendored) = 60 entries
```

All 266 symbols still resolve, so neither invariant throws. This is verbatim the scenario lines 363-371 describe — "run `export-baseline` from such a tree and the emptied baseline gets committed, after which every real member removal passes green forever" — and it is still live, on the very build Phase 4 is about to remove exports from.

It is a genuine improvement over the path test (143 members survive where 0 did), so this is not a regression. What makes it HIGH rather than MEDIUM is the pairing: a confident 10-line comment asserting the hole is closed, plus a test named after it that goes green (see the companion comment on `export-surface-extractor.test.ts`). That combination is how a guard gets trusted into disarming itself — which is the one failure this module exists to prevent.

The predicate is asking the wrong question. "Where does this file sit" and "did resolution pass through a `node_modules` segment" are both proxies; the question is "is this declaration part of *the bundle I was asked to read*". The entry file's own directory answers it directly and handles the sibling chunk for free:

```ts
function isExternalDeclaration(entryDir: string, declaration: ts.Declaration): boolean {
  return !declaration.getSourceFile().fileName.startsWith(entryDir)
}
```

`readExportSurface` already has the absolute entry path, so `dirname(entryDeclarationFile)` is in scope. `ioredis` still tests external (it resolves outside `dist/`), so the `does not pull members out of node_modules` case keeps passing.

_Architecture lens._

tests-package/export-surface-extractor.test.ts
```diff
@@ -0,0 +1,452 @@
...
+  })
+
+  test('a surface read from inside node_modules keeps its members', () => {
+    // The guard used to match `/node_modules/` in the file's own path, so a
+    // `dist/` that itself lives under one — a linked or vendored checkout, a
+    // workspace install, CI against an installed copy — lost every member and
+    // variant while still looking plausible. Neither invariant catches that:
+    // the declarations all resolve. Run `export-baseline` there and the
+    // emptied baseline gets committed, disarming the guard permanently.
+    const source = `export type Foo = { alpha: string; beta: number };
```
fatal10110, line 135: **HIGH (companion) — this fixture is a single self-contained `.d.ts`, and a program *root* is never flagged as an external library. So the test structurally cannot reproduce the failure its own comment describes: it goes green whether or not the real two-file bundle survives.**

The comment is an accurate description of a real and still-live bug (see the comment on `export-surface.ts:377`). The fixture cannot reach it. `dist/` is an entry plus a shared `redis-error-*.d.ts` chunk; the members live in the chunk, and only the chunk gets flagged. A one-file fixture has no chunk, so both sides come back identical for a reason that has nothing to do with the predicate under test.

Same fixture, written in the real artifact's shape — one relative re-export is the whole difference:

```
# two-file bundle: entry.d.ts = "export { type Foo, Bar } from './chunk.js'"
 normal: {"Bar":{"kind":"value","members":["go"]},"Foo":{"kind":"type","members":["type:alpha","type:beta"]}}
 nested: {"Bar":{"kind":"value"},"Foo":{"kind":"type"}}
 EQUAL?  false

# the fixture as committed here (single file)
 EQUAL?  true     <-- test passes
```

To be fair to it, the test is not vacuous in general: reverting `isExternalDeclaration` to `fileName.includes('/node_modules/')` does turn it red, so it does pin *something*. It just pins the half of the bug that was already fixed, and not the half that ships.

Splitting `source` into an entry and a chunk is the whole change, and it makes the test fail today — which is what you want before touching the predicate.

This is the same "untested invariant" argument this file's own header makes at lines 5-12, one layer further in: a test that cannot go red for the reason it names is the same vacuity as no test, minus the signal that one is missing.

_Architecture lens._

tests-package/export-surface.ts
```diff
@@ -97,68 +165,171 @@ function collectFromTypeNode(
...
+        }
+      }
 
-    if (
-      literal.kind === ts.SyntaxKind.TrueKeyword ||
-      literal.kind === ts.SyntaxKind.FalseKeyword ||
-      literal.kind === ts.SyntaxKind.NullKeyword
-    ) {
-      into.variants.add(literal.getText())
+      into.variants.add(sourceTextOf(constituent))
```
fatal10110, line 226: **MEDIUM — the empty-arm fallback makes an arm's *representation* depend on whether it currently happens to have a nameable member, so adding a field to such an arm reports a strictly additive change as BREAKING. That is the same failure class the `MemberNamespace` doc cites as the reason the prefix had to become unconditional, reintroduced two functions away.**

The fallback closes the real gap it was added for — I confirmed both shapes from the last round are now captured, and that removing the fallback reds `records a union arm that contributes no nameable members`. The cost is that an arm now moves between `variants` and `members` as it is edited:

```ts
export type A = { a: string }
export type B = { b: string }

export type T = (A & B) | 'lit'                  // -> {"variants":["'lit'","A & B"]}
export type T = (A & B & { z: number }) | 'lit'  // -> {"members":["type:z"],"variants":["'lit'"]}

findRemovals  -> ["T — union variant A & B gone"]     <-- BREAKING, for an additive edit
findAdditions -> ["T.type:z"]
```

Same for the index-signature arm, in the same direction:

```ts
export type T = { [k: string]: number } | 'lit'                 // -> variants ["'lit'","{ [k: string]: number }"]
export type T = { a: number; [k: string]: number } | 'lit'      // -> members ["type:a"], variant gone
findRemovals -> ["T — union variant { [k: string]: number } gone"]
```

The arm did not go anywhere and the type only widened. Line 105 argues, correctly, that a member's identity must not depend on "what else the symbol declares" — this is the same instability one level down, where an arm's identity depends on what else *the arm* declares.

Not reachable on the current build: scanning `export-surface.json` for a variant containing `{` or `&` returns zero hits, so this branch never fires against the real `dist/` today. That is also the argument for deleting it rather than fixing it — see the verdict comment.

If it stays, the stable form is to record the arm's source text *in addition to* its members rather than instead of them, so the arm keeps one identity across edits.

_Architecture lens._

tests-package/export-surface.ts
```diff
@@ -172,39 +343,147 @@ function checkerMembersOfConst(
...
+      continue
+    }
+
+    if (ts.isModuleDeclaration(declaration)) {
+      // `declare namespace fn { … }` merged onto a function or class. Its
+      // exported members are callable published surface.
+      const body = declaration.body
+
+      if (body && ts.isModuleBlock(body)) {
+        for (const name of namespaceMemberNames(body)) {
```
fatal10110, line 468: **MEDIUM — the new namespace branch adds its members with *no* prefix, which re-opens, for merged namespaces, exactly the type-side/value-side masking this PR's headline change closes for interfaces and type aliases.**

`namespaceMemberNames` returns bare strings and they go straight into `collected.members` alongside the class/function value side. A `declare namespace` can declare types, so the two namespaces collide again:

```ts
export declare class Foo { bar(): void; baz(): void }
export declare namespace Foo { type bar = string }
// -> {"kind":"value","members":["bar","baz"]}

export declare class Foo { baz(): void }              // method `bar` deleted — a hard break
export declare namespace Foo { type bar = string }
// -> {"kind":"value","members":["bar","baz"]}

findRemovals -> []
```

Deleting the method is invisible because the namespace's `type bar` still contributes the string `bar`. This is the `interface Coll` / `declare const Coll` scenario from the last round, in the branch added to answer the *other* half of that round's finding.

Two smaller things in the same function, both from hand-rolling a statement-kind allowlist where the compiler already has the answer:

- **A nested namespace is not pinned at all** — there is no `ts.isModuleDeclaration` case, so `declare namespace fn { namespace Inner { … } }` records nothing for `Inner`, not even its name. Deleting `fn.Inner` diffs to nothing.
- **Non-exported members are over-recorded.** `declare namespace fn { const priv: number; export const pub: number }` records both, but once a namespace has an explicit export, `priv` is not part of its surface. Removing an internal then reads as BREAKING.

`declaration.symbol.exports` answers all three at once, and the re-export forms for free:

```
checker on `declare namespace fn { const implicitOnly; export const explicit; namespace Inner { const deep } }`
  symbol.exports keys        -> ['implicitOnly', 'explicit', 'Inner']
  namespaceMemberNames(body) -> ['implicitOnly', 'explicit']
```

Not reachable today — `grep -c namespace dist/*.d.ts` is 0 in all three declaration files, same as enums were before this PR. Given that, deleting the branch is at least as defensible as fixing it; see the verdict comment.

_Architecture lens._

tests-package/export-surface.ts
```diff
@@ -97,68 +165,171 @@ function collectFromTypeNode(
...
+
+  if (ts.isLiteralTypeNode(inner) && ts.isStringLiteral(inner.literal)) {
+    const escaped = inner.literal.text
+      .replace(/\\/g, '\\\\')
+      .replace(/'/g, "\\'")
+
+    return `'${escaped}'`
+  }
+
+  return inner.getText().replace(/\s+/g, ' ').trim()
```
fatal10110, line 289: **MEDIUM — the quote fix is right and I confirmed it end to end, but canonicalising only *string literal type nodes* traded one source-spelling coupling for another: numeric arms regressed from canonical to raw spelling, and the whitespace collapse both over- and under-normalises. The sentence "which has no such degree of freedom" is the part that is wrong.**

First, the fix works. I flipped two `CommandFlag` arms in `src/core/command-definition.ts` to double quotes and rebuilt — the emitted declaration really does carry mixed quoting, so the hazard was real:

```
dist/redis-error-*.d.ts:814
type CommandFlag = "readonly" | "write" | 'denyoom' | 'admin' | 'noscript' | …
```

With the canonicalisation: `test:package` 52/52, zero removals. Neutering just the string branch on that same build: `CommandFlag — union variant 'readonly' gone` + `'write' gone`. The `.prettierrc` coupling is genuinely closed for union string arms.

**Numeric arms went the other way.** `main` recorded literals via `literal.text`, which TypeScript already normalises. `getText()` does not:

```
"1e3"        -> .text = "1000"      (main: canonical)     now: "1e3"
"0x10"       -> .text = "16"                              now: "0x10"
"1_000_000"  -> .text = "1000000"                         now: "1_000_000"
```

so rewriting a numeric arm in an equivalent spelling now reports BREAKING:

```
export type N = 1000 | 2   ->  export type N = 1e3 | 2     DIFFERS
export type N = 16 | 2     ->  export type N = 0x10 | 2    DIFFERS
```

And unlike the quote case there is **no CI backstop**. `prettier --check` + `singleQuote` is what made the quote bug only reachable via a config change; prettier preserves all three numeric spellings verbatim (I checked: `1e3 | 0x10 | 1_000_000 | 0b11 | 0o17` round-trips unchanged, it only lowercases hex digits). So this one is reachable by anyone typing `0x10` today. Only `RespVersion: 2 | 3` is numeric in the baseline, so it is not live — but it is strictly more reachable than the bug it replaced.

**`.replace(/\s+/g, ' ')` cuts both ways.** It is not enough for object-literal arms and too much for template-literal ones:

```
export type O = { [k: string]: number } | { [k:string]:number }
  -> variants ["{ [k: string]: number }","{ [k:string]:number }"]   # a reformat reads as BREAKING

export type T = `a  b` | `a b`
  -> variants ["`a b`"]                                             # two arms in, one out — deleting either is invisible
```

Template-literal whitespace is semantically significant, so the doc's "keep their source spelling, which has no such degree of freedom" is backwards for both kinds: numerics have several degrees of freedom and templates lose one they need.

All of this is one symptom: `sourceTextOf` pins a *spelling* where the guard wants an *identity*. `checker.typeToString` on the union's constituent types is already canonical — one quote style, no redundant parens, no source whitespace, numerics normalised — and would delete `sourceTextOf`, `unwrapParens`, the escaping and this whole class of finding in one move. See the verdict comment.

_Architecture lens._

tests-package/export-surface.ts
```diff
@@ -172,39 +343,147 @@ function checkerMembersOfConst(
...
+          continue
+        }
+
+        // With the initializer, not just the name: a `const enum` value is
+        // inlined into a consumer's build, so changing `A = 0` to `A = 5` is a
+        // break of the same class as `2 | 3` becoming `'2' | '3'` — which the
+        // union handling goes out of its way to make visible.
+        collected.members.add(
+          member.initializer
+            ? `${name}=${sourceTextOf(member.initializer)}`
```
fatal10110, line 455: **LOW — enum initializers bypass the quote canonicalisation added six lines up in this same commit, and implicit enum values — the more common spelling, and the one the rationale argues hardest for — are not pinned at all.**

The rationale is right: a `const enum` value is inlined into a consumer's build, so a value change is a break of the same class as `2 | 3` becoming `'2' | '3'`. Two gaps in carrying it out.

**String enum values keep the source's quote characters.** `sourceTextOf`'s canonical branch gates on `ts.isLiteralTypeNode(inner) && ts.isStringLiteral(inner.literal)`, but an enum initializer is a bare `ts.StringLiteral` *expression* — never wrapped in a `LiteralTypeNode` — so it falls through to raw `getText()`:

```ts
export declare const enum Kind { Hash = 'hash', List = 'list' }   // -> ["Hash='hash'","List='list'"]
export declare const enum Kind { Hash = "hash", List = "list" }   // -> ["Hash=\"hash\"","List=\"list\""]

findRemovals -> ["Kind.Hash='hash' — member gone", "Kind.List='list' — member gone"]

// the union path, same two spellings, correctly agrees:
export type Kind2 = 'hash' | 'list'  /  "hash" | "list"   -> findRemovals []
```

That is the `.prettierrc` coupling this commit introduced `sourceTextOf` to eliminate, still live for the construct the same commit just started recording — the two sitting side by side in one output.

**Implicit values are invisible.** They are the common spelling, and shifting them is the classic break:

```ts
export declare const enum L { Low, High }        // -> members ["High","Low"]
export declare const enum L { Mid, Low, High }   // -> members ["High","Low","Mid"]
// Low silently moved 0 -> 1, High 1 -> 2, both inlined into consumers' builds
findRemovals  -> []
findAdditions -> ["L.Mid"]     <-- reads as purely additive
```

`checker.getConstantValue(member)` gives the resolved value for both spellings and sidesteps the quoting question entirely, since it returns a `string | number` rather than source text. (It also handles `A = 1 << 2`, which currently records the expression text.)

No enum is exported today — `grep -c 'enum ' dist/*.d.ts` is 0 — which is the reason this is LOW and also the reason worth asking whether the branch should be here at all yet. See the verdict comment.

_Architecture lens._

fatal10110:
## Re-review of `bc41487` — not yet: one HIGH, then ship

Re-reviewed the delta `4fe72f6..bc41487` against a scratchpad clone at the PR head, built fresh. `test:package` is **52/52 green**, and the baseline round-trips to zero removals and zero additions on both entry points and both `.d.ts` and `.d.mts`.

No wire-compat lens on this review, same as last round: no `src/`, no protocol surface, nothing for a real `redis-server` to ground-truth.

### Prior findings

| # | Prior finding | Status |
|---|---|---|
| 1 | **HIGH** `isMerged` wrong in both directions | **Fixed**, and well covered — see red runs below |
| 2 | MED `/node_modules/` substring match | **Partly fixed** — the hole is still open. New HIGH |
| 3 | MED union arm contributing nothing is dropped | **Fixed**; the fallback introduces a new MEDIUM |
| 4 | MED raw `getText()` leaks quotes / parens / `.prettierrc` | **Fixed for strings** (verified end-to-end); numerics regressed. New MEDIUM |
| 5 | MED `Partial<Record<…>>` rationale does not reproduce | **Fixed** — header now cites the annotation-dropped case |
| 6 | MED `aliasSymbol` compared by name | **Fixed** — both directions genuinely covered |
| 7 | LOW enum values / `ModuleDeclaration` | **Addressed**, with new gaps in both branches |
| 8 | LOW builtin mechanism + error text | **Fixed** — header is now exact, error names the second cause |
| 9 | LOW stale scratch rationale + unguarded `after()` | **Fixed** |

### The HIGH, reproduced

Both directions, reconstructing the pre-`bc41487` shape rather than trusting the report:

- **Conditional prefix** (type-side prefixed only when a value half exists): **41/52**, and `adding a value half is purely additive, not a removal` fails with exactly `'Foo.a — member gone'`.
- **Interface members flattened into the value namespace**: **44/52**, failing `namespaces the interface spelling of a merge identically`.

Both match the reported numbers. The masking case the old predicate missed is now genuinely *caught*, not merely recorded differently:

```
before: ["name","other","type:name"]      after: ["other","type:name"]
findRemovals -> ["Coll.name — member gone"]
```

The unconditional rule is the right call, and the stability argument for it holds.

### Baseline churn — verified independently

The claim is what makes a 709-line baseline diff safe to wave through, so I checked it rather than reading it:

- **91 + 266 symbols**, identical symbol *sets*, not just counts.
- **Zero** `kind` changes.
- With `type:` / `static:` normalised away, member sets are identical for **every** entry — **0 substantive changes**.
- Member **counts** identical per entry (576 both sides), so nothing collapsed into a name that already existed and nothing split.
- **Variants byte-identical, unnormalised** (107 both sides) — so the quote canonicalisation, the paren unwrap and the empty-arm fallback each produced *no* baseline movement.

287 of 576 member strings gained a prefix. That is the entire diff. The claim holds.

### The two self-corrections

- **Const gate.** Confirmed in both directions. My first attempt reproduced your *failed* neutering exactly — comparing resolved symbols by name leaves the suite at 52/52, because the set carries the original name. The faithful revert (compare against **export** names, as the pre-#397 code did) reds exactly the two new tests, at `the const gate fires through a renamed re-export` and `the const gate does not fire on a mere name collision`. Your correction was the right diagnosis.
- **Empty-arm fallback.** No residue in the committed baseline: `CreateIoredisMockOptions` records `["type:cluster","type:databaseCount","type:seed"]` and **no** variants, so both arms contributed. Independently, patching the buggy set-growth form back in turns the real-baseline tests red — that baseline could not have been generated with it.

Spot-checks: each of the eight production changes has a test that genuinely goes red when it is neutered (quote canonicalisation, paren unwrap, empty-arm fallback in both its forms, `/node_modules/`, the const gate, the namespace branch, enum initializers). No vacuous tests found in the new file.

The `.prettierrc` decoupling I checked end-to-end rather than by fixture: flipping two `CommandFlag` arms in `src/` to double quotes and rebuilding really does emit mixed quoting into `dist/redis-error-*.d.ts:814`, and the suite stays 52/52. Neutering just the string branch on that same build prints the two breaking removals. That one is solidly fixed.

### New findings

| Sev | Site | Problem |
|---|---|---|
| **HIGH** | `export-surface.ts:377` | `isSourceFileFromExternalLibrary` still flags the **shared chunk** when `dist/` sits under `node_modules/`. `core` drops 454→143 members, 72→3 variants; no invariant throws |
| **HIGH** | `export-surface-extractor.test.ts:135` | The test named after that bug uses a single-file fixture; a program *root* is never flagged, so it cannot go red for its stated reason |
| MED | `export-surface.ts:226` | Empty-arm fallback makes an arm's representation depend on whether it currently has a member — an additive edit reports BREAKING |
| MED | `export-surface.ts:468` | Namespace members are added **unprefixed**, re-opening the masking hole the headline change closes |
| MED | `export-surface.ts:289` | Numerics regressed from canonical `.text` to raw spelling (`1e3` ≠ `1000`), with no prettier backstop; whitespace collapse over- and under-normalises |
| LOW | `export-surface.ts:455` | Enum string initializers bypass the canonicalisation added six lines up; implicit enum values not pinned at all |

Only the HIGH is live on today's build. The rest need a construct `/core` does not currently export — which is itself the point below.

### On scope — narrow it, and the cuts are deletions

You asked the right question, so here is a direct answer rather than "harden again".

**Judged against its actual job, the guard works.** I simulated a Phase 4 removal — deleted `SerialTurnQueue` from the `/core` entry and rebuilt. Exactly **one** of 52 tests failed:

```
BREAKING: 1 export(s) disappeared from "js-redis-server/core".
  - SerialTurnQueue (value) — gone
```

One clean, precise failure, no noise from the other 51. That is the job, and it is done.

**But look at where the bugs live.** Of the four `/core` breaks this guard was written for, #375 and #377 removed top-level symbols — level 1. #374 and #376 were member-level — level 2. **None of the four was variant-level.** Level 3 is speculative coverage for a break shape this package has never had.

Now the ledger. Across three rounds: the symbol list, the `kind` downgrade check, the two structural invariants and both diff functions have produced **zero** bugs. Every finding in every round — the union gap, `isMerged`, the const gate, the `node_modules` predicate twice, `''noscript''`, the `.prettierrc` coupling, the empty-arm fallback twice, the namespace masking, the enum quoting, the numeric regression — lives in the hand-rolled AST walk underneath levels 2 and 3. That is not a run of bad luck; it is a subsystem re-implementing, by hand, things the checker already answers.

So the guard is not too *large*, it is aimed slightly wrong: it pins **spellings** where it wants **identities**, and it pins shapes this package does not export.

**Cut now, in this PR — three branches that cannot fire against the real `dist/`** (0 enums, 0 namespaces, 0 `{`/`&` variants in the baseline, all verified):

1. the `ts.isEnumDeclaration` branch,
2. the `ts.isModuleDeclaration` branch and `namespaceMemberNames`,
3. the empty-arm text fallback.

That is roughly 70 lines and 5 tests, it deletes three of the six comments above outright, and it removes a branch that today *weakens* the guard (namespace masking) and one that today only generates false BREAKINGs. Put all three on the "what this does not cover" list — which is the honest state, and cheaper than carrying half-right implementations until the day something needs them.

**Then, as a follow-up, the structural narrowing:** replace `sourceTextOf` / `unwrapParens` / `isMemberBearing` with the checker — `type.isUnion() ? type.types.map(t => checker.typeToString(t)) : []`. `typeToString` is canonical by construction: one quote style, no redundant parens, no source whitespace, numerics normalised. It deletes the escaping, the quoting question, the paren question, the whitespace collapse and the numeric regression as a class, rather than one bug per round.

**What I would not cut:** the symbol list and `kind` (that is the job), `members`, `checkerMembersOfConst` (it is what pins `RedisValue`'s factory half and `FEATURE_GATES`), the two structural invariants, and the 15 diff tests — the diff is the one part of this tool that has never been wrong.

Keep `variants` too, but as literal *values* with a kind tag (`str:2` vs `num:2`) rather than source text. That preserves the `CommandFlag`-drops-an-arm catch and the `2 | 3` → `'2' | '3'` collision fix, and has no degrees of freedom to get wrong.

### Verdict

**Not yet — ship after the HIGH.** The fix is small: one predicate (`dirname(entryDeclarationFile)` instead of `isSourceFileFromExternalLibrary`) and one fixture split into two files so the test can fail for its stated reason. Everything else on this PR verified: both HIGH directions, all seven secondary fixes, the churn claim, and both self-corrections. The three suggested deletions would be a good thing to fold in while you are here, since they are removals and they close three of the six findings without new code.

fatal10110:
Closing unmerged: the export-surface guard is scaffolding, not product. It has consumed more review rounds than the code it guards. Not worth more time.

PR #407, opened 2026-09-21
fix: allow cluster SORT BY patterns real Redis accepts and gate the rules by profile (#387)

### LINKED ISSUE #387: ClusterPolicy rejects SORT BY nosort and non-glob BY patterns that real Redis allows; SORT cluster rules not version-gated
Found while re-reviewing PR #378, ground-truthed against real Redis 7.2.1 **and** a `redis:8.0` container (8.0.6). **Pre-existing on `main`.**

## Problem
`ClusterPolicy` (`src/core/execution-policies/cluster-policy.ts`, `getSortClusterPatternError`) rejects `SORT` invocations that real Redis accepts:

- `SORT key BY nosort` — the documented way to skip sorting. `nosort` contains no `*`, so it is treated as a cross-slot-risky pattern and denied.
- Any non-glob `BY` pattern generally — real Redis only denies patterns that could resolve to keys in another slot.

Both Redis 7.2.1 and 8.0.6 allow these.

## Why it matters now
PR #378 removes the node-redis facade's own client-side CROSSSLOT construction specifically so `ClusterPolicy` owns these rules. That is the right design — it makes this policy the single source of truth, which raises the cost of it being wrong.

## Related: version gating
The SORT cluster rules are **not** version-gated, but they differ across versions. Confirmed on 8.0.6: the long error wording (`... when keys formed by the pattern may be in different slots.`), `GET '#'` allowed, plus a hash-tag relaxation. 7.2.1 uses the short wording (`ERR BY option of SORT denied in Cluster mode.`). The repo's compatibility-profile system (`src/core/compatibility/`) exists for exactly this and should gate them.

## Acceptance
- `SORT key BY nosort` succeeds in cluster mode, matching real Redis.
- Non-glob `BY` patterns are allowed unless they genuinely may cross slots.
- The error wording and `GET '#'` handling are gated by profile, verified against both 7.2.1 and 8.0.x.

fatal10110:
## Summary

`ClusterPolicy.getSortClusterPatternError` refused `SORT` invocations real Redis
accepts, and applied Redis 8.0's rules to every compatibility profile.

Root cause: a `BY` pattern with no `*` was run through a cross-slot check, so the
documented `SORT key BY nosort` (and any other constant pattern) was denied; glob
patterns were compared by **hash-tag bytes**, so a pattern tagged with an untagged
source key (`SORT k BY {k}w_*`) was denied even though both hash to the same slot.

The check now mirrors `sortCommand()`:

- A `BY` pattern without `*` is constant — real Redis sets `dontsort`, never looks
  a weight key up, and never reaches the cluster guard. `runSort` applies the same
  rule (previously only the literal `nosort` short-circuited, so a constant weight
  key was still read and could raise a spurious `WRONGTYPE`), and `sortRoutingKeys`
  no longer routes on a constant `BY`, matching `sortGetKeys()`.
- Glob patterns go through a port of `patternHashSlot()` (`cluster.c`) compared
  against the sort key's slot.

## Version gating

Ground-truthed against cluster-enabled `redis:7.2.1`, `redis:7.4.0` and `redis:8.0.6`
containers, cross-checked against `src/sort.c` at `redis` 7.2.4/7.4.0/8.0.0 and
`valkey` 8.0.0/9.0.0. Two new `FeatureId`s:

| Gate | Redis | Valkey | Behavior below the gate |
| --- | --- | --- | --- |
| `sort.cluster-pattern-slot` | 7.4.0 | 8.0.0 | every `BY` glob and *every* `GET` pattern refused, with the short `ERR BY/GET option of SORT denied in Cluster mode.` wording |
| `sort.cluster-get-hash` | 8.0.0 | 9.0.0 | `GET '#'` is hashed like any other pattern, so it is refused |

Observed on the real servers:

```
7.2.1  SORT l BY nosort      -> 3 1 2          SORT l BY 'w_*'   -> ERR BY option of SORT denied in Cluster mode.
7.2.1  SORT l GET '#'        -> ERR GET option of SORT denied in Cluster mode.
7.4.0  SORT l BY '{l}w_*'    -> ok             SORT l GET '#'    -> ERR GET option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.
8.0.6  SORT l BY '{l}w_*'    -> ok             SORT l GET '#'    -> ok
8.0.6  SORT l BY w_foo       -> 3 1 2 (unsorted, cross-slot constant accepted)
```

Closes #387

## Test plan

- [x] `tests-integration/ioredis/sort-integration.test.ts` + node-redis twin: three
      new cases (`BY nosort`, constant `BY` in another slot, `BY` glob hash-tagged to
      an untagged source key) — red on mock before the fix (`ERR BY option of SORT
      denied ...`), green against real Redis before the fix
- [x] New `tests-integration/compatibility/sort-cluster-gates.test.ts` (mock-only)
      asserts both gates' wording and accept/refuse split on all 7 profiles
- [x] `npm test` — 342 pass / 0 fail
- [x] `npm run test:integration:mock` — 1125 pass / 0 fail
- [x] `npm run test:integration:compatibility:mock` — 13 pass / 0 fail on each of the
      7 profiles
- [x] `npm run test:package`, `npm run test:demo`, `npm run build`, `npm run lint`
- [x] Real backend (targeted, shared cluster — `clean:redis` deliberately skipped):
      sort + cluster + compatibility files, then `key/**`, `multi`, `scripts` —
      73 pass / 0 fail and 92 pass / 0 fail


src/core/compatibility/feature-gates.ts
```diff
@@ -25,4 +25,11 @@ export const FEATURE_GATES: Record<FeatureId, VersionGate> = {
   'hscan.novalues': { redis: '7.4.0', valkey: '9.0.0' },
   'xread.plus-id': { redis: '7.4.0' },
   'cluster.multi-db': { valkey: '9.0.0' },
+  // SORT BY/GET in cluster mode: compare each pattern's slot against the sort
+  // key's slot (`patternHashSlot()`) instead of refusing every pattern, and use
+  // the longer "...may be in different slots." wording — Redis 7.4 / Valkey 8.0.
+  'sort.cluster-pattern-slot': { redis: '7.4.0', valkey: '8.0.0' },
+  // SORT GET '#' (the element itself) is exempt from that slot comparison —
+  // Redis 8.0 / Valkey 9.0. Before that it is hashed like any other pattern.
+  'sort.cluster-get-hash': { redis: '8.0.0', valkey: '9.0.0' },
```
fatal10110: **HIGH (blocking) — wrong version boundary.**

The `GET '#'` exemption from the cluster slot comparison landed in redis **7.4.2** and valkey **8.0.2**, not 8.0.0 / 9.0.0. Because `PRESETS['redis-7.4']` resolves to 7.4.4 (`src/core/compatibility/profile.ts:57`), the shipped `redis-7.4` profile now contradicts the exact version it claims to model.

I bisected this on single-node clusters with `SORT l GET '#'`:

- redis 7.4.0 and 7.4.1 — refuse
- redis 7.4.2, 7.4.3, 7.4.4, 7.4.5, 7.4.11 — accept
- valkey 8.0.0 and 8.0.1 — refuse
- valkey 8.0.2 and later — accept

Concrete failure: on the `redis-7.4` profile, `SORT {t}:l GET '#'` returns `ERR GET option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.`, where real 7.4.4 returns the list elements.

Fix: `'sort.cluster-get-hash': { redis: '7.4.2', valkey: '8.0.2' }`. That still leaves `valkey-8.0` (which pins 8.0.0) denying, which is the correct behaviour for that version.

The same wrong boundary is written into three other places that need to move with it: `getHashProfiles` in `tests-integration/compatibility/sort-cluster-gates.test.ts:37`, the gate table row in `docs/API.md`, and the note in `docs/COMMANDS.md`. The new test currently asserts the bug rather than catching it.

A likely cause worth noting: this looks like it was ground-truthed on 7.4.**0** and then generalised to the profile name, but the `redis-7.4` preset is pinned at 7.4.**4**.

src/commands/keys.ts
```diff
@@ -700,7 +700,7 @@ function sortElements(
   args: SortArgs,
   db: RedisDatabase,
 ): Buffer[] {
-  if (args.by && isNoSortPattern(args.by)) {
+  if (args.by && isConstantSortPattern(args.by)) {
```
fatal10110, line 708: **MEDIUM-HIGH (blocking) — numeric sort has no tie-break, so equal scores keep insertion order.**

Real Redis' `sortCompare()` breaks equal numeric scores by falling through to `compareStringObjects()` on the elements themselves. The comparator below is `scored.sort((a, b) => a.score - b.score)` over a stable JS sort, so ties retain insertion order instead of being ordered by element.

This is not a corner case: missing weight keys give every element score 0, which is the common shape. Against real 8.0.6:

```
RPUSH {t}:l 3 1 2
SORT {t}:l BY '{t}:nw_*'   -> 1 2 3
```

(and the same when all the weight keys exist with equal values). This mock returns `["3","1","2"]`.

The bug is pre-existing in `sortElements`, but this PR is what makes it reachable in cluster mode for an untagged source key — it is exactly the `BY '{key}weight:*'` shape the new test "SORT BY a glob hash-tagged to an untagged source key" exercises. That test only passes because it happens to set both weight keys to distinct values.

Fix: fall back to `Buffer.compare(a.element, b.element)` when the scores are equal. Note that `ALPHA` combined with `BY` on missing weights is already correct — real Redis genuinely yields `cmp = 0` there and the current behaviour matches.

src/commands/keys.ts
```diff
@@ -700,7 +700,7 @@ function sortElements(
...
     return [...elements]
```
fatal10110, line 709: **MEDIUM — a SET source with a constant `BY` must be force-sorted when `STORE` or a script is involved.**

This early return honours `dontsort` unconditionally, but real Redis overrides it for unordered sources whose output has to be deterministic:

```c
if (dontsort && sortval->type == OBJ_SET && (storekey || CLIENT_SCRIPT)) {
    dontsort = 0;
    alpha = 1;
    sortby = NULL;
}
```

Against real 8.0.6:

```
SADD s c a b
SORT s BY nosort STORE d   -> a b c
EVAL "return redis.call('SORT','s','BY','nosort')" 0  -> a b c
```

This mock returns `c a b` in both cases. Lists are correctly *not* force-sorted, so the condition really is set-specific.

This is pre-existing for the literal `nosort`, but this PR widens the trigger from one literal to *every* constant `BY` pattern, so it becomes considerably easier to hit.

src/commands/keys.ts
```diff
@@ -818,13 +818,22 @@ function isSelfSortPattern(pattern: Buffer): boolean {
   return pattern.length === 1 && pattern[0] === 0x23
 }
 
-function isNoSortPattern(pattern: Buffer): boolean {
-  return pattern.toString().toLowerCase() === 'nosort'
+/**
+ * A BY pattern with no `*` is constant: every element resolves to the same
+ * weight key, so real Redis sets `dontsort` and skips both the lookup and the
```
fatal10110: **MEDIUM — the "no `*` means no lookup" rule this comment states is not applied to `GET`.**

The doc comment here describes the rule correctly, but `expandSortPattern`'s `index === -1` branch only implements it for `BY`. Real Redis applies it in `lookupKeyByPattern()` for both:

```c
p = strchr(spat,'*');
if (!p) return NULL;
```

so a `GET` pattern without `*` yields NULL for every element. Against real 8.0.6:

```
SORT l BY nosort GET r407:const   -> (nil) (nil) (nil)
```

This mock returns `["HELLO","HELLO","HELLO"]` (the constant key's value, repeated).

This matters more after the change: under the new guard, `patternHashSlot('{t}:const') === keySlot`, so a constant `GET` is now explicitly *accepted* in cluster mode where it was previously refused outright — the divergence is newly reachable rather than merely latent.

Fix: make `expandSortPattern` return null/NULL for a `GET` pattern with no `*`, not just for `BY`.

src/commands/keys.ts
```diff
@@ -818,13 +818,22 @@ function isSelfSortPattern(pattern: Buffer): boolean {
...
+ * sort. `BY nosort` is just the documented spelling of such a pattern — there
+ * is nothing special about the literal.
+ */
+function isConstantSortPattern(pattern: Buffer): boolean {
```
fatal10110: **MEDIUM — `readSortPattern` raises WRONGTYPE where real Redis silently returns NULL.**

Real `lookupKeyByPattern()` checks the resolved object's type and returns NULL for anything that is not a string; it never propagates an error. `readSortPattern` throws WRONGTYPE instead, so a single non-string key in the weight/get namespace turns the whole command into an error.

Against real 8.0.6 with `w_1` created as a list:

```
SORT l BY 'r407:w_*'    -> 1 3 2     (element 1 gets weight 0)
SORT l GET 'r407:w_*'   -> (nil) 5 1
```

This mock raises `WRONGTYPE Operation against a key holding the wrong kind of value` for both.

This PR fixes only the constant-`BY` variant of the problem (by never performing the lookup at all); glob `BY` and every `GET` path remain divergent. Fix: treat a non-string lookup result as NULL — score 0 for `BY`, a nil reply element for `GET`.

src/commands/keys.ts
```diff
@@ -818,13 +818,22 @@ function isSelfSortPattern(pattern: Buffer): boolean {
...
+  return !pattern.includes(0x2a)
```
fatal10110: **LOW — `*`-detection should stop at the first NUL, to match `strchr`.**

Real Redis locates the wildcard with `strchr(spat, '*')`, which stops at a NUL byte. This implementation is binary-safe and scans the whole buffer, so the two disagree on patterns containing an embedded NUL. Against real 8.0.6:

```
SORT l BY "a\x00b*"   -> 3 1 2   (treated as constant: no '*' before the NUL, dontsort)
SORT l BY "ab*"        -> 1 2 3   (treated as a glob)
```

This mock classifies `"a\x00b*"` as a glob, and in cluster mode denies it. Fix: search for `0x2a` only up to the first `0x00`. The same applies to the inline `args.by.includes(0x2a)` check in `src/core/execution-policies/cluster-policy.ts:156`.

For what it's worth, `patternHashSlot` itself is correctly length-based on both sides — real Redis passes an explicit length there, so no change is needed in the port.

src/commands/keys.ts
```diff
@@ -818,13 +818,22 @@ function isSelfSortPattern(pattern: Buffer): boolean {
...
 }
 
 function sortRoutingKeys(args: SortArgs): Buffer[] {
   const keys = [args.key]
-  if (args.by && !isNoSortPattern(args.by)) {
+  // A constant BY is never looked up, so it is not a key this command touches
+  // and must not take part in slot routing (real Redis routes SORT on its
+  // source key and STORE destination alone).
+  if (args.by && !isConstantSortPattern(args.by)) {
```
fatal10110: **MEDIUM — `sortRoutingKeys` should not report `BY`/`GET` patterns as keys at all.**

Real `sortGetKeys()` skips `BY` and `GET` entirely; the only keys `SORT` reports are the source key and, when present, the `STORE` destination. Against real 8.0.6:

```
COMMAND GETKEYS SORT l BY 'w_*' GET 'p_*'             -> l
COMMAND GETKEYS SORT l BY 'w_*' GET 'p_*' STORE dst   -> l dst
```

This mock reports `['l', 'w_*', 'p_*']`. After this PR the constant form is finally correct, but the glob form is still wrong, which is an odd halfway state.

The good news is that the fix is now trivial: the new `patternHashSlot()` guard in `cluster-policy.ts` performs the cross-slot check independently, so the routing hack this list was serving is no longer needed. `sortRoutingKeys` can be reduced to `args.key` plus `args.store`.

src/core/execution-policies/cluster-policy.ts
```diff
@@ -71,7 +72,11 @@ export function createClusterPolicy(
         transactionSlots.delete(ctx.session)
       }
 
-      const sortPatternError = getSortClusterPatternError(plan, topology)
+      const sortPatternError = getSortClusterPatternError(
```
fatal10110: **MEDIUM — the cluster pattern guard runs before redirection, so a stale-slot-map client gets a hard error instead of a `MOVED`.**

Real Redis performs slot ownership and cross-slot checks in `processCommand()`, which runs *before* `sortCommand()` ever parses `BY`/`GET`. On a real 2-node cluster:

```
(ask the wrong node)  SORT k BY 'w_*'                  -> MOVED 14276 <host>:<port>
SORT l BY 'w_*' STORE otherdst                          -> CROSSSLOT Keys in request don't hash to the same slot
```

This mock returns the BY-denied error in both cases. That is the worse failure mode of the two: a client with a stale slot map receives a terminal error instead of a redirect it could follow and retry, so it never converges.

The ordering is pre-existing, but it lives in exactly the function this PR rewrites, so it is worth correcting here: run the MOVED/ASK and CROSSSLOT determination before `getSortClusterPatternError`. One constraint to preserve — the pattern guard must still precede the WRONGTYPE check; I verified that ordering against a real server.

src/core/execution-policies/cluster-policy.ts
```diff
@@ -121,57 +150,78 @@ function getSortClusterPatternError(
...
-    Buffer.isBuffer(args.by) &&
-    sortPatternMayCrossSlot(args.key, args.by, topology)
-  ) {
-    return new RedisCommandError(
-      'BY option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.',
-    )
+  const comparesPatternSlots = profile.has('sort.cluster-pattern-slot')
+  const keySlot = topology.calculateSlot(args.key)
+
+  if (Buffer.isBuffer(args.by) && args.by.includes(0x2a)) {
```
fatal10110: **LOW — `BY` is always reported before `GET`, but real Redis reports in argument order.**

This checks `args.by` first and only then loops over `args.get`, so whichever option is syntactically first in the command line is irrelevant. Real Redis raises the error as it parses, i.e. in argument order. Against real 8.0.6:

```
SORT l GET 'n_*' BY 'w_*'   -> GET option of SORT denied in Cluster mode...
```

This mock returns the BY error. This holds for both the legacy and the long wording, so it is not fixed by the profile gate. Fix: iterate the parsed options in the order they appeared rather than checking `by` unconditionally first.

src/core/execution-policies/cluster-policy.ts
```diff
@@ -104,9 +109,33 @@ export function createClusterPolicy(
...
+ * A BY pattern with no `*` is constant on every version: real Redis sets
+ * `dontsort`, never looks a weight key up, and so never reaches the guard —
+ * which is why the documented `BY nosort` works in cluster mode.
+ */
 function getSortClusterPatternError(
   plan: CommandPlan,
   topology: RedisClusterTopology,
+  profile: CompatibilityProfile,
 ): RedisCommandError | null {
   if (plan.definition.name !== 'sort' && plan.definition.name !== 'sort_ro') {
```
fatal10110, line 148: **LOW — a trailing syntax error preempts the cluster denial.**

This guard operates on an already-fully-parsed `plan`, whereas real Redis aborts at the offending token as it parses left to right. So a syntax error *after* `BY` never gets the chance to mask the cluster denial in real Redis, but does here:

```
SORT l BY 'w_*' BADARG   real: BY option of SORT denied in Cluster mode...
                         mock: ERR syntax error
```

Real Redis does return `ERR syntax error` when the junk token *precedes* `BY`, so this is specifically about ordering, not about which error wins in general. Low impact — it only shows up on malformed commands — but it falls out of the same "guard after full parse" structure as the redirection-ordering comment above.

src/core/execution-policies/cluster-policy.ts
```diff
@@ -121,57 +150,78 @@ function getSortClusterPatternError(
...
-  source: Buffer,
+/**
+ * Port of `patternHashSlot()` from Redis `cluster.c`: the slot every key a
+ * glob pattern can match must belong to, or `-1` when that cannot be inferred.
+ *
+ * A wildcard or an escape before the closing brace makes the match set
+ * unbounded; a non-empty `{...}` tag pins the slot to the tag; anything else
+ * is a literal key and hashes as one.
+ */
+function patternHashSlot(
```
fatal10110, line 196: **LOW — code quality around the new guard.**

- **`patternHashSlot` has no unit test.** This is a freshly hand-ported five-branch algorithm and the `?`, `[`, `\\`, `{}`-sentinel and unterminated-`{` branches are entirely uncovered — `tests/core-cluster-policy.test.ts` does not mention `sort` at all, and the new integration test only exercises the plain-tag and cross-slot cases. These are cheap, pure-function table tests; please add them, since a silent regression here re-opens exactly the cluster-safety hole the guard exists to close.
- **Unchecked structural cast.** `plan.args as { key?: unknown; by?: unknown; get?: unknown }` (lines 144-148) is an unchecked structural cast over the non-exported `SortArgs`. Renaming `SortArgs.by` would compile cleanly and silently disable the entire guard. Export `SortArgs` and cast to it, so the compiler links the two.
- **Duplication.** `isSelfSortPattern` and the `*`-presence check are each implemented twice, once here and once in `src/commands/keys.ts`. They must stay in lockstep for the guard to mean anything — export the `keys.ts` versions and import them.

Minor nits elsewhere in the diff:

- `tests-integration/compatibility/sort-cluster-gates.test.ts`: the `ProfileName` union plus `process.env.REDIS_COMPAT ?? 'redis-8.0'` is copy-pasted from `profile-gates.test.ts`; hoist it into a shared test helper.
- The `if (!Buffer.isBuffer(pattern)) continue` at line 167 is unreachable given how `args.get` is built.
- The `!comparesPatternSlots` legacy short-circuit at line 170 is loop-invariant but sits inside the `get` loop; hoisting it above the loop reads better and makes the "legacy denies every GET" rule obvious.
- `assert.ok(redisClient)` in the new node-redis test is redundant — the value is non-nullable there.

fatal10110 (commented):
**Verdict: don't ship as-is.**

The core of this change is correct and a real improvement. The `patternHashSlot()` port is a faithful reproduction of the `cluster.c` original, and the constant-`BY` `dontsort` rule is exactly right: a pattern with no `*` is never looked up, so its slot is irrelevant and it must not participate in routing. I verified the new behaviour with a 25-pattern differential sweep against redis 8.0.6 — hash tags, nested and empty braces, unterminated braces, `?`/`[`/`\\` wildcards, untagged source keys — with **zero divergences**. `SORT_RO` parity and the standalone-mode `dontsort` behaviour were both confirmed against real servers as well.

The `sort.cluster-pattern-slot` gate (`redis 7.4.0` / `valkey 8.0.0`) is also **correct**, and is a genuine correction to issue #387, which claimed the split was 7.2 vs 8.0. I confirmed that redis 6.2.14, 7.0.15, 7.2.4 and 7.2.16 all emit the short `denied in Cluster mode.` wording and refuse even a same-tag `BY '{l}w_*'`.

Two things block shipping:

1. **The `sort.cluster-get-hash` version boundary is wrong** (`src/core/compatibility/feature-gates.ts`). The `GET '#'` exemption landed in redis 7.4.2 / valkey 8.0.2, not 8.0.0 / 9.0.0, so the shipped `redis-7.4` profile now contradicts the exact version it models. Details inline.
2. **The numeric sort comparator has no tie-break** (`src/commands/keys.ts`). Real Redis breaks equal scores by comparing the elements themselves; this PR is what first makes that path reachable in cluster mode for an untagged source key. Details inline.

The remaining nine comments are mostly pre-existing divergences that this PR either widens the trigger for or rewrites the surrounding code of — worth fixing here or tracking as follow-ups, but they are not what makes the change unshippable.

src/core/compatibility/feature-gates.ts
fatal10110: Confirmed and fixed in 03f5c9d. I re-bisected independently on my own single-node cluster-enabled containers (`SORT a387:l GET '#'`, all slots assigned, high ports so the shared cluster was untouched):

```
redis 7.4.0   ERR GET option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.
redis 7.4.1   ERR GET option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.
redis 7.4.2   1 2 3
redis 7.4.4   1 2 3
redis 8.0.6   1 2 3
valkey 8.0.0  ERR GET option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.
valkey 8.0.1  ERR GET option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.
valkey 8.0.2  1 2 3
valkey 9.0.0  1 2 3
```

You were right about the cause too — I ground-truthed 7.4.0 (whose `sort.c` has the long wording and the slot comparison but no `#` exemption) and then wrote the gate against the profile name rather than the preset version.

The gate is now `'sort.cluster-get-hash': { redis: '7.4.2', valkey: '8.0.2' }`, with the bisection recorded in a comment. `getHashProfiles` in the gate test is now `['redis-7.4', 'redis-8.0', 'valkey-9.0']` — `valkey-8.0` pins 8.0.0 and correctly still refuses — and the `docs/API.md` row and `docs/COMMANDS.md` note both spell out the patch versions and why the two profiles land on opposite sides.

src/commands/keys.ts
fatal10110: Confirmed and fixed in 03f5c9d. Reproduced on my own 8.0.6 container:

```
RPUSH a387:{t}l 3 1 2
SORT a387:{t}l BY 'a387:{t}nw_*'   -> 1 2 3      (all weights missing)
SORT a387:{t}l BY 'a387:{t}ew_*'   -> 1 2 3      (all weights = 5)
```

The comparator is now `a.score - b.score || Buffer.compare(a.element, b.element)`.

Your note about `ALPHA` is right as well, and I checked it before touching anything so I would not "fix" it by mistake — `sortCompare()` genuinely yields `cmp = 0` when both `cmpobj`s are NULL:

```
RPUSH a387:{t}al c a b
SORT a387:{t}al BY 'a387:{t}nw_*' ALPHA   -> c a b
```

Both are now asserted, on ioredis and node-redis, and both pass against real Redis as well as mock.

src/commands/keys.ts
fatal10110: Confirmed and fixed in 03f5c9d. Reproduced on 8.0.6, including the list control you mentioned:

```
SADD a387:{t}s c a b
SORT a387:{t}s BY nosort                       -> c a b
SORT a387:{t}s BY nosort STORE a387:{t}d       -> a b c
EVAL "return redis.call('SORT',KEYS[1],'BY','nosort')" 1 a387:{t}s  -> a b c
RPUSH a387:{t}ll c a b
SORT a387:{t}ll BY nosort STORE a387:{t}ld     -> c a b   (lists untouched)
```

`runSort` now takes the full `RedisExecutionContext` and runs the arguments through `forceDeterministicSetOrder()`, which drops the `BY` and forces `alpha` when the source is a set and either `STORE` is present or the command runs inside a script. `DESC` is deliberately left alone — the C override only clears `dontsort`/`sortby` and sets `alpha`.

The script half needed a signal that did not exist, so `RedisExecutionContext` gains `readonly inScript?: boolean`, set by `createLuaMonitorContext()` in `src/core/lua-runtime.ts` — the mirror of `CLIENT_SCRIPT`. Covered on both clients against real Redis and mock.

src/commands/keys.ts
fatal10110: Confirmed and fixed in 03f5c9d. On 8.0.6:

```
SET a387:{t}const HELLO
RPUSH a387:{t}l 3 1 2
SORT a387:{t}l BY nosort GET a387:{t}const   -> (nil) (nil) (nil)
```

`readSortPattern()` now resolves the wildcard position first and returns `null` when there is none, so the rule lives in the one place both `BY` and `GET` go through, exactly like `lookupKeyByPattern()`'s `if (!p) return NULL;`. `expandSortPattern` takes the already-resolved index and no longer has a no-wildcard branch at all.

Your point about this being newly reachable is what made it worth doing here rather than deferring: a same-slot constant `GET` now passes the guard, so without this the accepted-but-wrong case would have been introduced by this PR.

src/commands/keys.ts
fatal10110: Confirmed and fixed in 03f5c9d. Reproduced your exact shape on 8.0.6 — `w:1` created as a list, `w:2`=5, `w:3`=1, source `3 1 2`:

```
SORT a387:{t}l BY 'a387:{t}w:*'    -> 1 3 2
SORT a387:{t}l GET 'a387:{t}w:*'   -> (nil) 5 1
```

`readSortPattern()` now returns `null` for anything whose type is not `string` instead of throwing, so a non-string weight scores 0 for `BY` and yields a nil element for `GET`. `readSortSource()` still raises WRONGTYPE for the *source* key, which is correct — that check is `lookupKeyByPattern`'s caller, not `lookupKeyByPattern` itself.

Both shapes are asserted on ioredis and node-redis and pass against real Redis.

src/commands/keys.ts
fatal10110: Confirmed and fixed in 03f5c9d. Verified on 8.0.6 via `EVAL` so the NUL survived the wire:

```
EVAL "return redis.call('SORT',KEYS[1],'BY','a\0b*')" 1 a387:{t}l   -> 3 1 2
```

i.e. treated as constant (`dontsort`), not as a glob — so it is accepted in cluster mode, which is what we now do too.

Both call sites are fixed at once: the `*` search moved into `sortPatternWildcardIndex()` in the new `src/core/sort-patterns.ts`, which stops at the first `0x00`, and both `src/commands/keys.ts` and `cluster-policy.ts` import `isConstantSortPattern` from there (see also my reply on the duplication point). Covered by a unit test in `tests/core-cluster-policy.test.ts`.

And agreed on `patternHashSlot` — it takes an explicit length in C and stays length-based here; I left it alone.

src/commands/keys.ts
fatal10110: Confirmed and fixed in 03f5c9d. On 8.0.6:

```
COMMAND GETKEYS SORT a387:{t}l BY 'a387:{t}w_*' GET 'a387:{t}p_*'                        -> a387:{t}l
COMMAND GETKEYS SORT a387:{t}l BY 'a387:{t}w_*' GET 'a387:{t}p_*' STORE a387:{t}dst      -> a387:{t}l  a387:{t}dst
```

`sortRoutingKeys` is now `args.store ? [args.key, args.store] : [args.key]`, and you are right that the guard makes the old routing hack unnecessary — the pattern check runs on every SORT regardless of which node the command lands on. Asserted with `COMMAND GETKEYS` in the ioredis integration test, green on real and mock.

One knock-on worth flagging: the routing comment in `src/client-mocks/node-redis-mock.ts` explicitly cited `sortRoutingKeys` including the patterns as the reason it does no client-side cross-slot rejection. That reasoning still holds (the policy still owns rules the facade does not share), so I updated the wording rather than the behaviour.

src/core/execution-policies/cluster-policy.ts
fatal10110: Confirmed and fixed in 03f5c9d. Verified both halves against the real 6-node cluster (sending to a node that does not own the slot, so the redirect is not auto-followed):

```
redis-cli -p 30000 SORT 'a387:{t}l' BY 'a387:{t}w_*'                                 -> MOVED 15891 127.0.0.1:30002
redis-cli -p 30005 SORT 'a387:{t}l' BY 'a387:{t}w_*' STORE 'a387:{u}dst'             -> CROSSSLOT Keys in request don't hash to the same slot
```

`validateClusterSlot()` now runs before `getSortClusterPatternError()` in `beforeExecute`, so MOVED/ASK/CROSSSLOT/CLUSTERDOWN all win. The transaction slot-pin still happens after both, so a pattern-denied SORT does not pin a slot.

Your constraint holds automatically here: the guard is an `ExecutionPolicy.beforeExecute` hook, so it always precedes `execute()` and therefore the WRONGTYPE check inside `runSort`. There is a unit test for the ordering in `tests/core-cluster-policy.test.ts` ("reports MOVED before the pattern guard") since the integration harness always connects to the correct slot owner.

src/core/execution-policies/cluster-policy.ts
fatal10110: Confirmed, but deferring — this one wants a separate issue rather than a fix here.

Your repro reproduces on my 8.0.6 container (both options cross-slot, so either error is available):

```
SORT a387:{t}l GET '{other}n_*' BY '{other}w_*'   -> ERR GET option of SORT denied in Cluster mode when ...
```

and we return the BY error.

The reason I am not fixing it in this PR: the guard is an `ExecutionPolicy` hook, so it only ever sees `plan.args`, which is the *result* of `parseSort()` — `{ by, get, ... }` with no record of the order the options appeared in. Reconstructing the order from `plan.rawArgs` means re-tokenising the command inside the policy, and doing that correctly is not a one-liner: option values are opaque, so `SORT k GET BY BY GET` is a valid command whose first GET pattern is the literal `BY`. A naive scan gets that wrong and would trade a cosmetic mismatch for a wrong-pattern classification in the guard — which is the part that has to be right.

The clean fix is at the parser: have `parseSort()` keep the options in encounter order (or record the index each one was parsed at) and have the guard walk that. That is a schema-level change touching `SortArgs`' shape and every consumer, which does not belong in a cluster-guard fix.

Same root cause as the trailing-syntax-error comment below, so one issue covers both. Flagging for the coordinator to file rather than opening it myself.

src/core/execution-policies/cluster-policy.ts
fatal10110: Confirmed, deferring for the same structural reason as the option-order comment above.

Reproduced on 8.0.6, including your control showing it really is about position:

```
SORT a387:{t}l BY '{other}w_*' BADARG   -> ERR BY option of SORT denied in Cluster mode when ...
SORT a387:{t}l BADARG BY '{other}w_*'   -> ERR syntax error
```

We return `ERR syntax error` for both.

This cannot be fixed inside the policy at all: `parseSort()` throws `RedisSyntaxError` during `CommandExecutor.plan()`, so the command never reaches `beforeExecute` and the guard is not merely losing a race — it is never invoked. Making real Redis' left-to-right abort observable means the cluster check has to happen *during* parsing, which is a parser/policy layering change well outside a cluster-guard fix.

It shares a root cause with the option-order comment (the guard runs on a fully-parsed plan), so one issue covers both. Flagging for the coordinator to file rather than opening it myself.

src/core/execution-policies/cluster-policy.ts
fatal10110: All three main points fixed in 03f5c9d, plus most of the nits.

**`patternHashSlot` unit tests** — added a `cluster policy SORT pattern guard` describe block to `tests/core-cluster-policy.test.ts` (12 tests). It drives the real surface (`session.execute('sort', …)`) rather than the private function, and covers every branch you named: `?`, `[`, `\` before the tag can be read, the `{}` empty-tag sentinel, an unterminated `{`, a wildcard *after* a closed tag (accepted — `patternHashSlot` returns at the brace, matching real 8.0.6), the plain-tag accept, the cross-slot refusals, constant `BY`, `GET '#'`, the NUL rule, and MOVED-before-guard.

**Unchecked cast** — `SortArgs` is now exported from `src/commands/keys.ts` and the guard casts to it, so renaming a field is a compile error.

**Duplication** — `isSelfSortPattern` and the `*`-presence check now live in a new `src/core/sort-patterns.ts` and are imported by both sides. I deliberately did *not* export them from `keys.ts` for the policy to import: `src/core` should not depend on `src/commands` at runtime, and a neutral module keeps the layering intact. (The `SortArgs` import is `import type`, so it erases.) `patternHashSlot` stays in the policy since it needs the topology.

Nits:

- unreachable `Buffer.isBuffer(pattern)` guard — removed (typing `plan.args` as `SortArgs` made it provably dead).
- loop-invariant legacy short-circuit — hoisted; the legacy branch is now a single early block that reads "no glob BY, and no GET at all".
- redundant `assert.ok(redisClient)` in the new node-redis test — removed.
- shared `ProfileName` / `REDIS_COMPAT` helper — **not done**. It is a real duplication, but hoisting it means editing `profile-gates.test.ts` too, and other agents are working in this tree right now; I would rather not touch an unrelated test file in a cluster-guard PR. Happy to do it as a follow-up if you want it.

src/core/compatibility/feature-gates.ts
```diff
@@ -25,4 +25,13 @@ export const FEATURE_GATES: Record<FeatureId, VersionGate> = {
   'hscan.novalues': { redis: '7.4.0', valkey: '9.0.0' },
   'xread.plus-id': { redis: '7.4.0' },
   'cluster.multi-db': { valkey: '9.0.0' },
+  // SORT BY/GET in cluster mode: compare each pattern's slot against the sort
+  // key's slot (`patternHashSlot()`) instead of refusing every pattern, and use
+  // the longer "...may be in different slots." wording — Redis 7.4 / Valkey 8.0.
+  'sort.cluster-pattern-slot': { redis: '7.4.0', valkey: '8.0.0' },
+  // SORT GET '#' (the element itself) is exempt from that slot comparison.
+  // Bisected on single-node clusters: redis 7.4.0/7.4.1 refuse and 7.4.2+
+  // accept; valkey 8.0.0/8.0.1 refuse and 8.0.2+ accept. Before that it is
```
fatal10110, line 41: Verified independently, not taken on trust. My own single-node cluster containers, `SORT l GET '#'`:

```
redis  7.4.0  -> ERR GET option of SORT denied in Cluster mode when keys formed by the pattern may be in different slots.
redis  7.4.1  -> ERR GET option of SORT denied ...
redis  7.4.2  -> accepted
redis  7.4.4  -> accepted        <- what `redis-7.4` pins
redis  7.4.5 / 7.4.11 / 8.0.0 / 8.0.6 -> accepted
valkey 8.0.0  -> ERR GET option of SORT denied ...   <- what `valkey-8.0` pins
valkey 8.0.1  -> ERR GET option of SORT denied ...
valkey 8.0.2 / 8.0.3 / 8.0.4 / 8.1.0 / 9.0.6 -> accepted
```

So `{ redis: '7.4.2', valkey: '8.0.2' }` is exactly right, and it happens to land both presets on the correct side of their own patch version. Confirmed on the branch: `redis-7.4` now accepts `GET '#'` and `valkey-8.0` still denies it.

src/commands/keys.ts
```diff
@@ -733,9 +743,36 @@ function applySortLimit(
   return elements.slice(start, start + limit.count)
 }
 
-function runSort(args: SortArgs, db: RedisDatabase) {
+/**
+ * Mirrors `sortCommand()`'s determinism override: a constant `BY` normally
+ * means "do not sort", but an unordered SET source whose output has to be
+ * reproducible — it is written by STORE, or returned to a script — is
+ * force-sorted ALPHA with the `BY` dropped. Lists and zsets already have a
```
fatal10110: **Medium, pre-existing — but this comment's premise is currently false for zsets.**

"Lists and zsets already have a defined order" is true of real Redis, not of this codebase: `readSortSource` builds the zset case from `db.getSortedSet(key)!.members.values()`, which is the `Map`'s *insertion* order, not rank order. So the `dontsort` path that this function deliberately does not override returns the wrong order for zsets.

```
ZADD z 2 c 1 a 3 b          # insertion c,a,b ; rank a,c,b

real 8.0.6  SORT z BY nosort            -> a, c, b      (rank order, == ZRANGE)
mock        SORT z BY nosort            -> c, a, b      <-- wrong
real 8.0.6  SORT z BY nosort LIMIT 1 2  -> c, b
mock        SORT z BY nosort LIMIT 1 2  -> a, b         <-- wrong
mock        ZRANGE z 0 -1               -> a, c, b      (so only SORT's reader is off)
```

Same for `SORT z BY <any constant> STORE d`. Not a regression — `readSortSource` is untouched here and the pre-PR `BY nosort` was equally wrong — so a follow-up issue is fine. The fix is one line: have the zset branch of `readSortSource` go through `getSortedMembers` (`src/commands/zsets/helpers.ts`), the same accessor `ZRANGE` uses. Until then, consider softening this comment to say *lists* have a defined order, so the reader is not told an invariant the code does not hold.

src/core/execution-policies/cluster-policy.ts
```diff
@@ -87,6 +88,15 @@ export function createClusterPolicy(
         },
       )
 
+      const sortPatternError = getSortClusterPatternError(
```
fatal10110, line 94: **Medium, pre-existing — the MOVED half of my earlier finding is fixed, the `MULTI` half is not.**

Moving the guard after `validateClusterSlot` fixes MOVED and CROSSSLOT (verified: both now match real byte for byte). But `beforeExecute` also runs at *queue* time, and real Redis queues the command and only fails at `EXEC`:

```
real 8.0.6, owning node:
  MULTI                -> +OK
  SORT l BY 'w_*'      -> +QUEUED
  EXEC                 -> 1) (error) ERR BY option of SORT denied in Cluster mode when keys formed ...

mock pr-407, owning node:
  MULTI                -> +OK
  SORT l BY 'w_*'      -> -ERR BY option of SORT denied in Cluster mode when keys formed ...
  EXEC                 -> -EXECABORT Transaction discarded because of previous errors.
```

So a transaction that real Redis runs (with one failed element) is aborted wholesale here. MOVED at queue time *is* correct — real does the same — the difference is that the BY/GET denial is not a routing decision, it comes out of `sortCommand`'s argument scan.

This is the same root cause as the two deferrals you argued (correctly) cannot be fixed in the policy: the check does not belong in `beforeExecute` at all, it belongs in SORT's own argument parse, where left-to-right order, the syntax-error interaction and queue-vs-exec timing all come out right for free. That needs profile + topology in the parse context, which is a real piece of work — I would open one follow-up issue covering all three rather than treat them separately.

tests/core-cluster-policy.test.ts
```diff
@@ -213,6 +213,106 @@ describe('new cluster execution policy', () => {
...
+    ['question mark', '?x{tag-b}*'],
+    ['character class', '[ab]{tag-b}*'],
+    ['backslash escape', '\\x{tag-b}*'],
+  ] as const) {
+    test(`refuses a BY glob opening with a ${name}`, async () => {
+      assert.strictEqual(await sortError(KEY, 'BY', pattern), BY_DENIED)
+    })
+  }
+
+  test("treats '{}' as no tag and hashes the whole GET pattern", async () => {
```
fatal10110, line 274: **Low — this is the one `patternHashSlot` branch the table does not actually pin.**

`{}ids` / `{}other` exercises the empty-tag case, but it cannot tell `tagStart = -2` from `tagStart = -1`: there is no later brace in either pattern, so a port that simply reset the sentinel would pass this test unchanged. The discriminating shape is an empty tag *followed by* a real one:

```js
// -2 (correct): whole pattern is hashed -> different slot -> denied
// -1 (broken):  '{tag-b}' re-opens a tag -> same slot -> accepted
assert.strictEqual(await sortError(KEY, 'GET', '{}x{tag-b}y'), GET_DENIED)
```

I confirmed that is what real Redis does — on 8.0.6, `SORT r407:l GET '{}a{r407:l}b'` is refused even though the embedded tag is the sort key's own, and this branch refuses it too. So the implementation is right; only the coverage claim is a touch optimistic. One extra line in this test closes it.

Everything else in the table does pin its branch — I checked that each of `?x{tag-b}*`, `[ab]{tag-b}*` and `\\x{tag-b}*` would flip to accepted if the corresponding byte were dropped from the wildcard set.

tests/core-cluster-policy.test.ts
```diff
@@ -213,6 +213,106 @@ describe('new cluster execution policy', () => {
...
+    assert.strictEqual(await sortError(KEY, 'BY', '{tag-a}weight'), null)
+    assert.strictEqual(await sortError(KEY, 'BY', 'nosort'), null)
+  })
+
+  test("accepts GET '#' and refuses a cross-slot GET glob", async () => {
+    assert.strictEqual(await sortError(KEY, 'GET', '#'), null)
+    assert.strictEqual(await sortError(KEY, 'GET', '{tag-a}name:*'), GET_DENIED)
+  })
+
+  test('stops at the first NUL when looking for the wildcard', async () => {
```
fatal10110, line 305: **Low / tooling.** The pattern literal here contains a real `0x00` byte in the source file (I hex-dumped it to be sure — `'a' 0x00 'b' '*'`), and that makes the whole file binary as far as text tooling is concerned:

```
$ grep -n "first NUL" tests/core-cluster-policy.test.ts
Binary file tests/core-cluster-policy.test.ts matches
$ rg -n "first NUL" tests/core-cluster-policy.test.ts
binary file matches (found "\0" byte around offset 10520)
```

So ordinary `grep`/`rg` over this repo silently stops returning line matches for this file (`git grep` still works), and the byte is invisible in the GitHub diff. Writing the NUL as a **six-character escape sequence** instead — backslash, `u`, `0`, `0`, `0`, `0` — produces the identical Buffer while keeping the file plain text, and prettier/eslint are happy either way.

(Apologies for the earlier version of this comment: I pasted that escape sequence through a layer that interpreted it, so it arrived as the very NUL byte I was suggesting you remove — which rather makes the point about how easily this byte travels unnoticed.)

To be clear this is cosmetic, not a correctness risk: if the NUL were ever lost the pattern becomes `ab*`, which is denied, so the test fails loudly rather than silently passing.

src/core/lua-runtime.ts
```diff
@@ -120,6 +120,7 @@ function createLuaMonitorContext(
     server: ctx.server,
     session: ctx.session,
     executor: ctx.executor,
+    inScript: true,
```
fatal10110, line 125: Nit: with `inScript` added, `createLuaMonitorContext` is no longer just about the monitor — it is *the* execution context every `redis.call`/`pcall` runs under, and `inScript` is now load-bearing for a reply's contents (`SORT` over a set). `createLuaCallContext` (or `createScriptContext`) would stop the next reader assuming this is monitor plumbing they can skip.

The threading itself is clean and I traced it end to end: one producer here, one consumer at `keys.ts:761`, `executePlanSync` has no other caller, and the force-ALPHA fires for both `SORT` and `SORT_RO` inside `EVAL` exactly as real Redis does (`a,b,c` on both sides for each).

src/core/execution-policies/cluster-policy.ts
```diff
@@ -8,6 +8,9 @@ import {
 } from '../redis-error'
 import type { RedisClientSession } from '../redis-context'
 import type { RedisClusterTopology } from '../../state'
+import type { CompatibilityProfile } from '../compatibility'
+import { isConstantSortPattern, isSelfSortPattern } from '../sort-patterns'
+import type { SortArgs } from '../../commands/keys'
```
fatal10110: Note, not a request: `src/core/sort-patterns.ts` was added specifically to avoid a runtime `src/core` -> `src/commands` import, and then this line reintroduces the same direction as a type-only import. It is erased at compile time so there is no cycle and no bundling consequence — `tsc --noEmit` is clean — but the layering argument is now only half applied. If the intent is that `src/core` never names `src/commands`, `SortArgs`'s cluster-relevant shape (`key`, `by`, `get`) could live next to the predicates in `sort-patterns.ts` and `SortArgs` could extend it; if the intent was only "no runtime edge", this is fine as is and worth a one-line comment saying so.

fatal10110 (commented):
## Re-review of `03f5c9d` — all prior findings verified fixed

I re-ran the full battery against **real redis 8.0.6** (shared standalone + 3-master cluster) and against this branch's mock side by side, and I **re-bisected the version boundary myself** with my own single-node cluster containers rather than taking the reply's word for it. Both deferrals hold up on their merits. Three new items below, none of them blocking.

### Prior findings — status

| # | finding | status |
|---|---|---|
| 1 | `sort.cluster-get-hash` boundary (was 8.0.0/9.0.0) | **fixed** |
| 2 | numeric sort missing the lexicographic tie-break | **fixed** |
| 3 | WRONGTYPE from a non-string weight/GET key | **fixed** |
| 4 | constant `GET` dereferenced instead of nil | **fixed** |
| 5 | SET + constant `BY` + STORE/script not force-ALPHA | **fixed** |
| 6 | guard preempting MOVED / CROSSSLOT | **fixed** (see new item B for the MULTI half) |
| 7 | `plan.keys` reporting BY/GET patterns as keys | **fixed** |
| 8 | `BY` reported before `GET` | deferred — reasoning correct |
| 9 | trailing syntax error preempting the denial | deferred — reasoning correct |
| 10 | NUL-truncation of the `*` search | **fixed** |
| 11 | code quality (cast, duplicated predicates, no unit tests) | **fixed** (boilerplate hoist declined, fine) |

**Boundary re-bisect (my own containers, `cluster-enabled`, all slots, `SORT l GET '#'`):**

```
redis 6.2.14 / 7.0.15 / 7.2.4 / 7.2.16 -> ERR GET option of SORT denied in Cluster mode.        (legacy wording)
redis 7.4.0 / 7.4.1                    -> ERR GET ... when keys formed by the pattern may be in different slots.
redis 7.4.2 / 7.4.3 / 7.4.4 / 7.4.5 / 7.4.11 -> accepted
redis 8.0.0 / 8.0.6                    -> accepted
valkey 8.0.0 / 8.0.1                   -> ERR GET ... may be in different slots.
valkey 8.0.2 / 8.0.3 / 8.0.4 / 8.1.0 / 9.0.6 -> accepted
```

`{ redis: '7.4.2', valkey: '8.0.2' }` is right, and it lands every preset on the correct side — `redis-7.4` pins 7.4.4 (accept), `valkey-8.0` pins 8.0.0 (deny). Confirmed end to end on the branch: `redis-7.2` legacy wording, `redis-7.4` **accepts** `GET '#'`, `redis-8.0` accepts, `valkey-8.0` denies, `valkey-9.0` accepts.

**Behaviour battery, real 8.0.6 vs this branch's mock — every row matches:**

```
                                  real 8.0.6        mock (pr-407)
constant GET                      nil,nil,nil       [null,null,null]
BY glob, w_1 is a list            1,3,2             1,3,2
GET glob, w_1 is a list           nil,5,1           [null,5,1]
SET + const BY + STORE            a,b,c             a,b,c
SET + const BY in EVAL            a,b,c             a,b,c
SET + const BY, SORT_RO in EVAL   a,b,c             a,b,c
SET + const BY DESC STORE         c,b,a             c,b,a
SET + const BY LIMIT 0 2 STORE    a,b               a,b
LIST + const BY STORE             c,a,b             c,a,b
tie-break, weights missing        1,2,3             1,2,3
tie-break, weights all equal      1,2,3             1,2,3
ALPHA + missing weights           c,a,b             c,a,b
COMMAND GETKEYS SORT k BY GET     [k]               [k]
COMMAND GETKEYS ... STORE dst     [k,dst]           [k,dst]
BY "a\0b*" (NUL before the *)     3,1,2             3,1,2
```

Cluster ordering, real vs mock, both now identical: `SORT k BY 'w_*'` to a non-owning node -> `MOVED 8348 …`; `… STORE <other slot>` -> `CROSSSLOT`; the empty-tag discriminator `GET '{}a{k}b'` -> denied on both (so the `-2` sentinel is behaving, even though no test pins it — see inline).

**Deferrals — both sound.** `SORT k GET BY BY GET` really is legal (real 8.0.6 standalone returns three nils), so scanning `plan.rawArgs` for the first `BY`/`GET` token genuinely cannot work; ordering has to come out of `parseSort`. And the syntax-error case is not just unreachable in `beforeExecute` — it is unreachable in `execute` too, since `parseSort` throws inside `plan()`. Real Redis emits the denial *during* its left-to-right argument scan (`SORT l BY 'w_*' BADARG` -> BY denied, `SORT l BADARG BY 'w_*'` -> syntax error), so matching it means teaching the parser the cluster rule, not re-ordering the policy. Deferring that is the right call — but note it is the same root cause as new item B below: all three orderings fall out of the guard living in the policy instead of in SORT's argument scan. Worth one follow-up issue covering all three rather than three separate ones.

### New items (details inline)

- **A — medium, pre-existing:** a zset source with a constant `BY` is returned in *insertion* order, not rank order. `src/commands/keys.ts` inline.
- **B — medium, pre-existing:** inside `MULTI` the guard fires at queue time; real Redis replies `+QUEUED` and surfaces the error inside `EXEC`. `cluster-policy.ts` inline.
- **C/D — low:** one uncovered `patternHashSlot` branch and the raw NUL byte in the test file. `tests/core-cluster-policy.test.ts` inline.
- Two notes (naming, layering) inline on `lua-runtime.ts` and `cluster-policy.ts`.

### What I ran on this branch

`tsc --noEmit` clean, `eslint .` clean, `prettier --check` clean; `tests/core-cluster-policy.test.ts` 24/24; sort integration on mock 64/64; `sort-cluster-gates` 5/5 on each of `redis-7.2` / `redis-7.4` / `valkey-8.0` / `valkey-9.0`; and, most usefully, both new integration files **against the real backend** — ioredis **30/30**, node-redis **34/34**.

### Verdict

**Ship.** Items A and B are pre-existing and neither is made worse here; A only matters because the new doc comment asserts the opposite, so the cheapest resolution is a follow-up issue plus a one-word softening of the comment. This is a clean, well-evidenced round — the bisect in particular was reproduced independently and matched exactly.

tests/core-cluster-policy.test.ts
fatal10110: Good catch — fixed in f4cd14b. You were right that the old pair pinned nothing: neither `{}ids` nor `{}other` has a later brace, so `-1` and `-2` are indistinguishable there.

Ground-truthed your discriminator on the real 6-node cluster before adding it (key `a387:{t}l`, sent to its owning node):

```
SORT a387:{t}l GET '{}x{t}y'            -> ERR GET option of SORT denied in Cluster mode when keys formed ...
SORT a387:{t}l GET '{t}y'               -> (accepted)
SORT a387:{t}l GET '{}a{a387:{t}l}b'    -> ERR GET option of SORT denied in Cluster mode when keys formed ...
```

So the empty tag really does close the tag permanently rather than resetting it, even when the embedded tag is the sort key's own.

Added as its own test with both halves (the refusal *and* the `{tag-b}y` accept, so it fails if the guard simply started denying everything), and I mutation-checked it rather than trusting the green:

```
tagStart = -2  ->  25 pass / 0 fail
tagStart = -1  ->  24 pass / 1 fail   ("'{}' permanently closes the tag rather than resetting it")
```

Exactly one assertion moves, which is what a pinned branch should look like.

tests/core-cluster-policy.test.ts
fatal10110: Fixed in f4cd14b — and you were right about the byte being on disk. I had typed the six-character escape in the first place, but something between my editor and the file collapsed it, so what landed was a real `0x00`:

```
before:  b"or(KEY, 'BY', 'a\x00b*'), null"        b'\x00' in file -> True
after:   b"or(KEY, 'BY', 'a\\u0000b*'), null"     b'\x00' in file -> False
```

Given your warning I did the replacement at the byte level (explicit `0x5c` for the backslash) and asserted `b'\x00' not in data` afterwards, then re-checked the raw bytes again after prettier and again after the pre-commit hooks ran. All three checks are clean, and `grep -c "first NUL" tests/core-cluster-policy.test.ts` now returns `1` instead of "Binary file ... matches".

No behavioural change — same Buffer, and the 25-test file still passes.

src/core/lua-runtime.ts
fatal10110: Agreed — renamed to `createLuaCallContext` in f4cd14b, and added a short doc comment saying what it actually is: the context every `redis.call`/`redis.pcall` runs under, carrying the script's monitor sink *and* `inScript`. `createLuaMonitorContext` now appears nowhere in the tree.

src/core/execution-policies/cluster-policy.ts
fatal10110: Took your first option in f4cd14b: the shape now lives in `sort-patterns.ts` and `src/core` no longer names `src/commands` at all.

```ts
// src/core/sort-patterns.ts
export type ClusterSortArgs = {
  key: Buffer
  by?: Buffer
  get: readonly Buffer[]
}

// src/commands/keys.ts
export interface SortArgs extends ClusterSortArgs { … }
```

Reasoning, since you asked for it to be examined rather than just decided: my original justification for the type-only edge was "no runtime cycle, no bundling consequence", and that is true but it is not the property I actually cared about. What `sort-patterns.ts` exists to protect is the *direction* of dependency — a reader should be able to open anything in `src/core` and know it does not reach into the command layer. An erased import still makes `src/core` name `src/commands`, so it still costs a reader that guarantee, and it would still have to be unpicked by anyone extracting `core` on its own. The one thing the type-only edge bought me was the compile-time link the earlier review asked for (rename `SortArgs.by` → the guard fails to build), and `extends` gives me that for free while pointing the arrow the right way. So there was no trade-off to make; I had just stopped one step short.

Verified: `grep -rn "from '\.\./\.\./commands\|from '\.\./commands" src/core/` returns nothing, and `tsc --noEmit` is clean. The intent is recorded in a comment on `ClusterSortArgs` so the next person does not re-litigate it.

src/core/execution-policies/cluster-policy.ts
fatal10110: Reconsidered and taken, in f4cd14b. My reason for declining last round was that hoisting meant editing `profile-gates.test.ts` while sibling agents were active in this tree; those PRs are all through review now, so the reason has expired and the duplication has not.

`ProfileName` and the `REDIS_COMPAT` read now live in `tests-integration/utils.ts` as `ProfileName` / `activeProfile`, and both `profile-gates.test.ts` and `sort-cluster-gates.test.ts` import them. `expectedVersion` stays in `profile-gates.test.ts` — it is that test's own assertion data, not shared plumbing. The whole 7-profile matrix is still 13 pass / 0 fail per profile.

src/commands/keys.ts
fatal10110: Confirmed, and tracked as #418 — the ordering fix itself is deliberately out of scope for this PR.

I did take your last suggestion, since it is about a comment this PR introduced: `forceDeterministicSetOrder`'s doc no longer tells the reader that zsets have a defined order here, because they do not. It now says *lists* do, notes that real Redis leaves zsets alone for the same reason, and points at #418 for the `readSortSource` gap.

I re-ran the real-Redis half on the shared cluster before editing, so the comment is not asserting something I had not checked myself:

```
ZADD a387:{t}z 2 c 1 a 3 b      # insertion c,a,b ; rank a,c,b
real 8.0.6  SORT a387:{t}z BY nosort   -> a c b
real 8.0.6  ZRANGE a387:{t}z 0 -1      -> a c b
```

and the mock side is exactly as you described — `readSortSource`'s zset branch reads `members.values()`, i.e. `Map` insertion order, while `ZRANGE` goes through `getSortedMembers`.

src/core/execution-policies/cluster-policy.ts
fatal10110: Agreed on all three points, and tracked as #417 — not fixed here.

Your diagnosis matches what I concluded when I deferred the option-order and syntax-error findings last round: the check is in `beforeExecute` because that is where the topology and profile are reachable, but it is not a routing decision, so every place `beforeExecute`'s timing differs from `sortCommand`'s argument scan it goes wrong. Queue-vs-exec is the third face of that, and I had missed it — thanks.

Worth recording for whoever picks up #417: the guard also has to keep running *after* `validateClusterSlot` in whatever form it takes, because MOVED/CROSSSLOT genuinely do precede it at queue time as well as at execute time (verified on the real cluster both ways). So the fix is "move the BY/GET denial into SORT's own parse", not "move the whole thing".

tests/core-cluster-policy.test.ts
```diff
@@ -213,6 +213,116 @@ describe('new cluster execution policy', () => {
...
+    })
+  }
+
+  test("treats '{}' as no tag and hashes the whole GET pattern", async () => {
+    // keyHashSlot() ignores an empty tag, so pattern and key hash identically.
+    assert.strictEqual(await sortError('{}ids', 'GET', '{}ids'), null)
+    assert.strictEqual(await sortError('{}ids', 'GET', '{}other'), GET_DENIED)
+  })
+
+  test("'{}' permanently closes the tag rather than resetting it", async () => {
```
fatal10110, line 280: This is the right test, and I re-ran your mutation check rather than taking the numbers:

```
baseline (tagStart = -2)  -> # tests 25  # pass 25  # fail 0
mutated  (tagStart = -1)  -> # tests 25  # pass 24  # fail 1
                             not ok 9 - '{}' permanently closes the tag rather than resetting it
```

25/0 and 24/1 exactly as you reported, with this test the only one that moves — so it does discriminate the sentinel, which the `{}ids` / `{}other` pair could not.

Ground truth re-confirmed on the recreated cluster (redis 8.0.6), including the `SORT_RO` twin:

```
SORT    r407c:l GET '{}x{r407c:l}y'  -> ERR GET option of SORT denied in Cluster mode when keys formed ...
SORT_RO r407c:l GET '{}x{r407c:l}y'  -> ERR GET option of SORT denied in Cluster mode when keys formed ...
SORT    r407c:l GET '{r407c:l}y'     -> accepted (nil per element)
```

The second assertion earning its place as the control is a nice touch — without it the test would pass against an implementation that refused everything.

tests/core-cluster-policy.test.ts
```diff
@@ -213,6 +213,116 @@ describe('new cluster execution policy', () => {
...
+    assert.strictEqual(await sortError(KEY, 'BY', '{tag-a}weight'), null)
+    assert.strictEqual(await sortError(KEY, 'BY', 'nosort'), null)
+  })
+
+  test("accepts GET '#' and refuses a cross-slot GET glob", async () => {
+    assert.strictEqual(await sortError(KEY, 'GET', '#'), null)
+    assert.strictEqual(await sortError(KEY, 'GET', '{tag-a}name:*'), GET_DENIED)
+  })
+
+  test('stops at the first NUL when looking for the wildcard', async () => {
```
fatal10110, line 305: Confirmed gone, at byte level rather than by eye: I scanned all 14787 bytes of the file on disk for `0x00` — none — and the literal is now the six-character escape sequence (backslash, `u`, four zeros), which `Buffer.from` still turns into `61 00 62 2a`. Tooling is back to normal:

```
$ grep -c "first NUL" tests/core-cluster-policy.test.ts
1
$ rg -n "first NUL" tests/core-cluster-policy.test.ts
305:  test('stops at the first NUL when looking for the wildcard', ...
```

I also round-tripped the file through `prettier --write` independently: the escape survives as an escape and no NUL is reintroduced, so the pre-commit hook will not undo this. `prettier --check` / `eslint` / `tsc` clean.

And the test still means what it says — real 8.0.6 over raw RESP on the owning node, with the NUL byte put on the wire directly: `SORT r407c:l BY <a, NUL, b, asterisk>` returns `3 1 2` (constant, `dontsort`) while `BY "ab*"` is refused.

(Writing that escape sequence out in my previous version of this comment backfired — GitHub turned it back into the control character I was recommending you remove. Hence the prose spelling here.)

src/core/sort-patterns.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+/**
+ * The part of `SORT`'s parsed arguments the cluster guard reasons about.
+ * It lives here rather than in `src/commands/keys.ts` so `src/core` never
+ * names `src/commands` — not even in an erased `import type`, which would
+ * leave the layering only half applied. `SortArgs` extends it, so the two
+ * stay linked and a rename there is a compile error here.
+ */
+export type ClusterSortArgs = {
```
fatal10110, line 18: Good call taking the first option — and the compile-time link is real, I mutation-tested it rather than reading the comment:

```
rename ClusterSortArgs.by -> byPattern (this file only)
  -> 13 tsc errors, all in src/commands/keys.ts
     TS2353 on the parseSort object literal, TS2339 on each args.by
```

So the failure mode my first review wanted is now enforced, and the dependency arrow points the right way (I checked the stronger claim too: `src/core` has no import of `src/commands` of any kind — not a value import, not `import type`, not `require`).

One nuance worth recording, since it is the kind of thing that decays quietly. Because `by` is optional, the *additive* rename still type-checks: declare `byPattern?: Buffer` on `SortArgs` and switch `keys.ts` to it, leaving the inherited `by?` declared and unused, and `tsc` reports **0 errors** — the guard then reads a field nobody writes. What catches that today is the behaviour net, not the type: the same mutation fails **5 of the 25** guard unit tests. Between the two it is closed, so nothing to do here; just do not thin those tests on the grounds that the type now guarantees it.

src/commands/keys.ts
```diff
@@ -733,9 +742,38 @@ function applySortLimit(
...
 }
 
-function runSort(args: SortArgs, db: RedisDatabase) {
+/**
+ * Mirrors `sortCommand()`'s determinism override: a constant `BY` normally
+ * means "do not sort", but an unordered SET source whose output has to be
+ * reproducible — it is written by STORE, or returned to a script — is
+ * force-sorted ALPHA with the `BY` dropped. Lists have a defined order, so
+ * only sets are overridden. (Real Redis leaves zsets alone for the same
+ * reason; `readSortSource` reads them in insertion rather than rank order
```
fatal10110, line 751: Checked this clause by clause against both the code and real Redis, since a comment that over-corrects is as misleading as the one it replaced:

- "Lists have a defined order, so only sets are overridden" — matches the guard below (`db.getType(args.key) !== 'set'`) and matches C's `sortval->type == OBJ_SET`.
- "Real Redis leaves zsets alone for the same reason" — re-verified on 8.0.6: `SORT z BY nosort` -> `a c b`, identical to `ZRANGE`, i.e. rank order and not force-sorted.
- "`readSortSource` reads them in insertion rather than rank order here" — exactly the defect, and #418 describes it accurately.

Neither over- nor under-claims: it does not imply the mock is correct, and it does not imply the gap is cosmetic. Softening this without being asked was the right call — the comment arrived with this PR, so shipping it as written would have shipped a false invariant.

tests-integration/utils.ts
```diff
@@ -319,3 +319,20 @@ export function bufferClient<T extends { withTypeMapping: (m: never) => T }>(
...
+export type ProfileName =
+  | 'redis-6.2'
+  | 'redis-7.0'
+  | 'redis-7.2'
+  | 'redis-7.4'
+  | 'redis-8.0'
+  | 'valkey-8.0'
+  | 'valkey-9.0'
+
+export const activeProfile = (process.env.REDIS_COMPAT ??
```
fatal10110, line 337: Verified this did not change behaviour for `profile-gates.test.ts`, which neither of us owns: the union here is byte-identical to the one it replaced (same seven members, same order), this expression is character-identical to the one removed, and nothing anywhere in `tests-integration/` *writes* `REDIS_COMPAT` — it is only read here and in `test-config.ts`, and set by `package.json`'s per-profile loop before the process starts. So evaluating it at `utils.ts` module scope cannot see a different value.

Empirically, same results either side of the change:

```
redis-6.2   03f5c9d [8 pass 0 fail]   f4cd14b [8 pass 0 fail]
redis-7.2   03f5c9d [8 pass 0 fail]   f4cd14b [8 pass 0 fail]
redis-7.4   03f5c9d [8 pass 0 fail]   f4cd14b [8 pass 0 fail]
valkey-9.0  03f5c9d [8 pass 0 fail]   f4cd14b [8 pass 0 fail]
```

Leaving `expectedVersion` in the test was the right split — it is that test's assertion data, and `Record<ProfileName, string>` keeps exhaustiveness enforced, so a member dropped from the shared union is still a compile error there.

fatal10110 (commented):
## Final pass on `f4cd14b` — clean, nothing regressed

I re-ran every claim rather than reading the summary, including the mutation check. All five items land, and I found nothing new in the PR.

### Verified

**1. The `{}` sentinel test genuinely discriminates.** I reproduced your mutation check independently on an extracted `f4cd14b` tree:

```
baseline (tagStart = -2)   -> # tests 25  # pass 25  # fail 0
mutated  (tagStart = -1)   -> # tests 25  # pass 24  # fail 1
                              not ok 9 - '{}' permanently closes the tag rather than resetting it
```

Exactly your 25/0 and 24/1, and the new test is the only one that moves. Ground truth re-confirmed on the freshly recreated cluster: `SORT r407c:l GET '{}x{r407c:l}y'` -> refused, `GET '{r407c:l}y'` -> accepted (nil elements), and the same refusal via `SORT_RO`.

**2. The NUL really is gone.** Scanned the file on disk byte-wise — no `0x00` anywhere in its 14787 bytes; the literal is the six-character escape. Tooling agrees:

```
$ grep -c "first NUL" tests/core-cluster-policy.test.ts
1
$ rg -n "first NUL" tests/core-cluster-policy.test.ts
305:  test('stops at the first NUL when looking for the wildcard', ...
```

I also round-tripped the file through `prettier --write` myself: the escape survives, no NUL reintroduced. `prettier --check`, `eslint .` and `tsc --noEmit` are clean on the final tree.

**3. `SortArgs extends ClusterSortArgs` does preserve the link** — and I checked both directions rather than assuming:

- Renaming the field where it is declared (`ClusterSortArgs.by` -> `byPattern`, `src/core/sort-patterns.ts` only) -> **13 `tsc` errors**, all in `keys.ts` (`TS2353` on the `parseSort` literal, `TS2339` on each `args.by`). That is the failure mode my first review asked for, and it now fires.
- The one mutation that still type-checks is an *additive* rename — declaring `byPattern?: Buffer` on `SortArgs` and switching `keys.ts` to it while the inherited optional `by` stays declared and unused: **0 tsc errors**, because `by` is optional so neither a missing- nor an excess-property check trips. But that mutation fails **5 of the 25** unit tests, so the guard tests catch what the type cannot. Between the two layers the hole is closed; worth knowing the type alone does not close it, in case the tests are ever thinned.

Also confirmed your stronger claim: `src/core` contains **no import of `src/commands` of any kind** — not a value import, not an `import type`, not a `require`. The only occurrences of the word in `src/core` are prose and unrelated identifiers. `createLuaMonitorContext` appears nowhere in the repo.

**4. The `utils.ts` hoist is behaviour-preserving for `profile-gates.test.ts`.** The union is byte-identical to the one it replaced (same seven members, same order) and `activeProfile` is the same expression; nothing in `tests-integration/` ever *writes* `REDIS_COMPAT` (only `package.json`'s per-profile loop and `test-config.ts`'s read), so moving the read to `utils.ts` module scope cannot change what it sees. Empirically, `profile-gates.test.ts` gives identical results before and after on four profiles:

```
redis-6.2   03f5c9d [8 pass 0 fail]  f4cd14b [8 pass 0 fail]  SAME
redis-7.2   03f5c9d [8 pass 0 fail]  f4cd14b [8 pass 0 fail]  SAME
redis-7.4   03f5c9d [8 pass 0 fail]  f4cd14b [8 pass 0 fail]  SAME
valkey-9.0  03f5c9d [8 pass 0 fail]  f4cd14b [8 pass 0 fail]  SAME
```

`expectedVersion: Record<ProfileName, string>` still forces exhaustiveness, so a member silently dropped from the shared union would be a compile error in that file — the hoist keeps its own safety net.

**5. The softened comment now matches the code, in both directions.** "Lists have a defined order, so only sets are overridden" matches `db.getType(args.key) !== 'set'` and matches C's `sortval->type == OBJ_SET`; "real Redis leaves zsets alone for the same reason" is right (re-verified: `SORT z BY nosort` -> `a c b`, identical to `ZRANGE`, i.e. rank order and *not* force-sorted); and "`readSortSource` reads them in insertion rather than rank order here" is exactly the defect, no more and no less. It does not over-claim that the mock is correct, nor under-claim by implying the gap is cosmetic. #418 exists and describes it accurately.

### No regressions

```
unit                            356/356
compatibility, all 7 profiles    13/13 each
sort integration (mock)          64/64
sort integration (REAL)          ioredis 30/30, node-redis 34/34
scripts (mock / REAL)            24/24 / 24/24      <- covers the lua-runtime rename
cluster integration (mock)       10/10
tsc --noEmit / eslint / prettier clean
```

I also re-ran the mock-vs-real cluster probe across all five relevant profiles. Every row still matches real 8.0.6 — including the asymmetry a shared-predicate refactor could easily have flattened: `BY "a\0b*"` is **accepted** (constant, `strchr` stops at the NUL) while `GET "a\0b*"` is **refused** (the GET guard goes through the length-based `patternHashSlot`, which does see the `*`). Confirmed on the owning node over raw RESP on both sides.

### One note, on #418 rather than on this PR

The issue's repro block pairs `ZADD z 1 a 3 c 2 b` with real output `a, c, b`. With those scores real 8.0.6 returns `a, b, c`; `a, c, b` is the output for `ZADD z 2 c 1 a 3 b` (the spelling from my review comment). Both spellings demonstrate the bug — the scores and the expected output just got crossed when the issue was written. Nothing to change here; flagging so whoever picks up #418 does not chase the mismatch.

### Verdict

**Ship.** Three rounds, every finding closed with ground-truthed evidence, and the two items that could not be fixed in this layer were deferred with correct reasoning and a follow-up issue. The mutation check in particular is the right instinct — it is what turns "the branch is covered" into something checkable.

PR #408, opened 2026-09-21
fix: decode flat-pairs as RESP3 tuples on RESP3 connections (#385)

### LINKED ISSUE #385: flat-pairs decoded flat in both client mocks; real node-redis returns [field, value] tuples under RESP3
Found while reviewing PR #378. **Pre-existing on `main`** — equally wrong before and after that PR.

## Problem
Both in-memory clients decode a `flat-pairs` reply into a flat array. Real node-redis returns `[field, value]` **tuples** when the connection is in RESP3.

The shared decoder (`decodeRedisValue`, added in #378; previously duplicated in `node-redis-mock.ts` and `in-memory-client.ts`) flattens unconditionally:

```ts
case 'flat-pairs':
  return value.entries.flatMap(([k, v]) => [decode(k), decode(v)])
```

**Failure scenario:** a consumer written against real node-redis in RESP3 iterates `for (const [field, value] of reply)` and gets characters/undefined against the mock. Affects the commands that reply with flat-pairs — check `XRANGE`/`XREAD` entry fields, `CONFIG GET`, `HRANDFIELD ... WITHVALUES`, `ZRANDMEMBER ... WITHSCORES`.

## Proposal
Make the shape protocol-dependent: flat under RESP2, tuples under RESP3 — matching what each real client does. The decoder already takes an options object (`pushShape`, `narrowBigInt`, `error`, `returnBuffers`), so this is a natural fifth knob.

## Acceptance
A RESP3 decode test per affected command, asserting the tuple shape against what real node-redis produces.

fatal10110:
## Summary

`decodeRedisValue` flattened every `flat-pairs` reply unconditionally, so both socketless clients (`InMemoryRedisClient` and the node-redis facade) handed back `[member, score, …]` even on a connection that had negotiated RESP3. A consumer written against real node-redis — which speaks RESP3 by default as of `redis@6` — doing `for (const [field, value] of reply)` reads characters out of that flat array.

`DecodeRedisValueOptions` gains a fifth knob, `flatPairsShape: 'flat' | 'tuples'`, plus a `flatPairsShapeFor(respVersion)` helper that holds the RESP2/RESP3 rule in one place. Both socketless clients derive it from the RESP version their `ClientSession` negotiated, read per reply so a `HELLO 3` switches the shape mid-connection the way a real client's parser does. The default stays RESP2/flat; the wire encoder, the server and every command implementation are untouched, and nothing is compatibility-gated — this is protocol-dependent, not Redis-version-dependent.

In the node-redis facade `decodeReply` now requires the version explicitly (a silent default is what produced this bug); the curated methods that immediately narrow a reply to a scalar or a flat string array use a separate `decodeScalarReply`, since no `flat-pairs` can reach them. The cluster facade reads the version off the node session that served the command — `HELLO` has no keys, so it lands on `masters[0]` and leaves the other nodes' sessions alone, the same per-node caveat `sessionForCommand` already documents for pub/sub.

### Real-client ground truth

Probed with `node-redis@6` and `ioredis@5` against real Redis 8.0.6, using `sendCommand` / `.call` so no per-command `transformReply` is in the way:

| command | node-redis `RESP: 2` | node-redis `RESP: 3` (v6 default) |
| :-- | :-- | :-- |
| `ZRANGE z 0 -1 WITHSCORES` | `["a","1","b","2"]` | `[["a",1],["b",2]]` |
| `ZRANDMEMBER z -3 WITHSCORES` | `["b","2","a","1","a","1"]` | `[["b",2],["b",2],["a",1]]` |
| `ZPOPMIN z 2` | `["a","1","b","2"]` | `[["a",1],["b",2]]` |
| `ZUNION 1 z WITHSCORES` | `["a","1","b","2"]` | `[["a",1],["b",2]]` |
| `HRANDFIELD h -3 WITHVALUES` | `["f2","v2","f1","v1","f1","v1"]` | `[["f2","v2"],["f2","v2"],["f1","v1"]]` |

The two clients do **not** agree: `ioredis@5` has no RESP3 — passing `protocol: 3` still negotiates `proto 2` against the same server — so a real ioredis always sees the flat shape and needs no switch.

Two of the shapes the issue asked to check turn out **not** to be `flat-pairs` and are unchanged, confirmed against the same server:

- `XRANGE` / `XREAD` entry fields stay a flat array on both protocols (`[["1-1",["fa","va","fb","vb"]]]`); the implementation builds them with `RedisValue.array`.
- `CONFIG GET` is a `map` reply (`RedisValue.map`), an object on RESP3.

## Test plan

- [x] New `tests/flat-pairs-resp3.test.ts` covers ZRANGE / ZRANDMEMBER / ZUNION / ZPOPMIN `WITHSCORES` and `HRANDFIELD WITHVALUES` on both socketless clients, at RESP2 and after `HELLO 3`, plus an empty-reply case and a RESP3 `MULTI`/`EXEC` replay. Red before the fix (the 3 RESP3 cases failed, the RESP2 ones passed), green after.
- [x] `tests/core/decode-redis-value.test.ts` gains the option-level assertion: both clients' option constants default to `'flat'`, `flatPairsShapeFor` maps 2→flat / 3→tuples, and the decoder produces each shape.
- [x] `npm test` — 349 pass, 0 fail.
- [x] `npm run test:integration:mock` — 1114 pass, 0 fail.
- [x] `npm run build`, `npm run lint`, `prettier --check` — clean.
- [x] Targeted real-Redis run (`TEST_BACKEND=real`, node-redis zset + hash suites) passes; these integration tests drive the *real* clients over TCP and so never touch the changed code path. The behavioral check against real Redis is the ground-truth table above.

Deliberately **not** in scope (separate pre-existing divergences in the same decoder, worth their own issues): the socketless clients decode a `map` / `map-pairs` reply to an object regardless of protocol, where real node-redis at RESP2 returns a flat array; and a sorted-set score decodes to a `number` on both protocols, where real node-redis at RESP2 returns a string.

Closes #385


fatal10110 (commented):
## Verdict: **don't ship as-is**

The core decoder change is correct, and it is unusually well-evidenced — the real-client ground truth quoted in the PR description was independently confirmed, command by command, against `node-redis@6` and a live Redis 8.0.6. The RESP2/RESP3 `flat-pairs` rule itself, its single home in `flatPairsShapeFor`, and the `InMemoryRedisClient` and `NodeRedisMockClient` wiring are all right.

What blocks it is what the change does on the way through the cluster facade. On `main` every node returned uniformly flat, so a `WITHSCORES` reply had one shape regardless of routing. After this PR the shape of that reply depends on which node owns the key and on scheduling — a `Promise.all` of two `ZRANGE`s can hand back tuples for a key that must be flat and flat for a key that must be tuples. That is a new high-severity nondeterminism that did not exist before, and it is worth stressing that it is *introduced* here rather than merely uncovered.

Findings 1 and 2 are one bug wearing two hats and share a single fix of roughly fifteen lines: hold one client-level protocol version on the facade, set it when `run()` observes a `HELLO`/`RESET`, and replay it onto each node session in `sessionFor()`. That removes the per-reply read of `lastRespVersion` entirely, so the race in finding 1 disappears along with the per-node split in finding 2. Everything below that is medium or low and can be sequenced afterwards.

### Verified-correct claims

These were checked rather than taken on trust, and all hold:

- The flat-pairs producer enumeration is complete — there are exactly two producers, `scorePairs` in `helpers.ts` and `hashEntriesResult` in `hashes.ts`.
- `XRANGE` is genuinely not a flat-pairs reply, so leaving it alone is right.
- `flatPairsShapeFor` is total over `RespVersion = 2 | 3` and has an exhaustive fallback.
- `RESET` correctly demotes the session back to RESP2.
- The `decodeScalarReply` assumption that no flat-pairs reply can reach it holds across all 21 call sites.
- The claim that `ioredis@5` has no RESP3 was independently confirmed: five different option spellings all negotiate `proto 2`, there is no `RESP3` string anywhere in the shipped build, and `createIoredisMock` is not in the decoder's path at all.

### On the failing integration test

The single unexplained `test:integration:mock` failure mentioned in the PR description has been diagnosed and is **not** caused by this PR. It is a pre-existing 20 ms hash-field-TTL race, now filed as issue #411.

### Findings outside the diff

None — every finding below anchors to a line inside the diff. Two of them point at code a little outside their hunk (the stale comment near line 106, and `teardown()` near line 736); each is raised on the nearest in-diff line and says so explicitly.

src/client-mocks/node-redis-mock.ts
```diff
@@ -738,9 +765,21 @@ export class NodeRedisMockCluster extends CommandRunner {
...
+   */
+  protected get respVersion(): RespVersion {
+    return this.lastRespVersion
+  }
+
   protected async run(args: NodeRedisCommandArgument[]): Promise<RedisValue> {
     const session = this.sessionForCommand(args)
-    return runOnSession(session, args, this.closed)
+    const value = await runOnSession(session, args, this.closed)
+    this.lastRespVersion = session.protocolVersion
```
fatal10110: **HIGH (blocking) — `lastRespVersion` is read a microtask after it is written, so concurrent commands cross-contaminate each other's reply shape.**

`NodeRedisMockCluster.run()` has no `runExclusive` guard — the standalone client has one at around line 394, but the cluster path does not — so `await runOnSession(...)` interleaves freely across concurrent callers. This line then stores the protocol version of whichever node happened to finish last, and `decodeReply(value, this.respVersion)` in `sendCommand` (line 179) reads it one microtask later. Between those two points another command's completion can overwrite the field.

The result is that reply shape becomes a function of scheduling. Confirmed in both directions on a 3-master mock cluster after `HELLO 3`:

```
sequential k2 (masters[0], RESP3): [["a",1],["b",2]]
sequential k0 (other node, RESP2): ["a",1,"b",2]
concurrent  k2: ["a",1,"b",2]      <- k2 lost its tuples
reversed    k0: [["a",1],["b",2]]  <- k0 gained tuples it must not have
```

A plain `Promise.all` of two `ZRANGE ... WITHSCORES` calls is enough to trigger it, and it fails in the direction that silently corrupts data rather than throwing.

The fix is shared with the next finding: hold a single client-level protocol version on the facade, set when `run()` observes a `HELLO`/`RESET`, and replay it onto each node session in `sessionFor()`. That deletes the per-reply read of `lastRespVersion` altogether, and the race goes with it.

src/client-mocks/node-redis-mock.ts
```diff
@@ -738,9 +765,21 @@ export class NodeRedisMockCluster extends CommandRunner {
     return this
   }
 
+  /**
+   * The version negotiated on the node session that served the last command.
+   * There is no single connection to read here: `HELLO` has no keys, so it
+   * lands on `masters[0]` and leaves the other nodes' sessions where they were
+   * — the same per-node caveat {@link sessionForCommand} documents for pub/sub.
+   */
+  protected get respVersion(): RespVersion {
```
fatal10110, line 867: **HIGH (blocking) — the negotiated protocol only reaches one node, so a RESP3 cluster client sees a mix of flat and tuple replies.**

The doc comment above correctly identifies that `HELLO` is keyless and therefore routes to `masters[0]`, but it treats the consequence as an acceptable caveat rather than a bug. It is a real divergence: real node-redis propagates the `RESP` setting to *every* node client when it builds the slot map (see `@redis/client/dist/lib/cluster/cluster-slots.js`), so a real RESP3 cluster client gets tuples for every key without exception. Here the other nodes' sessions stay on RESP2 forever.

Confirmed sequentially after `HELLO 3` on a 3-master mock cluster — the reply shape simply tracks key ownership:

```
{ k0: FLAT, k2: TUPLES, k3: TUPLES, k4: FLAT, ... }
```

What makes this worse than a documented limitation is that the facade answers `{"proto":3}` to the caller, so nothing in the observable behaviour signals that only one node actually switched. A consumer writing `for (const [field, value] of reply)` against a connection that reported RESP3 will iterate characters out of a flat array for any key that happens to land on another node.

**Combined fix for this and the previous finding (~15 lines):** hold one client-level protocol version on the facade, set it when `run()` observes a `HELLO`/`RESET`, and replay it onto each node session in `sessionFor()` so every session is brought up to the client's version before it serves a command. The per-reply `lastRespVersion` read then becomes unnecessary and can be deleted, which also closes the race in finding 1.

src/client-mocks/node-redis-mock.ts
```diff
@@ -652,6 +677,7 @@ export class NodeRedisMockMulti {
     this.settled = true
 
     const result = await this.runTransaction(this.queued)
+    const respVersion = this.respVersion()
```
fatal10110: **MEDIUM — `exec()` applies one protocol version to every queued reply, so a `HELLO` inside a `MULTI` retroactively reshapes the replies that preceded it.**

`this.respVersion()` is read once here, after the whole transaction has run, and the resulting value is then applied uniformly to every item in the loop below. If the queued commands include a `HELLO 3`, the items queued *before* it get decoded under RESP3 even though a real server produced them under RESP2.

Confirmed by A/B against real `node-redis@6` and Redis 8.0.6, running `MULTI; ZRANGE z 0 -1 WITHSCORES; HELLO 3; ZRANGE z 0 -1 WITHSCORES; EXEC` on a RESP2 connection:

```
REAL item0 (before HELLO 3): ["a","1","b","2"]     MOCK item0: [["a",1],["b",2]]   <- wrong
REAL item2 (after  HELLO 3): [["a",1],["b",2]]     MOCK item2: [["a",1],["b",2]]   <- right
```

The session layer already solves this correctly on a per-item basis: `src/core/client-session.ts` tracks `sawProtocolSwitch` and pre-encodes each item under the version in force at the time. The facade then decodes `.value` and discards that work. Reaching for the per-item version the session already computed, instead of a single post-hoc read, would line this up with real behaviour.

Worth noting that the common case is fine — real node-redis' `EXEC` nesting was verified independently (RESP2 `[["a","1","b","2"],...]`, RESP3 `[[["a",1],["b",2]],...]`) and the mock matches whenever no `HELLO` is queued. This only bites on a protocol switch mid-transaction, which is why it is medium rather than blocking.

src/core/decode-redis-value.ts
```diff
@@ -109,9 +132,9 @@ export function decodeRedisValue(
       return out
```
fatal10110, line 131: **MEDIUM — issue #385 is not fully closed by this PR: the `map` arm has the same bug one case-label away.**

This is the `map` / `map-pairs` arm, immediately above the `flat-pairs` arm the PR fixes. The PR description says `CONFIG GET` "is a map reply" and leaves it there, which is true on the RESP3 axis but misses where the actual divergence lives: it is at **RESP2**, in the `map` dimension. A `map` reply is an object on RESP3 but a flat array on the RESP2 wire, and the decoder returns an object at both.

Confirmed by comparing `node-redis@6` `sendCommand` against the mock:

| | real RESP2 | mock RESP2 | real RESP3 | mock RESP3 |
|---|---|---|---|---|
| `HGETALL` | `["f1","v1","f2","v2"]` | `{f1:...}` :x: | `{f1:...}` | :white_check_mark: |
| `CONFIG GET maxmemory` | `["maxmemory","0"]` | `{maxmemory:"0"}` :x: | `{maxmemory:"0"}` | :white_check_mark: |
| `XREAD` | array of pairs | object :x: | object | :white_check_mark: |

This is the same bug class as the one being fixed, and the fix would have the same shape — a `mapShape` knob derived from the negotiated version.

One caveat for whoever picks it up: node-redis' curated `hGetAll()` and `configGet()` helpers return objects at *both* protocols, because their `transformReply` normalises. So a `mapShape` knob must not reach `CommandRunner.hGetAll()`, only the raw `sendCommand` path. Tracked separately as issue #414 — flagging it here so the next reader does not assume #385 is closed.

src/client-mocks/node-redis-mock.ts
```diff
@@ -226,7 +236,8 @@ abstract class CommandRunner extends EventEmitter {
   }
 
   async hGetAll(key: string): Promise<{ [field: string]: string }> {
-    const reply = decodeReply(await this.run(['HGETALL', key]))
+    const value = await this.run(['HGETALL', key])
+    const reply = decodeReply(value, this.respVersion)
     // decode() turns a RESP2 map reply into a plain object already.
```
fatal10110: **MEDIUM — this comment asserts something the PR's own evidence disproves, and it now sits next to a correct one.**

"decode() turns a RESP2 map reply into a plain object already" is stated as a fact about the protocol, but it is only true of *this* decoder, not of RESP2. Real node-redis at RESP2 returns `["f1","v1","f2","v2"]` for `HGETALL`; the object shape is RESP3 behaviour (see the table in the finding on `src/core/decode-redis-value.ts`). The comment reads as a justification for the cast below it, which makes it actively load-bearing for the next reader.

The same applies to the comment around line 106 — "mirrors node-redis RESP2 defaults" — which is outside this hunk but is the same claim and is false for the same reason. I am raising it here because line 106 is not in the diff.

Both of these are now adjacent to the newly-added and genuinely correct protocol-aware comments introduced by this PR, which makes the contradiction harder to spot and more likely to mislead. Since the PR is already touching the protocol story in this file, rewording them to describe what the decoder does — rather than what RESP2 does — would be cheap and would leave the file internally consistent.

src/in-memory-client.ts
```diff
@@ -51,6 +52,9 @@ export const IN_MEMORY_DECODE_OPTIONS: DecodeRedisValueOptions = {
   // is `["message", channel, payload]`); keep the type tag so push-mode
   // consumers see the same shape a real client would.
   pushShape: 'tagged',
+  // The RESP2 default. Overridden per reply from the session's negotiated
+  // version, so a connection that sent `HELLO 3` reads `[k, v]` tuples.
+  flatPairsShape: 'flat',
```
fatal10110: **LOW — `flatPairsShape: 'flat'` is dead in both constants, and the test that guards it would not notice a regression.**

Every real call site now spreads over this value: `InMemoryRedisClient.decode()` overrides it with `flatPairsShapeFor(this.session.protocolVersion)`, and the node-redis facade's `decodeReply` does the same. The only code that ever observes the constant's value is `decodeScalarReply`, which by construction only handles values that can never be flat-pairs. So the field is initialised, spread over, and never read for its meaning — in this constant and in `NODE_REDIS_DECODE_OPTIONS` alike.

The consequence is a test that cannot fail in a useful way: `assert.strictEqual(NODE_REDIS_DECODE_OPTIONS.flatPairsShape, 'flat')` in `tests/core/decode-redis-value.test.ts` (around lines 145-146) would stay green even if the per-session override were removed entirely and every RESP3 reply came back flat. It asserts the default, not the behaviour.

Not worth blocking on, but the comment here explains a default that nothing depends on, which will read as meaningful to the next person who touches it. The design finding on `DecodeRedisValueOptions` suggests a way to make the whole field go away.

tests/flat-pairs-resp3.test.ts
```diff
@@ -0,0 +1,244 @@
...
+        [],
+      )
+      assert.deepStrictEqual(
+        await c.command('HRANDFIELD', 'missing', 2, 'WITHVALUES'),
+        [],
+      )
+    })
+  })
+
+  describe('node-redis facade', () => {
```
fatal10110, line 162: **LOW — the new suite covers the two clients that already worked and skips the one that received a new mechanism.**

The style of this file is right — `node:test` plus `node:assert`, real clients rather than stubs, and the ground-truth transcript in the header comment is exactly the sort of evidence that makes a protocol test reviewable. The gap is in what it reaches. It covers `InMemoryRedisClient` and `NodeRedisMockClient`, both of which take the straightforward path of reading their own session's version. It never touches `NodeRedisMockCluster`, which is the one client that got a genuinely new mechanism in this PR — and the one where that mechanism is broken.

Beyond that, there is no `HELLO 2` downgrade case, no `RESET` case, and no concurrency case anywhere in the file. The practical effect is that findings 1, 2 and 3 are all invisible to this suite: each would be caught by roughly four lines.

Concretely, the missing cases are a cluster client issuing `HELLO 3` and then reading two keys that hash to different nodes; the same two reads under `Promise.all`; a `HELLO 3` followed by `HELLO 2` (and by `RESET`) asserting the shape goes back to flat; and a `MULTI` with a `HELLO 3` queued in the middle. Adding the cluster ones alongside the fix would keep the regression from coming back.

src/core/decode-redis-value.ts
```diff
@@ -41,12 +42,34 @@ export type DecodeRedisValueOptions = {
...
   error: (text: string, code?: string) => Error
   /** Return `Buffer`s for bulk-string/verbatim replies instead of utf8 strings. */
   returnBuffers?: boolean
 }
 
+/**
+ * The {@link DecodeRedisValueOptions.flatPairsShape} a connection speaking
+ * `version` must decode to — the one place the RESP2/RESP3 rule lives.
+ */
+export function flatPairsShapeFor(
```
fatal10110: **LOW (design) — passing the `RespVersion` itself would be simpler than passing a shape derived from it, and it generalises where this does not.**

`encodeRedisValue` already takes `{ version: RespVersion }`, so the decoder taking a derived enum is an asymmetry within the same module. Putting `version?: RespVersion` on `DecodeRedisValueOptions` instead carries exactly the same information — the decoder can branch on it directly at the `flat-pairs` arm — and lets `flatPairsShapeFor` and both `...NODE_REDIS_DECODE_OPTIONS, flatPairsShape: ...` spreads be deleted outright.

The reason this is worth raising now rather than later is that it does not stay a one-knob question. The `map`/`map-pairs` divergence raised separately needs the same protocol bit, and so does the RESP2 score-type divergence. Taken the current way, each of those adds another option (`mapShape`, `doubleShape`) plus another `...For()` helper plus another spread at every call site — three knobs, three helpers, and every caller re-deriving all of them from one value it already has. Taken as `version`, they are three `if`s inside the decoder and no call-site change at all.

Non-blocking, and reasonable to do as part of whichever follow-up lands the `map` fix rather than in this PR.

src/client-mocks/node-redis-mock.ts
```diff
@@ -707,6 +733,7 @@ export class NodeRedisMockCluster extends CommandRunner {
   private readonly masters: ClusterNodePipeline[]
   private readonly sessions = new Map<string, ClientSession>()
   private readonly replicationLinks: { close(): void }[]
+  private lastRespVersion: RespVersion = 2
```
fatal10110: **LOW — this new per-connection field is not cleared in `teardown()`.**

`lastRespVersion` is added here as mutable per-connection state, but `teardown()` (around line 736 in the current file, just outside this hunk) does not reset it to `2` the way it clears the connection's other state. Today this is latent rather than a live bug — the field is overwritten on the next `run()` before anything reads it — so it only matters if a future change ever reads the version before the first command of a reused connection.

Raising it because AGENTS.md specifically calls out per-connection teardown as the place new state has to be registered, so this is the repo's own convention rather than a general tidiness point. One line in `teardown()`.

Note that if the combined fix for findings 1 and 2 is taken, this field disappears and is replaced by a single client-level version — which will need the same teardown registration.

docs/TESTING.md
```diff
@@ -255,6 +255,13 @@ Three flavours, depending on which client you want to look like:
 | `createNodeRedisMock`  | `node-redis`    | a hand-written facade mirroring node-redis' public surface |
 | `createInMemoryClient` | our own bespoke | a thin client that returns native JS replies, no RESP      |
 
+The two hand-rolled clients (`createNodeRedisMock`, `createInMemoryClient`)
+start on RESP2 and follow a `HELLO 3` the way a real connection does, so the
+pair-shaped replies change with the protocol: `WITHSCORES` / `WITHVALUES` come
+back flat (`['a', 1, 'b', 2]`) on RESP2 and as tuples (`[['a', 1], ['b', 2]]`)
```
fatal10110, line 261: **LOW (pre-existing, no regression) — this documents a RESP2 value real node-redis never produces.**

`ZSCORE` and `WITHSCORES` decode to numbers on both protocols, where real node-redis returns *strings* at RESP2: the real RESP2 reply is `["a","1","b","2"]`, not `['a', 1, 'b', 2]`. That divergence predates this PR and this PR does not regress it — RESP3 is now byte-for-byte correct and RESP2 is exactly as it was.

The reason to mention it is that this paragraph, and the new test's RESP2 expectations (`['a', 1, 'b', 2]`), both write down the wrong RESP2 value as though it were ground truth. The rest of the surrounding evidence in this PR is real-client-verified, so a reader has every reason to trust these numbers too, and the divergence gets entrenched rather than merely tolerated. The header comment in `tests/flat-pairs-resp3.test.ts` does flag it explicitly, which is good; this doc paragraph does not.

A sentence here noting that the scores are numbers where real node-redis at RESP2 gives strings would keep the docs honest without changing any behaviour. Tracked as issue #414.

src/client-mocks/node-redis-mock.ts
fatal10110: Fixed in c20d539, together with the next finding.

Reproduced first, on a 3-master mock cluster after `HELLO 3`, and got exactly what you did — including the direction that invents tuples:

```
sequential per key : {"k0":"FLAT","k1":"FLAT","k2":"TUPLES","k3":"TUPLES","k4":"FLAT","k5":"FLAT"}
concurrent k2/k0   : FLAT FLAT
concurrent k0/k2   : TUPLES TUPLES
```

`NodeRedisMockCluster` now holds one client-level `clientRespVersion` instead of the per-reply `lastRespVersion`, so `respVersion` no longer depends on which node finished last, and the write-then-read-a-microtask-later window is gone rather than narrowed. The same probe after the fix:

```
sequential per key : {"k0":"TUPLES","k1":"TUPLES","k2":"TUPLES","k3":"TUPLES","k4":"TUPLES","k5":"TUPLES"}
concurrent k2/k0   : TUPLES TUPLES
concurrent k0/k2   : TUPLES TUPLES
```

I did not add `runExclusive` to the cluster path: serializing every cluster command to fix a decode bug would be a real behavioural change for a facade that deliberately lets independent nodes run in parallel, and with the version no longer per-node there is nothing left for the interleaving to corrupt. Covered by a `Promise.all` test in both orders in `tests/flat-pairs-resp3.test.ts`.

src/client-mocks/node-redis-mock.ts
fatal10110: Agreed and fixed in c20d539. You are right that the doc comment treated a divergence as a caveat — and the point that the facade reports `proto: 3` while only one node honours it is what settles it: there is no way for a consumer to notice.

Reproduced the key-dependent shape before changing anything (`{k0:FLAT, k1:FLAT, k2:TUPLES, k3:TUPLES, k4:FLAT, k5:FLAT}` after a single `HELLO 3`), and it is uniform `TUPLES` afterwards.

Implementation follows your suggestion with one deliberate difference: the replay onto other node sessions is **lazy**, done in `syncProtocol()` on each node's next command rather than eagerly fanned out at `HELLO` time. A session that has not run a command has not produced a reply to mis-shape, so the two are indistinguishable from outside, and the lazy form also covers sessions created *after* the switch, which an eager fan-out would miss. `sessionFor()` therefore stays synchronous and unchanged.

The version is replayed by sending a real `HELLO <n>` on that node's session rather than poking `setProtocolVersion`, which keeps it going through the same command path a real per-node handshake would.

`clientRespVersion` is reset in `teardown()` (your separate finding on that field).

src/client-mocks/node-redis-mock.ts
fatal10110: Confirmed and fixed in c20d539.

I re-ran your A/B against Redis 8.0.6 with `node-redis@6` and also checked the reverse direction and the `HELLO` item itself, since that decides where the boundary sits:

```
connection RESP2, queued HELLO 3
  item0 : ["a","1"]
  item1 (the HELLO reply) is OBJECT (RESP3 shape)
  item2 : [["a",1]]

connection RESP3, queued HELLO 2
  item0 : [["a",1]]
  item1 (the HELLO reply) is ARRAY (RESP2 shape)
  item2 : ["a","1"]
```

So the switch takes effect *at* the `HELLO`'s own reply, not after it — which matches what `ClientSession.appendTransactionValue` already does, since it encodes with `this.protocolVersion` read after the command ran.

I did not reach into the session for the per-item versions in the end. What it computes is a pre-encoded `Buffer` per item, not a version per item, so exposing it would have meant adding transaction-scoped state to a core class for one facade's benefit. Instead `runTransactionSpan` returns a `replyVersions` array replayed from the queue the facade already owns, and `exec()` decodes item *i* under `replyVersions[i]`.

Two things I checked rather than assumed while doing that:

- `RESET` does not need modelling: real Redis runs it immediately instead of queueing it, and it aborts the transaction — `MULTI; RESET; EXEC` gives `ERR EXEC without MULTI` on 8.0.6.
- The replay is validated rather than trusted: if the version it predicts does not match the one actually in force after `EXEC` (a rejected `HELLO`, or any future command that moves the protocol), every item falls back to the final version, which is the behaviour this had before.

New test in `tests/flat-pairs-resp3.test.ts` asserts item0 flat / item2 tuples for a `HELLO 3` queued mid-MULTI, and that the connection is left on RESP3 afterwards.

src/core/decode-redis-value.ts
fatal10110: Agreed on the substance, deferring the fix to #414 as you suggest.

Your table matches what I measured independently while writing this PR — `HGETALL` and `CONFIG GET maxmemory` are flat arrays at RESP2 through `sendCommand` and objects at RESP3 — which is why I called it out in the PR description and filed it. Worth adding one more row from my own probe: `XREAD`'s top level has the same split (array of `[stream, entries]` pairs at RESP2, object keyed by stream name at RESP3), and it is `map-pairs` rather than `map`, so both arms need it.

Taking your `hGetAll`/`configGet` caveat seriously: a `mapShape` knob must not reach the curated methods. I have not built it here, but I did make the refactor below pay for it — `DecodeRedisValueOptions` now carries `version` rather than a derived shape, so #414 adds `if (options.version === 3)` inside the `map` arm and changes no call site.

I have also reworded the `hGetAll` comment that asserted the object shape was RESP2 behaviour (your next finding), pointing at #414 instead, so the file no longer claims #385 settled this.

src/client-mocks/node-redis-mock.ts
fatal10110: Both fixed in c20d539.

Line 106 (`NodeRedisReply`) now just says "Native JS value a reply decodes to" — the RESP2 claim added nothing that the type did not already say, and was false.

The `hGetAll` comment now describes what the decoder does and why the cast is still right, rather than making a claim about the protocol:

```ts
// This decoder renders a `map` reply as a plain object at either protocol,
// which is what node-redis' own `hGetAll` transformReply produces — not
// what RESP2 puts on the wire, where a map is a flat array (see #414).
```

That keeps it load-bearing for the cast (the curated method *is* correct at both protocols, because of the transformReply) while pointing at the raw-`sendCommand` divergence instead of denying it.

src/in-memory-client.ts
fatal10110: You are right on both halves, and taking the design suggestion below removed the field entirely — fixed in c20d539.

`flatPairsShape` is gone from both constants and from the options type. `DecodeRedisValueOptions` is now `ClientDecodeOptions & { version: RespVersion }`: the constants carry only what genuinely belongs to the *client*, and the connection supplies `version` at decode time. So there is no longer a default to explain, and nothing initialised-then-always-overridden.

Your point about the test was the sharper one — it asserted the default rather than the behaviour, and would have stayed green with the override deleted. It is replaced by a test that decodes the same `flat-pairs` reply under `version: 2` and `version: 3` **through both clients' option objects**, so it fails if either client's decoding stops depending on the protocol. The behaviour is then covered end-to-end by the client-level cases in `tests/flat-pairs-resp3.test.ts`, which now include the cluster facade.

`decodeScalarReply` is the one remaining place that passes a version it cannot observe; it says so.

tests/flat-pairs-resp3.test.ts
fatal10110: Fair hit — the suite tested the easy clients and skipped the one carrying the new mechanism. All four gaps are covered in c20d539.

Added to `tests/flat-pairs-resp3.test.ts`:

**`describe('node-redis cluster facade')`** over a 3-master mock with six keys, enough to land on every master:
- `HELLO 3` reaches every node — asserts the shape per key, which is the test that fails on the old code with `{k0:FLAT, k2:TUPLES, …}`.
- concurrency — `Promise.all([zrange('k2'), zrange('k0')])` and the same pair reversed, both directions, which is what caught the cross-contamination.
- `HELLO 2` downgrade, after touching a non-`masters[0]` key first so more than one node session is actually live and on RESP3 before the downgrade.
- a RESP2 baseline, so "uniformly flat" is asserted rather than assumed.

**Standalone**: `HELLO 2` and `RESET` both returning to the flat shape, and a `HELLO 3` queued mid-MULTI asserting item0 flat / item2 tuples.

All four are new-code-only guards; I verified the pre-fix behaviour by probe rather than by reverting, and the probe output is in my replies on findings 1 and 2.

src/core/decode-redis-value.ts
fatal10110: Persuaded, and done here rather than deferred — c20d539. The argument that decided it is the one about #414: taking it later means paying the migration anyway, with two more knobs and two more `...For()` helpers written first.

`DecodeRedisValueOptions` is now:

```ts
export type DecodeRedisValueOptions = ClientDecodeOptions & {
  version: RespVersion
}
```

`flatPairsShapeFor` is deleted, the `flat-pairs` arm branches on `options.version === 3`, and the symmetry with `encodeRedisValue({ version })` is called out where the split happens.

One deliberate departure from your sketch: `version` is **required**, not optional. An optional `version` defaulting to 2 reintroduces exactly the failure mode that produced #385 — a protocol-dependent shape that silently picks RESP2 when a call site forgets. Making it required means the per-client constants can no longer be complete options objects on their own, hence the `ClientDecodeOptions` split, which turned out to be worth having on its own terms: it names the line between "this client presents replies differently" (`pushShape`, `narrowBigInt`, `error`) and "this connection negotiated a protocol", and it is what lets the option-level test iterate both clients' constants against both versions.

Cost is a spread at each of the two decode sites and a `version: 2` in `decodeScalarReply`, which is commented as unobservable. Net: one exported helper and one option gone.

src/client-mocks/node-redis-mock.ts
fatal10110: Good catch on the convention — fixed in c20d539, and as you predicted the field itself is gone.

`lastRespVersion` was replaced by the client-level `clientRespVersion` from the combined fix for findings 1 and 2, and that one *is* registered in `teardown()`:

```ts
this.sessions.clear()
// Per-connection state goes back to its handshake default alongside the
// sessions it belongs to.
this.clientRespVersion = 2
```

Which matters slightly more than it did for the old field: `clientRespVersion` is read by `syncProtocol()` *before* a session runs anything, so a stale value here would be observable rather than latent if the facade ever grew a reconnect.

docs/TESTING.md
fatal10110: Fair — I wrote that paragraph and it presents a value I had already measured to be wrong as though it were ground truth. Fixed in c20d539; the paragraph now ends:

> Note the scalars inside those replies are not yet fully faithful at RESP2: scores decode to numbers where real node-redis at RESP2 hands back strings (`['a', '1', 'b', '2']`), and a map reply decodes to an object where RESP2 puts a flat array on the wire — both tracked in #414.

I kept the `['a', 1, 'b', 2]` example rather than replacing it with the real-node-redis value, since the paragraph is documenting what *these clients* return and a reader needs to recognise what they will actually see; the caveat is what makes it honest. Same reasoning as the header comment in the test, which you noted already flags it.

src/client-mocks/node-redis-mock.ts
```diff
@@ -738,9 +845,38 @@ export class NodeRedisMockCluster extends CommandRunner {
...
   protected async run(args: NodeRedisCommandArgument[]): Promise<RedisValue> {
     const session = this.sessionForCommand(args)
-    return runOnSession(session, args, this.closed)
+    await this.syncProtocol(session)
+    const value = await runOnSession(session, args, this.closed)
+    // A HELLO (or RESET) just switched this node's session — adopt it as the
+    // client's version so every other node is brought along on its next
+    // command. Nothing else can move a session's protocol, so this is a no-op
+    // for ordinary commands.
+    this.clientRespVersion = session.protocolVersion
```
fatal10110: **HIGH (blocking, new in `c20d539`) — this line writes the stale version back, so an ordinary command racing a `HELLO` loses the `HELLO` permanently.**

The client-level `clientRespVersion` is the right design and it does fix both blocking findings — I re-ran my probes and got uniform `TUPLES` across a 10-key sweep, in both `Promise.all` orders. But this line adopts `session.protocolVersion` after *every* command, not just after one that moved the protocol. A concurrent command routed to a node that has not synced yet will therefore write its stale version over the one `HELLO` just negotiated.

Sequence, all in one `Promise.all`:

1. `HELLO 3` → keyless → `masters[0]`. `syncProtocol` is a no-op, the command runs, `session0.protocolVersion = 3`.
2. `ZRANGE k0 …` → `master-1`. `syncProtocol` sees `session1.protocolVersion === clientRespVersion === 2` and returns early — step 1 has not landed yet.
3. Step 1 completes: `clientRespVersion = 3`.
4. Step 2 completes: `clientRespVersion = session1.protocolVersion` = **2**. The `HELLO` is gone.

From then on `syncProtocol` faithfully drags every node *back* to RESP2, so the client is wedged on the flat shape while having answered `{"proto":3}` to the caller.

Reproduced on a 3-master mock cluster, 10 trials per ordering:

```
hello-first : HELLO 3 silently lost in 10/10 trials
cmd-first   : HELLO 3 silently lost in 0/10 trials
```

Deterministic, not a rare interleaving. The downgrade direction goes the same way — `HELLO 2` raced against a `ZRANGE` leaves the client on `TUPLES`:

```
HELLO 2 downgrade raced -> k2 after: TUPLES (want FLAT)
```

This is a wider blast radius than the bug it replaced. The old `lastRespVersion` corrupted a single reply and self-corrected on the next command; this corrupts the client's whole protocol state and *persists* until another `HELLO`.

It also answers the question behind declining `runExclusive`. I think declining is right — serializing a deliberately-parallel facade to fix a decode bug is the wrong trade, and I am not asking for it. But "with the version no longer per-node there is nothing left for the interleaving to corrupt" is not quite true: `clientRespVersion` is itself per-client mutable state written after an await, so it is exactly what the interleaving corrupts now. The fix is to make the write conditional rather than to serialize — only adopt the version when *this* command moved *this* session:

```ts
const before = session.protocolVersion
const value = await runOnSession(session, args, this.closed)
if (session.protocolVersion !== before) {
  this.clientRespVersion = session.protocolVersion
}
return value
```

That keeps the parallelism, keeps `RESET` working (it moves the session, so it is adopted), and makes a non-switching command incapable of writing anything. `protocolSwitchedBy()`, already in this file for the MULTI replay, is an alternative predicate, but the before/after compare is stronger because it also catches `RESET` and anything future without a second list to maintain.

Worth noting the new `Promise.all` test at `tests/flat-pairs-resp3.test.ts:347` does not reach this: it awaits `HELLO 3` to completion *before* the `Promise.all`, so both raced commands are ordinary ones. Racing the `HELLO` itself against one ordinary command on another node is the missing case, and it fails today.

src/client-mocks/node-redis-mock.ts
```diff
@@ -591,6 +621,68 @@ export class NodeRedisMockClient extends CommandRunner {
...
+  before: RespVersion,
+  after: RespVersion,
+): RespVersion[] {
+  const versions: RespVersion[] = []
+  let current = before
+  for (const args of queued) {
+    current = protocolSwitchedBy(args) ?? current
+    versions.push(current)
+  }
+  return current === after ? versions : queued.map(() => after)
```
fatal10110: **LOW–MEDIUM (new in `c20d539`) — the mispredict fallback discards per-item versions it had already computed correctly, and is observably wrong when it fires.**

The `replayProtocolSwitches` design is good and the per-item replay is right — I verified it end to end against Redis 8.0.6 and the mock now matches in both directions (see my note in the review body). This is only about the `: queued.map(() => after)` branch.

A mispredict does not mean "every item's version is unknown". It means the *tail* diverged. `versions` is still exactly right up to the point of divergence, and this line throws all of it away in favour of a single version that is known to be wrong for the earlier items.

It is reachable, and when it fires the per-item path would have been right. `MULTI` with two `HELLO`s where the second fails:

```
ZRANGE z 0 -1 WITHSCORES      -> ran at RESP2, must be FLAT
HELLO 3                       -> succeeds, session now v3
ZRANGE z 0 -1 WITHSCORES      -> ran at RESP3, TUPLES
HELLO 2 AUTH u p              -> FAILS (bad auth), session stays v3
ZRANGE z 0 -1 WITHSCORES      -> ran at RESP3, TUPLES
```

The predictor produces `[2, 3, 3, 2, 2]`, so `current` is `2`; the actual version after `EXEC` is `3`. `current !== after`, the fallback fires, and every item decodes under `3`:

```
errorIndexes = [3]
MOCK items   = TUPLES, object, TUPLES, object, TUPLES
EXPECTED     = FLAT,   object, TUPLES, <err>,  TUPLES
post-EXEC ZRANGE = TUPLES   (v3 confirms the HELLO 2 did fail)
```

`versions[0]` was `2` — correct — and the fallback overwrote it with `3`. Without the fallback this case is right.

I could not construct a case where the fallback *improves* on `versions`. For the single-`HELLO` failures it fires on (`HELLO 9`, `HELLO 3 EXTRA`, `HELLO 3 AUTH bad`) it happens to produce the same answer the per-item array would have, because the only prediction was the failed one. So on the evidence it is neutral in the cases it was written for and wrong in the case above.

Suggestion: drop the fallback and return `versions`. A mispredict means an unknown protocol-moving command exists, and in that situation a per-item replay that is right up to the divergence is strictly better than a uniform value that is wrong before it. If you would rather keep a guard, reconcile only the tail — keep `versions[i]` for `i` below the last predicted switch and use `after` from there on — or make the mismatch throw in development, since today it can only be reached by a `HELLO` that fails, and silently degrading is what makes it hard to notice.

Either way this branch has no test, which is worth fixing given it is the one part of the replay that changes behaviour rather than computing it.

fatal10110 (commented):
## Re-review of `c20d539`

I re-ran my own repros against the new head rather than reading the diff, and independently re-verified the real-Redis facts the implementation now depends on.

**Nine of ten prior findings are genuinely fixed; the tenth is deferred to #414 with an honest comment. Two new findings, one blocking** — the cluster fix over-corrected in a way the new tests do not reach. Both are inline.

### Prior findings

| # | Finding | Status |
|---|---|---|
| 1 | HIGH — cluster `lastRespVersion` cross-contamination | **Fixed, verified** |
| 2 | HIGH — `HELLO` reaching only `masters[0]` | **Fixed, verified** |
| 3 | MEDIUM — `exec()` applying one version to every item | **Fixed, verified against real Redis** |
| 4 | MEDIUM — `map`/`map-pairs` at RESP2 | Deferred to #414, comment now accurate — reasonable |
| 5 | MEDIUM — two false comments | Fixed |
| 6 | LOW — dead `flatPairsShape` + test asserting it | Fixed (field deleted outright) |
| 7 | LOW — test coverage gaps | Fixed (cluster, concurrency, downgrade, RESET, MULTI) |
| 8 | LOW — `version` vs derived shape | Adopted; making it **required** is the better call |
| 9 | LOW — `teardown()` registration | Fixed |
| 10 | LOW — docs entrenching the RESP2 scalars | Fixed |

**1 and 2 — verified.** Same 3-master probe as before:

```
[A] sequential k2: [["a",1],["b",2]]      (was TUPLES)
[A] sequential k0: [["a",1],["b",2]]      (was FLAT)
[A] concurrent k2/k0: TUPLES TUPLES       (was FLAT FLAT)
[A] reversed   k0/k2: TUPLES TUPLES       (was TUPLES TUPLES - invented)
[A] sweep k0..k9: all TUPLES
```

**3 — verified against real Redis, both directions.** The "`HELLO` takes effect at its own reply" fact the replay is built on holds, and the mock now matches it exactly:

```
REAL start=RESP2 queued HELLO 3: item0=FLAT   item1=proto3(map)   item2=TUPLES
REAL start=RESP3 queued HELLO 2: item0=TUPLES item1=proto2(array) item2=FLAT
MOCK start=RESP2 queued HELLO 3: item0=FLAT                       item2=TUPLES
MOCK start=RESP3 queued HELLO 2: item0=TUPLES                     item2=FLAT
```

The `RESET`-needs-no-modelling claim also holds — `RESET` is not queueable, so it cannot appear mid-span.

### New findings

- **Inline on `node-redis-mock.ts:860` — HIGH, blocking.** `run()` adopts `session.protocolVersion` after every command, so an ordinary command racing a `HELLO` writes the stale version back and the `HELLO` is lost for the whole client, permanently. 10/10 deterministic. Roughly a 3-line fix, and it does **not** require `runExclusive`.
- **Inline on `node-redis-mock.ts:668` — LOW–MEDIUM.** The `replyVersions` mispredict fallback discards correct per-item versions; reachable, and wrong when it fires.

### On declining `runExclusive`

I think declining is correct and I am not asking for it — serializing a deliberately-parallel facade to fix a decode bug is the wrong trade. But the stated reason, "with the version no longer per-node there is nothing left for the interleaving to corrupt", does not hold: `clientRespVersion` is per-client mutable state written after an await, which is precisely what the interleaving now corrupts. The right answer is a conditional write, not a lock. Detail inline.

### Lazy propagation — scrutinised, sound

Aside from the write-back bug above, the lazy `syncProtocol` design holds up:

- A session syncs on its next command, **before** it produces a reply, so there is no window in which a live session answers under the wrong protocol. The two-sessions-disagree state is real but unobservable, exactly as argued.
- There is no MOVED-redirect path in this facade (`sessionForCommand` routes by computed slot and lets the node's `ClusterPolicy` surface the error), so no session is created mid-command. The lazy form's advantage over an eager fan-out — covering sessions created after the switch — is real for `sessionFor()` growth.
- Replaying via a real `HELLO` rather than `setProtocolVersion` is the better choice for fidelity. It does put an extra command in that node's stream, which would be visible to a per-node `MONITOR` or command stats; not reachable through this facade today (`MONITOR`/`SUBSCRIBE` are keyless and land on `masters[0]`, the one node that never needs syncing), so no action — noting it because #410 is touching the MONITOR feed.
- Sub-case checked: a node session wedged in subscriber mode would make `syncProtocol`'s `HELLO` throw and fail an unrelated command. Not reachable (the only way to wedge a session is keyless `SUBSCRIBE` → `masters[0]`). Real Redis rejects `HELLO` in subscriber mode with the identical wording, so the model is faithful either way:
  ```
  -ERR Can't execute 'hello': only (P|S)SUBSCRIBE / (P|S)UNSUBSCRIBE / PING / QUIT / RESET are allowed in this context
  ```

### `version` required — no caller broken

`tsc --noEmit` clean. Three `decodeRedisValue` call sites, all updated. Neither `DecodeRedisValueOptions`/`ClientDecodeOptions` nor the two option constants are re-exported from `src/index.ts`, so requiring `version` is not a public API break. Agreed that an optional default of `2` would reinvite #385; the `ClientDecodeOptions` split is a real improvement and names a line that was previously blurred.

### Tests

- `npm test` equivalent: **355/355 pass** — matches your number.
- Mock integration: **1113/1114**, one failure — `HPERSIST, HTTL and HPTTL report and clear hash field TTLs`.

That failure is **not** yours, and it is almost certainly the "one unexplained failure that did not reproduce" from the first round. It is a pre-existing 20-millisecond timing race, which I root-caused last round and which reproduced here again unprompted. `hpExpire(key, 'soon', 20)` followed by more round-trips than 20 ms allows under load: the field expires and `HTTL` returns `-2`, failing `assert.ok(seconds[2] >= 0)` at `tests-integration/node-redis/hash/field-persist.test.ts:59`. Its sibling is `hIncrBy` returning `1` instead of `2` at `field-expire.test.ts:202` for the same reason. Deterministic with a 30 ms stall:

```
immediately HTTL soon = [ 0 ]   HINCRBY ctr = 2
after 30ms  HTTL soon = [ -2 ]  HINCRBY ctr = 1
```

Last round: 0 failures in 6 unloaded runs, 10 failures across 12 runs under CPU contention, only ever these two tests. Four files affected (ioredis and node-redis mirrors of each). Worth its own issue — raise the TTL to ~500 ms. No file under `tests-integration/` touches the socketless clients, so this PR's code is never exercised by that suite.

### Verdict

**Don't ship yet — one blocking regression, and it is small.** Everything else in `c20d539` is a clear improvement, the real-Redis grounding for the MULTI replay is solid, and taking the `version` refactor now rather than deferring was the right call. The `clientRespVersion` write-back is a conditional away from correct; with that and a test that races `HELLO` against an ordinary command, I would be happy with this.

src/client-mocks/node-redis-mock.ts
fatal10110: Confirmed and fixed in 9bf3a26. You are right on all three counts, including the correction to my "nothing left to corrupt" argument — `clientRespVersion` is post-await mutable state, so it is precisely what the interleaving corrupts now. I should have noticed that when I wrote the line.

Reproduced your numbers before changing anything, 10 trials per ordering on a 3-master mock:

```
hello-first : HELLO 3 lost in 10/10 trials
cmd-first   : HELLO 3 lost in 0/10 trials
HELLO 2 downgrade raced -> k2 after: TUPLES (want FLAT)
```

and after:

```
hello-first : HELLO 3 lost in 0/10 trials
cmd-first   : HELLO 3 lost in 0/10 trials
HELLO 2 downgrade raced -> k2 after: FLAT (want FLAT)
```

I checked the before/after compare rather than taking it verbatim, and it is the right shape — with one ordering detail worth stating, since it is the part that could have gone wrong: `before` has to be captured **after** `syncProtocol()`, not before it. `syncProtocol` moves the session itself, so capturing earlier would make every first-command-after-a-switch look like a protocol-moving command and re-adopt a version the client already holds. Harmless today, but it would have put the unconditional write back in a subtler form.

I took the before/after compare over `protocolSwitchedBy()` for the reason you give — it catches `RESET` and anything future without a second list to maintain — and the `RESET` path is covered by an existing test.

New test `a HELLO raced against a command on another node is not lost` puts the `HELLO` **inside** the `Promise.all` and runs both orderings, then sweeps every key; `a HELLO 2 raced …` does the downgrade direction. Both are red on c20d539 and green here — I verified that by checking out the old file rather than assuming:

```
c20d539 + new tests : 15 tests, 12 pass, 3 fail
  not ok - a failed HELLO inside MULTI does not move the shape
  not ok - a HELLO raced against a command on another node is not lost
  not ok - a HELLO 2 raced against a command on another node is not lost
9bf3a26             : 15 tests, 15 pass, 0 fail
```

One residual I am leaving alone and want to be explicit about: a reply produced by a command that was **in flight** while the `HELLO` landed is still decoded under whichever version the client holds when `sendCommand` reads it. That is inherent to racing a protocol switch — a real node-redis connection has the same ambiguity, its parser being per-connection — and the important part is that it is now confined to that one reply instead of persisting in the client's state.

src/client-mocks/node-redis-mock.ts
fatal10110: Agreed that the fallback is wrong, and it is gone in 9bf3a26 — but not by deleting it, because deleting it alone introduces the mirror bug. Details, since you asked me to validate rather than take it.

Your case reproduces exactly. Before:

```
errorIndexes: [3]
items: TUPLES, object, TUPLES, object, TUPLES     <- item0 ran at RESP2
```

Where I think the "it is neutral in the cases it was written for" half does not hold is when the failed `HELLO` is **not** last. Put it at the head and the fallback is the only thing saving the result:

```
MULTI
  HELLO 3 AUTH u p            -> WRONGPASS, session stays v2
  ZRANGE z 0 -1 WITHSCORES    -> ran at RESP2, must be FLAT
EXEC
```

The predictor gives `[3, 3]`; actual after is `2`. With the fallback: `[2, 2]` → `object, FLAT`, correct. Without it: `[3, 3]` → item1 decodes as TUPLES, wrong. Your three probes (`HELLO 9`, `HELLO 3 EXTRA`, `HELLO 3 AUTH bad`) all produce the same answer either way for the reason you give — the mispredicted index is the error item itself, which never reaches `decodeReply` — but that stops being true as soon as anything is queued behind the failure. On this server all three fail distinctly, so the case is easy to build: `WRONGPASS invalid username-password pair or user is disabled.`, `NOPROTO unsupported protocol version`, `ERR syntax error`.

So the real defect is upstream of the fallback: the predictor was **guessing** that a syntactically-plausible `HELLO` succeeded. It does not have to guess — a queued command that failed has an error in its own slot of `EXEC`'s array, and that is exactly when the protocol did not move. `replayProtocolSwitches` now takes the EXEC reply and skips any `HELLO` whose item came back an error:

```ts
if (items[index]?.kind !== 'error') {
  current = protocolSwitchedBy(args) ?? current
}
```

That makes the replay exact for everything this facade models, which is what lets the fallback go: with failed `HELLO`s accounted for, the only thing left that could diverge is a protocol-moving command the facade does not know about, and there your argument is decisive — a replay right up to the divergence beats a uniform value wrong before it. I dropped it rather than reconciling the tail, since with an unknown switcher the divergence point is unknown too, so "last predicted switch" is not a sound boundary.

Both shapes now come out right:

```
reviewer's case : FLAT, object, TUPLES, <err>, TUPLES      (was TUPLES, …)
failed HELLO first : object, FLAT                          (unchanged, still right)
```

Covered by `a failed HELLO inside MULTI does not move the shape`, which asserts both orderings and is red on c20d539.

src/client-mocks/node-redis-mock.ts
```diff
@@ -738,9 +853,45 @@ export class NodeRedisMockCluster extends CommandRunner {
...
+    const value = await runOnSession(session, args, this.closed)
+    // Adopt the version only when *this* command moved *this* session — a
+    // HELLO or a RESET. Writing it unconditionally would let an ordinary
+    // command racing a HELLO put its own stale version back: commands here run
+    // concurrently, so a command that started before the HELLO landed sees a
+    // session still on the old protocol and would undo the switch for the
+    // whole client. Comparing before/after instead of matching on the command
+    // name keeps RESET (and anything future) working without a second list of
+    // protocol-moving commands to maintain.
+    if (session.protocolVersion !== before) {
```
fatal10110, line 887: **Verified fixed — and the ordering point you raised is real, my suggested snippet was wrong.**

NEW-1 no longer reproduces. Same probe as before, 10 trials per case:

```
hello-first : HELLO 3 lost 0/10   (was 10/10)
cmd-first   : HELLO 3 lost 0/10
downgrade   : HELLO 2 lost 0/10   (was 10/10)
```

On `before` being captured *after* `syncProtocol()` — you are right and I was wrong. I tested it rather than reasoning about it: I took this file, moved the `const before` one line up so it is captured before the sync, and re-ran the downgrade race.

```
before captured BEFORE syncProtocol : HELLO 2 lost 10/10
before captured AFTER  syncProtocol : HELLO 2 lost  0/10   (as shipped)
```

The mechanism is worth writing down because it is not obvious from the diff. With `before` captured too early, a node whose only protocol change came from `syncProtocol` itself reports `protocolVersion !== before` and writes back the version it was *synced to*. If a `HELLO` in the other direction landed in between, that write reverts it — the identical failure mode as NEW-1, just sourced from the sync rather than from a stale ordinary command. Capturing after the sync makes a pure sync invisible to the adopt-check, which is exactly what you want: only a change caused by *this* command counts.

Comparing before/after rather than matching command names is also the better of the two options I offered, for the reason you give — `RESET` comes along for free and there is no second list to keep in sync.

No new issue here.

src/client-mocks/node-redis-mock.ts
```diff
@@ -591,6 +620,77 @@ export class NodeRedisMockClient extends CommandRunner {
...
+  queued: readonly NodeRedisCommandArgument[][],
+  before: RespVersion,
+  result: RedisValue,
+): RespVersion[] {
+  const items =
+    result.kind === 'array' || result.kind === 'set' ? result.items : []
+  const versions: RespVersion[] = []
+  let current = before
+  queued.forEach((args, index) => {
+    if (items[index]?.kind !== 'error') {
```
fatal10110, line 671: **Verified fixed, and you were right to override my recommendation — I checked your mirror case rather than taking it.**

I built my own suggestion as a variant to test it: this file with the `items[index]?.kind !== 'error'` guard removed and `return versions` kept — i.e. "delete the fallback, keep the per-item replay", which is what I asked for. Then I ran both your case and mine against it and against what shipped.

```
                              MY RECOMMENDATION       SHIPPED 9bf3a26        CORRECT
MULTI; HELLO 3 AUTH u p; Z    [object, TUPLES]  WRONG [object, FLAT]         [object, FLAT]
Z; HELLO 3; Z; HELLO 2        [FLAT,object,TUPLES,    [FLAT,object,TUPLES,   [FLAT,object,TUPLES,
   AUTH u p; Z                 object,FLAT]     WRONG  object,TUPLES]         object,TUPLES]
```

So your mirror case is real, and my recommendation would have been a regression — it gets *both* cases wrong, including the one I raised it for. Your reasoning about why is correct: "neutral in the cases it fires on" only held while the mispredicted index happened to be the error item, which never reaches `decodeReply`, and it stops holding the moment anything is queued behind the failure. Fixing it upstream so the replay is exact, and only then dropping the fallback, is strictly better than either of the things I proposed. Agreed too that "reconcile the tail" was unsound — with an unknown switcher you do not know where the divergence starts.

**On the inference this now rests on** — "an error in slot *i* means command *i* did not move the protocol". I tried to break it and could not. Every `HELLO` failure mode I could construct leaves the protocol where it was, on real Redis 8.0.6 and here, including the ones that fail *after* the protover has been parsed:

```
real :6399, each on a fresh RESP2 connection, then ZRANGE ... WITHSCORES
  HELLO 3 AUTH baduser badpass  -> -WRONGPASS ...                          reply *2 $1 a $1 1   (RESP2)
  HELLO 3 SETNAME "bad name"    -> -ERR Syntax error in HELLO option 'name' reply *2 $1 a $1 1   (RESP2)
  HELLO 3 EXTRA                 -> -ERR Syntax error in HELLO option 'EXTRA' reply *2 $1 a $1 1  (RESP2)
  HELLO 9                       -> -NOPROTO unsupported protocol version   reply *2 $1 a $1 1   (RESP2)
  HELLO 3 (control, succeeds)   -> %7 ... proto :3                         (RESP3 map)
```

Unauthenticated `HELLO 3` against the auth server behaves the same — `-NOAUTH`, and the connection is still RESP2 after a successful `AUTH`. This repo matches on all four: protocol moved `no` in every case.

I also probed the index-alignment half of the inference, since `items[index]` assumes EXEC's array is 1:1 with the queue:

```
unknown cmd queued          -> ErrorReply EXECABORT      (replay result unused, exec() returns before it)
runtime error mid-queue     -> [FLAT, <err>, object, TUPLES]   HELLO not skipped by someone else's error
two failing HELLOs then Z   -> [object, object, FLAT]
HELLO 3 as the last item    -> [FLAT, object]
```

All correct. The one case that would break it — a command that moves the protocol *and then* errors — does not exist in Redis today for `HELLO`, and any future one would fail `protocolSwitchedBy` anyway. No new issue.

fatal10110 (commented):
## Third-pass review of `9bf3a26`

**Both new findings are genuinely fixed, and I found nothing further. Ship it.**

I re-ran my own repros rather than taking the claims, and on the one point where you overrode my recommendation I built my recommendation as a working variant and tested it against yours. You were right. Details are inline on the two lines that changed.

### NEW-1 — fixed

```
hello-first : HELLO 3 lost 0/10   (was 10/10)
cmd-first   : HELLO 3 lost 0/10
downgrade   : HELLO 2 lost 0/10   (was 10/10)
```

The `before`-after-`syncProtocol` ordering you flagged is real and material — inverting it loses `HELLO 2` in 10/10 trials. My original snippet did not cover it. Inline.

### NEW-2 — fixed, and my recommendation was wrong

Your mirror case reproduces. I built "delete the fallback, keep the per-item replay" as an actual variant of this file: it returns `[object, TUPLES]` where the correct answer is `[object, FLAT]`, and it also gets wrong the very case I raised the finding for. The shipped error-skip gets both right. Inline, with the ground-truth for the inference it now rests on.

### Scrutiny that came back clean

- **The error-skip inference.** Four `HELLO` failure modes on real Redis 8.0.6 — `WRONGPASS`, bad `SETNAME`, trailing garbage, `NOPROTO` — all leave the connection on RESP2, including the ones that fail after the protover is parsed. Unauthenticated `HELLO 3` on the auth server likewise. This repo matches all five. Index alignment holds across `EXECABORT`, a mid-queue runtime error, two failing `HELLO`s, and `HELLO` as the last item.
- **Regression tests bite, and pin the right things.** I repeated your source-swap independently — `c20d539`'s `node-redis-mock.ts` under `9bf3a26`'s tests — and got your numbers: **12 pass / 3 fail**, **15/15** on the shipped head. The three that go red are exactly the three that should:
  ```
  not ok - a failed HELLO inside MULTI does not move the shape          (NEW-2)
  not ok - a HELLO raced against a command on another node is not lost  (NEW-1, upgrade)
  not ok - a HELLO 2 raced against a command on another node is not lost (NEW-1, downgrade)
  ```
  The MULTI test pins both competing directions — the tail-failure case I raised and the head-failure case you raised — which is the right shape for a test guarding a decision that went back and forth.

### The disclosed in-flight residual — I could not substantiate it

Worth reporting what I found, because it is less bad than your disclosure suggests. I tried to make it produce a wrong answer and could not: `Promise.all([HELLO 3, ...5 reads])`, 40 trials against the cluster facade and 40 against real `node-redis@6`/Redis 8.0.6, gave byte-identical, completely stable results — all five reads under the new protocol in every single trial, mock and real alike. Adding a 30-cycle event-loop warm-up did not shake it either.

So on the evidence I have, the mock and real node-redis agree here, and I cannot show a divergence. Two honest caveats: I did see a non-reproducible 3/2 split in one early exploratory run before I controlled the setup, and I could not recreate it in 80 subsequent trials — I am recording it rather than asserting it. And strictly the bound is the *set of commands in flight* when the switch lands, not one reply, so the wording is a little tighter than the thing it describes. Neither changes the disposition: the residual is inherent to racing a protocol switch, it is disclosed, and it is not worth engineering against.

### Tests

- Unit: **358/358** — matches your number.
- Mock integration: **1113/1114**. The one failure is `hash writes clear or preserve field TTLs like Redis`, which is the *other* sibling of the pre-existing 20 ms `HPEXPIRE` race I root-caused in the first pass — not yours, and not this PR's code (nothing under `tests-integration/` touches the socketless clients). It passes **3/3 in isolation** and only fails under full-suite load, which is the signature. Still worth its own issue: `hpExpire(key, …, 20)` followed by more round-trips than 20 ms allows, across four files (`field-expire` and `field-persist`, ioredis and node-redis mirrors). Raising the TTL to ~500 ms fixes all four.
- `tsc --noEmit` clean.

### Verdict

**Ship.** Three passes in, the protocol story in this facade is now the best-evidenced part of the file: a client-level version with a conditional adopt, an exact per-item MULTI replay grounded in real-server behaviour rather than inference, and regression tests that actually go red on the previous commit. The two things I pushed back on in round two — declining `runExclusive` and keeping some form of fallback handling — you resolved better than I proposed in both cases.

src/client-mocks/node-redis-mock.ts
fatal10110: You are right, and my wording above was wrong — corrected in 48e2112.

"Confined to that one reply" describes a serialized client. This one is not: commands on different nodes run concurrently by design, and each reads `clientRespVersion` only when its *own* reply is decoded. So every command already in flight when the `HELLO` lands can come back in the other protocol's shape — the bound is the in-flight set at that moment, not one reply. Thank you for measuring it rather than letting the claim stand; it is exactly the kind of sentence that gets quoted later as if it had been established.

The `clientRespVersion` doc now states it properly: what the bound actually is, that it is inherent rather than an artefact of this facade (real node-redis has the same ambiguity, its parser belonging to the connection rather than to the individual reply), and what the conditional write in `run()` is therefore for — stopping the window outliving the switch, not closing it.

While in there I also wrote down the `before` ordering constraint you independently verified by inverting it. It was load-bearing and existed only in a PR thread:

```ts
// Captured *after* syncProtocol, which moves the session itself: reading
// it earlier would make the first command on a newly-synced node look like
// a protocol-moving one and re-adopt a version the client already holds.
const before = session.protocolVersion
```

Comment-only commit, no behaviour change: `npm test` 358/358, `test:integration:mock` 1114/1114, `tests/flat-pairs-resp3.test.ts` 15/15, build and lint clean.

Leaving the RESP2 `map`/`map-pairs` and score-type divergences to #414 as agreed.

fatal10110 (commented):
## Final pass on `48e2112`

**Comment-only confirmed and accurate. Ship.**

I verified the diff is genuinely non-executable rather than taking it: every changed line in `src/client-mocks/node-redis-mock.ts` is a comment line (15 added, 1 replaced), no other file is touched, `tsc --noEmit` is clean and `tests/flat-pairs-resp3.test.ts` is 15/15 on this head. Skipping a real-backend run this pass is sound for exactly that reason — comments cannot change runtime behaviour, and the previous head's targeted real runs still describe this one.

**The `clientRespVersion` doc now states the real bound.** "Bounded by the set of commands in flight at that moment, not by one" is the correction I raised, and the tightened sentence above it — "two concurrent commands *on different nodes*" — is now right rather than merely less wrong: that is precisely the case the client-level read fixes, and it no longer implies anything about a `HELLO` in flight. Saying the conditional write "stops that window outliving the switch, not closes it" is an accurate description of what it does.

**On "real node-redis has the same ambiguity":** I could not falsify it. Across 40 trials against the cluster facade and 40 against real `node-redis@6`/Redis 8.0.6, both gave identical, stable results in every trial. The mechanism argument also holds — node-redis's decoder state belongs to the connection, and a raw `sendCommand(['HELLO', …])` is not its supported way to switch protocols in the first place (that is the `RESP` constructor option). Fair as stated.

**One nit on the `before` comment**, inline: the ordering constraint is stated correctly but its consequence is described as the benign outcome rather than the one that makes it load-bearing. Non-blocking.

**No stale copy of the old claim elsewhere** — confirmed by grep across `docs/`, `src/` and `tests/`; `docs/TESTING.md` describes the clients following a `HELLO` and makes no concurrency or single-reply claim, so there was nothing to correct there.

Four passes in, this is in good shape: the two blocking cluster bugs, the MULTI replay, and now the comments that describe them all landed with real-server evidence behind each one.

src/client-mocks/node-redis-mock.ts
```diff
@@ -738,9 +864,48 @@ export class NodeRedisMockCluster extends CommandRunner {
...
+    return this.clientRespVersion
+  }
+
   protected async run(args: NodeRedisCommandArgument[]): Promise<RedisValue> {
     const session = this.sessionForCommand(args)
-    return runOnSession(session, args, this.closed)
+    await this.syncProtocol(session)
+    // Captured *after* syncProtocol, which moves the session itself: reading
+    // it earlier would make the first command on a newly-synced node look like
+    // a protocol-moving one and re-adopt a version the client already holds.
```
fatal10110, line 876: **The bound is right; this sentence describes the harmless half of what it prevents.**

The ordering constraint is worth recording here — it was load-bearing and lived only in a PR thread, so putting it in the code is the right call. The mechanism is stated correctly: `syncProtocol` moves the session, so capturing earlier makes a pure sync indistinguishable from a real switch.

The consequence is understated, though. "Re-adopt a version the client already holds" is the benign outcome — a no-op write. What makes the ordering load-bearing is the case where the client has *moved on*: if a `HELLO` in the other direction lands between the sync and this write, re-adopting reverts it, and the switch is lost for the whole client. That is the same failure as the one the conditional below is there to prevent, arriving from the sync instead of from a stale ordinary command.

I proved it by inversion when checking the previous commit — moving this capture one line up, above `syncProtocol`, loses a `HELLO 2` in **10/10** trials, against 0/10 as written:

```
before captured BEFORE syncProtocol : HELLO 2 lost 10/10
before captured AFTER  syncProtocol : HELLO 2 lost  0/10   (as shipped)
```

As written, a reader who trusts the comment would conclude the ordering is a tidiness matter and that moving it is safe. Something closer to "…and write that version back, clobbering a concurrent `HELLO` in the other direction — the same failure the conditional below guards against" would carry the actual stake.

Comment-only, non-blocking — the code is correct either way.

Issue #384, opened 2026-09-21
Error text: subscribed-mode error says 'config' not 'config|get'; unknown-command hex-dumps binary names


fatal10110:
Two small error-text divergences found while reviewing PR #377 against real redis-server 7.2.1. **Both pre-existing on `main`.**

## 1. Subscribed-mode error names the parent command, not the subcommand
`src/core/execution-policies/subscribed-policy.ts:12` builds the message from `plan.definition.name`, so a container command reports only its parent name.

```
In RESP2 subscribed mode, sending: CONFIG GET maxmemory
REAL 7.2.1: ERR Can't execute 'config|get': only (P|S)SUBSCRIBE / ...
MOCK:       ERR Can't execute 'config': only (P|S)SUBSCRIBE / ...
```

Same root cause as the keyspace-notification subcommand-naming issue filed alongside this one — `definition.name` cannot express a dispatched subcommand.

## 2. Unknown-command error mishandles binary bytes
`src/core/redis-error.ts` (`UnknownRedisCommandError`) renders a binary command name by hex-dumping it, and uses a different truncation model than Redis. Real Redis echoes the raw bytes and truncates args at a fixed count/length.

Worth checking the exact real behavior for: a command name containing non-UTF8 bytes, a very long name, and more than 20 args.

## Acceptance
Both messages match real Redis byte-for-byte, pinned by raw-TCP tests (`tests-integration/raw-tcp/`), which is where exact wire text belongs per `AGENTS.md`.

Issue #359, opened 2026-09-21
refactor: drop unused afterExecute/onStream policy hooks and dedupe sync/async executor paths


fatal10110:
## Problem
`ExecutionPolicy` defines three hooks — `beforeExecute`, `afterExecute`, `onStream` (`src/core/execution-policies/index.ts`). All 4 shipped policies (auth, cluster, subscribed-mode, transaction) implement **only** `beforeExecute`. `afterExecute` / `onStream` are exercised only by two tests (`tests/core-command-executor.test.ts:350, 459`).

Supporting them forces a lot of machinery in `src/core/command-executor.ts`:
- `ensureNonThenableStream` (re-wraps thenable streams so `await` doesn't unwrap them)
- `assertSyncPolicyResult` + `MaybePromise` (rejects async hooks on the Lua path)
- the two near-identical pipelines `executePlanInternal` (l.188-246) and `executePlanSyncInternal` (l.270-323), which differ only in `await`
- `executeRawWithPlan` has a single internal caller (`executeRaw`)
- `createMonitorDeferredContext` (l.381-410) hand-copies every `ctx` field just to override `monitor`

## Proposal
1. Remove `afterExecute` and `onStream` from `ExecutionPolicy` (YAGNI; re-add when a real policy needs them). Drop the helpers above.
2. With only sync-or-async `beforeExecute` left, factor the shared parts of the async and sync paths (`beforeExecute` loop, transaction-dirty handling, error→`RedisResult` mapping, notify-command save/restore) into small helpers so each path is ~15 lines.
3. Inline `executeRawWithPlan` into `executeRaw`.
4. Replace the manual ctx copy with `Object.create(ctx, { monitor: { value: … } })` (keeps the live `db` getter that #94 depends on) — or add an explicit `withMonitor(ctx, monitor)` helper.

## Compatibility
`ExecutionPolicy` is exported via `js-redis-server/core` → technically breaking for anyone with custom policies using the removed hooks. Land in a minor with a changelog note (package is 0.x).

## Acceptance
- Update the two tests that use the removed hooks (delete or rewrite as `beforeExecute` tests).
- All other unit/integration tests pass; `command-executor.ts` shrinks by ~120 lines.

PR #409, opened 2026-09-21
fix: enforce proto-max-bulk-len in SETRANGE and APPEND (#383)

fatal10110:
## Summary

`SETRANGE` never checked `proto-max-bulk-len`, so the issue's repro
(`SETRANGE k 536870911 xx`) replied `:536870913` and allocated ~512MB inside the
test process, where real Redis replies
`-ERR string exceeds maximum allowed size (proto-max-bulk-len)` and creates no
key. `APPEND` had the same gap.

Investigating the two follow-up questions in the issue:

- **`CONFIG GET/SET proto-max-bulk-len`** was modelled, but only as an inert
  entry in the defaults map — `CONFIG SET` stored any string and nothing read
  it. It is now a behaviour-driving field on `RedisServerState` (the same
  pattern `notify-keyspace-events` already uses): `CONFIG GET` reports the live
  value, `CONFIG SET` parses Redis memory values (`1048576`, `1mb`, `512MB`, …)
  and enforces Redis' own `[1048576, 9223372036854775807]` bounds with both real
  failure messages, and `APPEND`/`SETRANGE` check against it.
- **`APPEND`** is affected and is now guarded. Matching Redis
  (`checkStringLength` in `t_string.c`), the check only applies when the key
  already holds a value — `APPEND` to a missing key is never size-checked.
- **`GETRANGE`** is **not** affected: it only reads an existing value, so Redis
  applies no limit there. Verified against 7.2.1 and 8.0.6 that
  `GETRANGE k 0 536870911` and `GETRANGE k 0 9223372036854775807` both just
  return the value.

Two adjacent int64 divergences found while ground-truthing are fixed in the same
commit, because they produce the *wrong error* for exactly the inputs this issue
is about:

- `SETRANGE`'s offset is an int64 in Redis but was parsed as a safe integer, so
  `SETRANGE k 9223372036854775807 xx` answered
  `ERR value is not an integer or out of range` instead of the size error.
- `GETRANGE`'s `start`/`end` are int64 and clamped, but were likewise parsed as
  safe integers, so `GETRANGE k 0 9223372036854775807` errored instead of
  returning the value.

Ordering is preserved to match Redis: `WRONGTYPE` and a negative offset are both
reported before the size check, and an empty `SETRANGE` value short-circuits
before it (so `SETRANGE k 536870911 ""` still replies `0` and creates no key).

No compatibility gate was added: the error wording and the CONFIG bounds are
identical on every profile this repo supports (spot-checked on redis 7.2.1 and
[...]

PR #410, opened 2026-09-21
fix: microsecond MONITOR timestamps and real CONFIG subcommand wording (#388)

### LINKED ISSUE #388: CONFIG unknown-subcommand wording diverges; MONITOR timestamps millisecond-quantized where real Redis is microsecond
Two error/behavior divergences found while re-reviewing PRs #377 and #378 against real Redis 7.2.1. **Both pre-existing on `main`.**

## 1. `CONFIG <unknown-subcommand>` wording
The mock's reply for an unrecognized `CONFIG` subcommand does not match real Redis. Worth capturing the exact real text (including the `Try CONFIG HELP.` suffix and quoting style) and pinning it in `tests-integration/raw-tcp/`, which is where exact wire text belongs per `AGENTS.md`.

Related and already filed as #384: the subscribed-mode rejection says `'config'` where real Redis says `'config|get'`. Same root cause — `plan.definition.name` cannot express a dispatched subcommand — so these may fix together.

## 2. MONITOR timestamps are millisecond-quantized
Real Redis emits MONITOR lines with **microsecond** precision (`<unix-seconds>.<6 digits>`). The mock derives its timestamp from `Date.now()`, so the last three digits are always zero.

**Failure scenario:** a test asserting sub-millisecond ordering or parsing the fractional field for precision sees an implausible pattern; anything diffing mock MONITOR output against real capture shows a systematic difference on every line.

`src/state/monitor-feed.ts` / `src/core/command-executor.ts` (`timestampMs`) are the relevant paths. `process.hrtime.bigint()` offset against a captured wall-clock origin would give real sub-millisecond resolution if that is judged worth it; otherwise the field could at least be documented as quantized.

## Acceptance
Both behaviors match real Redis byte-for-byte, pinned by raw-TCP tests — or, for the MONITOR case, an explicit documented decision to keep millisecond resolution.

fatal10110:
## Summary

Both halves of #388, both pre-existing on `main`.

**MONITOR timestamps (part 2).** Real Redis stamps every `MONITOR` line from `gettimeofday()`, so the six fractional digits carry genuine microsecond resolution. The mock derived its timestamp from `Date.now()`, which looks the same on the wire (still six digits) but quantizes every line to a whole millisecond — the last three digits were always `000`, a systematic difference on *every* line when diffing against a real capture.

The issue left the call open: implement real sub-millisecond resolution, or document the field as quantized. **Implemented it**, because the cost is one small module-local helper and the "document it" alternative leaves an assertion that can never be written (you cannot pin sub-millisecond ordering or a fractional field against a clock that does not have one), while the wire format already promises the precision. `monitorTimestampMicros()` anchors a coarse `Date.now()` epoch against `process.hrtime.bigint()` and recaptures the anchor if the two drift more than a second apart, so it keeps nanosecond-derived resolution without slowly diverging from wall-clock time across NTP steps or suspend/resume. Captured from `redis-server` 7.2.1 for reference:

```
1790019912.925375 [0 172.17.0.1:50762] "ECHO" "a388:0"
1790019912.925532 [0 172.17.0.1:50762] "ECHO" "a388:1"
1790019912.925538 [0 172.17.0.1:50762] "ECHO" "a388:2"
```

`RedisMonitorCommandEvent.timestampMs` becomes `timestampMicros` (a plain `number` is exact for microsecond epoch values until the year 2255), and the seconds/fraction split moves into `formatMonitorTimestamp()` — integer arithmetic rather than `(micros / 1e6).toFixed(6)`, which can round the last digit once the value passes a double's ~15 significant digits. The touch in `src/core/command-executor.ts` is deliberately one line plus its import, since that file is contended by the pending refactor queue (#359/#365/#366/#367).

**CONFIG unknown-subcommand wording (part 1).** Fixed here rather than deferred to #384. #384's root cause is that `plan.definition.name` cannot express a *dispatched* subcommand, which is a general mechanism problem. This divergence is not that: the wrong text was a literal template string inside `configCommand.execute`, one branch away from the correct one, and fixing it introduces no subcommand-naming mechanism. Captured from real Redis:

| | reply |
|---|---|
| real 7.2.1 / 8.0.6 | `ERR unknown subcommand 'BOGUS'. Try CONFIG HELP.` |
| mock (before) | `ERR Unknown CONFIG subcommand or wrong number of arguments for 'BOGUS'. Try CONFIG HELP.` |

The subcommand is echoed with the casing the client sent. Identical on 7.2.1 and 8.0.6, so **no compatibility-profile gate applies** — this was never version-specific, it was always wrong.

The neighbouring arity errors (`CONFIG GET` -> `wrong number of arguments for 'config|get' command`, `CONFIG` -> `... for 'config' command`) already match real Redis and are now pinned as regression coverage, but nothing here changes how a subcommand gets *named* — the subscribed-mode `'config'` vs `'config|get'` case in #384 is untouched and still needs that mechanism.

Closes #388

## Test plan

- [x] `tests-integration/raw-tcp/command-errors-config.test.ts` (new, 4 tests) — exact bytes of the unknown-subcommand reply, casing preservation, trailing args, and the container's wrong-arity reply. Red on mock before the fix (3 of 4 failed on the wording), green against real Redis from the start.
- [x] `tests-integration/raw-tcp/monitor-protocol.test.ts` — new `#388` test pipelines 24 `PING <marker>` commands in one write so they land inside the same millisecond, then asserts at least one monitor line has a non-`000` microsecond field, that all of them sit near wall-clock time, and that they are non-decreasing. Verified red by temporarily restoring `Date.now() * 1000`, which failed with *all 24 monitor timestamps were millisecond-quantized: 1790020196.836000, 1790020196.836000, ...*; green against real Redis before and after.
- [x] `tests/state-core.test.ts` — two unit tests for the zero-padding / boundary cases of `formatMonitorTimestamp` and the wall-clock-tracking + sub-millisecond behavior of `monitorTimestampMicros`. Unit rather than integration because a live feed cannot be made to emit a chosen microsecond value.
- [x] `npm test` — 344 pass, 0 fail.
- [x] `npm run test:integration:mock` — 1119 pass, 0 fail.
- [x] `npm run test:integration:real` — 1027 pass, 0 fail (68 cancelled; the real backend was shared with three other agents during this run). A second full real run showed a handful of unrelated failures (hash field-TTL, cluster `READONLY` replica routing, `BLPOP` thundering herd, plus cascading `cancelledByParent`); re-running those suites targeted gave 41 pass / 0 fail, confirming contention on the shared server rather than a regression. None of them touch MONITOR or CONFIG.
- [x] `npm run build`, `npm run lint` — clean.
- [x] Docs checked (`README.md`, `docs/COMMANDS.md`, `docs/ARCHITECTURE.md`, `docs/API.md`) — none quote the old CONFIG error string, name the `timestampMs` field, or claim millisecond MONITOR precision, so nothing went stale.


src/commands/config.ts
```diff
@@ -205,8 +205,13 @@ export const configCommand = defineCommand({
...
     }
 
+    // Matches the generic unknown-subcommand reply real Redis sends for a
+    // container command, echoing the subcommand with the casing the client
+    // sent. Captured byte-for-byte from redis-server 7.2.1 and 8.0.6 (identical
+    // on both, so no compatibility gate applies) and pinned by
+    // tests-integration/raw-tcp/command-errors-config.test.ts (#388).
     throw new RedisCommandError(
-      `Unknown CONFIG subcommand or wrong number of arguments for '${args.subcommand}'. Try CONFIG HELP.`,
+      `unknown subcommand '${args.subcommand}'. Try CONFIG HELP.`,
```
fatal10110: **1. HIGH (blocking) — this wording IS version-specific, and the comment above it asserts the opposite.**

Real Redis changed this template at 7.0.0. Reproduced on a `redis:6.2` container (`redis_version:6.2.24`):

```
6.2.24  CONFIG BOGUS -> -ERR Unknown subcommand or wrong number of arguments for 'BOGUS'. Try CONFIG HELP.
6.2.24  CONFIG GET   -> -ERR Unknown subcommand or wrong number of arguments for 'GET'. Try CONFIG HELP.
7.0.15  CONFIG BOGUS -> -ERR unknown subcommand 'BOGUS'. Try CONFIG HELP.
7.0.15  CONFIG GET   -> -ERR wrong number of arguments for 'config|get' command
```

So the old mock string was exactly right for 6.2, and the premise that "this was never version-specific, it was always wrong" does not hold.

This matters because the repo ships a `redis-6.2` profile (`src/core/compatibility/profile.ts`) and sweeps it in CI via `test:integration:compatibility:mock`. It already has a gate for this identical wording boundary — `'client.setinfo.unknown-subcommand-error': { redis: '7.0.0', valkey: '7.2.0' }` — consumed by `clientSetInfoUnavailableError` in `src/commands/connection.ts`, which switches between exactly these two templates.

Concrete failure: `createRedisServer({ compatibility: 'redis-6.2' })` plus `CONFIG BOGUS` now emits the 7.0+ text where real 6.2 emits the capital-U form. Net effect is an improvement on the default `redis-8.0` profile and a regression on `redis-6.2`. Valkey profiles are unaffected, since Valkey forked at 7.2.

Fix: route this through a profile-aware helper mirroring `clientSetInfoUnavailableError`, gated on the same 7.0.0 / 7.2.0 boundary, and correct the two comments — this one and the header of `tests-integration/raw-tcp/command-errors-config.test.ts`, which makes the same "identical on both, so no compatibility-profile gate applies" claim.

src/commands/config.ts
```diff
@@ -205,8 +205,13 @@ export const configCommand = defineCommand({
       return configRewrite(args.args)
     }
 
+    // Matches the generic unknown-subcommand reply real Redis sends for a
+    // container command, echoing the subcommand with the casing the client
+    // sent. Captured byte-for-byte from redis-server 7.2.1 and 8.0.6 (identical
+    // on both, so no compatibility gate applies) and pinned by
+    // tests-integration/raw-tcp/command-errors-config.test.ts (#388).
     throw new RedisCommandError(
```
fatal10110: **4. MEDIUM — the new error string is missing Redis's `%.128s` truncation of the echoed subcommand.**

Real Redis formats the echoed subcommand with `%.128s`, so anything longer is cut at 128 bytes. The boundary was verified exact against 8.0.6 and 7.2.1: at length 128 the mock and real replies are byte-identical; at 129 they diverge — real echoes 128 `X` characters, the mock echoes all 129. With a 300-character subcommand name the mock reflects all 300.

This is precisely the "different truncation model than Redis" that issue #384 part 2 describes. The separability argument for this PR holds for subcommand *naming*, but not for the *echo model* — the new string inherits that defect class. Applying `%.128s`-equivalent truncation when building the message closes it.

src/commands/config.ts
```diff
@@ -205,8 +205,13 @@ export const configCommand = defineCommand({
...
-      `Unknown CONFIG subcommand or wrong number of arguments for '${args.subcommand}'. Try CONFIG HELP.`,
+      `unknown subcommand '${args.subcommand}'. Try CONFIG HELP.`,
     )
```
fatal10110: **5. MEDIUM — non-UTF-8 subcommand bytes are mangled to U+FFFD before they reach this message.**

`schema: { subcommand: t.string() }` resolves to `readToken(...).toString()` in `command-schema.ts`, which is a UTF-8 decode, so invalid bytes become replacement characters and the reply no longer echoes what the client sent:

```
real 8.0.6 : -ERR unknown subcommand '\xff\xfe\xfd'. Try CONFIG HELP.
mock       : -ERR unknown subcommand '\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd'. Try CONFIG HELP.
```

The mock's error body is 6 bytes longer than real for that input. This PR adds a suite whose stated purpose is pinning these bytes, and this case is not covered. Reading the raw subcommand token as a Buffer for the echo (rather than the decoded string) would fix both the bytes and the test gap.

src/commands/config.ts
```diff
@@ -205,8 +205,13 @@ export const configCommand = defineCommand({
       return configRewrite(args.args)
     }
 
+    // Matches the generic unknown-subcommand reply real Redis sends for a
+    // container command, echoing the subcommand with the casing the client
+    // sent. Captured byte-for-byte from redis-server 7.2.1 and 8.0.6 (identical
+    // on both, so no compatibility gate applies) and pinned by
+    // tests-integration/raw-tcp/command-errors-config.test.ts (#388).
```
fatal10110: **10. LOW — `CONFIG HELP` is unimplemented, so this new error message points at itself.**

`execute` has four branches (get/set/resetstat/rewrite), while `introspection.subcommands` advertises `commandSubcommandInfo('config|help', 2)`. The result:

```
CONFIG HELP        mock: -ERR unknown subcommand 'HELP'. Try CONFIG HELP.
CONFIG HELP extra  mock: -ERR unknown subcommand 'HELP'. Try CONFIG HELP.
real 8.0.6         HELP -> 11-element array;  HELP extra -> wrong number of arguments for 'config|help' command
```

This is pre-existing and not caused by the PR. It is worth raising here because the PR adds a suite dedicated to CONFIG error wording that covers `BOGUS`, `bogus`, `BoGuS`, trailing args and zero args — and omits the one subcommand the message itself names. At minimum, add a test pinning the current `CONFIG HELP` behavior so the gap is visible; implementing the branch would be better.

src/commands/config.ts
```diff
@@ -205,8 +205,13 @@ export const configCommand = defineCommand({
       return configRewrite(args.args)
     }
 
+    // Matches the generic unknown-subcommand reply real Redis sends for a
```
fatal10110: **13. LOW (cross-cutting) — this wording is duplicated across 15 sites with three variants live at once.**

`git grep "nknown subcommand" -- src/commands/` returns 15 hits across `command.ts`, `config.ts`, `connection.ts`, `pubsub.ts`, `scripts.ts`, `streams/xgroup.ts` and `streams/xinfo.ts`, plus two error classes in `core/redis-error.ts` — with three different wordings in production simultaneously.

Findings 1, 3 and 4 are all direct consequences of that duplication. A single profile-aware helper — `unknownSubcommandError(container, subcommand, profile)` — would fix the 6.2 gate, XGROUP and the 128-byte truncation in one place, instead of once per call site.

Separately, this site throws a bare `RedisCommandError` rather than a subclass, which runs against the AGENTS.md convention for error types. Both points are filed as issue #413.

src/core/command-executor.ts
```diff
@@ -11,7 +11,10 @@ import {
 import type { RedisExecutionContext } from './redis-context'
 import { RedisResult } from './redis-result'
 import { isResponseStream, ResponseStream } from './response-stream'
-import type { RedisMonitorCommandEvent } from '../state/monitor-feed'
+import {
```
fatal10110: **6. MEDIUM — this is the first *runtime* import from `core/` into `state/`, and it points the wrong way.**

On `origin/main`, all 10 `from '../state'` imports under `src/core/` are `import type`, so the core layer had no emitted dependency on the state layer at all. This PR flips one of them to a value import in order to obtain a wall-clock helper — and that helper has nothing to do with state: `monitorTimestampMicros()` touches no `RedisServerState`, no keyspace, no feed. It is a clock.

The "one line plus its import" framing is accurate in volume (+5/-2); the objection is the direction, not the size. A `src/core/clock.ts` exporting `monitorTimestampMicros` would keep this edge type-only and put the helper at the layer that actually calls it, with `state/monitor-feed.ts` re-exporting it if the `/core` surface needs to keep the name.

fatal10110 (commented):
**Verdict: don't ship as-is.** Blocking: findings 1 and 2.

The MONITOR half of this PR is genuinely correct and well-tested. Mutation testing confirms the red-before-green claims, and the unit monotonicity assertion has real teeth — it catches a 10 microsecond inversion. The CONFIG half is the problem: it rests on a factual claim that is false, and it ships an ungated version-specific string plus a source comment and a test header that assert the opposite.

## On separability from #384

The author's argument that the CONFIG fix is separable from #384 is correct. #384's root cause is `plan.definition.name` in `subscribed-policy.ts`, a policy-layer problem with no subcommand in scope, whereas this fix sits inside `execute()` where `args.subcommand` is already available, and it introduces no subcommand-naming mechanism of its own. #384 part 1 still reproduces unchanged on this branch. The caveat is that findings 4 and 5 below show the new string does carry #384 part 2's *other* defect class — the echo/truncation/binary model.

## Findings outside the diff

**2. HIGH (blocking) — `CHANGELOG.md`.** There is no `BREAKING (/core)` entry for the `timestampMs` -> `timestampMicros` rename on `RedisMonitorCommandEvent`. That type is exported from `src/internal.ts`, which `tsup.config.ts` builds as the published `js-redis-server/core` subpath, so this is a public API break. `CHANGELOG.md`'s own preamble makes the entry mandatory, and `[Unreleased] / Removed` already carries three such entries (#375, #376, #377). This is sharper than it looks: the immediately preceding commit on `main`, `92a0d5f chore: drop the /core export-surface guard (#406)`, removed the CI check that caught exactly this class of break, and says verbatim — "What stays: CHANGELOG.md, where intentional /core removals are recorded." A JS consumer reading `event.timestampMs` now gets `undefined` and therefore `NaN` timestamps, with nothing in the release notes to warn them. The rename itself is clean: `git grep timestampMs` returns zero hits on the branch and no doc names the field. Please add the entry.

**3. MEDIUM — `src/commands/streams/xgroup.ts`.** XGROUP still carries the old wording on the default profile. Probing all container commands against real 8.0.6 and this PR's mock, the mock now matches real for CONFIG, CLIENT, XINFO, COMMAND, ACL, SCRIPT, FUNCTION, PUBSUB and SLOWLOG, and still diverges only for XGROUP:

```
real 7.2.1 / 8.0.6 : -ERR unknown subcommand 'BOGUS'. Try XGROUP HELP.
mock (PR head)     : -ERR Unknown subcommand or wrong number of arguments for 'BOGUS'. Try XGROUP HELP.
```

CONFIG was one of two stragglers and this PR fixes one of them. (OBJECT/CLUSTER/LATENCY/MEMORY are not implemented standalone at all and fall through to `unknown command` — a separate, larger gap.) Tracked together with finding 1 in issue #413.

**11. LOW — `src/core/resp-encoder.ts`, `sanitizeErrorText`.** `/[\r\n]+/g -> ' '` collapses runs, where real Redis uses `sdsmapchars(s, "\r\n", "  ", 2)`, a 1:1 character map. So `CONFIG "a\r\nbc"` yields `'a  bc'` (two spaces) on real Redis and `'a bc'` (one space) on the mock. Protocol-safe either way, but the byte count differs — which matters for a suite whose stated purpose is pinning exact bytes. Dropping the `+` fixes it.

## One note on the test evidence

`tests-integration/raw-tcp/monitor-protocol.test.ts` currently cannot be run whole against the real backend: test 5, "skips cluster commands rejected before local execution", times out at both 25s and 60s. Its first 352 lines are byte-identical to `origin/main`, so this PR does not cause the hang — but it does mean the "green against real Redis" claim for that file is not reproducible as stated.

src/state/monitor-feed.ts
```diff
@@ -34,6 +44,65 @@ export class RedisMonitorFeed {
...
+  const drift =
+    derivedMicros > wallClockMicros
+      ? derivedMicros - wallClockMicros
+      : wallClockMicros - derivedMicros
+
+  if (drift <= RESYNC_THRESHOLD_MICROS) {
+    return Number(derivedMicros)
+  }
+
+  wallClockOriginMicros = wallClockMicros
```
fatal10110: **7. MEDIUM — the resync branch breaks the monotonicity this function's own JSDoc promises.**

Discarding `derivedMicros` and returning `wallClockMicros` outright means the derived clock is not monotonic at the resync boundary. Reproduced by stubbing `Date.now` and running the real implementation:

```
baseline+1                       1790022786.405079
after -5s wall step (re-anchor)  1790022781.405000   <- 5.000079s BACKWARDS
```

Two consequences:

1. A backwards wall-clock step larger than 1s makes the MONITOR feed emit a line whose timestamp precedes the previous one. Real Redis's `gettimeofday()` does the same thing, so this is arguably faithful — but the JSDoc and both new tests assert an invariant the code does not actually hold. Either the docs and tests should be relaxed to match, or the branch should clamp forward.

2. Every resync emits one millisecond-quantized line (`...405000` above) — which is the exact defect this PR exists to remove, reintroduced on the recovery path, and covered by no test.

Also worth noting: the 1-second tolerance band means the mock's MONITOR timestamp can sit up to a second away from wall-clock time indefinitely, where real Redis is always exact.

src/state/monitor-feed.ts
```diff
@@ -34,6 +44,65 @@ export class RedisMonitorFeed {
...
+}
+
+/**
+ * Render a microsecond epoch timestamp the way real Redis writes it on a
+ * `MONITOR` line: `<unix-seconds>.<6 digits>`. Split with integer arithmetic
+ * rather than `(micros / 1e6).toFixed(6)`, which can round the last digit once
+ * the value exceeds a double's ~15 significant digits.
+ */
+export function formatMonitorTimestamp(timestampMicros: number): string {
+  const seconds = Math.floor(timestampMicros / 1_000_000)
```
fatal10110: **9. LOW — `formatMonitorTimestamp` emits malformed strings for non-integer input, and it is now `/core`-public.**

```
1695000000000000.5 -> "1695000000.0000.5"    (two decimal points)
1.5                -> "0.0001.5"
NaN                -> "NaN.000NaN"
```

`padStart` cannot fix a fractional string, so the output is not a valid timestamp at all. This is unreachable from `monitorTimestampMicros()`, which always returns an integer — but both this helper and `RedisMonitorCommandEvent` are published API via `src/internal.ts`, so an external caller can hit it. A `Math.trunc` on entry closes it.

For the record, all integral inputs are correct: 0, 1, 999999, 1000000 and the second boundaries all pad to exactly 6 digits, verified.

tests-integration/raw-tcp/monitor-protocol.test.ts
```diff
@@ -352,8 +352,101 @@ describe(`Raw TCP MONITOR protocol (${testRunner.getBackendName()})`, () => {
...
+    const frames: Buffer[] = []
+    for (let i = 0; i < sampleCount; i++) {
+      frames.push(commandFrame('ping', `${marker}:${i}`))
+    }
+    actor.write(Buffer.concat(frames))
+    for (let i = 0; i < sampleCount; i++) {
+      await actor.readRawFrame()
+    }
+
+    const before = Date.now()
```
fatal10110: **8. MEDIUM — the wall-clock lower bound is loose only because `before` is captured too late.**

`const before = Date.now()` runs after `actor.write(...)` and after all 24 reply reads, so every timestamp under test was stamped strictly *before* this line executes. That is exactly why the assertion below needs a 60-second fudge factor to pass.

Moving this capture above the `actor.write(...)` makes `before` a genuine lower bound, and the assertion can then tighten to `before - 1_000` — symmetric with the existing `after + 1_000` upper bound, and a far stronger check.

tests-integration/raw-tcp/monitor-protocol.test.ts
```diff
@@ -352,8 +352,101 @@ describe(`Raw TCP MONITOR protocol (${testRunner.getBackendName()})`, () => {
...
+      assert.ok(
+        millis >= before - 60_000 && millis <= after + 1_000,
+        `monitor timestamp ${value.text} is not near wall-clock time`,
+      )
+    }
+
+    // Monotonic across the feed, like a real capture.
+    for (let i = 1; i < timestamps.length; i++) {
+      assert.ok(
+        timestamps[i].text >= timestamps[i - 1].text,
```
fatal10110: **12. LOW — assertion strength, and a string comparison that happens to work.**

Mutation testing results for this suite: `Date.now() * 1000` (the original #388 regression) fails both the unit and raw-TCP tests; dropping `padStart(6, '0')` fails both; a non-monotonic -10 microsecond jitter fails the unit test but **passes** this raw-TCP one; -3 ms fails both. So this monotonicity check needs roughly a 3 ms inversion to fire — low sensitivity rather than vacuous, and the unit test covers the gap.

The genuinely weak assertion is the unit test's `Math.abs(sample / 1000 - Date.now()) < 1000`, whose tolerance is exactly `RESYNC_THRESHOLD_MICROS / 1000` — the implementation guarantees it by construction, so it can never fail.

Separately, `timestamps[i].text >= timestamps[i - 1].text` compares timestamp *strings*. That is sound today (unix seconds stay 10 digits until 2286) but wrong by construction, and `match[2]` is already captured, so a numeric comparison is two characters away.

src/commands/config.ts
fatal10110: Confirmed and fixed in ea64adb — you are right and my claim was wrong. I reproduced it myself on throwaway containers (`redis:6.2` on 7389, `redis:7.0` on 7390) rather than taking it on trust:

```
6.2.24  CONFIG BOGUS -> -ERR Unknown subcommand or wrong number of arguments for 'BOGUS'. Try CONFIG HELP.
7.0.15  CONFIG BOGUS -> -ERR unknown subcommand 'BOGUS'. Try CONFIG HELP.
8.0.6   CONFIG BOGUS -> -ERR unknown subcommand 'BOGUS'. Try CONFIG HELP.
```

So the string I deleted was correct for 6.2, and my PR was a regression on that profile.

Added `error.unknown-subcommand-wording` (`{ redis: '7.0.0', valkey: '7.2.0' }`) mirroring `client.setinfo.unknown-subcommand-error`, and routed CONFIG through a new `unknownSubcommandError(container, subcommand, profile)` in `src/commands/helpers.ts` rather than a CONFIG-private branch, so #413 has something to migrate the other call sites onto. Both false comments are corrected — this one and the header of `command-errors-config.test.ts`.

`tests-integration/compatibility/profile-gates.test.ts` now asserts both sides of the boundary. I checked it has teeth by stubbing the gate to always-true: red on `redis-6.2` (2 failures), green on the other six. Full sweep is 10 pass / 0 fail on all seven profiles.

One thing the gate does **not** cover, so it is not mistaken for complete: the *arity* path. On 6.2, `CONFIG RESETSTAT extra` also uses the legacy template (`Unknown subcommand or wrong number of arguments for 'RESETSTAT'. Try CONFIG HELP.`), where 7.0+ says `wrong number of arguments for 'config|resetstat' command`. The mock throws the 7.0+ form unconditionally. Pre-existing, unchanged by this PR, and it belongs with #413/#384 rather than here.

src/commands/config.ts
fatal10110: Fixed in ea64adb. I verified the boundary independently before implementing, and it turns out the truncation is itself version-gated:

```
echoed length for a name of   127  128  129  300
6.2.24                        127  128  129  300    (no truncation at all)
7.0.15                        127  128  128  128
8.0.6                         127  128  128  128
```

So `%.128s` arrived with the same 7.0 change as the wording, and it is gated together with it rather than applied unconditionally — applying it on the 6.2 profile would have been a fresh divergence in the other direction.

Implemented as `raw.subarray(0, 128)` on a Buffer so the cut is at 128 **bytes**, not 128 code points; that is part of why CONFIG's schema field became `t.bulk()` (see my reply on finding 5). `command-errors-config.test.ts` pins 1/127/128 echoed whole and 129/300 cut to 128 — 5 pass / 0 fail on mock and against the real 8.0 backend.

src/commands/config.ts
fatal10110: **Deferred, with the schema half done.** Your diagnosis is right and I reproduced the 6-byte difference, but the remaining fix is not reachable from this layer.

I did change CONFIG's schema from `t.string()` to `t.bulk()` in ea64adb, so `args.subcommand` is now the raw `Buffer` and the lossy UTF-8 decode no longer happens during parsing. That was needed for the 128-**byte** truncation anyway.

What blocks the rest: `RedisCommandError` carries a `string`, and `formatError` / `sanitizeErrorText` in `src/core/resp-encoder.ts` operate on strings before the frame is built. There is no path from a `Buffer` to raw bytes on the wire without making the error-message pipeline byte-oriented end to end. `toString('latin1')` does not rescue it either — `\xff\xfe\xfd` would then UTF-8-encode to 6 bytes instead of today's 9, still not the 3 real Redis sends.

That is precisely #384 part 2 ("`UnknownRedisCommandError` renders a binary command name by hex-dumping it") — the same missing mechanism, owned by a different lane, and `resp-encoder.ts` is being edited concurrently. Rewriting a shared encoder inside a MONITOR-timestamp PR is the wrong trade. I have deliberately left no test asserting the current wrong bytes, so nothing here has to be un-pinned when #384 lands.

src/commands/config.ts
fatal10110: **Deferred, and deliberately without a test** — I want to be explicit about why rather than adding a weak one.

This suite runs against **both** backends. A test pinning the mock's current `-ERR unknown subcommand 'HELP'. Try CONFIG HELP.` would fail against real 8.0.6, which returns an 11-element array; and a test asserting the real reply means *implementing* `CONFIG HELP`, whose payload varies by version — a feature, not an error-wording fix. A `skip: backend === 'real'` test pinning known-wrong behaviour struck me as worse than none, since it reads as approval of the current output.

Your underlying point stands: a suite dedicated to `CONFIG` error wording that omits the one subcommand the message itself names is a genuine gap. I have not filed an issue because that was not delegated to me — flagging it here so it is on the record for whoever picks up #413, which already owns this family.

src/commands/config.ts
fatal10110: Agreed, and partly acted on. ea64adb adds `unknownSubcommandError(container, subcommand, profile)` to `src/commands/helpers.ts`, holding the profile gate and the `%.128s` truncation in one place, so #413 becomes a mechanical migration rather than a design exercise.

I deliberately migrated **only CONFIG**. Doing one or two more of the fifteen ad hoc — XGROUP being the obvious candidate, see my reply on finding 3 — would leave the codebase in a *third* state, which is worse than the two it is in now. The helper is the part worth landing here because this PR needs it regardless.

On the bare `RedisCommandError`: I kept it, so the helper matches the fourteen sites it is meant to absorb. Introducing a subclass now means either a new `/core` export that #413 immediately reshapes, or two inconsistent error types for one message. Better done once, in the sweep, where the convention can be applied uniformly.

src/core/command-executor.ts
fatal10110: Fixed in 04c48e5 — you are right, and the direction was the part I had not thought about. Moved to `src/core/clock.ts`.

I checked your premise rather than assuming it:

```
$ grep -rn "from '\.\./state" src/core/ | grep -v "import type"
(no output)
```

That is the state on `main` and the state again now; my commit had been the single exception. The helper genuinely belongs in `core/` on its own merits too — it touches no `RedisServerState`, keyspace or feed, and `command-executor.ts` is its only caller.

`/core` keeps both names, now re-exported from `./core/clock` instead of through the state block, so the published surface is unchanged. I did *not* add a re-export from `state/monitor-feed.ts`, since that would recreate the edge in the opposite direction for no benefit; `RedisMonitorCommandEvent`'s doc comment points at the clock module by path instead.

src/state/monitor-feed.ts
fatal10110: **Kept the behaviour, fixed the claims** — 04c48e5. This is the finding I spent longest on, so here is the reasoning.

On your point 1, I took the option you flagged as "arguably faithful". Real Redis stamps from `gettimeofday()`, which steps backwards on an NTP correction; a MONITOR capture taken across such a step is not ordered on real Redis either. Clamping forward would make the mock *more* ordered than the thing it mocks, and any test written against real behaviour would then be unreproducible here. So the code stays and the documentation stops overclaiming:

- the JSDoc now scopes monotonicity explicitly to "while the wall clock is stable", states that a step larger than the threshold produces a backwards timestamp, and names `gettimeofday()` as the reason;
- both of your consequences are listed as accepted costs, including the up-to-1s offset from true wall-clock time between resyncs, which I had not documented at all.

On your point 2 — the quantized line on the recovery path — you are right that it was covered by no test. There is now one: it stubs `Date.now`, steps it -5s and then +5s, and asserts the anchor is recaptured in both directions. I verified it has teeth by deleting the resync branch, which makes it fail. What that line *cannot* be is unquantized: at a resync a freshly floored `Date.now()` is genuinely all the precision available, and manufacturing sub-millisecond digits by carrying over the previous phase would be inventing precision rather than measuring it. So it is documented as a known cost of a path that only a >1s clock step reaches.

The two existing tests no longer assert an invariant the code does not hold — the unit one is retitled to "have sub-millisecond resolution" with the monotonicity check scoped to the stable-clock case it actually exercises.

src/state/monitor-feed.ts
fatal10110: Fixed in 04c48e5 — `Math.trunc` on entry, and thank you for confirming the integral cases pass, which saved me re-deriving them.

Added coverage for exactly the case you found:

```ts
assert.strictEqual(formatMonitorTimestamp(1_695_000_000_000_000.5), '1695000000.000000')
assert.strictEqual(formatMonitorTimestamp(1.5), '0.000001')
```

The `/core`-public framing is what made this worth fixing rather than noting — unreachable from `monitorTimestampMicros()` today, but the guard costs one call and the failure mode is a silently malformed timestamp rather than a throw.

tests-integration/raw-tcp/monitor-protocol.test.ts
fatal10110: Fixed in 04c48e5, exactly as suggested. You identified the cause correctly — the 60-second fudge was not a considered tolerance, it was me papering over a lower bound I had captured too late.

`const before = Date.now()` now sits immediately above `actor.write(...)`, so it genuinely brackets the window in which every line was stamped, and the assertion is `before - 1_000`, symmetric with the existing `after + 1_000`. Green on mock and against the real 8.0 backend.

tests-integration/raw-tcp/monitor-protocol.test.ts
fatal10110: Both fixed in 04c48e5, and the mutation table was the most useful thing in this review — thank you for running it.

**The string comparison.** Fixed properly rather than with the two-character change. `collectMonitorTimestamps` now returns a `MonitorTimestamp` with the whole value as integer microseconds (`seconds * 1_000_000 + micros`, comfortably inside the safe-integer range), and the ordering check compares that. The sensitivity goes from ~3 ms to 1 µs, so this assertion is no longer carried by the unit test.

**The vacuous unit assertion.** You are right that `Math.abs(sample / 1000 - Date.now()) < 1000` is guaranteed by construction — the resync threshold is exactly 1 s, so any value that could violate it triggers a re-anchor that puts it back inside the band. Deleted rather than re-tuned: what it was reaching for (the epoch is real wall-clock time, not a raw monotonic counter) is now covered directly by the new `Date.now`-stubbing resync test, which pins the anchor within 1 ms of a stepped clock in both directions.

The `Number.isSafeInteger` check stayed — that one is not free, it is what keeps the microsecond-as-`number` decision honest.

fatal10110:
Thanks — this was a genuinely good review, and finding 1 was a real error on my part, not a judgement call. Three commits pushed, no force-push:

- `ea64adb` — CONFIG wording gated on the 7.0 boundary (findings 1, 4, partial 5, helper for 13)
- `04c48e5` — clock moved to `src/core/clock.ts`, resync semantics documented rather than overclaimed (findings 6, 7, 8, 9, 12)
- `3b36d55` — CHANGELOG (finding 2)

I replied on each of the ten inline threads. Covering the three that had nowhere to land:

### 2. CHANGELOG — fixed (`3b36d55`)

You are right that this is a silent break, and the `#406` framing is what made it land: the guard that would have caught it was removed one commit before mine, explicitly on the grounds that the CHANGELOG is what remains. Added under `[Unreleased] / Removed` with the migration mapping (`event.timestampMs` -> `event.timestampMicros / 1000`) and an explicit note that a stale reader gets `undefined` and therefore `NaN` rather than a type error. Also added a `Fixed` entry for the CONFIG gate, and `[#410]` / `[#413]` link refs.

### 3. XGROUP — deferred to #413, confirmed still broken

Verified on my own throwaway containers rather than taking the numbers on trust:

```
real 6.2.24   -> Unknown subcommand or wrong number of arguments for 'BOGUS'. Try XGROUP HELP.
real 7.0.15   -> unknown subcommand 'BOGUS'. Try XGROUP HELP.
real 8.0.6    -> unknown subcommand 'BOGUS'. Try XGROUP HELP.
mock (PR head)-> Unknown subcommand or wrong number of arguments for 'BOGUS'. Try XGROUP HELP.
```

So your finding stands, and with the new `unknownSubcommandError()` helper it is a one-line change. I still left it. Fixing one arbitrary site out of the remaining fourteen puts the codebase in a third state — some gated, some 7.0-hardcoded, some 6.2-hardcoded — which is harder to reason about than the two states it is in now, and #413 exists precisely to sweep them together. It is also a different command family from the one this PR is about. Happy to be overruled if you would rather take the straggler now.

### 11. `sanitizeErrorText` — deferred, confirmed

Reproduced on the PR head:

```
real 7.0.15 / 8.0.6 : -ERR unknown subcommand 'a  bc'. Try CONFIG HELP.   (two spaces)
mock                : -ERR unknown subcommand 'a bc'. Try CONFIG HELP.    (one space)
```

Your reading of `sdsmapchars(s, "\r\n", "  ", 2)` as a 1:1 map is correct and dropping the `+` is the right fix. I am not doing it in this PR: `src/core/resp-encoder.ts` is on every error reply in the codebase, it is being edited concurrently by another lane, and a one-character change with that blast radius belongs in a PR where it is the subject and can be swept with the rest of the error-text work. Flagging it for #413.

### On the test evidence

You were right to call out that "green against real Redis" was not reproducible. The containers had failed in the interim (two cluster nodes OOM-killed, then a restart blocked by stale `nodes-*.conf` in an anonymous volume) — and, to be straight about my own part in it, my earlier real-backend runs went through `npm run test:integration:real`, which chains `clean:redis` and therefore `FLUSHALL`s the shared instances. That was mine and it was wrong.

On the recreated containers, `monitor-protocol.test.ts` including test 5 is now **13 pass / 0 fail** against real. Everything since has been targeted single files, never the suite script.

Full results in my report; the short version is unit 345/0, mock integration 1122/0, the seven-profile sweep 10/0 each, and the two touched raw-TCP files 18/0 against real 8.0.

src/commands/helpers.ts
```diff
@@ -131,6 +133,58 @@ export function parsePositiveExpireToken(
...
+): RedisCommandError {
+  const raw = Buffer.isBuffer(subcommand) ? subcommand : Buffer.from(subcommand)
+
+  if (!profile.has('error.unknown-subcommand-wording')) {
+    return new RedisCommandError(
+      `Unknown subcommand or wrong number of arguments for '${raw.toString()}'. Try ${container} HELP.`,
+    )
+  }
+
+  const echoed = raw.subarray(0, SUBCOMMAND_ECHO_LIMIT).toString()
```
fatal10110, line 192: **LOW (new, introduced by this commit)** — when the 128-byte cut lands mid-UTF-8-sequence, `.toString()` turns the partial byte into U+FFFD and the reply comes out **2 bytes longer** than real Redis.

Real Redis cuts at 128 bytes and emits the raw partial byte. Probed with `CONFIG <127 x 'X' + U+00E9>` (129 bytes):

```
real 7.0.15 / 8.0.6 : echoed 128 bytes, tail ... 58 58 58 c3
mock (this commit)  : echoed 130 bytes, tail ... 58 58 ef bf bd
```

`raw.subarray(0, 128)` correctly yields 127 `X` plus `0xc3`, then `.toString()` decodes the orphaned lead byte to U+FFFD (3 bytes).

Two things worth noting rather than fixing here:

1. This is **not** a regression — before this commit the mock echoed all 300 bytes, so 128±2 is strictly closer. It is a residual gap.
2. It is not covered by your non-UTF-8 deferral as written. That deferral is scoped to *invalid* input; this fires on **well-formed UTF-8**, because the truncation itself creates the split. A 400-byte name of 2-byte characters is perfectly valid input, and roughly half of such names will split. Worth one sentence on #384 part 2 so the byte-oriented error pipeline is scoped to cover it — a `Buffer`-carrying `RedisCommandError` fixes both at once.

The aligned case is already right: `CONFIG <200 x U+00E9>` (400 bytes) cuts cleanly at 128 bytes on both real and mock, tail `c3 a9 c3 a9 c3 a9`.

src/commands/helpers.ts
```diff
@@ -131,6 +133,58 @@ export function parsePositiveExpireToken(
...
+ * Note this covers the *unknown subcommand* path only. On 6.2 the same template
+ * also served wrong-arity replies for a known subcommand, where 7.0+ says
+ * `wrong number of arguments for 'config|get' command`; that arity path is a
+ * separate divergence and is not handled here (#413).
+ *
+ * There are 15 hand-rolled copies of this message across the command modules
+ * with three wordings live at once; #413 tracks migrating them onto this
+ * helper. Only CONFIG is routed through it so far.
+ */
+export function unknownSubcommandError(
```
fatal10110, line 179: **Note for #413, not a defect** — since this is billed as the migration target for the other fourteen sites, two things about the signature are worth recording now, while it has one caller and is cheap to change.

**1. It covers roughly nine of the fifteen sites, not all fifteen.** Your JSDoc correctly scopes it to the *unknown subcommand* path. But several of the sites still carrying the old template are not that path — `command.ts:126`, `connection.ts:733`, `connection.ts:1104` and `pubsub.ts:391` use the 6.2 wording for arity/unsupported cases, where 7.0+ says `wrong number of arguments for 'x|y' command` instead. Those need the arity helper you flagged as #413's other half, not this one. Splitting the #413 checklist along that line up front will save a round of wrong migrations.

**2. `subcommand: Buffer | string` is the right call, and the first non-`execute()` caller will need the Buffer arm.** `xgroup.ts:137` lives inside the schema parser, where the local `subcommand` is already `input[index]?.toString().toUpperCase()`. Migrating it by passing that variable would echo `BOGUS` for a client that sent `bogus` — a fresh casing divergence, since real Redis echoes verbatim. The correct call is `unknownSubcommandError('XGROUP', input[index] ?? Buffer.alloc(0), ctx.profile)`. Confirmed `ParseContext` carries `profile`, so the helper is reachable from a schema parser — worth stating on #413, because the ergonomics there differ from `execute()`.

Minor third: `error.unknown-subcommand-wording` and the existing `client.setinfo.unknown-subcommand-error` now describe overlapping behaviour at the identical `{ redis: '7.0.0', valkey: '7.2.0' }` boundary. When #413 reaches `connection.ts:716`, the older gate should probably fold into this one rather than both surviving.

src/core/clock.ts
```diff
@@ -0,0 +1,88 @@
...
+ * rather than `(micros / 1e6).toFixed(6)`, which can round the last digit once
+ * the value exceeds a double's ~15 significant digits.
+ *
+ * `Math.trunc` guards the published `/core` entry point: a fractional input
+ * would otherwise produce a string with two decimal points
+ * (`1695000000000000.5` -> `"1695000000.0000.5"`). {@link monitorTimestampMicros}
+ * never returns one.
+ */
+export function formatMonitorTimestamp(timestampMicros: number): string {
+  const micros = Math.trunc(timestampMicros)
```
fatal10110: **LOW** — the `Math.trunc` guard closes the fractional case (verified: removing it turns `formats monitor timestamps with six fractional digits` red), but `NaN` and `Infinity` still produce a malformed string:

```
formatMonitorTimestamp(NaN)       -> "NaN.000NaN"
formatMonitorTimestamp(Infinity)  -> "Infinity.000NaN"
```

`Math.trunc` is the identity on both, then `micros - seconds * 1_000_000` is `NaN`.

The JSDoc above only promises to guard the fractional case, so this is consistent rather than wrong — but since the reason for guarding at all is that this is a published `/core` entry point an external caller can hand anything, `Number.isFinite` would cover the same surface for the same one line. Purely optional; `monitorTimestampMicros()` can produce neither.

fatal10110 (commented):
## Re-review of 3b36d55 — both blocking findings fixed; ship

Re-verified against my own throwaway containers (`redis:6.2` -> 6.2.24, `redis:7.0` -> 7.0.15, `redis:8.0.6`) on free high ports, all removed. Shared infra untouched.

### Prior findings

| # | Finding | Status |
|---|---|---|
| 1 | **BLOCKING** — wording is version-specific, ungated, two false comments | **Fixed**, verified |
| 2 | **BLOCKING** — no `BREAKING (/core)` CHANGELOG entry | **Fixed** |
| 3 | XGROUP still on the 6.2 wording | Deferred to #413 — see below |
| 4 | Missing `%.128s` truncation | **Fixed**, and correctly gated |
| 5 | Non-UTF-8 bytes mangled to U+FFFD | Half fixed (`t.bulk()`); deferral accepted |
| 6 | First runtime `core/` -> `state/` import | **Fixed**, verified |
| 7 | Resync breaks the claimed monotonicity | Behaviour kept, claims corrected — accepted |
| 8 | `before` captured after the traffic | **Fixed** |
| 9 | `formatMonitorTimestamp` malformed on non-integer | **Fixed** |
| 10 | `CONFIG HELP` unimplemented | Deferred; reasoning accepted |
| 11 | `sanitizeErrorText` collapses CR/LF runs | Deferred; correct lane |
| 12 | Weak assertions | **Fixed** |
| 13 | 15 duplicated literals | Helper landed; #413 filed |

### Blocking 1 — fixed, and the gate has teeth

The gate reproduces my ground truth exactly. Mock replies per profile, byte-for-byte against real:

```
redis-6.2   -ERR Unknown subcommand or wrong number of arguments for 'BOGUS'. Try CONFIG HELP.   == real 6.2.24
redis-7.0   -ERR unknown subcommand 'BOGUS'. Try CONFIG HELP.                                    == real 7.0.15
redis-8.0   -ERR unknown subcommand 'BOGUS'. Try CONFIG HELP.                                    == real 8.0.6
valkey-8.0  -ERR unknown subcommand 'BOGUS'. Try CONFIG HELP.
```

I did not take the "I verified the gate has teeth" claim on trust. Four mutations, each caught:

| mutation | result |
|---|---|
| gate forced always-true | `redis-6.2` **2 fail**, other profiles green |
| drop the 128-byte truncation | `redis-8.0` **1 fail**, `redis-6.2` green |
| delete the re-anchor branch | unit **1 fail** (`re-anchors on a wall-clock step`) |
| remove the `Math.trunc` guard | unit **1 fail** (`formats monitor timestamps`) |

### The new truncation claim — CONFIRMED, independently

This was not in my review and the `t.bulk()` change rests on it, so I probed all three versions directly with raw RESP frames. Echoed `X` count for a name of length N:

```
            N=127  N=128  N=129  N=300
6.2.24        127    128    129    300     <- no truncation at all
7.0.15        127    128    128    128
8.0.6         127    128    128    128
```

The claim is right on both halves: 6.2 does not truncate, 7.0+ cuts at exactly 128, boundary exact. Coupling the truncation to the same gate as the wording is correct — applying `%.128s` on the 6.2 profile would have been a fresh divergence in the other direction.

It is also genuinely **bytes**, not code points, which justifies `t.bulk()`. Confirmed with multi-byte input — real Redis cuts mid-character and emits the raw partial byte:

```
CONFIG <127 x 'X' + U+00E9>   (129 bytes)
  6.2.24  echoed 129 bytes, tail ... 58 58 c3 a9
  7.0.15  echoed 128 bytes, tail ... 58 58 58 c3    <- split mid-sequence
  8.0.6   echoed 128 bytes, tail ... 58 58 58 c3
```

### `t.bulk()` side effects — none found

`t.bulk()` and `t.string()` share the identical parse body (`readToken(input, index)`, `nextIndex: index + 1`); they differ only in the `.toString()`. So arity, parse-failure behaviour and the bare-`CONFIG` reply are unchanged — confirmed by the raw-TCP suite passing against both backends. `args.subcommand` has exactly one other reader (`.toString().toLowerCase()` at `config.ts:192`), `keys()` is unaffected, and `introspection` is a static table. `tsc --noEmit` and `eslint` are clean.

### Clock move — type-only claim verified

```
$ git grep -n "from '\.\./state" pr-410 -- src/core/ | grep -v "import type"
(no output)
```

All 10 `core/` -> `state/` imports are `import type` again. `src/core/clock.ts` imports nothing from `state/`. The `/core` published surface is unchanged — both names still exported from `src/internal.ts`, just from `./core/clock`.

### Finding 7 — argument accepted

Clamping forward would make the mock strictly more ordered than `gettimeofday()`, which is the wrong direction for a compatibility mock. The scoped JSDoc plus the new stub test is the right resolution. The "quantized resync line is unfixable" argument is also correct: at a resync the freshly floored `Date.now()` genuinely is all the precision available, and carrying the previous phase forward would invent it.

### Test 5 — confirmed

`monitor-protocol.test.ts` now runs **13/13 against the real backend**, including `skips cluster commands rejected before local execution`, which hung at both 25s and 60s last round. The broken-container-state explanation holds. Combined raw-TCP run against real: **18/18**.

### Verification I ran

| suite | result |
|---|---|
| `tests/state-core.test.ts` + both raw-TCP files, mock | 32 pass / 0 fail |
| `profile-gates.test.ts`, all 7 profiles | 10/10 each |
| both raw-TCP files, **real** backend | 18 pass / 0 fail |
| CONFIG-adjacent suites (ioredis + node-redis config, keyspace-notifications, standalone-auth) | 41 pass / 0 fail |
| `tsc --noEmit`, `eslint` | clean |

I did not run the full real suite, so the aggregate counts in the PR body remain unverified by me.

---

## New findings

Two inline, plus the XGROUP call below. All low severity; none blocking.

## On the XGROUP deferral — you invited being overruled, so: the argument is wrong, the outcome is fine

The stated reason does not hold. Counting **live wire wordings on the default profile**:

- before this PR: form A (7.0+) at 8 sites, form B (6.2) at CONFIG and XGROUP;
- after this PR: form A at 8 sites, form A via the gate at CONFIG, form B at XGROUP;
- with XGROUP migrated: form A at 8, form A via the gate at 2, form B at **zero**.

Migrating XGROUP does not create a third wording — it removes the last site emitting the wrong one. The third *implementation* state (gated helper) already exists; this PR created it. And on the default profile XGROUP is simply wrong today.

That said, I am not going to push it, for two reasons that are better than the one given:

1. It needs its own test, which is real scope.
2. It is **not** the one-liner it looks like. `xgroup.ts:137` sits inside the schema parser, where `subcommand` is already `input[index]?.toString().toUpperCase()` — so a naive migration would echo `BOGUS` for a client that sent `bogus`, a fresh casing divergence. The correct call passes the raw token: `unknownSubcommandError('XGROUP', input[index] ?? Buffer.alloc(0), ctx.profile)`. `ParseContext` does carry `profile`, so the helper is reachable there — worth recording on #413 since this is the first non-`execute()` call site.

Other deferrals: **non-UTF-8 echo** — agreed, `RedisCommandError` carrying a `string` is the real blocker and it is #384 part 2's lane. **`CONFIG HELP`** — agreed; a `skip: backend === 'real'` test pinning known-wrong output would read as approval. **`sanitizeErrorText`** — agreed, `resp-encoder.ts` is not this PR.

---

## Verdict: **ship**

Both blocking findings are genuinely fixed and I verified the fixes rather than the claims. The new version-gating claim, which I had missed entirely, is correct on both halves and is the better fix — you found a real divergence in my review's blind spot. The gate, truncation, re-anchor and `Math.trunc` guards all fail under mutation, so the coverage is real. The two new findings below are edge cases that do not block.

Correcting my own record: my review asserted `%.128s` was version-invariant. It is not, and a wrong gate plus an unneeded schema change was the exact risk — you checked instead of implementing what I said.

src/commands/helpers.ts
fatal10110: **Left it, widened the wording** — `242c700`. You are right that my deferral said "non-UTF-8" and this fires on perfectly well-formed input, so the wording excluded the case by accident. Reproduced both sides with `'A'.repeat(127) + 'é' + 'B'.repeat(20)`:

```
real 8.0.6 : 128 echoed bytes, 172 total — 127 'A' plus the raw 0xC3, cut mid-character
here       : 130 echoed bytes, 174 total — 127 'A' plus U+FFFD (EF BF BD)
```

I considered a third option you did not list — trimming back to the last whole UTF-8 boundary, which would echo 126–127 bytes. I rejected it: it is still not what real does (128 with a partial byte), it adds code that is wrong in a *new* way, and it creates a third behaviour to unwind when the byte pipeline lands. Being 2 bytes over with no extra machinery is easier to reason about than being 1–2 bytes under with some.

So the comment now states what the code actually does:

> Known gap, shared with the non-UTF-8 case below: when the 128-byte cut lands *inside* a multi-byte character — including a perfectly well-formed one — real Redis emits the raw partial byte, while `toString()` here yields U+FFFD. For `'A'.repeat(127) + 'é' + ...` real replies with 128 echoed bytes (172 total) and this replies with 130 (174).

Both cases are also written into the test-file header as gap 1, with the same numbers. Point taken on the root of last round's blocker — the comment claiming more than the code is the failure mode I am trying not to repeat.

src/core/clock.ts
fatal10110: Taken — `242c700`. `Number.isFinite(timestampMicros) ? Math.trunc(timestampMicros) : 0`, same line, and the JSDoc now names all three bad inputs instead of only the fractional one, so the comment and the code agree.

Coverage added alongside the fractional cases:

```ts
assert.strictEqual(formatMonitorTimestamp(Number.NaN), '0.000000')
assert.strictEqual(formatMonitorTimestamp(Number.POSITIVE_INFINITY), '0.000000')
assert.strictEqual(formatMonitorTimestamp(Number.NEGATIVE_INFINITY), '0.000000')
```

You flagged it as optional, but "the reason for guarding at all is that a `/core` caller can hand it anything" is the whole argument — a guard that covers one of the three ways to violate the contract is the awkward middle.

src/commands/helpers.ts
fatal10110: This is the most valuable thing in the review and it was about to be lost in a PR thread — written up on #413 as a comment, with attribution to this review, so whoever picks it up finds it: https://github.com/fatal10110/js-redis-server/issues/413

All three points accepted, and your split at the arity line matches what I hit building it. I had already scoped the JSDoc to the unknown-subcommand path and noted the arity divergence (`CONFIG RESETSTAT extra` on 6.2 uses the legacy template, on 7.0+ it is `wrong number of arguments for 'config|resetstat' command`), but I had not gone through the fifteen sites to work out *which* ones that excludes. Your four — `command.ts:126`, `connection.ts:733`, `connection.ts:1104`, `pubsub.ts:391` — make the checklist concrete.

The `xgroup.ts:137` point is the one I am most glad you caught. Passing the already-upper-cased local would have echoed `BOGUS` for a client that sent `bogus`, and that is exactly the class of silent divergence this PR started as. Your `input[index] ?? Buffer.alloc(0)` form is right, and confirming `ParseContext` carries `profile` removes the main doubt I had about whether the helper was reachable from a schema parser at all.

On folding `client.setinfo.unknown-subcommand-error` into the new gate: agreed in principle, and I left it alone deliberately. Collapsing two gates is a change to the published gate table with its own doc row, and doing it while `connection.ts` still has its own copy of the message would leave a gate named for a behaviour that no longer lives where the name says. Better as the last step of #413, once `connection.ts:716` is migrated — I have said so in the #413 comment.

I have added two constraints of my own there that only show up once you try to use the helper.

src/commands/config.ts
fatal10110: **Update — leaving it unimplemented, but it is now recorded rather than invisible** (`242c700`).

I went and captured what implementing it would actually take, because "leave it" without that is just a shrug. Real's 11-element reply is not version-stable:

```
6.2.24   ...  "    Prints this help."
7.0.15   ...  "    Prints this help."
7.2.16   ...  "    Print this help."     <- changes here
7.4.11   ...  "    Print this help."
8.0.6    ...  "    Print this help."
```

And the arity path forks too: `CONFIG HELP extra` is `wrong number of arguments for 'config|help' command` on 7.0+, but the legacy `Unknown subcommand or wrong number of arguments for 'HELP'. Try CONFIG HELP.` on 6.2.

So a byte-exact `CONFIG HELP` needs a **third** gate at 7.2 for the verb, plus the profile-aware arity path — which is the same arity mechanism I deferred to #413/#384. That is a feature with its own compat surface, and folding it into a PR about MONITOR timestamps is how scope creep produces a third half-migrated state.

What I did instead: the full captured payload and both boundaries are now in the header of `command-errors-config.test.ts` as an explicit "known gap 2", and repeated on #413. Whoever picks it up does not have to re-derive any of it — the version sweep is the expensive part and it is done.

Your original reasoning about why no test can go here still holds unchanged, so there is still no test.

fatal10110:
Cleanup pushed as `242c700`. Replies are on the four relevant threads; the short version, since two of the items were judgement calls you left to me:

**`sanitizeErrorText` — took it.** You were right that the blast radius was the only reason to hesitate, and with the other lanes through review that reason is gone. Dropping the `+` changed one existing unit assertion, `formats RESP2 double edge values and sanitizes error lines`, which had pinned the collapsing behaviour — corrected to real's bytes (`-ERR bad  frame`, two spaces) with the `sdsmapchars` reasoning in a comment. Added a raw-TCP case that runs on both backends and verified it goes red under the old regex. Full mock integration is 1123/0 and the four error-text-sensitive raw-TCP suites are 25/0 against real, so the wide change is covered rather than assumed.

**UTF-8 split — left it, widened the wording.** Reproduced both sides (real 128 echoed bytes / 172 total, here 130 / 174). I also considered trimming back to the last whole character boundary and rejected it: still not what real does, wrong in a new way, and a third behaviour to unwind when the byte pipeline lands. The comment now states the gap instead of implying exactness — that was the root of the earlier blocker and I would rather over-correct on it.

**`CONFIG HELP` — left it, but recorded properly.** I swept five real versions first, which turned up something that settles the question: the reply's last line changes at 7.2 (`Prints this help.` -> `Print this help.`), and `CONFIG HELP extra` forks between the arity error and the legacy template at 7.0. So byte-exact implementation needs a third gate plus the deferred arity mechanism. The captured payload and both boundaries are in the test-file header and on #413, so the expensive part is done for whoever takes it.

**`Number.isFinite` — took it,** with NaN/±Infinity coverage.

The #413 notes are up: https://github.com/fatal10110/js-redis-server/issues/413#issuecomment-5772684716 — your three points, attributed, plus three more that only surface once you try to use the helper (truncation has to live inside the gated branch; the byte-fidelity gap is inherited by every migrated site; and the 6.2 template doubles as the arity message, which is what makes the arity half overlap #384).

Results: unit 345/0, mock integration 1123/0, seven-profile sweep 10/0 each, real 8.0 on the four raw-TCP suites 25/0. Build, lint and `tsc --noEmit` clean. No `clean:redis`, no suite scripts against the shared backend.

tests-integration/raw-tcp/command-errors-config.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        `-ERR unknown subcommand '${'X'.repeat(128)}'. Try CONFIG HELP.\r\n`,
+      )
+    }
+  })
+
+  // Real Redis sanitizes an error body with `sdsmapchars(s, "\r\n", "  ", 2)`,
+  // a 1:1 character map, so a `\r\n` in the echoed subcommand becomes TWO
+  // spaces. Collapsing the run to one is protocol-safe but changes the byte
+  // count, which is exactly what this suite exists to pin.
+  test('CR and LF in the echoed subcommand are mapped 1:1, not collapsed', async () => {
```
fatal10110, line 136: **Verified, with one gap worth a third line.** The 1:1 mapping is correct and version-invariant — I probed nine CR/LF shapes against 6.2.24, 7.0.15, 7.2.1, 7.4.0 and 8.0.6, and this server now matches all five on every one:

```
input        real (all 5)   this server
a\rb      -> 'a b'          'a b'
a\nb      -> 'a b'          'a b'
a\r\nb    -> 'a  b'         'a  b'
a\n\rb    -> 'a  b'         'a  b'
a\r\r\rb  -> 'a   b'        'a   b'
a\n\n\nb  -> 'a   b'        'a   b'
a\r\nb\r\nc -> 'a  b  c'    'a  b  c'
\r\nX\r\n -> '  X  '        '  X  '
\r\n      -> '  '           '  '
```

**The gap:** the two cases asserted here are a `\r\n` pair and a lone `\n`. Neither pins a **lone `\r`** — and that is the one with protocol consequence, since a stray CR is what can sit next to a frame terminator. I confirmed it escapes: replacing the implementation with

```js
return value.replace(/\r\n/g, '  ').replace(/\n/g, ' ')
```

— which leaves a lone `\r` untouched in the error body — passes `core-resp-encoder`, `command-errors-config` and `protocol-errors` in full (17 pass / 0 fail). One more line closes it, and real's answer is already pinned above:

```js
await expectReply(
  conn,
  ['CONFIG', 'a\rb'],
  "-ERR unknown subcommand 'a b'. Try CONFIG HELP.\r\n",
)
```

Low severity — the shipped regex is correct, this only hardens the pin.

Separately, on whether four real suites was adequate coverage for a change that touches every error reply: it was, and I widened it anyway. `sanitizeErrorText` only alters bytes for a message containing a CR or LF, and no static error message in the repo contains one (`git grep` over `src/commands/` and `src/core/redis-error.ts` finds none) — so the blast radius is exactly the paths that echo client input. I ran **13** error-text-sensitive raw-TCP suites against real 8.0.6 (all six `command-errors-*`, `protocol`, `protocol-errors`, `resp3-wire`, `pubsub-protocol`, `standalone-auth`, `script-resp3-replies`, `monitor-protocol`): **69 pass / 0 fail**. And the fix generalizes — nine of the ten container commands now match real byte-for-byte on `<CONTAINER> "a\r\nb"`.

src/commands/helpers.ts
```diff
@@ -131,6 +133,68 @@ export function parsePositiveExpireToken(
...
 }
 
+/**
+ * Real Redis formats the echoed subcommand with `%.128s`, so a longer name is
+ * cut at 128 bytes. Verified exact against 7.0.15 and 8.0.6 for single-byte
+ * names: at 128 the reply is byte-identical to the mock's, at 129 and 300 real
+ * echoes 128 characters. Redis 6.2 has no truncation at all — a 300-byte name
+ * comes back whole.
+ *
+ * Known gap, shared with the non-UTF-8 case below: when the 128-byte cut lands
```
fatal10110, line 143: **The numbers check out.** I re-measured `'A'.repeat(127) + 'é'` on real 8.0.6 and on this server: real echoes 128 bytes, this echoes 130, and the `172` / `174` totals are consistent once "total" is read as the error text without the frame terminator (`-ERR unknown subcommand '` is 25 bytes, `'. Try CONFIG HELP.` is 19, so 128 + 44 = 172 and 130 + 44 = 174). I briefly thought it was off by two before noticing I was counting the CRLF. No change needed.

**One nuance the wording doesn't capture, worth a clause.** "128 echoed bytes against this server's 130" reads as a fixed +2, but the gap varies with *where* the cut lands, because Node emits one U+FFFD (3 bytes) per incomplete sequence regardless of how many orphan bytes it replaces:

```
input                       real 8.0.6            this server
127 x 'A' + U+00E9    echoed 128, tail ...c3      echoed 130, tail ...efbfbd   (+2)
127 x 'A' + U+1F600   echoed 128, tail ...f0      echoed 130, tail ...efbfbd   (+2)
126 x 'A' + U+1F600   echoed 128, tail ...f09f    echoed 129, tail ...efbfbd   (+1)
125 x 'A' + U+1F600   echoed 128, tail ...f09f98  echoed 128, tail ...efbfbd   (+0)
200 x U+00E9          echoed 128, tail ...c3a9    echoed 128, tail ...c3a9     (exact)
```

The last-but-one row is the one to record: when three orphan bytes are replaced by one 3-byte U+FFFD, **the reply is the same length as real's but different bytes**. A future test for #384 part 2 that checks only the echoed length would pass on that input while the bytes still diverge. Worth saying "0 to 2 bytes longer, and in one case the same length with different bytes" rather than a single figure, so whoever closes #384 part 2 writes a byte assertion rather than a length one.

The direction is right in every case — this server is never *shorter* than real, since U+FFFD is 3 bytes and an orphan run is 1 to 3 — so the comment does not under-claim.

I also agree with rejecting the trim-to-character-boundary option. Echoing 126-127 bytes is not real's answer either, and it would be a third state to unwind when the byte pipeline lands.

fatal10110 (commented):
## Final pass on 242c700 — everything checks out; ship

Verified on my own throwaway containers at 6.2.24, 7.0.15, 7.2.1, 7.4.0 and 8.0.6, all removed afterwards. Shared infra untouched.

### 1. `sanitizeErrorText` — the highest-risk item, and it is correct

You asked for more than the single `\r\n` pair, so I probed nine shapes against all five versions. Real Redis's mapping is strictly 1:1 and **version-invariant** — identical on 6.2 through 8.0 — and this server now matches on every one:

```
input          real (all 5 versions)   this server
a\rb        -> 'a b'                   'a b'
a\nb        -> 'a b'                   'a b'
a\r\nb      -> 'a  b'                  'a  b'
a\n\rb      -> 'a  b'                  'a  b'
a\r\r\rb    -> 'a   b'                 'a   b'
a\n\n\nb    -> 'a   b'                 'a   b'
a\r\nb\r\nc -> 'a  b  c'               'a  b  c'
\r\nX\r\n   -> '  X  '                 '  X  '
\r\n        -> '  '                    '  '
```

So three-or-more runs, lone `\r`, lone `\n`, reversed `\n\r`, multiple separated runs and leading/trailing runs are all right.

**Was the four-suite real run adequate?** Yes, and here is the reasoning rather than just the result. `sanitizeErrorText` only changes bytes for a message that contains a CR or LF, and **no static error message in this repo contains one** — `git grep` across `src/commands/` and `src/core/redis-error.ts` finds no literal newline in any error text. So the blast radius is exactly the paths that echo client input, which is what you picked. I widened it anyway: **13** error-text-sensitive raw-TCP suites against real 8.0.6 (all six `command-errors-*`, `protocol`, `protocol-errors`, `resp3-wire`, `pubsub-protocol`, `standalone-auth`, `script-resp3-replies`, `monitor-protocol`) — **69 pass / 0 fail**.

The fix also generalizes beyond CONFIG. Nine of the ten container commands now match real byte-for-byte on `<CONTAINER> "a\r\nb"` (see the XINFO note below for the tenth-and-a-half).

One coverage gap in the new test, inline.

### 2. The corrected unit assertion pins real's bytes, not a new guess

Both assertions are right:

- `RedisValue.error('bad\r\nframe', 'ERR')` -> `-ERR bad  frame\r\n` — two spaces, matching real's `a\r\nb -> 'a  b'` on all five versions.
- `RedisValue.error('one\nline', 'ERR')` -> `-ERR one line\r\n` — one space, matching real's lone-LF row.

Good instinct correcting it rather than deleting it: it had been pinning the collapsing behaviour, so it was a real assertion pointed at the wrong answer.

### 3. Both `CONFIG HELP` boundaries — CONFIRMED, and this settles the deferral

I swept the same five versions independently. Both claims hold exactly:

```
              elements   last payload line              CONFIG HELP extra
6.2.24           11      '    Prints this help.'        -ERR Unknown subcommand or wrong number of arguments for 'HELP'. Try CONFIG HELP.
7.0.15           11      '    Prints this help.'        -ERR wrong number of arguments for 'config|help' command
7.2.1            11      '    Print this help.'         -ERR wrong number of arguments for 'config|help' command
7.4.0            11      '    Print this help.'         -ERR wrong number of arguments for 'config|help' command
8.0.6            11      '    Print this help.'         -ERR wrong number of arguments for 'config|help' command
```

The wording boundary is at **7.2** and the arity fork is at **7.0**, as reported — and note we hit them on different patch levels (you used 7.2.16 / 7.4.11, I used 7.2.1 / 7.4.0), so the boundaries are version boundaries and not patch artefacts. Element count is 11 throughout.

That means a byte-exact `CONFIG HELP` needs a third gate for the payload wording **plus** the profile-aware arity path already deferred — two gates and a mechanism for one subcommand. That is a feature with its own compat surface, not an error-wording fix, and deferring it is right. Recording the full payload and both boundaries in the test-file header is the correct way to leave it.

### 4. `Number.isFinite` — covers the gap, changes nothing else

`Number.isFinite(x) ? Math.trunc(x) : 0` is the identity on every finite input, so nothing moves for real timestamps, and the existing cases (0, 1, 999999, second boundaries, `1695000000000000.5`, `1.5`) all still pass. Removing the guard turns `formats monitor timestamps with six fractional digits` red, so the new `NaN` / `±Infinity` assertions have teeth.

### 5. Mutation checks

| mutation | result |
|---|---|
| restore the collapsing regex `/[\r\n]+/g` | **2 fail** — the new raw-TCP case and the corrected unit assertion |
| drop the `Number.isFinite` guard | **1 fail** |
| an impl handling `\r\n` and lone `\n` but **not** a lone `\r` | **0 fail** — escapes; see inline |

### 6. Regressions — none

| suite | result | your claim |
|---|---|---|
| `tests/**` unit | 345 pass / 0 fail | 345/345 ✓ |
| mock integration, full | **1123 pass / 0 fail** | 1123/1123 ✓ |
| 13 raw-TCP suites vs real 8.0.6 | 69 pass / 0 fail | superset of your 25/25 |
| `tsc --noEmit`, `eslint` | clean | ✓ |

Every count you reported reproduces.

### One thing for #413 that neither of us had: `XINFO` has XGROUP's defect too

My re-review flagged `xgroup.ts:137` as needing the Buffer arm because its local `subcommand` is already `.toUpperCase()`d, and your #413 comment carries that. `xinfo.ts:101` has the **same** bug, and it is easier to miss because its *wording* is already correct, so a sweep that greps for the old template will skip it:

```
XINFO "a\r\nb"   real 8.0.6 : -ERR unknown subcommand 'a  b'. Try XINFO HELP.
                 this server: -ERR unknown subcommand 'A  B'. Try XINFO HELP.
XGROUP "a\r\nb"  real 8.0.6 : -ERR unknown subcommand 'a  b'. Try XGROUP HELP.
                 this server: -ERR Unknown subcommand or wrong number of arguments for 'A  B'. Try XGROUP HELP.
```

Both are the two schema-parser sites, both need `input[index]` rather than the uppercased local. Worth adding XINFO to the #413 checklist alongside XGROUP — right now it reads as already-done.

(The CR/LF halves of those two lines are correct on both, incidentally — that is the `sanitizeErrorText` fix generalizing.)

---

## Verdict: **ship**

Three low findings from my re-review, two of them things you could reasonably have left, and you took the riskiest one and did it properly — including fixing the assertion it broke instead of deleting it, and mutation-checking the result. The two you left are left correctly: the UTF-8 split genuinely needs the byte pipeline and the trim-to-boundary alternative would have been wrong in a new way; `CONFIG HELP` needs two gates and a mechanism, which your five-version sweep now demonstrates rather than asserts.

Both new claims I was asked to check — the 7.2 wording boundary and the 7.0 arity fork — are correct, and they were found by going and looking rather than by reasoning from the source. Same for the version-gated truncation last round. That is twice now that checking beat implementing what the review said.

Remaining: one inline test-coverage gap (lone `\r`), one wording nuance on the echo-limit comment, and XINFO for #413. None blocking.

fatal10110:
Merged `origin/main` (`c5df778`, #409) in as `851bc33` — merge commit rather than a rebase, so the history stays reviewable. `mergeable=MERGEABLE` now.

### What actually conflicted

Two files, both add/add rather than semantic:

**`src/commands/config.ts`** — one hunk, the `./helpers` import line. #409 added `INT64_MAX`, this branch added `unknownSubcommandError`; resolved as the union. Everything else in the file merged cleanly, and I verified both sides survived rather than trusting that: #409's `MEMORY_UNITS` Map, `parseMemoryValue` and `configSetFailed` are intact, as are this branch's `subcommand: t.bulk()` schema and the gated `unknownSubcommandError` throw.

**`tests-integration/compatibility/profile-gates.test.ts`** — not in the trial merge, but it conflicted here. Git interleaved two independently added tests into one broken hunk (my `if (!supportsUnknownSubcommandWording())` opening spliced onto #409's `CONFIG SET` body). Rebuilt as the union from the merge stages rather than hand-splicing the markers: #409's failure-wording/overflow test plus my two unknown-subcommand tests, and both profile helpers. 11 tests per profile now, from 10 on each side.

### Do the gates interact?

**No.** Three gates now sit at the identical `{ redis: '7.0.0', valkey: '7.2.0' }` boundary, but on mutually exclusive control-flow paths:

| gate | fires in | reachable when |
|---|---|---|
| `config.set.failure-message` | `configSet()` | subcommand **is** `set` |
| `config.memory-value.reject-overflow` | `parseMemoryValue()`, via `configSet()` | subcommand **is** `set` |
| `error.unknown-subcommand-wording` | `execute()`'s final throw | subcommand is **not** get/set/resetstat/rewrite |

Neither can shadow the other — the unknown-subcommand throw is only reached after every known subcommand has been ruled out, and the CONFIG SET gates are only reached after `set` has been matched. The seven-profile sweep exercises both sides of all three.

I checked the `feature-gates.ts` / `profile.ts` auto-merge rather than assuming it, as asked: all four entries are present in both the `FeatureId` union and the `FEATURE_GATES` table. Worth noting the type system had my back here anyway — `Record<FeatureId, VersionGate>` fails to typecheck if the union and the table diverge, so a half-merged gate could not have compiled.

### Do the two helpers overlap?

**No, and I am not merging them.** `configSetFailed(profile, name, value, detail)` is file-local and produces CONFIG SET's *failure detail*; `unknownSubcommandError(container, subcommand, profile)` is shared and produces the *container dispatch* error. Disjoint messages, disjoint call paths, disjoint audiences — #413 migrates the latter across 15 sites and has nothing to say about the former. Consolidating them would be a design change smuggled into a conflict resolution, so it is not in this commit.

### Also in this push: the lone `\r`

Good catch, and it was a real hole. The two assertions I added last round (`a\r\nbc`, `x\ny`) are both passed by an implementation that maps the `\r\n` pair and `\n` alone but leaves a bare `\r` — which is the one case with protocol consequence, since an unmapped `\r` ends the error line early and the client resynchronizes mid-frame. Verified by mutating `sanitizeErrorText` to exactly that (`.replace(/\r\n/g, '  ').replace(/\n/g, ' ')`): **red**, where before it was 17/0 green.

Captured from real 8.0.6 and now pinned: `'a\rb'` -> `'a b'`, `'p\r\rq'` -> `'p  q'`, `'m\n\nn'` -> `'m  n'`.

### Results

| run | result |
|---|---|
| 7-profile compatibility sweep | **11 pass / 0 fail on each of 7** |
| `npm test` (unit) | **348 pass, 0 fail** |
| `npm run test:integration:mock` | **1159 pass, 0 fail** |
| real 8.0: `command-errors-config`, `monitor-protocol`, `protocol-errors`, `command-errors-keyspace`, `proto-max-bulk-len` | **26 pass, 0 fail** |
| mutation: pair-then-LF sanitizer | **red** (1 fail) |
| `npm run build`, `npm run lint`, `tsc --noEmit` | clean |

I included #409's own `proto-max-bulk-len` raw-TCP suite in the real run, since that is the code my resolution touched most.

Issue #412, opened 2026-09-22
test: socketless client mocks have no integration coverage at all


fatal10110:
Found while reviewing PR #408. This is the structural reason bugs like #385 survive.

## Problem
**No file under `tests-integration/` uses `createNodeRedisMock`, `createIoredisMock` or `createInMemoryRedis`.** The integration suites all drive a real TCP server (mock or real backend) through the real client libraries. The socketless client facades — the ones this package exists to provide — are covered only by a handful of unit tests in `tests/`.

Consequences, all observed:
- #385 (flat-pairs decoded flat under RESP3) shipped and survived, because no integration test ever asked a socketless client for a `WITHSCORES` reply.
- PR #408's review found a scheduling-dependent RESP-version race in `NodeRedisMockCluster` that its own new unit tests could not see.
- `npm run test:integration:mock` passing tells you nothing about `createNodeRedisMock`.

The facades are also where divergence is most likely, because they hand-decode replies instead of going through the wire encoder/parser — so they reimplement, per client, what the real client libraries do.

## Proposal
Give the socketless clients the same integration treatment the TCP path gets. Options, roughly in order of value per effort:

1. Parameterize the existing `tests-integration/node-redis/**` harness over a third backend — `mock` (TCP), `real`, and `socketless` — so every existing node-redis assertion also runs against `createNodeRedisMock`. Same for ioredis. This reuses hundreds of existing assertions rather than writing new ones.
2. If (1) is too invasive, add a focused socketless suite covering the reply-shape surface: `WITHSCORES`/`WITHVALUES`, `HGETALL`, `CONFIG GET`, `XRANGE`/`XREAD`, `MULTI`/`EXEC` nesting, pub/sub pushes, cluster routing — at RESP2 and RESP3.
3. Cover `NodeRedisMockCluster` specifically. It has the most machinery and the least coverage.

Note `test-config.ts`'s `TEST_BACKEND` switch is the natural place for (1).

## Acceptance
- A reply-shape divergence in a socketless client fails an integration test, not just review.
- `NodeRedisMockCluster` is exercised by more than one test.

Issue #398, opened 2026-09-21
CI: redis-cluster container intermittently exits (1) at startup, failing -real jobs before any test runs (3 sightings)


fatal10110:
Observed **three separate times** in a single working session, across different PRs and different jobs.

## Symptom
A `-real` integration job fails at the **Start Redis (cluster + standalone)** step, before any test executes:

```
Container js-redis-server-redis-standalone-auth-1  Healthy
Container js-redis-server-redis-standalone-1       Healthy
container js-redis-server-redis-cluster-1 exited (1)
##[error]Process completed with exit code 1.
```

Both standalone containers report healthy; only the cluster container exits. Always cleared by a plain re-run of the failed job.

## Sightings
1. PR #378, `raw-real`
2. `main` at `cad826e`, `node-redis-real` (run 35627448738)
3. PR #394, `ioredis-real`

Three in one session suggests this is frequent, not exotic.

## Why it matters
- **It is indistinguishable at a glance from a real failure.** Someone checking a red PR sees a failed `-real` job and has to open logs to discover no test ran.
- It erodes trust in the check signal — the same erosion that let the browser demo stay broken for two months (fixed in #391) and that #394 is trying to prevent.
- If the `-real` jobs ever become required checks for branch protection, this becomes a recurring merge blocker.

## Investigation starting points
- `docker-compose.test.yml`'s `redis-cluster` service: what its entrypoint does and why it exits 1. Cluster formation (`--cluster create`) is timing-sensitive and may be racing its own node startup.
- Whether the container has a healthcheck at all, and whether `--wait` is actually waiting for cluster *formation* rather than process start.
- Whether the exit is deterministic under load (CI runners are noisy neighbours) — a retry or a proper readiness gate may be all that is needed.
- Capture `docker compose logs redis-cluster` on failure; the current workflow discards it, which is why three sightings produced no diagnostic detail.

## Related
- #395 — `clean:redis` is a silent no-op in CI, so the real-backend suites never start from a clean keyspace. Different root cause, same blast radius: unreliable `-real` jobs.

## Acceptance
- The failure is diagnosed (logs captured on failure at minimum).
- Either the race is fixed with a real readiness gate, or startup is retried, so a `-real` job's red genuinely means a test failed.

fatal10110:
New evidence from a local run, which may point at the root cause — and suggests the CI symptom is one of two faces of the same design flaw.

## Observed locally
[...]

PR #421, opened 2026-09-22
fix: node-redis mock close path throws ClientClosedError like the real client (#389)

### LINKED ISSUE #389: node-redis mock: double quit() does not throw ClientClosedError; close-path lifecycle diverges from real client
Found while re-reviewing PR #378. **Pre-existing on `main`** — the node-redis facade's `quit()` / `disconnect()` lifecycle diverges from the real client.

## Problem
Calling `quit()` twice on `NodeRedisMockClient` does not behave like real node-redis:

- **Real node-redis**: the second call throws `ClientClosedError`, and `'end'` is emitted exactly **once**.
- **Mock**: does not throw, and the event behavior differs.

A test in `tests/client-mocks/node-redis-mock.test.ts` currently pins the mock's behavior, so the divergence is enshrined.

**Failure scenario:** teardown code that defensively calls `quit()` in both a `finally` and an `afterEach` passes against the mock and throws against the real client — or an `'end'`-counting listener sees a different count.

## Scope
Worth auditing the whole close path together: `quit()`, `disconnect()`, `destroy()`, and the `'end'` / `'error'` emissions, against real node-redis v6. `ClientClosedError` should come from the `redis` package when present, the way the facade already resolves `WatchError` / `ErrorReply` / `MultiErrorReply` (see `ensureRedisErrors`).

## Acceptance
Double `quit()` throws `ClientClosedError`; `'end'` fires exactly once; the existing test is updated to pin the real behavior rather than the mock's.

fatal10110:
## Summary

`NodeRedisMockClient`/`NodeRedisMockCluster` let you close an already-closed
client silently, so teardown that defensively calls `quit()` in both a
`finally` and an `afterEach` passed against the mock and threw against the
real client. A test pinned that divergence rather than the real behaviour.

### Ground truth first

Before changing anything I drove **real node-redis v6.0.0** against a local
`redis-server` and recorded the whole close path. Every combination behaves
identically:

| sequence | result |
| --- | --- |
| `quit()` → `quit()` | 2nd throws `ClientClosedError('The client is closed')` |
| `quit()` → `disconnect()` | throws `ClientClosedError` |
| `disconnect()` → `quit()` | throws `ClientClosedError` |
| `disconnect()` → `disconnect()` | throws `ClientClosedError` |
| `destroy()` → `quit()` / `disconnect()` / `destroy()` | throws `ClientClosedError` (synchronously for `destroy()`) |
| command after any close | throws `ClientClosedError` |
| never-connected client | all three close methods throw `ClientClosedError` |

In every case `'end'` fired **exactly once** and `'error'` **never** fired on a
clean close. First-call return values: `quit()` → `'OK'`, `disconnect()` →
`undefined`, `destroy()` → `undefined` (sync).

### What changed

- `ClientClosedError` is resolved from the `redis` package the same way
  `WatchError` / `ErrorReply` / `MultiErrorReply` already are, via
  `ensureRedisErrors`, with a shape-compatible local fallback. A synchronous
  `clientClosedError()` helper reads the already-resolved cache, since
  `destroy()` must throw synchronously.
- Both clients gained an `assertOpen()` guard on `quit()` / `disconnect()` /
  `destroy()`. Because the guard runs *before* teardown, `'end'` stays a
  once-per-close signal and a clean close still emits no `'error'`.
- `runOnSession` now throws `ClientClosedError` for a command on a closed
  client instead of a generic `Error('node-redis mock client is closed')`.
- `docs/TESTING.md` documents the close-path contract and the
  `instanceof ClientClosedError` teardown idiom.

The `'end'`-counting test that previously noted the mock diverged now pins the
real once-only semantics, and the suite's own `afterEach` teardown was updated
to tolerate `ClientClosedError` — the same thing real-client users must do.

Closes #389

## Test plan

- [x] Ground-truth script drove real node-redis@6 against a real
      `redis-server` for double-quit, quit→disconnect, disconnect→quit,
      double-disconnect, destroy→quit/disconnect/destroy, command-after-close
      and never-connected, recording `'end'`/`'error'` counts for each
- [x] New unit tests red before the fix (6 failures), green after
- [x] Double `quit()` throws `ClientClosedError`; `'end'` fires exactly once
- [x] Same assertions for the cluster facade
- [x] `npm test` (383 tests) passes
- [x] `npm run build` and `npm run lint` clean

Integration suites are untouched: nothing under `tests-integration/` imports
the facade (`src/index.ts` only re-exports it), so the blast radius is the
unit tests above.


src/client-mocks/node-redis-mock.ts
```diff
@@ -99,6 +113,17 @@ async function ensureRedisErrors(): Promise<RedisErrorConstructors> {
   return resolvedRedisErrors
 }
 
+/**
+ * The error real node-redis throws for any call on a closed client — closing
+ * one that is already closed included. Synchronous because `destroy()` is:
+ * {@link ensureRedisErrors} has already run by the time any client exists
+ * (`createNodeRedisMock` awaits it), so the real class is available here.
+ */
+function clientClosedError(): Error {
```
fatal10110: **[high] The "already resolved by the time any client exists" claim is false — `NodeRedisMockCluster.create()` is a public export that never awaits `ensureRedisErrors()`.**

The JSDoc right above this line asserts the cache is always warm:

> `ensureRedisErrors` has already run by the time any client exists (`createNodeRedisMock` awaits it), so the real class is available here.

That holds only for the `createNodeRedisMock()` path. But `src/index.ts:47-48` exports `NodeRedisMockClient` and `NodeRedisMockCluster` as **values**, and `NodeRedisMockCluster.create({ masters })` is a synchronous static that constructs a fully working cluster without ever touching `ensureRedisErrors()`. In a process where `createNodeRedisMock()` is never called, `resolvedRedisErrors` stays `undefined`, `?? FALLBACK_REDIS_ERRORS` kicks in, and every close throws a `FacadeClientClosedError` — which fails the exact `instanceof` check this PR's docs and tests prescribe.

Reproduced on this head (`af9d3f9`), clean process, `redis@6.0.0` installed and importable:

```ts
import { ClientClosedError } from 'redis'
import { NodeRedisMockCluster } from '../src/index'
// createNodeRedisMock() deliberately never called
const cl = NodeRedisMockCluster.create({ masters: 3 })
await cl.quit()
await cl.quit()   // -> catch
```
```
2nd quit ctor                = FacadeClientClosedError
instanceof ClientClosedError = false
message                      = "The client is closed"
docs/TESTING.md idiom would  = RETHROW -> BROKEN
```

**Failure scenario:** a consumer follows the teardown idiom this PR adds to `docs/TESTING.md` (`if (!(err instanceof ClientClosedError)) throw err`) against a cluster built via the exported class. The guard does not match, the error is rethrown, and the `afterEach` fails with a confusing `Error: The client is closed` that *looks* like the thing they just handled. `redis` being installed makes no difference — the cache was simply never populated.

This is not a theoretical race: it is a deterministic miss on a supported public entry point. Two workable fixes:
- Resolve the classes synchronously at module load (`createRequire('redis')` in a try/catch) so `clientClosedError()` never has to guess; or
- keep the async resolve but have the `NodeRedisMockClient` / `NodeRedisMockCluster` constructors kick it off and make `clientClosedError()` the only thing that needs it — plus name the fallback class `ClientClosedError` (see the separate comment on line 75) so at least `constructor.name` and `util.inspect` degrade gracefully.

The same hole applies to a direct `new NodeRedisMockClient({...})`, which is also exported.

src/client-mocks/node-redis-mock.ts
```diff
@@ -912,21 +957,31 @@ export class NodeRedisMockCluster extends CommandRunner {
   // teardown is fully synchronous (no push-drain loop to await) — unlike the
   // standalone client whose teardown awaits its pub/sub drain.
   async quit(): Promise<string> {
+    this.assertOpen()
```
fatal10110: **[high] The cluster close contract is modelled on the standalone client, but real `RedisCluster` does neither of these things: it never throws `ClientClosedError` on a redundant close, and it never emits `'end'`.**

The standalone half of this PR is correct — I ground-truthed all of it. The cluster half copies a contract that the real class does not have. From `node_modules/@redis/client@6.0.0`:

`dist/lib/cluster/index.js:370-389` — `quit()`/`disconnect()` are *deprecated aliases*, and all four close methods delegate to the slots object:
```js
/** @deprecated Use `close` instead. */
quit() { return this._self._slots.quit(); }
/** @deprecated Use `destroy` instead. */
disconnect() { return this._self._slots.disconnect(); }
close() { ...; return this._self._slots.close(); }
destroy() { ...; return this._self._slots.destroy(); }
```

`dist/lib/cluster/cluster-slots.js:580-595` — the shared close path. Note `Promise.allSettled`, and the emitted event name:
```js
async #destroy(fn) {
    this.#isOpen = false;
    const promises = [];
    for (const client of this.#clients()) { promises.push(fn(client)); }
    ...
    this.#resetSlots();
    this.nodeByAddress.clear();
    this.#reconnectionTracker.clear();
    await Promise.allSettled(promises);
    this.#emit('disconnect');
}
```

Two concrete divergences this PR introduces:

1. **No `ClientClosedError` on a redundant close.** There is no open-check here at all — after the first close `#resetSlots()` has emptied the node list, so a second `quit()`/`disconnect()`/`destroy()` iterates nothing and resolves. And even if a node client did throw, `Promise.allSettled` swallows it. Grep confirms the class is not reachable from the cluster path at all:
   ```
   $ grep -rn "ClientClosedError" node_modules/@redis/client/dist/lib/cluster/
   (no matches)
   ```
2. **The event is `'disconnect'`, not `'end'`.** `NodeRedisMockCluster.quit()/disconnect()/destroy()` each `this.emit('end')`. Real `RedisCluster` emits only `'connect'`, `'error'` and `'disconnect'`:
   ```
   $ grep -rn "#emit(" node_modules/@redis/client/dist/lib/cluster/cluster-slots.js
   65:  this.#emit('connect');
   153: this.#emit('error', err);
   326: this.#emit('error', err);
   520: this.#emit('error', err);
   551: (destroy) this.#emit('disconnect');
   595: (#destroy) this.#emit('disconnect');
   ```
   The one `'end'` in that directory (`cluster-slots.js:437`) is a `.once('end', …)` *listener* on an individual node client, not an emission on the cluster.

**Failure scenario:** a user writes cluster teardown against the real client — `cluster.on('disconnect', …)` plus an unguarded double `close()` in a `finally` — and it works. Ported to the mock, the `'disconnect'` listener never fires and the second close throws `ClientClosedError`. The reverse is worse: they write it against the mock (guarding for `ClientClosedError` and listening on `'end'`, exactly as the new test at `tests/client-mocks/node-redis-mock.test.ts:570` and `docs/TESTING.md` tell them to), it passes, and against a real cluster the `'end'` listener is dead code and the guard is never exercised.

The new cluster test is named *"the close path matches the standalone client"* — which is precisely the assumption that does not hold. `#412` flags the cluster class as the least-covered here, and this is an instance of it: the guard was applied by symmetry rather than ground-truthed.

Note the pre-existing `'end'` emission is not this PR's doing, but the PR now pins it in a test and inherits it into a documented contract, so it is worth settling here rather than enshrining it. At minimum the cluster's `assertOpen()` should go, and `emit('end')` should become `emit('disconnect')`.

src/client-mocks/node-redis-mock.ts
```diff
@@ -494,15 +519,29 @@ export class NodeRedisMockClient extends CommandRunner {
...
+   */
+  private assertOpen(): void {
+    if (this.closed) {
+      throw clientClosedError()
+    }
+  }
+
   /** Gracefully close: tear down pub/sub + the command session. */
   async quit(): Promise<string> {
+    this.assertOpen()
```
fatal10110, line 664: **[medium] `quit()` is the *graceful* close: real node-redis drains every pending command before closing. The mock rejects them all.**

`#389` asked for the whole close path to be audited against real node-redis v6, and this is the part that got missed. The redundant-close semantics this PR adds are right — I verified all of them — but the first, *successful* `quit()` still behaves like a hard close.

Ground truth, `redis@6.0.0` against a real `redis-server`, 2000 commands issued and then closed without awaiting them:

```js
const ps = []
for (let i = 0; i < 2000; i++) ps.push(c.get('pr421:bulk:' + i).then(() => 'ok', e => 'ERR:' + e.constructor.name))
const closeRes = await c[method]()
```
```
[REAL quit() with 2000 pending]       close=resolved:OK        resolved=2000/2000 errors={}                                      events=["end"]
[REAL disconnect() with 2000 pending] close=resolved:undefined resolved=0/2000    errors={"ERR:DisconnectsClientError":2000}     events=["end"]
[REAL destroy() with 2000 pending]    close=sync-ok            resolved=0/2000    errors={"ERR:DisconnectsClientError":2000}     events=["end"]
```

That is the whole point of `quit()` vs `disconnect()`: `quit()` waits, the other two flush. The mock on this head does not distinguish them:

```
[MOCK quit mid-command]  quit: resolved:"OK" | in-flight: REJECTED:ClientClosedError|msg="The client is closed" | events: ["end"]
```

`teardown()` sets `this.closed = true` in its synchronous prefix before `session.close()`, so a command issued *before* `quit()` but not yet dispatched reads `this.closed === true` at `runOnSession` and is rejected retroactively. The close is applied to commands that were already in flight when it was requested.

**Failure scenario:** the common fire-and-forget-then-quit shape —
```js
client.set('k', 'v')        // not awaited
await client.quit()
```
— resolves cleanly against real node-redis (the SET lands, then the socket closes). Against the mock the SET rejects, and because nothing awaits it, it surfaces as an unhandled rejection that fails the test run from somewhere unrelated to the line that caused it. The `quitIfOpen()` helper this PR adds to the test file makes the shape *more* likely, since teardown now closes unconditionally.

This is pre-existing rather than introduced here — but `docs/TESTING.md:352` now states "The close path matches real node-redis v6", which makes it a documented guarantee. Either drain the session in `quit()` (await in-flight work before flipping `closed`), or narrow the doc claim to the redundant-close semantics that were actually verified.

src/client-mocks/node-redis-mock.ts
```diff
@@ -512,6 +551,7 @@ export class NodeRedisMockClient extends CommandRunner {
    * The push-drain loop settles on the next tick via the abort signal.
    */
   destroy(): void {
+    this.assertOpen()
```
fatal10110: **[medium] `destroy()` / `disconnect()` flush in-flight commands with `DisconnectsClientError`, not `ClientClosedError`.**

The PR's comment on `runOnSession` says "Real node-redis rejects a command on a closed client with this exact error, same as it does a redundant `quit()`/`disconnect()`/`destroy()`." That is true for a command issued **after** the close completes — I verified it for all three. It is *not* true for a command that was already in flight when the close landed; real node-redis uses a different class for that case.

`node_modules/@redis/client@6.0.0/dist/lib/client/index.js:1150`:
```js
this._self.#queue.flushAll(new errors_1.DisconnectsClientError());
```

Ground truth against a real `redis-server`:
```
[REAL destroy mid-command] inflight => rejected:DisconnectsClientError|name="Error"|msg="Disconnects client" | events: ["end"]
```
vs. this head:
```
[MOCK destroy mid-command] destroy: returned undefined | in-flight: REJECTED:ClientClosedError|msg="The client is closed" | events: ["end"]
```

Two distinct errors, and node-redis exports both (`DisconnectsClientError` is in the root `redis` export surface alongside `ClientClosedError`).

**Failure scenario:** a test harness that hard-kills a client mid-request and asserts on the rejection — `catch (e) { if (e instanceof DisconnectsClientError) ignore() }`, or just a log line that greps for `Disconnects client` — passes against real node-redis and fails against the mock, with the mock's error looking plausible enough that the mismatch reads as a flake rather than a mock gap.

This is the "a fallback that is subtly different is worse than none" hazard applied to the wire errors: the PR moved this path from an obviously-fake `Error('node-redis mock client is closed')` to a real node-redis class, which is an improvement, but it is now the *wrong* real class for this particular transition, and the wrongness is much harder to notice.

Fix: plumb a reason through `teardown()` so the in-flight flush uses `DisconnectsClientError` for `disconnect()`/`destroy()`, and keep `ClientClosedError` for commands issued after the client is already closed. (`quit()` should not reject them at all — see the comment on line 536.)

src/client-mocks/node-redis-mock.ts
```diff
@@ -494,15 +519,29 @@ export class NodeRedisMockClient extends CommandRunner {
 
   // --- lifecycle -----------------------------------------------------------
 
+  /**
+   * Real node-redis rejects every close call on an already-closed client with
+   * `ClientClosedError` — `quit()`/`disconnect()` as a rejection, `destroy()`
+   * synchronously — so defensive teardown that closes twice must see the same
+   * error here. 'end' therefore stays a once-per-close signal.
+   */
+  private assertOpen(): void {
```
fatal10110, line 647: **[medium] Deferring `connect()` is not neutral any more — `closed` is now a one-way latch, so `connect()` resolves successfully on a permanently dead client.**

The PR notes that real node-redis can `connect()` again after `quit()` and left the facade's no-op `connect()` (line 404) alone. I checked whether that deferral is safe against the new guards, and it is not quite: nothing ever resets `closed`, so `assertOpen()` turns it into a latch.

Real node-redis v6 against a real `redis-server`:
```
[REAL reconnect after quit] connect() resolved | get => resolved:null | 2nd quit resolved | events: ["end","end"]
```
The client genuinely comes back, serves commands, and emits a **second** `'end'` on the next close.

This head:
```
[MOCK reconnect after quit] connect: resolved:"this"
                            get:     REJECTED:ClientClosedError|msg="The client is closed"
                            quit2:   REJECTED:ClientClosedError|msg="The client is closed"
                            events:  ["end"]
```

So `await client.connect()` resolves — signalling success — and then every single command and every close on that "connected" client throws. A resolved `connect()` that hands back an unusable client is a worse failure mode than an explicit throw, because the error surfaces one call later, at a site that has nothing to do with the lifecycle mistake.

For completeness on what changed: `origin/main` was also broken here (`get` after reconnect threw `Error: node-redis mock client is closed`), but its unguarded `quit()` happened to resolve `'OK'` and emit the second `'end'`, accidentally matching real node-redis on the event count. This PR trades that accident for a throw. Net it is still the right direction — the double-`quit()` bug in `#389` is real and the fix is correct — but the reconnect path went from accidentally-right to definitely-wrong, so "leave `connect()` alone" is no longer a free deferral.

Cheapest honest options, in order of preference:
1. Have `connect()` reset `closed = false` and rebuild the session, matching real node-redis (and then a second close legitimately emits a second `'end'`, as the real client does).
2. Failing that, make `connect()` on a closed client **throw** `ClientClosedError` rather than resolve, so the failure lands on the offending line.
3. At minimum, document in `docs/TESTING.md` that the facade's `connect()` is a no-op and a closed client cannot be reopened — right now the doc says the close path matches real node-redis, which a reader will reasonably extend to this.

Unrelated but adjacent, in case it informs the choice: real node-redis rejects a second `connect()` on an *open* client with `Error: Socket already opened`; the facade no-ops that too. Pre-existing, not this PR's problem.

src/client-mocks/node-redis-mock.ts
```diff
@@ -69,10 +70,19 @@ class FacadeMultiErrorReply extends FacadeErrorReply {
   }
 }
 
+// Mirrors node-redis' own class exactly — message included, and (like the real
+// one) it leaves `name` at 'Error'.
+class FacadeClientClosedError extends Error {
```
fatal10110: **[low] The `name` claim checks out, but "mirrors node-redis' own class exactly" does not — `constructor.name` and `util.inspect` both leak `Facade`. Renaming the class fixes it for free.**

Credit where due: I verified the deliberate choice in the comment, and it is correct. Real node-redis does **not** set `name`, so both classes report `'Error'` — `node_modules/@redis/client@6.0.0/dist/lib/errors.js:29-33`:
```js
class ClientClosedError extends Error {
    constructor() {
        super('The client is closed');
    }
}
```
(Worth noting the sibling `FacadeWatchError` / `FacadeMultiErrorReply` above *do* set `name`, which real node-redis also does not — so this class is the consistent one and those two are the outliers. Out of scope here.)

Axis-by-axis comparison of the two classes, same process, `redis@6.0.0`:

| axis | real `ClientClosedError` | `FacadeClientClosedError` | |
| :-- | :-- | :-- | :-- |
| `message` | `"The client is closed"` | `"The client is closed"` | match |
| `name` | `"Error"` | `"Error"` | match |
| `String(err)` | `"Error: The client is closed"` | `"Error: The client is closed"` | match |
| `err.stack` first line | `"Error: The client is closed"` | `"Error: The client is closed"` | match |
| `JSON.stringify(err)` | `{}` | `{}` | match |
| `Object.keys(err)` | `[]` | `[]` | match |
| **`err.constructor.name`** | `"ClientClosedError"` | **`"FacadeClientClosedError"`** | **differs** |
| **`util.inspect(err)`** | `"ClientClosedError: The client is closed"` | **`"FacadeClientClosedError: The client is closed"`** | **differs** |
| `instanceof ClientClosedError` | `true` | `false` | differs (inherent) |

The `instanceof` gap is unavoidable for a fallback and is fine. The other two are not: `util.inspect` is what `console.error(err)` and every node:test assertion dump prints, and `constructor.name` is a common duck-type check. Both are one-word fixes — name the class `ClientClosedError`:

```ts
class ClientClosedError extends Error {
  constructor() {
    super('The client is closed')
  }
}
```

It is module-private, so there is no collision with the imported real class, and `FALLBACK_REDIS_ERRORS` reads more honestly too. This matters more than it looks given the bug on line 122, where the fallback is reachable even with `redis` installed — at which point a user staring at `FacadeClientClosedError: The client is closed` in a test dump has no idea why.

docs/TESTING.md
```diff
@@ -349,6 +349,24 @@ await client.sendCommand(['HSET', 'h', 'f1', 'a']) // escape hatch
 await client.quit() // tears down the in-memory state
 ```
 
+The close path matches real node-redis v6: `quit()` resolves `'OK'`,
```
fatal10110: **[low] "The close path matches real node-redis v6" is broader than what was verified — three sub-claims here do not hold, and the idiom below needs a caveat.**

What this paragraph asserts is mostly right, and I confirmed each piece against `redis@6.0.0` driving a real `redis-server`:

```
[quit->quit]               quit => resolved:"OK"       | quit  => ClientClosedError "The client is closed" | events: ["end"]
[quit->disconnect]         quit => resolved:"OK"       | disc  => ClientClosedError                        | events: ["end"]
[disconnect->quit]         disc => resolved:undefined  | quit  => ClientClosedError                        | events: ["end"]
[disconnect->disconnect]   disc => resolved:undefined  | disc  => ClientClosedError                        | events: ["end"]
[destroy->quit]            dest => returned:undefined  | quit  => ClientClosedError                        | events: ["end"]
[destroy->destroy]         dest => returned:undefined  | dest  => SYNC-THROW ClientClosedError             | events: ["end"]
[quit->destroy]            quit => resolved:"OK"       | dest  => SYNC-THROW ClientClosedError             | events: ["end"]
```
No `'error'` in any permutation, `'end'` exactly once, `destroy()` genuinely throwing from a bare `try`/`catch` with no `await`. All eight permutations match the mock. That part of the PR is solid.

The sentence overreaches on three points, all detailed in separate comments:
1. **`quit()` is not graceful in the mock.** Real `quit()` resolves all 2000 pending commands; the mock rejects them with `ClientClosedError` (line 536).
2. **In-flight commands at `disconnect()`/`destroy()` get the wrong class** — real node-redis flushes them with `DisconnectsClientError('Disconnects client')` (line 554).
3. **The cluster does not match at all** — real `RedisCluster` emits `'disconnect'`, never `'end'`, and does not throw `ClientClosedError` on a redundant close (line 960). As written, this paragraph sits in a section that covers both facades, so a reader will apply it to `createNodeRedisMock({ cluster })` too.

Suggest narrowing to what was verified, e.g. *"Closing an already-closed standalone client throws `ClientClosedError` …"*, and saying explicitly that `quit()` does not wait for in-flight commands the way the real client does.

Separately, on the snippet below: `import { ClientClosedError } from 'redis'` only works when `redis` is actually installed. The facade exists precisely so it can be used without it, and in that case the thrown error is the module-private fallback, which this `instanceof` will not match — so the documented idiom silently inverts (rethrows instead of swallowing) in exactly the configuration the fallback was written for. Worth one sentence: *"requires `redis` to be installed; without it the facade throws a structurally identical stand-in that `instanceof` will not match."* See also line 122 — today that stand-in is reachable even **with** `redis` installed, which is the bug that makes this caveat urgent rather than academic.

(Ignoring the textual overlap with #427 in this file — different region, and no semantic conflict: #427 only touches reply-shape decoding, nothing in the close path.)

src/client-mocks/node-redis-mock.ts
```diff
@@ -912,21 +957,31 @@ export class NodeRedisMockCluster extends CommandRunner {
...
   }
 
   destroy(): void {
+    this.assertOpen()
     this.teardown()
     this.emit('end')
   }
 
+  /** See {@link NodeRedisMockClient.assertOpen} — same node-redis contract. */
+  private assertOpen(): void {
```
fatal10110: **[low] `assertOpen()` and `closed` are duplicated verbatim across the two classes — both already extend `CommandRunner`.**

This is a byte-for-byte copy of `NodeRedisMockClient.assertOpen()` (line 528), and the `private closed = false` fields at lines 384 and 881 are duplicated the same way. Both classes extend `CommandRunner` (line 185), which is exactly the shared base this belongs on:

```ts
abstract class CommandRunner extends EventEmitter {
  protected closed = false
  protected assertOpen(): void {
    if (this.closed) throw clientClosedError()
  }
}
```

That also removes the `/** See {@link NodeRedisMockClient.assertOpen} — same node-redis contract. */` pointer, which is the tell that the duplication was already noticed.

Worth resolving in the same pass as the cluster comment on line 960 though — if the cluster's close contract turns out to be genuinely different from the standalone client's (it is: `'disconnect'` not `'end'`, and no `ClientClosedError`), then hoisting the guard would be the wrong move and the two `assertOpen()` bodies should not be unified after all. Whichever way that lands, the current state — identical code plus a comment asserting the contracts are the same, when they are not — should not survive it.

Minor, while here: the cluster's `teardown()` guard (`if (this.closed) return`) is now unreachable through every caller, same as the standalone one. The standalone version at least got a JSDoc explaining why it is kept deliberately; the cluster's did not.

fatal10110:
Pushed `94efd40` addressing the review. Both blockers fixed; 5 of the 6 non-blocking findings fixed, 1 deferred with a follow-up issue and the docs narrowed accordingly.

### Blockers

**1 — `clientClosedError()` cold cache (HIGH).** Correct, and worse than stated: `NodeRedisMockClient` is exported too, so `create()` was not the only cold entry point. Rather than chase entry points I made resolution eager — `void ensureRedisErrors()` now runs at module load, so the cache is warm by the time any caller has awaited anything. The false JSDoc is gone; it now states the real precondition and the residual `instanceof` caveat, which is also documented in TESTING.md per finding 7. Fix 2 independently removes the cluster's dependency on this error entirely.

**2 — cluster close contract backwards (HIGH).** Confirmed, twice. The shared cluster fixture on :30000 was `cluster_state:fail`, so I stood up my own throwaway 3-node cluster on 7911-7913 and probed it, and separately read `cluster/cluster-slots.js`. Both agree:

```
### cluster: double quit()
  - quit#1 -> resolved undefined
  - quit#2 -> resolved undefined
  events: {"end":0,"error":0,"connect":1,"disconnect":2}
```

- `'disconnect'`, never `'end'` — no `'end'` emit exists anywhere under `cluster/`.
- Redundant close never throws: `#destroy()` resets the slot/node maps first, so the second call finds nothing, awaits `Promise.allSettled([])` and emits `'disconnect'` **again**. The event is once per close *call*, not once per transition.
- All three resolve `undefined` — the cluster's `quit()` returns `#destroy()`'s promise, not `'OK'`. My mock was returning `'OK'`; that is fixed too.

Implementation and test both rewritten. One deliberate deviation is kept and labelled as such in code, test and docs: a command on a closed cluster throws `ClientClosedError`, where real v6 throws `TypeError: Cannot read properties of undefined (reading 'replicas')` from its own just-reset slot map. That is an upstream crash, not a contract.

### Also fixed

**3 — `quit()` now drains.** `closed` is claimed synchronously so *new* commands are refused, but the teardown is queued through the same FIFO `runExclusive` lock, so commands already issued run first — the queue-a-QUIT-command behaviour. Test issues 500 un-awaited commands before `quit()` and asserts zero rejections. The open/closed decision moved from execution time to *issue* time, which is what made the old code retroactively kill queued commands. `multi()` got the same treatment.

**4 — `DisconnectsClientError`.** Added to the resolved-error set and thrown by `disconnect()`/`destroy()` for in-flight commands, via a `flushPendingWith` hook set synchronously so nothing slips past.

**6 — class renamed** to `ClientClosedError` (and the new `DisconnectsClientError` likewise), so `constructor.name` and `util.inspect` no longer leak `Facade`. Comment corrected to claim only what is true.

**7 — docs narrowed.** The close path is now a per-method table, split single-client vs cluster, with the `instanceof`-inverts-without-`redis` caveat and the known gap called out.

**8 — hoisted.** `closed` moved to `CommandRunner`. `assertOpen()` is no longer duplicated at all: after fix 2 the cluster does not throw, so it only exists on the single client.

### Deferred — #440

**5 — `connect()` cannot re-open a closed client.** You are right that it is not safe and that `main` was accidentally closer. Real node-redis reconnects, serves commands and emits a second `'end'`; my probe confirms `end: 2`. Making that work needs `RedisServerState.close()` to stop being terminal (re-arm active expiry, clear the flag) — a state-layer change outside this PR's file ownership and used by everything.

Interim: `connect()` on a closed client now rejects with `ClientClosedError` instead of silently resolving a dead client — a loud failure rather than a silent lie. Pinned by a test explicitly named a known gap so it cannot regress unnoticed, cross-referenced from code, test and docs, and tracked in **#440** with a fix sketch and the probe output.

### Verification

`npm test` 387/387, `npm run build`, `npm run lint`, prettier all clean. Throwaway cluster on 7911-7913 shut down; shared fixtures on 6399/6400/30000-30005 untouched (I did hit an unrelated authenticated service on 6502 while picking ports — `NOAUTH` refused it, nothing happened).

src/client-mocks/node-redis-mock.ts
```diff
@@ -494,15 +596,44 @@ export class NodeRedisMockClient extends CommandRunner {
...
+   * queue, so every command already issued runs to completion first — 2000
+   * pending commands all resolve. Queueing the teardown through the same
+   * {@link runExclusive} lock reproduces that: the lock is FIFO, so by the time
+   * teardown runs, nothing issued before the `quit()` is left.
+   */
   async quit(): Promise<string> {
+    this.assertOpen()
+    // Claim the closed flag synchronously so commands issued *after* this point
+    // are refused (assertOpen in run()) while the ones before it still drain.
+    this.closed = true
```
fatal10110, line 667: **[medium] New in the fix: a draining `quit()` can no longer be escalated to a hard close. `destroy()` during a stuck `quit()` throws without flushing, so `quit()` and the blocked command hang forever.**

Setting `closed = true` here, before the drain, means a later `destroy()`/`disconnect()` fails `assertOpen()` and throws **before** `flushInFlight()` runs. Real node-redis runs the flush first: `destroy()` is `#queue.flushAll(new DisconnectsClientError())` followed by `#socket.destroy()`, and only the second step throws.

Ground truth, `redis@6.0.0` against a real `redis-server` (private instance on 17814). `BLPOP 0` blocks, `quit()` waits behind it, and `destroy()` escalates:
```
[REAL BLPOP 0 + quit + destroy] race= TIMEOUT | destroy= SYNC THROW ClientClosedError: The client is closed
                                | quit= rej:DisconnectsClientError: Disconnects client
                                | blpop= rej:DisconnectsClientError: Disconnects client
```
The same sequence on this head (`94efd40`):
```
[MOCK BLPOP 0 + quit + destroy] race= TIMEOUT | destroy= SYNC THROW ClientClosedError: The client is closed
                                | quit= STILL HANGING | blpop= STILL HANGING
```
It is not limited to blocking commands. With 200 plain `GET`s pending, calling `quit()` without awaiting it and then `destroy()` gives this on real node-redis:
```
[REAL quit(unawaited)->destroy] quit=rej:DisconnectsClientError  destroy=SYNC-THROW ClientClosedError  pending={"ERR:DisconnectsClientError: Disconnects client":200}
[MOCK quit(unawaited)->destroy] quit=ok:OK                        destroy=SYNC-THROW ClientClosedError  pending={"ok":200}
```
(`disconnect()` in place of `destroy()` gives the same result, since real `disconnect()` is `Promise.resolve(this.destroy())`.)

**Failure scenario:** the usual "graceful, then force" teardown
```ts
await Promise.race([client.quit(), timeout(1000)]).catch(() => {})
try { client.destroy() } catch {}
```
works against real node-redis: the forced close rejects everything that is stuck and the process exits. Against the mock, the `destroy()` is swallowed as a `ClientClosedError`, `teardown()` never runs, the owned `RedisServerState` is never closed (its active-expiry timer stays live), and the test run hangs.

Fix: in `destroy()`/`disconnect()`, when a `quit()` is draining (`closed && !tornDown`), set `flushPendingWith` and reject the drain **before** throwing `ClientClosedError`, the same order real node-redis uses. Worth a test that uses a `BLPOP 0` so it cannot pass by accident.

src/client-mocks/node-redis-mock.ts
```diff
@@ -99,6 +123,40 @@ async function ensureRedisErrors(): Promise<RedisErrorConstructors> {
   return resolvedRedisErrors
 }
 
+// `createNodeRedisMock()` awaits resolution, but `NodeRedisMockClient` is also
+// constructible directly, so start resolving at module load too: by the time
+// any caller has awaited anything at all, the cache is warm and the *real*
+// classes are what get thrown. Without this a directly-constructed client could
+// throw the local fallback, and `err instanceof ClientClosedError` imported
+// from `redis` would silently be false.
+void ensureRedisErrors()
```
fatal10110: **[medium] Eager resolution narrows the cold-cache window but does not close it. In the shipped CJS build, the documented idiom still rethrows on a public construction path, and every consumer now loads `redis` at import time.**

The comment says the cache is warm "by the time any caller has awaited anything at all". It is not. `void ensureRedisErrors()` starts a dynamic `import('redis')`, and that takes longer than a macrotask to settle, even when `redis` is already in the module cache.

`NodeRedisMockClient` is exported from `.`, and `RedisServerState` + `createRedisCommandExecutor` are exported from `js-redis-server/core` (`tsup.config` maps `core` to `src/internal.ts`), so direct construction is a supported public path. Against the built `dist/` of this head (`npx tsup`), in a fresh process:
```js
const { ClientClosedError } = require('redis')
const { NodeRedisMockClient } = require('js-redis-server')
const { RedisServerState, createRedisCommandExecutor } = require('js-redis-server/core')
const c = new NodeRedisMockClient({ state: new RedisServerState({ databaseCount: 16 }), executor: createRedisCommandExecutor(), ownsState: true })
await c.set('k', 'v'); await c.quit(); await c.quit()   // -> catch
```
```
[dist CJS, direct new NodeRedisMockClient, 2 awaits in] 2nd quit ctor=ClientClosedError instanceof real=false
  docs idiom `if (!(err instanceof ClientClosedError)) throw err` => RETHROWN
```
Here is how long the window stays open, measured with a closed cluster and `get()` after close:
```
source (tsx):  sync=false  after await=false  after setImmediate=false  after setTimeout(10)=true
dist CJS:      after await=false  after setImmediate=false
dist ESM:      after await=false  after setImmediate=true
```
The rename to `ClientClosedError` does fix `constructor.name` and `util.inspect`, which was finding 6 from the first review. `instanceof` is still false on this path, and that is what the idiom in `docs/TESTING.md` relies on. The docs caveat also says the fallback only applies "if `redis` is not installed", which the output above disproves.

**New side effect:** the package declares `"sideEffects": false`, but loading `index` now imports `redis` (about 270 ms for `require('redis')` here) for every consumer, including ones that only use `createIoredisMock`:
```
[ioredis-only consumer] @redis/client loaded before: false after import+50ms: true
```

A fix that closes the window instead of racing it: resolve synchronously. Use `createRequire(import.meta.url)('redis')` in a try/catch (or plain `require` in the CJS build), and call it lazily from `closeError()` on first use. That removes the module-load side effect, and `destroy()` can stay synchronous. At minimum, the two comments and the docs caveat should describe the window instead of saying it does not exist.

docs/TESTING.md
```diff
@@ -349,6 +349,53 @@ await client.sendCommand(['HSET', 'h', 'f1', 'a']) // escape hatch
...
+The **single client's** close path follows real node-redis v6:
+
+| | behaviour |
+| --- | --- |
+| `quit()` | resolves `'OK'`; **graceful** — commands already issued still run |
+| `disconnect()` | resolves `undefined`; flushes in-flight commands with `DisconnectsClientError` |
+| `destroy()` | returns `undefined` synchronously; same flush as `disconnect()` |
+| `'end'` | emitted exactly **once** per close |
+| `'error'` | never emitted by a clean close |
+| any call on a closed client | throws `ClientClosedError` (`'The client is closed'`) — a rejection from `quit()`/`disconnect()`, synchronous from `destroy()` |
```
fatal10110: **[low] On a closed client, real `disconnect()` throws `ClientClosedError` synchronously. It does not return a rejected promise.**

I missed this in the first review because my harness caught sync throws and rejections the same way. The row above puts `disconnect()` with `quit()` as "a rejection". The real client's code is `node_modules/@redis/client@6.0.0/dist/lib/client/index.js:1116`:
```js
disconnect() {
    return Promise.resolve(this.destroy());
}
```
`this.destroy()` throws before `Promise.resolve` is reached. Ground truth, measured with `try { const p = c.disconnect(); … } catch`:
```
[REAL quit then disconnect()]       SYNC THROW ClientClosedError: The client is closed
[REAL disconnect then disconnect()] SYNC THROW ClientClosedError: The client is closed
[REAL destroy then disconnect()]    SYNC THROW ClientClosedError: The client is closed
[REAL quit then quit()]             returned promise -> REJECTED ClientClosedError: The client is closed
```
This head (`async disconnect()`):
```
[MOCK quit then disconnect()]       returned promise -> REJECTED ClientClosedError: The client is closed
```
**Failure scenario:** `client.disconnect().catch(() => {})` in teardown passes against the mock and throws an uncaught `ClientClosedError` against real node-redis, because a sync throw never reaches `.catch`. `await` inside `try` behaves the same either way, so this only affects the promise-chain style.

Fix: make `disconnect()` a non-`async` method that calls `assertOpen()` before returning a promise. Update this row, and change the existing `assert.rejects(() => …disconnect(), ClientClosedError)` assertions to `assert.throws`.

src/client-mocks/node-redis-mock.ts
```diff
@@ -908,23 +1064,37 @@ export class NodeRedisMockCluster extends CommandRunner {
...
   // teardown is fully synchronous (no push-drain loop to await) — unlike the
-  // standalone client whose teardown awaits its pub/sub drain.
-  async quit(): Promise<string> {
+  // standalone client, whose teardown awaits its pub/sub drain.
+
+  async quit(): Promise<void> {
     this.teardown()
-    this.emit('end')
-    return 'OK'
+    this.emit('disconnect')
```
fatal10110: **[low] The cluster emits `'disconnect'` synchronously inside `quit()`/`disconnect()`. Real `RedisCluster` emits it after the call returns (only `destroy()` emits synchronously).**

Real `quit()`/`disconnect()`/`close()` go through `async #destroy()`, which emits after `await Promise.allSettled(...)`. `destroy()` emits inline. Ground truth on a private 3-node cluster (17811-17813), logging `const p = c[m](); log('after-call-sync'); await p; log('after-await')`:
```
[REAL cluster ordering quit]       ["after-call-sync","disconnect","after-await"]
[REAL cluster ordering disconnect] ["after-call-sync","disconnect","after-await"]
[REAL cluster ordering close]      ["after-call-sync","disconnect","after-await"]
[REAL cluster ordering destroy]    ["disconnect","after-call-sync","after-await"]
```
This head:
```
[MOCK cluster ordering quit]       ["disconnect","after-call-sync","after-await"]
[MOCK cluster ordering disconnect] ["disconnect","after-call-sync","after-await"]
[MOCK cluster ordering destroy]    ["disconnect","after-call-sync","after-await"]
```
**Failure scenario:** code that sets state after calling `cluster.quit()` and reads it in a `'disconnect'` handler (e.g. `closing = true` for "expected vs unexpected disconnect" logging) works against real node-redis and misfires against the mock. This is a small fix: `await Promise.resolve()` before the emit in `quit()`/`disconnect()`, and keep `destroy()` synchronous.

Everything else about the rewritten cluster contract matches what I measured on that cluster: every close resolves `undefined`, `destroy()` returns `undefined` synchronously, a redundant close never throws, `'disconnect'` fires once per call, and `'end'` never fires. The one deliberate deviation (`ClientClosedError` instead of the upstream `TypeError: Cannot read properties of undefined (reading 'replicas')`, which is a rejection, not a sync throw) also looks right to me.

fatal10110:
## Re-review of `94efd40` (previous review: `af9d3f9`)

Ground truth for this pass is `redis@6.0.0` driving a private `redis-server` 7.2.1 on 17814 and a private 3-node cluster on 17811-17813. I also re-read `@redis/client/dist/lib/{client,cluster}` and built `dist/` for this head to test the shipped artifacts.

### Status of the previous 8 findings

| # | was | status | notes |
|---|---|---|---|
| 1 | **high**: cold error-class cache | **not fixed (narrowed)** | Eager `void ensureRedisErrors()` does not warm the cache after an `await` or a `setImmediate`. In the built CJS output, a directly constructed `NodeRedisMockClient` (public via `.` + `./core`) that runs `set` → `quit` → `quit` still gets an error that is not `instanceof` the real class, so the documented idiom rethrows. Details in the inline comment on line 132. |
| 2 | **high**: cluster close contract | **fixed** | Verified on a real cluster: every close resolves `undefined` (the extra divergence you found, confirmed), a redundant close never throws, `'disconnect'` fires once per call, `'end'` never fires, and 200 in-flight commands reject with `ClientClosedError` on both mock and real. **The deliberate deviation is accepted**: real v6 rejects with `TypeError: Cannot read properties of undefined (reading 'replicas')`, and copying that crash would help nobody. One small ordering nit is inline on line 1087. |
| 3 | medium: `quit()` doesn't drain | **fixed** | Mock: 2000/2000 pending commands resolve, which matches real. A `MULTI` pending at `quit()` resolves `["OK","1"]`, also matching real. A pending command that fails does not make `quit()` reject. The fix introduced a new hang, inline on line 622. |
| 4 | medium: wrong in-flight error class | **fixed** | `disconnect()`/`destroy()` reject 2000/2000 pending commands with the **real** `DisconnectsClientError`, matching real. |
| 5 | medium: `connect()` on a closed client | **deferred to #440, interim accepted** | Real node-redis reopens after both `quit()` and `destroy()` (verified: `get` works, `"end","end"`). A loud `ClientClosedError` rejection, pinned by a test named as a known gap, is a reasonable stopgap while `RedisServerState.close()` is terminal. |
| 6 | low: fallback naming | **fixed** | `constructor.name` and `util.inspect` now read `ClientClosedError`. `instanceof` is still false on the fallback path, which is covered by #1. |
| 7 | low: docs overclaim | **mostly fixed** | The per-method table and the single-client vs cluster split are good. Two inaccuracies remain: `disconnect()` throws synchronously (inline on `docs/TESTING.md:363`), and the caveat that the fallback applies only "if `redis` is not installed" is disproved by #1. |
| 8 | low: duplicated `assertOpen`/`closed` | **fixed** | `closed` now lives on `CommandRunner`, and only the single client guards. |

| sev | where | finding |
|---|---|---|
| **medium** | `node-redis-mock.ts:622` | A `destroy()`/`disconnect()` during a draining `quit()` throws **before** flushing. With a `BLPOP 0` pending, `quit()` and the `BLPOP` hang forever. Real node-redis rejects both with `DisconnectsClientError`. This breaks the "graceful, then force" teardown pattern. |
| **medium** | `node-redis-mock.ts:132` | Eager resolution doesn't close the window (see #1). It also adds an import-time side effect despite `"sideEffects": false`: every consumer, including ioredis-only ones, now loads `redis` (about 270 ms). |
| low | `docs/TESTING.md:363` | Real `disconnect()` on a closed client throws **synchronously** (`Promise.resolve(this.destroy())`), so `client.disconnect().catch(...)` crashes on real node-redis but not on the mock. |
| low | `node-redis-mock.ts:1087` | The cluster emits `'disconnect'` synchronously inside `quit()`/`disconnect()`. Real emits it after the call returns. `destroy()` is correctly synchronous. |

Pre-existing and not posted: after close, cluster commands lazily re-create node sessions that are never closed (0 → 3 sessions after 3 `get`s). The mock cluster has no v6 `close()`. On an open client, `connect()` is a no-op, where real throws `Socket already opened`.

### Checks
`node-redis-mock.test.ts` + `decode-redis-value.test.ts`: **51/51 pass**. `tsc --noEmit` and `eslint` are clean. CI is green apart from `ioredis-real`, which is the known #398 cluster-startup failure (fixed by #428) and is not counted.

### Verdict: **not yet**
Both original blockers are addressed in substance: the cluster is correct now, and the standalone drain and flush match real node-redis. Two things are still needed:
1. **The escalation hang (line 622).** A hung test process is the worst failure a mock can produce, and this one is new with the drain fix.
2. **Close the cold-cache window deterministically (line 132).** The documented `instanceof` idiom still rethrows on a public path in the shipped build.

The two low findings are small and could go into the same push. With 1 and 2 fixed, this is a ship.

fatal10110:
Pushed **`bf39def`**, which addresses the re-review of `94efd40`. Both blockers and both lows are fixed, and nothing is deferred. It also fixes one more pre-existing divergence that turned up while probing.

Ground truth for this pass: `redis@6.0.0` against a private `redis-server` on 17960 (since shut down), plus `@redis/client/dist/lib/client/{index,socket}.js` and `cluster/cluster-slots.js`.

### 1. Error classes now load in time, with no import-time work (blocker)

- **Change.** `void ensureRedisErrors()` and the async `import()` are gone. A synchronous `redisErrors()` resolves the classes lazily, via `createRequire(__filename)('redis')`, and caches them. `CommandRunner`'s constructor calls it, so every construction path (`createNodeRedisMock()`, a direct `new NodeRedisMockClient`, or `NodeRedisMockCluster.create()`) has the real classes before it can run, decode or close anything. That also covers the decode path's direct cache read without touching it.
- **Identity.** `redis` and `@redis/client` are CJS-only (no `exports` map). I checked that `require` and a native ESM `import` return the same `ClientClosedError`/`DisconnectsClientError` objects, so `instanceof` holds for ESM consumers too.
- **Build.** I added tsup `shims: true`. The ESM bundle gets a pure `fileURLToPath(import.meta.url)` for `__filename`, which loads nothing; CJS uses its native `__filename`. Audit of `dist/`: no static or dynamic `redis` import in any bundle, and a single lazy require in each root bundle (`core` doesn't include the facade).
- **Tests on the shipped build**, in `tests-package/dual-format.test.ts`, each in a fresh child process per format:
  - importing the package does not load `redis`;
  - your repro (a client assembled from `.` + `./core`, set → quit → quit, then `destroy()`) gets the real classes.

  Against a build of `94efd40` these fail 3 of 4, matching your report: CJS gets the stand-in class, and both formats load `redis` on import. At `bf39def` all 4 pass.
- **In-source tests.** Two new unit files, each in its own process for a clean slate: one checks import has no side effects; the other runs cold-cache closes with zero awaits before the first close.
- **Docs.** The fallback caveat is corrected: it applies only when `redis` genuinely can't be required, or predates a class (v4 has no `MultiErrorReply`).

### 2. A stuck `quit()` can be forced closed again (blocker)

I added a *closing* state, separate from *closed*. While `quit()` drains, new work and a second `quit()` get `ClientClosedError`, but `destroy()`/`disconnect()` can escalate. The drain now races an escalation promise, so a `BLPOP 0` ahead of it can't hold `quit()` forever.

**Two points where real v6 goes further than the review describes:**

- The forced `destroy()`/`disconnect()` rejects the pending `quit()` and `BLPOP` with `DisconnectsClientError`, as you said. **It also throws `ClientClosedError` synchronously itself.** Real `destroy()` runs `#queue.flushAll(...)` first, then `#socket.destroy()`, which throws because `socket.quit()` had already set `#isOpen = false` before awaiting. I mirror that. Otherwise graceful-then-forced teardown passes on the mock and throws on the real client, which is the #389 class of bug.
- **No `'end'` fires** on that path (`end=0`), because real node-redis never reaches `destroySocket()`. It also **leaks the socket**: after the forced close, the server still showed the connection blocked in `BLPOP`, and the process never exited. I did **not** reproduce the leak; the facade still tears everything down, so nothing hangs. This is documented in the source and TESTING.md.

The test you asked for (`BLPOP 0` pending → `quit()` → `destroy()`/`disconnect()`) asserts the synchronous throw, that both promises reject with `DisconnectsClientError`, `end === 0`, and that nothing hangs. Two more tests cover commands still queued behind the stuck drain, and refusal of new work while closing. The blocking tests carry an explicit `timeout`, since `npm test` has none, so a regression fails and names itself instead of stalling CI.

**Also fixed: a pre-existing divergence neither review caught.** A command still *executing* at `destroy()` with no `quit()` pending (a parked `BLPOP`) rejected with a leaked internal `Error: The operation was aborted`. Real v6 gives `DisconnectsClientError` and `end=1`; I verified this against the live server. The earlier 2000/2000 checks only covered commands still queued. A `flushable()` wrapper now maps any unit of work that is unstarted or unfinished when the flush lands to the flush error, which matches `flushAll()` rejecting everything unanswered. `MULTI` goes through the same wrapper.

### 3. `disconnect()` throws synchronously (low)

`disconnect()` is now `this.destroy(); return Promise.resolve()`, mirroring real `Promise.resolve(this.destroy())`. It is not `async`, which would turn the throw into a rejection. Verified live: a redundant `disconnect()` throws synchronously, `'end'` fires before `disconnect()`/`destroy()` return, and `'end'` fires before `quit()` resolves, not inside it. Tests and docs are updated, including the note that `client.disconnect().catch(...)` does not catch it.

### 4. Cluster `'disconnect'` timing (low)

`quit()`/`disconnect()` now go through `closeAsync()`: synchronous teardown, then `await Promise.allSettled([])`, then emit. That mirrors `#destroy()`, so callers never see the event before the call returns, and always see it before the promise settles. `destroy()` still emits in-line. A test pins both orderings.

### Checks

- `npm test`: 398/398 (includes `tsc --noEmit`)
- `npm run test:package`: 8/8
- `eslint` and `prettier --check`: clean

I ran no `TEST_BACKEND=real` suites. Every probe process and my private server are stopped, and shared fixtures were only read from, never written to.

Not addressed, from your pre-existing list: sessions lazily re-created after a cluster close, the missing v6 `close()` on the cluster mock, and `connect()` being a no-op on an open client. All are outside this PR's close-path scope; happy to file issues if you want them tracked.

src/client-mocks/node-redis-mock.ts
```diff
@@ -69,34 +72,95 @@ class FacadeMultiErrorReply extends FacadeErrorReply {
...
-            MultiErrorReply: redis.MultiErrorReply,
-          }
-        : FALLBACK_REDIS_ERRORS
   } catch {
-    resolvedRedisErrors = FALLBACK_REDIS_ERRORS
-  }
-  return resolvedRedisErrors
+    return FALLBACK_REDIS_ERRORS
+  }
+  return redis?.WatchError &&
```
fatal10110, line 132: **[low] The error-class check is all-or-nothing. On `redis` 4.0 through 4.6, one missing class (`MultiErrorReply`) also discards the real `ClientClosedError`, `DisconnectsClientError` and `WatchError`. The claim "v4 has no `MultiErrorReply`" is only partly true.**

`peerDependencies` allows `redis@^4`. The v4 line splits in two:

- **`redis` 4.7.x** (`@redis/client` 1.6.x) exports all five classes this check needs. From unpkg, `@redis/client@1.6.0/dist/lib/errors.js`:
  ```
  class AbortError class WatchError class ConnectionTimeoutError class ClientClosedError class ClientOfflineError class DisconnectsClientError class SocketClosedUnexpectedlyError class RootNodesUnavailableError class ReconnectStrategyError class ErrorReply class MultiErrorReply
  ```
  `@redis/client@1.6.1/dist/index.js:25` has `__exportStar(require("./lib/errors"), exports)`, and `redis@4.7.1` re-exports `@redis/client`. So 4.7 passes this check, and the note on line 116 and the one in `docs/TESTING.md` ("as v4 predates `MultiErrorReply`") are wrong for current v4.
- **`redis` ≤ 4.6.x** (`@redis/client` ≤ 1.5.x, e.g. `redis@4.6.15` → `@redis/client@1.5.17`) is missing only `MultiErrorReply`:
  ```
  @redis/client@1.5.0: class AbortError class WatchError class ConnectionTimeoutError class ClientClosedError class ClientOfflineError class DisconnectsClientError class SocketClosedUnexpectedlyError class RootNodesUnavailableError class ReconnectStrategyError class ErrorReply
  ```
  The `&&` chain then replaces **every** class with a local stand-in. That includes `ClientClosedError`, which the documented teardown idiom depends on, even though the installed `redis` exports it.

I reproduced this with a `redis` stub that has the 4.6 export surface (the real v6 module with `MultiErrorReply` removed). I used a consumer-style install of this head's `dist/`, running the same `set` → `quit` → `quit` repro:
```
redis-4.6-shape CJS: ctor=ClientClosedError redisInstalled=true instanceofConsumerRedis=false
```
**Failure scenario:** a project pinned to `redis@4.6` copies the `docs/TESTING.md` idiom (`if (!(err instanceof ClientClosedError)) throw err`). A redundant close is rethrown, and the only hint is a stand-in error that looks exactly like the real one. The same happens to `err instanceof WatchError` after a failed `WATCH`.

Fix: fall back one class at a time. Use `redis.X ?? FallbackX` for each constructor, so a missing `MultiErrorReply` costs only `MultiErrorReply`. Then correct the comment on line 116 and the docs sentence.

Unrelated hardening nearby: the bare `catch {}` around `createRequire(__filename)('redis')` also catches a `ReferenceError` if `__filename` is ever missing (for example if `shims: true` is dropped from `tsup.config.ts`, or the source is run as native ESM). That would quietly switch to the stand-ins. The new package tests would catch it for the published ESM bundle, but catching only `err.code === 'MODULE_NOT_FOUND'` would make any other failure visible.

fatal10110:
## Re-review, round 3: `bf39def` (previous: `94efd40`)

I checked this round against `redis@6.0.0` driving my own `redis-server` 7.2.1 on 18421. The cluster ordering comparison reuses the real-`RedisCluster` measurements from round 2 (same library version, my own 3-node cluster). I also installed this head's built `dist/` into consumer-style `node_modules` layouts and tested it there.

### Round-2 findings

| Round-2 finding | Status | Evidence |
|---|---|---|
| **medium**: a stuck `quit()` could not be forced closed (hang) | **fixed** | `BLPOP 0` → `quit()` → `destroy()`: **mock and real now match** on everything except the leak. `destroy()` and a following `disconnect()` both throw `ClientClosedError` synchronously. `quit()` and `BLPOP` reject with `DisconnectsClientError`. No `'end'` fires. With 50 pending `GET`s in place of the `BLPOP` the result is the same. |
| **medium**: error classes not loaded in time, and `redis` loaded on import | **fixed** | Details in the next section. |
| low: `disconnect()` on a closed client must throw synchronously | **fixed** | Throws synchronously on both mock and real. For `disconnect`/`destroy`/`quit`, the order of `'end'` against the call returning and its promise settling is identical on mock and real. |
| low: cluster emitted `'disconnect'` too early | **fixed** | The mock now emits after `quit()`/`disconnect()` return and before their promise settles, and inline from `destroy()`. That is exactly the order I measured on a real `RedisCluster` in round 2. |

**Accepted: copy the throw, not the leak.** I reproduced the real-client leak myself. After `BLPOP 0` → `quit()` → `destroy()`, the server still shows the connection blocked:
```
id=4 name=pr421esc age=3 flags=b cmd=blpop
```
The real client reports `isOpen=false` while still holding a `Socket` handle, and after about 8 s the process was still running (I killed it). The throw and the missing `'end'` are behaviour callers can see and code against, so the mock is right to copy them. The leaked socket is an upstream resource bug; copying it would only make test runs hang. The mock tears everything down, and its process exits on its own.

**Also verified:** a parked `BLPOP` hard-closed with no `quit()` pending now rejects with `DisconnectsClientError`, matching real (`["end"]` in both).

### The error-class loading blocker

I installed this head's built `dist/` into consumer-style `node_modules`. Each layout ran a client built from `.` + `./core`, then `set` → `quit` → `quit`, in both CJS and ESM:

| layout | CJS | ESM |
|---|---|---|
| npm flat | `instanceof` real = true | true |
| pnpm-style (`.pnpm` store + symlinks) | true | true |
| consumer without `redis` | stand-in, no throw | stand-in, no throw |
| nested physical copy of `redis` inside the package | false | false |

`createRequire(__filename)` resolves `redis` from the **package's own location**, not the consumer's. For a peer dependency that is correct: in flat and pnpm layouts it reaches the consumer's copy. The nested-copy case is inherent to having duplicate installs; a static import or the old dynamic import would behave the same way (and so would an `npm link`ed checkout).

Two more checks here:
- **ioredis-only consumers:** `@redis/client` is not loaded after import, and still not after running `createIoredisMock` and `createInMemoryClient` commands. It loads only once `createNodeRedisMock` is called, in both CJS and ESM.
- **`"sideEffects": false` is now honest:** there is no module-level work left. The ESM shim only computes `__filename`, and the `core` bundles don't reference `createRequire` at all. The browser demo imports `src/in-memory-client` and never reaches this module.

### New finding (posted inline)

| sev | where | finding |
|---|---|---|
| low | `node-redis-mock.ts:132` | The error-class check is all-or-nothing. `redis` ≤ 4.6.x (`@redis/client` ≤ 1.5.x) lacks only `MultiErrorReply`, and that one gap discards the real `ClientClosedError`/`DisconnectsClientError`/`WatchError` too, so the documented `instanceof` idiom fails on those versions (reproduced with a stub that has the 4.6 export surface). Separately, the "v4 has no `MultiErrorReply`" note is wrong: `redis` 4.7.x exports it. Fix: fall back one class at a time. |

### Not addressed, not blocking
Cluster node sessions are re-created after close, the cluster mock has no v6 `close()`, and `connect()` does nothing on an open client. All three are pre-existing and none was introduced here. #440 stays deferred, and that is accepted.

### Checks
- **Local:** the `tests/client-mocks/*` + `decode-redis-value` suites pass **83/83**, and the package suite passes **8/8**. `tsc --noEmit`, `eslint` and `prettier --check` are all clean.
- **CI:** all green on this head, including `ioredis-real`, `node-redis-real` and both `package` jobs. I ran no real-backend suites against the shared fixtures.

### Verdict: **ship**
Both round-2 blockers and both round-2 lows are fixed. I verified each claimed real-node-redis behaviour on my own server, not from the implementer's description. The one new finding is low severity, affects only `redis` ≤ 4.6, and is a small fix. It can go in before merge or as a follow-up.

Issue #373, opened 2026-09-21
tracking: architecture simplification pass (design-level)


fatal10110:
Second-pass **design/architecture** simplifications, from a whole-`src/` review on `main`. The first pass (line-level dead code / duplication) is #358–#365.

## Issues
- [ ] #366 — remove `ResponseStream`; deliver SUBSCRIBE/MONITOR frames via the session push queue (~250 lines)
- [ ] #367 — move node identity onto `RedisServerState`; one executor for all nodes (~80 lines, 11 files)
- [ ] #368 — single `createBackend()` behind TCP / ioredis / node-redis / in-memory front ends (~300 lines; fixes option-name and `compatibility` drift)
- [ ] #369 — one turn queue per server instead of per database (~60 lines; matches real Redis)
- [ ] #370 — derive `COMMAND INFO` arity/keys from definitions (**compat bug**: mock reports arity `-1` / firstKey `0` for `lpush`, `zadd`, `hset`, …)
- [ ] #371 — `notify-keyspace-events` as a `Set` (~90 lines)
- [ ] #372 — drop unused `RedisServerStateOptions` injection; resolve compat profile once (~40 lines)

## Suggested order
1. #366 first — shrinks executor, adapter and both client front ends.
2. #367, then #368 (much smaller once there is one executor and no `nodeRole`); #372 alongside #368.
3. #369, #370, #371 are independent — any time. #370 is the only user-visible bug fix.

Several change exports from `js-redis-server/core` (`ResponseStream`, `onStream`, `RedisServerStateOptions`) — batch those into one minor with a changelog note (package is 0.x). Combine with the first-pass breaking items (#359, #360, #364).

Issue #368, opened 2026-09-21
refactor: introduce a single createBackend() shared by TCP, ioredis, node-redis and in-memory front ends


fatal10110:
## Problem
There is no single "backend" object. Five entry points each assemble the same pieces — state, executor, cluster nodes, replication links, seeding, flush, teardown — and each has its own standalone-vs-cluster branch:

- `createRedisServer` and `createRedisMock` (`src/mock.ts`) — `createRedisMock` is `createRedisServer` + `url`/`seed`/`flush`, both wrapping the same `createStandalonePipeline` + `Resp2Server`
- `createRedisCluster` / `RedisCluster` (`src/cluster-server.ts`)
- `createIoredisMock` (`src/client-mocks/ioredis-mock.ts`) + `InMemoryNodeRegistry` + `VirtualConnectionTracker`
- `createNodeRedisMock` (`src/client-mocks/node-redis-mock.ts`) — builds its own state/executor and its own cluster session routing
- `createInMemoryRedis` (`src/in-memory-client.ts`)

(`cli.ts` goes through `createRedisServer`, so it is fine.)

This duplication has already produced **drift** (verified):
- Option names: TCP mock cluster uses `cluster: { masters, replicas }`; ioredis mock uses `cluster: { masters, replicasPerMaster }`; `RedisClusterOptions` uses `replicasPerMaster`.
- DB count: `createRedisServer` takes `databaseCount` (standalone) vs `databasesPerNode` (cluster); `createRedisMock`'s cluster path accepts **neither** (`createClusterMock` never forwards `databasesPerNode`).
- Capability gaps: `createIoredisMock` and `createNodeRedisMock` do **not** accept/forward `compatibility` at all (only `createInMemoryRedis` and the TCP factories do).
- Four separate teardown implementations (`RedisCluster.close`, ioredis standalone/cluster `attachTeardown`, node-redis `teardown`, mock `close`), each remembering to close replication links and states.

## Proposal
Introduce one internal `createBackend({ cluster?: { masters, replicas? }, databaseCount?, compatibility?, basePort? })` returning `{ nodes, topology, seed(), flush(), close() }`. A standalone server is a one-node backend.

Front ends become thin adapters over it:
- TCP: `listen(backend)` → `Resp2Server` per node (also backfills real ports)
- ioredis: connector class over `backend.nodes` (the node registry becomes `backend.resolve(host, port)`)
- node-redis: session routing via `backend.topology`
- in-memory client: one session on node 0

[...]

Issue #367, opened 2026-09-21
refactor: move node identity (id/role) onto RedisServerState; share one executor across nodes


fatal10110:
## Problem
Node identity is threaded through the transport/session plumbing instead of living on the node's `RedisServerState` (which is already one-per-node):

- `nodeRole` is passed through **11 files**: `Resp2Server` → `attachSession` → `ClientSession` → `RedisExecutionContext`, plus `virtual-connection`, `node-registry`, `ioredis-mock`, `cluster-server`, `command-executor` (ctx copy), `lua-runtime` (ctx copy).
- It has **one consumer**: `src/commands/connection.ts:800`, the `role` field of the `HELLO` reply.
- `localNodeId` is captured in closures: `createClusterPolicy({ localNodeId })` and `createClusterCommands(node.id)` (`src/cluster.ts:127`). This forces a **separate executor + command registry per cluster node**, so `extraCommands` / `policies` exist on `createRedisCommandExecutor` only to serve cluster mode.

## Proposal
1. Add `nodeId?: string` to `RedisServerState` (set by the cluster builder). Role = `state.clusterTopology.getNode(nodeId)?.role ?? 'master'`.
2. Drop `nodeRole` from `Resp2Server`, `attachSession`, `ClientSession`, `virtual-connection`, `NodePipeline` and `RedisExecutionContext`; `HELLO` reads role from `ctx.server`. The manual ctx-copy lines in `command-executor.ts` and `lua-runtime.ts` disappear.
3. `createClusterPolicy()` takes no options — it resolves the local node from `ctx.server`. Its per-session pinned-slot `WeakMap` can stay (sessions are per node).
4. `CLUSTER`/`READONLY`/`READWRITE` become ordinary registered commands reading `ctx.server`; **one executor** serves every node and standalone mode. Remove the `extraCommands`/`policies` plumbing if nothing else uses it.

## Open question
Standalone mode currently has no `CLUSTER` command (→ unknown command). Real Redis replies `ERR This instance has cluster support disabled`. With a shared registry, gate on "topology has nodes" and return that error — verify wording against a real server; this is a small behaviour improvement, but call it out in the PR.

## Acceptance
- `grep -rn nodeRole src/` returns only the `RedisServerState`/HELLO code.
- `createRedisCommandExecutor` is called once per backend, not once per node.
[...]

Issue #366, opened 2026-09-21
refactor: remove ResponseStream; deliver SUBSCRIBE/MONITOR frames via the session push queue


fatal10110:
## Problem
There are two out-of-band delivery channels for server-initiated frames, and every layer has to handle both:

1. **`ResponseStream`** (`src/core/response-stream.ts`) — has exactly **two producers**:
   - `src/commands/pubsub.ts:344` `framesResult()` — wraps a *finite array* of confirmation frames (multi-channel `SUBSCRIBE a b c`) in an async generator, only because `RedisResult` carries a single frame.
   - `src/commands/monitor.ts:85` — the live MONITOR feed.
2. **Session push queue** (`ClientSession.enqueuePush` / `readPushes`) — already carries pub/sub `message`/`pmessage`/`smessage` frames, i.e. the same kind of data as MONITOR lines.

Because both exist, the stream branch is duplicated in every consumer:
- `command-executor.ts`: stream handling in both async and sync paths, `onStream` hook, `ensureNonThenableStream`, `assertSyncCommandResult` stream branch (~10 refs)
- `client-session.ts`: `executeTransaction` closes a stream and substitutes an error entry; `registerResponseStreamCleanup` / `resetResponseStreams`
- `resp2/session-adapter.ts`: `spawnStreamDrain`, `drainStream`, `activeStreams`, **plus** the separate `writeSessionPushes` loop
- `in-memory-client.ts:87` `pushes()`: ~50-line hand-written merge of two async iterators (stream frames + push queue)
- `client-mocks/node-redis-mock.ts`: `drainSubscribeAck`
- `CommandDefinition.execute` return type is `RedisResult | Promise<RedisResult> | ResponseStream`

## Proposal
Delete `ResponseStream`; make the push queue the single channel.

- **SUBSCRIBE-family**: the reply is the first confirmation frame; remaining frames go out in order. Two viable options — pick after a spike:
  - (a) `RedisResult.preEncoded(...)` with all confirmation frames concatenated (already the mechanism `EXEC` uses; `session.protocolVersion` is known at execute time), or
  - (b) push the extra frames via `deferPushesUntilAfterReply()` so they follow the reply.
  Ordering of confirmation frames vs. the reply, and RESP2 vs RESP3, is the risk here — cover with raw-tcp tests.
[...]

PR #422, opened 2026-09-22
refactor: drop unused afterExecute/onStream policy hooks (#359)

### LINKED ISSUE #359: refactor: drop unused afterExecute/onStream policy hooks and dedupe sync/async executor paths

fatal10110:
## Summary

Refactor of `src/core/command-executor.ts`, plus one **user-visible bug fix**
to keyspace notifications, called out under "Behaviour" below. Everything else
is structure.

`ExecutionPolicy` declared three hooks, but all four shipped policies (auth,
cluster, subscribed-mode, transaction) implement only `beforeExecute`. A grep
across `src/` confirms `afterExecute` and `onStream` had **no** production
consumer — the only callers were two tests. Supporting them nevertheless forced
a result/stream rewriting loop into both executor paths plus
`assertSyncPolicyResult` + `MaybePromise` to reject an async hook on the Lua
path.

What changed:

1. **`ExecutionPolicy` is down to one hook**, `beforeExecute` (still
   sync-or-async). `MaybePromise` and `assertSyncPolicyResult` are gone; the
   Lua path's "policy hook cannot run asynchronously from scripts" guard is now
   six inline lines in `executePlanSyncInternal`, with the same error message,
   and is covered by a new test that never existed before.
2. **The async and sync pipelines are deduped** onto three shared module-level
   helpers — `applyPolicyShortCircuit` (transaction-dirty on a short-circuit
   error), `tagNotifyCommand` (notify-command save/restore) and
   `executionErrorResult` (`RedisCommandError` -> `RedisResult`, rethrow
   anything else). Each path is now the `beforeExecute` loop plus the command
   invocation.
3. **`executeRawWithPlan` is inlined into `executeRaw`**, its only caller. The
   `plan` field it returned was discarded at every call site. It was a *public*
   method on the exported `CommandExecutor`, so this is a `/core` break and is
   documented as one in `CHANGELOG.md`.
4. **`createMonitorDeferredContext` uses `Object.create(ctx, { monitor })`**
   instead of hand-copying every context field. This is the part worth a close
   look: `ctx.db` is a **live getter** (a queued `SELECT N` runs mid-EXEC and
   every later command must resolve the currently selected database — #94), so
   a plain spread would freeze it.

   The returned context has exactly one own property, which makes "never spread
   or key-enumerate this context" a load-bearing invariant. Nothing in `src/`
   does either today; the requirement is stated explicitly in the comment so it
   stays that way.

## Behaviour: a user-visible keyspace-notification fix

Separate from the hook removal, and **visible to clients**. An earlier revision
of this description called it "latent-only". That was wrong, and the review
reproduced the bug on `main`.

The executor names write events through a per-database tag
(`activeNotifyCommand`). The old code resolved `ctx.db` when tagging but
resolved it *again* in the `finally`. `ctx.db` is a live getter, so `SELECT`
(tagged on the database selected *before* it) restored the tag onto the one it
switched *to*, leaving the old database tagged `select`. My "latent" reasoning
failed on two points:

- Every later command **restores** the tag it saved, which puts the stale
  `select` back each time, so it is never cleared.
- `MOVE` and `COPY … DB` write into a database the executor **never tags**, so
  their events there took the stale name.

Verified on the mock (with and without the fix) and on a private real Redis
7.2.1, with `notify-keyspace-events KEA`:

```
SELECT 1; SET k v; MOVE k 0
  real Redis 7.2.1: __keyevent@1__:move_from k   __keyevent@0__:move_to k
  main:             __keyevent@1__:del k         __keyevent@0__:select k
  this PR:          __keyevent@1__:del k

SELECT 1; SET s v; COPY s c DB 0
  real Redis 7.2.1: __keyevent@0__:copy_to c
  main:             __keyevent@0__:select c
  this PR:          (nothing)
```

The keyspace channel mirrors each line (`__keyspace@0__:k select` on `main`).
Because the tag lives on the database rather than the connection, one client's
`SELECT` mislabelled another client's `MOVE`. The review also reproduced it
through `MULTI`/`EXEC` and `EVAL`.

`tagNotifyCommand` now returns a closure bound to the database it actually
tagged. This **removes the wrong event but does not add the right ones**:
`MOVE` and `COPY … DB` still omit `move_from`/`move_to`/`copy_to`. That gap
predates this PR and is out of scope. So is a second, known limit of the
save/restore, now stated in the `tagNotifyCommand` docblock: it is only correct
for LIFO nesting (Lua `redis.call`). Parked commands resumed out of order, such
as two BLPOPs on db1 resumed in arrival order, still leave db1 permanently
tagged `blpop`. Both are being filed as follow-ups.

Documented under `### Fixed` in `CHANGELOG.md`.

### What is deliberately *not* removed

`ensureNonThenableStream` stays. Its old doc comment framed it as `onStream`
machinery, but it also guards a stream returned by a **command definition**:
`executePlanInternal` is `async`, so `return thenableStream` would be unwrapped
by the implicit resolution and the stream would never reach the transport. That
is what `does not await thenable response streams`
(`tests/core-command-executor.test.ts`) pins, and that test is not one of the
two the issue asked to rewrite.

Note for whoever picks up #366: it now guards **only** the third-party
`defineCommand` surface. Both shipped stream producers (`src/commands/monitor.ts`,
`src/commands/pubsub.ts`) return plain object literals with no `then`, so
grepping `src/` for an internal caller finds nothing — that is expected, not
evidence the helper is dead. The comment says so in place.

Because of all this, `command-executor.ts` shrinks by ~50 lines rather than the
~120 the issue estimated: the estimate counted `ensureNonThenableStream` plus
the JSDoc the three new shared helpers carry. The duplication the issue was
actually aimed at is gone — the two `executePlan*Internal` bodies went from
59 + 54 lines to 23 + 28.

Closes #359

## Compatibility

Two `/core` breaks, both flagged **BREAKING (`/core`)** in `CHANGELOG.md` under
`[Unreleased] / Removed`; package is 0.x.

- The two policy hooks. `ExecutionPolicy` is not re-exported by name, but its
  shape is reachable through `CommandExecutorOptions.policies`. Since the hooks
  were optional, a policy that still declares them compiles and runs — it
  silently stops rewriting.
- `CommandExecutor.executeRawWithPlan()`, a public method on a class that *is*
  exported from `src/internal.ts`. The changelog gives the one-line migration
  and notes there is no replacement for the discarded `plan` half.

`MaybePromise` and `RawExecutionResult` were never re-exported, so neither is a
break.

## Test plan

- [x] `tests/core-command-executor.test.ts`: `executes sync plans through sync
      policy hooks` rewritten from `afterExecute` to a `beforeExecute`
      short-circuit on the sync path.
- [x] `supports response streams and stream policy hooks` -> `returns response
      streams untouched` (the `onStream` observer dropped).
- [x] New `rejects async policy hooks in sync execution` pins the async-hook
      rejection message that `assertSyncPolicyResult` used to own.
- [x] New `keeps the live db getter when deriving the monitor context`
      reproduces #94 through the monitor path. **Verified red** — swapping
      `Object.create` back to a spread fails it, and the write lands in the
      pre-SELECT database.
- [x] New `restores the notify-command tag on the database it tagged`, plus a
      `(sync path)` twin through `executePlanSync` on a fresh fixture. **Both
      verified red.** Restoring through `ctx.db` instead of the captured
      database leaves db0 tagged `select-then-write`.
- [x] New wire-level test `does not name a cross-database write after an
      earlier SELECT`, with ioredis and node-redis twins in
      `tests-integration/*/key/keyspace-notifications.test.ts`. It runs
      `SELECT 1; SET; COPY … DB 0; MOVE … 0` and asserts that db0 publishes no
      `select` event on either channel. A sentinel `SET` on db0 from a separate
      client replaces a sleep. Delivery to one subscriber is ordered, so once
      the sentinel arrives every earlier event has too, and it proves the
      subscription is live, so the empty-list assertion can't pass vacuously.
      **Red on the mock without the fix** (both clients), green with it.
- [x] Those two notification files pass **20/20 against real Redis 7.2.1**.
      The harness spawned a private `redis-server` child on a free port
      (`REDIS_STANDALONE_PORT` unset) and killed it afterwards. No shared
      fixture was touched.
- [x] `npm run build`, `npm run lint`, Prettier clean.
- [x] `npm test`: 382/382 pass.
- [x] `npm run test:integration:mock`: 1183/1183 pass.
- [ ] The rest of `TEST_BACKEND=real` was not run. Real mode talks to an actual
      Redis and never executes this executor code. The node-redis real suites
      also `FLUSHALL` the shared cluster, so CI is the place for that signal.

## Notes for the follow-ups

Ordered ahead of #366 / #367 / #368 in #373 and kept intentionally narrow —
`ResponseStream` itself is untouched, so #366 still owns deleting it (and with
it `ensureNonThenableStream`, `isResponseStream` in this file, and the
`ExecutorResult` union).

One correction to an earlier revision of this description, so the chain does not
inherit a false premise: I had claimed the old hand-copy "silently dropped
`inScript`" and that `Object.create` fixes it. The drop was real but
**unreachable** — `createLuaCallContext` (`src/core/lua-runtime.ts`) sets
`monitor.defer = true` unconditionally, and `createMonitorDeferredContext`
returns early on `defer`, so an `inScript` context never reaches the copy. No
behaviour depended on it and no test was passing because of it. The
live-`db`-getter argument for `Object.create` is unaffected and stands on its
own.


src/core/command-executor.ts
```diff
@@ -388,24 +313,17 @@ function createMonitorDeferredContext(
...
-    ...(ctx.transactionReplay
-      ? { transactionReplay: ctx.transactionReplay }
-      : {}),
-    ...(ctx.nodeRole ? { nodeRole: ctx.nodeRole } : {}),
+  // Override `monitor` on a prototype link rather than copying the context:
+  // `db` is a *live getter* on the session context (a queued `SELECT N` runs
+  // mid-EXEC and every later command must resolve the currently selected
+  // database — issue #94), and a spread would freeze it to the database that
+  // was selected when this context was built.
+  return Object.create(ctx, {
```
fatal10110, line 327: **[medium · architecture]** The prototype link makes the derived context carry exactly **one own property**, which is a new and unadvertised constraint on every downstream consumer.

The old hand-copy produced a plain object with all nine fields as own properties. `Object.create` produces an object whose only own key is `monitor`; everything else resolves through the prototype:

```
own keys of derived : [ 'monitor' ]
spread of derived   : {"monitor":{"deferredEvents":[]}}
derived.server      : S          <- property access still fine
live db getter      : DB:1       <- getter stays live, as claimed
```

So `{...ctx}`, `Object.assign({}, ctx)`, `Object.keys/entries/values(ctx)`, `JSON.stringify(ctx)` and `structuredClone(ctx)` now silently yield a near-empty object where they used to yield a full context.

I grepped all of `src/` and there is **no such call site today** — the wire-compat reviewer independently confirmed the same, so this is not a live bug. The concern is forward-compat, and it is sharpened by the fact that this PR is link 1 of a four-PR chain (#359 → #366 → #367 → #368/#372) that all rewrite this file:

- `createLuaCallContext` (`src/core/lua-runtime.ts:118`) is itself a hand-copy of every named field. It works through the prototype today, but it is precisely the function a later refactor would "simplify" to `{...ctx, inScript: true, monitor: {...}}` — which would then drop `db`, `server`, `session`, `executor`, `signal` and `park` in one stroke, with no type error, because `Object.create`'s return is `any` and the `as RedisExecutionContext` cast here already suppresses the only signal TypeScript could give.

Worth either taking the issue's alternative (`withMonitor(ctx, monitor)` returning a real object) or keeping `Object.create` and adding one line to the comment stating the invariant explicitly, e.g. *"the result is prototype-linked — never spread or enumerate a context, copy named fields"*. A cheap belt-and-braces option is a unit test asserting `Object.keys(derived).length === 1` is **not** relied on, i.e. that `derived.session === ctx.session` after a spread-free copy.

src/core/command-executor.ts
```diff
@@ -388,24 +313,17 @@ function createMonitorDeferredContext(
...
     monitor: {
-      ...ctx.monitor,
-      deferredEvents: [],
+      value: { ...ctx.monitor, deferredEvents: [] },
```
fatal10110, line 329: **[medium · architecture + wire-compat]** The descriptor omits `writable` and `configurable`, so both default to **`false`**. The old hand-copy created `monitor` as an ordinary writable, configurable property — this is a silent semantics change inside a PR whose stated contract is "no behaviour change".

```
monitor descriptor  : writable=false configurable=false enumerable=true
assign monitor      : THREW TypeError: Cannot assign to read only property 'monitor' of object '#<Object>'
delete monitor      : THREW TypeError: Cannot delete property 'monitor' of #<Object>
```

These throw rather than no-op because the whole codebase is ESM/TS strict mode.

Nothing assigns to or deletes `ctx.monitor` today (verified by grep across `src/`, and independently by the wire-compat reviewer), so there is no live failure. But `RedisMonitorContext` is part of the `/core` surface, `ExecutionPolicy` and `defineCommand` both hand a context to third-party code, and #366/#367 are about to move this machinery again. A reviewer reading `Object.create(ctx, { monitor: { value, enumerable } })` will not naturally register that they have just frozen the field.

The fix is one line and costs nothing:

```ts
monitor: {
  value: { ...ctx.monitor, deferredEvents: [] },
  enumerable: true,
  writable: true,
  configurable: true,
},
```

src/core/command-executor.ts
```diff
@@ -388,24 +313,17 @@ function createMonitorDeferredContext(
...
-    },
-    server: ctx.server,
-    session: ctx.session,
-    executor: ctx.executor,
-    ...(ctx.transactionReplay
-      ? { transactionReplay: ctx.transactionReplay }
-      : {}),
-    ...(ctx.nodeRole ? { nodeRole: ctx.nodeRole } : {}),
+  // Override `monitor` on a prototype link rather than copying the context:
+  // `db` is a *live getter* on the session context (a queued `SELECT N` runs
```
fatal10110, line 317: **[low · architecture]** This comment — and the corresponding paragraph in the PR description ("stops the copy from silently dropping fields the way the old literal dropped `inScript`") — rests on a premise that is **unreachable in practice**. The `Object.create` change is still fine; the *justification* is wrong and should not be carried into #366/#367.

It is true that the old object literal did not copy `inScript`. But an `inScript` context can only be produced by `createLuaCallContext` (`src/core/lua-runtime.ts:118`, the only writer of that field), and that function unconditionally sets:

```ts
monitor: {
  ...ctx.monitor,
  defer: true,          // <-- always
  clientAddress: 'lua',
},
```

`createMonitorDeferredContext` returns `ctx` untouched at its very first guard when `ctx.monitor?.defer` is set. So an `inScript: true` context never reaches the copy at all, and the field was never actually dropped.

I proved this empirically rather than by reading: I instrumented the function entry on this branch and ran the script suites.

```
$ # probe at createMonitorDeferredContext entry, before both guards
$ TEST_BACKEND=mock node --import tsx --test tests/commands-scripts.test.ts tests/lua-runtime-isolation.test.ts
  11 PROBE inScript=true defer=true disabled=undefined subs=0
```

Every one of the 11 `inScript` contexts that reached the function had `defer=true` and bailed at the first guard. Zero reached the copy.

Two consequences worth recording:
1. **Good news** — this confirms the "pure refactor" claim: there is no `ctx.inScript` behaviour change, and no test was passing because of the old omission. The only reader, `forceDeterministicSetOrder` in `src/commands/keys.ts:762` (SORT determinism), is unaffected either way.
2. The real argument for `Object.create` is just "don't hand-maintain a field list", which is a good enough reason on its own. Suggest rewording to that, and dropping the `inScript` claim — otherwise #366/#367 may "preserve" an invariant that was never load-bearing, or conversely assume `defer: true` can be relaxed in `createLuaCallContext` without consequence.

src/core/command-executor.ts
```diff
@@ -193,55 +175,21 @@ export class CommandExecutor {
...
+          return applyPolicyShortCircuit(plan, ctx, policyResult)
         }
       }
 
-      // Tag the database with the active command so keyspace notifications can
-      // name write events after the originating command. Saved/restored so
-      // nested (Lua redis.call) and parked/interleaved commands stay correct.
-      const prevNotifyCommand = ctx.db.activeNotifyCommand
-      ctx.db.activeNotifyCommand = plan.definition.name
+      const previousNotifyCommand = tagNotifyCommand(plan, ctx)
```
fatal10110: **[low · architecture + wire-compat]** Both review lanes landed on this independently. Extracting the *save* half into a named helper while leaving the *restore* inline cements an asymmetry that the rest of this PR is specifically about: `ctx.db` is a **live getter**.

`tagNotifyCommand` resolves `ctx.db` at tag time; the `finally` on line 189 (and line 243 on the sync path) resolves `ctx.db` **again**, after `execute()` has run. For any command that changes the selected database, those are two different `RedisDatabase` objects:

```
SELECT 1
  tagNotifyCommand -> ctx.db === db0 ; db0.activeNotifyCommand = 'select' ; returns null
  execute()        -> session.selectDatabase(1)
  finally          -> ctx.db === db1 ; db1.activeNotifyCommand = null
  net effect       -> db0.activeNotifyCommand stays 'select' forever; db1 got a tag that belonged to db0
```

This is **pre-existing** — `origin/main` lines 206-207/232 read `ctx.db` twice in exactly the same way — so it is not a regression and not a blocker. It is also currently benign, which both lanes verified rather than assumed: every write/delete path re-tags before firing, and `evict` events hardcode `'expired'` in `KeyspaceNotifier.resolve` (`src/state/keyspace-notifier.ts:352-358`) instead of consulting `activeNotifyCommand`. Live check against the mock on this branch confirmed `__keyevent@0__:expired expkey`, not `__keyevent@0__:select`.

Raising it here because this PR is the one that gives the behaviour a name and a docblock, and that docblock now asserts more than the code delivers:

> *"nested (Lua `redis.call`) and parked/interleaved commands depend on the save/restore"*

— true, but it quietly omits that the save/restore does not survive a database switch. The chain-safe shape is to have the helper capture the database object and hand back a restore closure, so the two halves can't drift:

```ts
function tagNotifyCommand(plan: CommandPlan, ctx: RedisExecutionContext): () => void {
  const db = ctx.db                       // resolve once
  const previous = db.activeNotifyCommand
  db.activeNotifyCommand = plan.definition.name
  return () => { db.activeNotifyCommand = previous }
}
```

Fine to defer to #366/#367 if you'd rather keep this PR strictly behaviour-preserving — but then the docblock should say the restore targets whichever database is selected *at completion*.

CHANGELOG.md
```diff
@@ -26,6 +26,19 @@ so the PR body is not a durable home for a breaking-change note.
 
 ### Removed
 
+- **BREAKING (`/core`)** The `afterExecute` and `onStream` hooks are gone from
+  `ExecutionPolicy` ([#359]). None of the four shipped policies (auth, cluster,
+  subscribed-mode, transaction) ever implemented them — only tests did — and
+  supporting them forced the executor to keep a synchronous mirror of its policy
+  chain plus a thenable-stream re-wrap for every hook result.
```
fatal10110: **[low · wire-compat + architecture]** This paragraph is inaccurate on two counts, and the `/core` break list below it is incomplete. Since the file's own preamble says the PR body is not a durable home for breaking-change notes, this text is the artifact people will actually read.

**1. The thenable-stream re-wrap was not removed.** `ensureNonThenableStream` is retained verbatim (line 463 of the new `command-executor.ts`) and is still invoked on every command-returned stream (line 186). The PR only re-pointed its doc comment from `onStream` to command definitions — which the PR description is upfront about, but the CHANGELOG is not.

**2. The synchronous policy mirror was not caused by these hooks and still exists.** `executePlanSyncInternal` still walks the full policy chain synchronously (lines 217-248), because the Lua `redis.call` path requires it. What actually went away is `assertSyncPolicyResult` + `MaybePromise`, replaced by six inline lines.

So the sentence "supporting them forced the executor to keep a synchronous mirror of its policy chain plus a thenable-stream re-wrap for every hook result" describes two costs, neither of which this PR recovers. Suggest narrowing it to what was genuinely removed: the `afterExecute`/`onStream` rewrite loops in both paths, `assertSyncPolicyResult`, and `MaybePromise`.

**3. Missing from the `/core` break list: `CommandExecutor.executeRawWithPlan()`.** The PR description argues that deleting `RawExecutionResult` is safe because it "was never re-exported from `src/internal.ts`" — correct for the *type*. But `executeRawWithPlan` was a **public method on `CommandExecutor`**, and `CommandExecutor` *is* exported from `src/internal.ts`:

```ts
export { CommandExecutor, type ExecutorResult } from './core/command-executor'
```

Anyone on `js-redis-server/core` calling `executor.executeRawWithPlan(...)` — the one documented way to get the resolved `CommandPlan` back alongside the result — now gets a `TypeError: executor.executeRawWithPlan is not a function` at runtime, and a compile error on TS. That deserves a line in the same `Removed` block. (The inlining itself is correct — I diffed the `try` boundary against `origin/main` and it is identical, with `this.plan()` and `await this.executePlan()` both inside the `try` in each version.)

src/core/command-executor.ts
```diff
@@ -484,11 +402,63 @@ const CLUSTER_PRE_EXECUTION_ERROR_CODES = new Set([
...
+  if (shouldDirtyTransaction(plan, ctx)) {
+    ctx.session.markTransactionDirty()
+  }
+
+  return RedisResult.error(err.message, err.code)
+}
+
+/**
+ * A command may return an object that is both a {@link ResponseStream} and
+ * thenable (e.g. an async wrapper). Callers `await` the executor result, and
```
fatal10110, line 484: **[low · architecture]** Keeping `ensureNonThenableStream` is the right call and the mechanism in the rewritten comment is accurate — `executePlanInternal` is `async`, so `return thenableStream` really would be unwrapped by the implicit promise resolution and the stream would never reach the transport. No objection to the deviation from the issue.

One clarification worth folding into the comment, so #366 has the full picture when it decides whether this dies with `ResponseStream`: **no shipped command actually returns a thenable stream.** The only two stream producers are plain non-thenable object literals:

- `src/commands/monitor.ts:86`
- `src/commands/pubsub.ts:350`

For both, `'then' in stream` is `false` and this function returns the argument untouched. The only thing exercising the re-wrap path is the synthetic command in `tests/core-command-executor.test.ts` ("does not await thenable response streams").

That makes this a guard on the **third-party `defineCommand` surface** rather than on internal callers — which is a legitimate reason to keep it, but a different one from "a command definition returns one". Saying so explicitly ("guards streams returned by *user-supplied* command definitions; no built-in command returns a thenable stream") prevents #366 from reading the current wording, grepping for an internal thenable producer, finding none, and concluding the function is dead.

docs/ARCHITECTURE.md
```diff
@@ -200,13 +198,11 @@ Two execution paths share this same plan:
 
 ## Execution policies
 
-An [`ExecutionPolicy`](../src/core/execution-policies/index.ts#L9) wraps every
-command with three optional hooks:
+An [`ExecutionPolicy`](../src/core/execution-policies/index.ts#L9) guards every
```
fatal10110: **[low · nit]** This line's anchor is now off by two, and this PR is what moved it. `ExecutionPolicy` is declared at **line 7** of `src/core/execution-policies/index.ts` on this branch — dropping the `ResponseStream` import and the `MaybePromise` alias shifted it up from line 9.

Since the line is being edited anyway ("wraps every" → "guards every"), worth fixing the fragment to `#L7` in the same stroke.

While you're here, the two anchors just above are also stale, and this PR widened the gap rather than causing it:

| doc link | points at | actually at (this branch) |
|---|---|---|
| `command-executor.ts#L34` (line 182) | `CommandExecutor.plan()` | 34 is inside the import block |
| `command-executor.ts#L65` (line 190) | `executePlan` | **159** |
| `command-executor.ts#L116` (lines 193, 496) | `executePlanSync` | **209** |

Those were already wrong before the PR (`executePlan` was at 177 and `executePlanSync` at 262 on `origin/main`), so fixing them is optional here. But `command-executor.ts` is about to be rewritten three more times by #366/#367/#368 — line-number fragments into it will be wrong again within the week. Consider dropping the `#L` fragments from these five links entirely, or pointing them at a stable symbol instead.

CHANGELOG.md
```diff
@@ -26,6 +26,39 @@ so the PR body is not a durable home for a breaking-change note.
...
+
+  `beforeExecute` is unchanged and still the place to short-circuit a command
+  (queue / redirect / reject); it may still be async on the network path, and
+  is still rejected when it returns a promise under `redis.call`. A custom
+  policy that rewrote results or wrapped streams has no drop-in replacement —
+  do it inside the command definition, or wrap `CommandExecutor`. Because the
+  hooks were optional, a policy object that still declares them compiles and
+  runs, silently doing nothing.
+
+- **BREAKING (`/core`)** `CommandExecutor.executeRawWithPlan()` is removed
```
fatal10110, line 44: **[medium · wire-compat + architecture] New in re-review: the `tagNotifyCommand` fix is user-visible, so it needs a `### Fixed` entry here.**

The PR body's new "Behaviour" section calls the fix *"latent-correctness only, not user-visible: every command re-tags before it runs"*. That premise fails twice:
- Every command **restores** its saved tag on exit, which puts a stale value right back.
- `MOVE` (`src/commands/keys.ts:459`) and `COPY … DB n` (`:538`) write into a target database the executor **never tags**, via `targetDb.set(...)`. That database's leftover tag names the event, because `KeyspaceNotifier.resolve` emits `WRITE_EVENT_OVERRIDES[tag] ?? tag` for a write.

Ground-truthed over TCP against **real Redis 7.2.1**, the `main` mock and this PR's mock, with `notify-keyspace-events KEA` and a `PSUBSCRIBE __key*__:*` subscriber:

```
SELECT 1; SET k v; MOVE k 0
  real Redis : __keyevent@1__:move_from k | __keyevent@0__:move_to k
  main       : __keyevent@1__:del k | __keyspace@0__:k select | __keyevent@0__:select k
  this PR    : __keyevent@1__:del k

SELECT 1; SET k v; COPY k k2 DB 0
  real Redis : __keyevent@0__:copy_to k2
  main       : __keyevent@0__:select k2
  this PR    : (nothing)
```

It also leaks **across connections**: client A's `SELECT` made client B's `MOVE` publish `select`. The same leak shows up via `MULTI; SELECT 1; SET …; EXEC` and via `EVAL` calling `redis.call('SELECT', …)`. This PR is clean in every case. I reproduced the same result in-process (`ClientSession` + `createRedisCommandExecutor`): on `main`, `db0` is left tagged `select`; on this head it is `null`.

So this is a real, client-observable bug fix. It deserves a line under `### Fixed` (line 160 on this head), and the PR body's Behaviour section should be corrected to match. The file's own preamble says the PR body is not a durable home for this kind of note. The new test's comment ("mis-name a later write event") already says as much. Suggested wording:

> A `SELECT` no longer leaves the previously selected database tagged with the `select` command, which made a later `MOVE` / `COPY … DB` into that database (from any connection) publish a bogus `select` keyspace event.

(The PR head's missing `move_to`/`copy_to`, and `del` where real Redis sends `move_from`, is a pre-existing gap on both trees and not this PR's to fix. No issue tracks it today; worth filing.)

src/core/command-executor.ts
```diff
@@ -484,11 +410,78 @@ const CLUSTER_PRE_EXECUTION_ERROR_CODES = new Set([
...
+    ctx.session.markTransactionDirty()
+  }
+
+  return result
+}
+
+/**
+ * Tag the database with the active command so keyspace notifications can name
+ * write events after the originating command, and return the undo. Callers run
+ * it in a `finally` — nested (Lua `redis.call`) and parked/interleaved commands
```
fatal10110: **[low · architecture + wire-compat] New in re-review: this docblock still overclaims for parked commands.** A permanent stale tag remains reachable on this head. Not a regression (`main` is strictly worse) and not a blocker.

The sentence *"parked/interleaved commands depend on the save/restore"* implies the save/restore is **correct** for them. The closure pins *which database* gets restored, but not the *order*. A per-database save/restore is a stack and is only correct if tags unwind LIFO. Two blocking commands parked on the same database unwind in the order they are **served**:

```
this PR (acf5cbc), in-process: A and C on db1, B on db0
> A: BLPOP q 0   (parks)       tags: db1= blpop        A saved prev = null
> C: BRPOP q2 0  (parks)       tags: db1= brpop        C saved prev = blpop
> B: SET k v ; MOVE k 1
  EVENT __keyevent@1__:brpop k          <-- parked C's tag names B's write
> B: LPUSH q x   (serves A)    tags: db1= null         A restores null
> B: LPUSH q2 y  (serves C)    tags: db1= blpop        C restores "blpop" -> stuck
> B: SELECT 0; SET k9 v; MOVE k9 1
  EVENT __keyevent@1__:blpop k9         <-- stale forever
```

The wire-compat lens confirmed this over TCP (real Redis emits `move_to`), and found that **`FLUSHALL` does not clear the stuck tag**. On `origin/main` the same run ends with **both** databases stuck (`db0= blpop db1= select`), so this PR strictly improves things and the fix itself is correct.

Asks:
1. **Docblock:** say the save/restore is correct for *nested* commands (Lua `redis.call`, EXEC replay), which unwind LIFO. Say it is **not** correct for independently parked commands, which resume in serve order, and point at a follow-up issue.
2. **Follow-up, not this PR:** a mutable database-wide tag is the wrong shape. A parked command's tag is visible to other connections' writes for the whole wait. #366/#367 could have the command name travel with each mutation event, or set it only around each synchronous `execute` segment. Cross-db writers (`MOVE`, `COPY … DB`) would then need to tag their target explicitly to emit `move_to`/`copy_to`.

tests/core-command-executor.test.ts
```diff
@@ -287,6 +323,41 @@ describe('new command executor core', () => {
...
+  })
+
+  test('restores the notify-command tag on the database it tagged', async () => {
+    // `ctx.db` is a live getter, so a command that switches databases mid-flight
+    // must not have its keyspace-notification tag restored onto the *new*
+    // database — that would leave the original tagged forever and mis-name a
+    // later write event.
+    const { executor, ctx, server } = createSelectMidCommandFixture()
+
+    await executor.executeRaw('select-then-write', [], ctx)
```
fatal10110, line 355: **[low · architecture + wire-compat] New in re-review: this regression test covers only the async path, and asserts only on an internal field.**

1. **The sync path is unpinned.** The closure is used at both call sites, `executePlanInternal` (L182/189) and `executePlanSync` (L236/243). This test only drives `executeRaw`. Reverting just the sync call site to `ctx.db.activeNotifyCommand = …` would pass the whole suite. That is exactly the Lua shape (`EVAL` → `redis.call('SELECT', 1)`) where the wire lens saw the leak on `main`. Please add the same asserts on a fresh fixture via:
   ```ts
   executor.executePlanSync(executor.plan('select-then-write', []), ctx)
   ```
2. **It asserts on `activeNotifyCommand`, not on what a client sees.** I confirmed it goes red when the closure restores via `ctx.db`, so it does discriminate. But the user-visible bug is the bogus event. A wire-level companion would lock that in and survive the #366/#367 rework of this field. For example, an integration test with `notify-keyspace-events KEA`: `SELECT 1; SET k v; COPY k k2 DB 0` must **not** publish `__keyevent@0__:select`.

src/core/command-executor.ts
```diff
@@ -25,11 +25,6 @@ import {
...
 export type ExecutorResult = RedisResult | ResponseStream
 
-export type RawExecutionResult = {
-  result: ExecutorResult
-  plan?: CommandPlan
-}
-
 export type CommandExecutorOptions = {
   registry: CommandRegistry
   policies?: readonly ExecutionPolicy[]
```
fatal10110, line 30: **[low · architecture] New in re-review, outside the diff (anchored on the nearest in-diff line):** the `CommandExecutor` class docblock at **L41-43** still says the policy chain gives policies a chance to *"short-circuit, rewrite, or wrap the result."* This PR removed rewrite (`afterExecute`) and wrap (`onStream`). It should read just "short-circuit (queue / redirect / reject)", matching the new `ExecutionPolicy.beforeExecute` doc and `AGENTS.md`.

CHANGELOG.md
```diff
@@ -26,6 +26,39 @@ so the PR body is not a durable home for a breaking-change note.
...
+  method is a break).
+
+  ```
+  (await executor.executeRawWithPlan(cmd, args, ctx)).result
+    -> await executor.executeRaw(cmd, args, ctx)
+  ```
+
+  There is no replacement for the `plan` half. A caller that wants the
+  `CommandPlan` builds it with the still-public `executor.plan(cmd, args)` and
+  passes it to `executor.executePlan(plan, ctx)` — note that `plan()` *throws*
```
fatal10110: **[low · architecture] New in re-review: the migration note undersells what you lose by switching from `executeRaw` to `plan()` + `executePlan()`.**

- `plan()` throws on **any** planning error: unknown command, arity, *and* argument parsing (`parseCommandArgs`, `keys()`), not just "unknown command or arity".
- Going through `plan()` + `executePlan()` also skips `rawCommandErrorResult` (`command-executor.ts` L122-143). A caller inside `MULTI` loses the transaction-dirty marking for a bad queued command, and the `EXEC foo` → `EXECABORT` discard.

Suggested wording: *"`plan()` throws on any planning error (unknown command, arity, argument parse) and bypasses the MULTI-dirty / EXECABORT handling `executeRaw` applies. Catch `RedisCommandError` yourself if you need either."*

fatal10110:
## Re-review of `acf5cbc` (prior review: `f7578aa`)

Two lenses: Redis wire-compat, ground-truthed against **real Redis 7.2.1** with main/PR mocks over TCP, and Node architect. Plus my own in-process probes and tests on a throwaway worktree.

**Verification on `acf5cbc`:** 131/131 across the 9 executor, script, transaction, keyspace and cluster-policy suites. The new `restores the notify-command tag…` test is **confirmed red** when the closure restores via `ctx.db` instead of the captured database.

### Prior findings

| # | Finding | Status |
|---|---|---|
| 1 | [med] Prototype-linked ctx: spread/enumeration hazard | **Fixed.** INVARIANT comment at L321-326; no spread/`Object.keys`/`assign` on a ctx anywhere in `src/` |
| 2 | [med] `monitor` descriptor non-writable/non-configurable | **Fixed.** `writable: true, configurable: true` at L331-332 |
| 3 | [low] `inScript` premise unreachable | **Fixed.** Correction note in PR body, reasoning matches `lua-runtime.ts` |
| 4 | [low] `tagNotifyCommand` restores onto the live `ctx.db` | **Code fixed, write-up wrong.** Closure at L439-450, used by both paths. But the PR body's "latent-only / not user-visible" is false: see inline |
| 5 | [low] CHANGELOG inaccuracies + missing `executeRawWithPlan` break | **Fixed.** One wording nit on the migration note (inline) |
| 6 | [low] `ensureNonThenableStream` justification | **Fixed.** L479-484 |
| 7 | [nit] `#L9` anchor | **Fixed.** `#L7`. The three pre-existing stale anchors were left deliberately; fine |

### New findings (inline)

| Sev | Where | Finding |
|---|---|---|
| **medium** | `CHANGELOG.md:44` | The notify-tag fix is **user-visible**. On `main`, `SELECT 1; SET k v; MOVE k 0` publishes a bogus `__keyevent@0__:select k` (real Redis: `move_to`). It also happens across connections, via MULTI/EXEC and via EVAL. Needs a `### Fixed` entry, and the Behaviour section needs correcting |
| low | `command-executor.ts:431` | Docblock implies save/restore handles parked commands. It only handles LIFO nesting. Two parked BLPOPs served FIFO leave db1 tagged `blpop` **permanently** (survives FLUSHALL). Pre-existing, not a regression |
| low | `core-command-executor.test.ts:355` | New test covers only the async path, and asserts on an internal field. Add an `executePlanSync` case, ideally plus a wire-level check |
| low | `command-executor.ts:30` → L41-43 | Class docblock still says policies can "rewrite, or wrap" |
| low | `CHANGELOG.md:58` | Migration note: `plan()` throws on any parse error, and bypasses MULTI-dirty/EXECABORT handling |

**Regressions vs `main`:** none. Wherever the PR differs from `main` it either matches real Redis or is wrong in the same way `main` already is. Nested Lua and EXEC replay are strictly better.

### Pre-existing gaps, not this PR (suggest filing; none are tracked today)
- A parked blocking command's tag stays on the database for the whole wait, and out-of-order resumption leaves it stuck. The structural fix is to carry the command name with each mutation event instead of a mutable per-db tag. That is a natural fit for #366/#367.
- `MOVE` never emits `move_from`/`move_to` and `COPY … DB` never emits `copy_to` (target write untagged; `move` missing from `DELETE_EVENT_OVERRIDES`).
- Blocking/move pops are named after the command, not the operation (`blpop`→`lpop`, `brpop`→`rpop`, `blmove`/`lmove`/`rpoplpush`→push+pop, `lmpop`/`zmpop`/`bzmpop`/`bzpopmin`→`lpop`/`zpopmin`/`zpopmax`).
- Two clients blocked on one key: the mock serves the **newest** waiter first; real Redis serves the oldest (`A BLPOP q 0; C BLPOP q 0; LPUSH q x; LPUSH q y` → real A=x C=y, mock A=y C=x).

### Verdict: **not yet, for documentation only**
The code is correct and strictly better than `main`. The one required change is a `### Fixed` CHANGELOG entry plus a corrected Behaviour section: this PR fixes a wire-visible bogus keyspace event, and the file's own preamble says the PR body is not a durable home for that. The LOWs can ride in the same push. After that, ship.

fatal10110:
## Re-review (round 3) of `96b4acc` (previous: `acf5cbc`)

**Verdict: ship.** All five round-2 findings are fixed, and I found nothing new, so there are no inline comments this round.

### How I verified it
I used a throwaway worktree at `96b4acc`. For the real backend, `setupRawStandalone()` spawned a private `redis-server` child; I did not touch the shared fixtures.
- Unit (`core-command-executor`): **18/18**. Mock integration (both `keyspace-notifications` files): **20/20**. Real Redis, private child: **20/20**.
- **Red check:** I reverted the closure to restore via `ctx.db`. Four tests failed: the async and sync unit tests, and the new wire test on **both** ioredis and node-redis. I restored the fix afterwards.
- `tsc --noEmit`, eslint and prettier are clean on the changed files.

### Round-2 findings
| Sev | Finding | Status |
|---|---|---|
| medium | Notify-tag fix is user-visible and needs a `### Fixed` entry | **Fixed.** The CHANGELOG entry has the correct traces. It says plainly that the fix removes the bogus `select` without adding `move_to`/`copy_to`. The PR body's Behaviour section now retracts "latent-only" |
| low | Docblock overclaims for parked commands | **Fixed.** It now states the LIFO-only limit, with the two-BLPOP example (L428-446) |
| low | Test covered only the async path / an internal field | **Fixed.** The sync-path unit test goes through `executePlanSync` on a fresh fixture. Wire tests on both clients assert that db0 publishes no `select` on either channel |
| low | Class docblock said "rewrite, or wrap" | **Fixed.** Now "short-circuit it — queue, redirect, or reject" (L41-43) |
| low | Migration note | **Fixed.** Covers any planning error, and the skipped MULTI-dirty/EXECABORT handling |

### Wire-test soundness (the coordinator asked me to check this)
- **Ordering:** the actor's COPY/MOVE events are published during those commands, before their replies return. The sentinel `SET` is sent only after the MOVE reply arrives, so its event is published strictly later. The keyspace and keyevent messages for one mutation go out back-to-back from the same notifier call. All of it goes to the same PSUBSCRIBE connection, which delivers in order.
- **Client side:** the collector is registered before the sentinel waiter. That is the same `pmessage` emitter in ioredis, and insertion order of the `EventBus` listener `Set` in node-redis. So every prior event is recorded before the waiter resolves.
- **Not vacuous:** the sentinel is a real db0 keyevent, so it proves the subscription is live and notifications are on. `waitForEvent` **rejects** after 1 s rather than resolving false. The red check above shows the test catches the bug on both clients.

### Non-blocking nit
The CHANGELOG ("tracked separately") and the docblock ("tracked as a follow-up") refer to follow-ups that don't have issue numbers yet. Worth adding the links once the coordinator files them. That can happen in a later PR.

Issue #59, opened 2026-06-09
Bug: TTL command rounds up (ceil) instead of truncating toward zero


artur-ma:
## Description

The `TTL` command uses `Math.ceil` to convert the remaining milliseconds to seconds. Real Redis truncates (floor for positive remainders), so 1500 ms remaining → `TTL` returns `1`, not `2`.

## Root Cause

`src/commands/keys.ts`:
```ts
Math.max(0, Math.ceil((expiresAt - Date.now()) / 1000))
```

## Expected behaviour (real Redis)

```
SET k v PX 1500
TTL k   → (integer) 1   (truncated, not rounded up)
```

## Fix

```ts
Math.max(0, Math.floor((expiresAt - Date.now()) / 1000))
```

`PTTL` is not affected (it already returns milliseconds directly).

PR #423, opened 2026-09-22
fix: stop hash field-TTL tests racing a 20ms TTL (#411)

### LINKED ISSUE #411: Hash field-TTL tests flake under load: 20ms TTLs lose a race with their own round-trips
Found while reviewing PR #409 and PR #408. **Pre-existing on `main`** — reproduced on unmodified `main` with both PRs' changes reverted.

## Problem
Two hash field-TTL tests fail intermittently in `test:integration:mock`. They are the long-standing "random" mock-suite flake, and the cause is now pinned: a **20 millisecond field TTL** followed by more client round-trips than 20 ms allows under load.

- `tests-integration/ioredis/hash/field-persist.test.ts:59` — `hpExpire(key, 'soon', 20)`, then `HTTL` returns `-2` (field already gone), failing `assert.ok(seconds[2] >= 0)` → `false == true`
- `tests-integration/ioredis/hash/field-expire.test.ts:202` — `hpExpire(key, ['replace','counter','float'], 20)`, then `counter` expires and `HINCRBY` recreates it from 0 → `1`, failing `assert.strictEqual(..., 2)` → `1 !== 2`

Both have node-redis mirrors, so **four files** are affected.

## Evidence
- 6 runs with an unloaded machine: 1114/1114 pass, 0 failures.
- 12 runs under 4x parallel CPU contention: **10 failures, exactly these two tests**, varying between them.

Deterministic reproduction — inserting a 30 ms stall between the `hpExpire` and the assertion:
```
immediately   HTTL soon = [ 0 ]    HINCRBY ctr = 2
after 30ms    HTTL soon = [ -2 ]   HINCRBY ctr = 1     <- exactly the observed failures
```

## Why it matters
This is the second flake class eroding trust in the test signal, alongside the SCAN flakes in #395. A suite that fails on a loaded CI runner and passes on a quiet one teaches people to re-run rather than read.

## Proposal
Raise the TTL to something a round-trip cannot outrun (~500 ms), or make the assertions tolerant of `-2` / a recreated counter. Raising the TTL is preferred — the tests are about TTL semantics, not about racing the clock.

Worth auditing the other `hpExpire`/`pexpire`-based tests for sub-100ms TTLs while fixing.

## Acceptance
- The four files no longer depend on winning a 20 ms race.
- `test:integration:mock` passes under deliberate CPU contention (e.g. 4x parallel load), not just on an idle machine.

fatal10110:
## Summary

Two hash field-TTL tests (and their node-redis twins) set a **20 ms** field TTL and then
made more client round-trips than 20 ms survives on a loaded machine, so the field expired
mid-test and the assertions failed — the long-standing `test:integration:mock` flake in #411.

The tests are about TTL semantics, not about racing the clock. Nothing here loosens an
assertion: no tolerance for `-2`, no tolerance for a recreated counter, no widened equality.
The fix is in three parts.

### 1. The field must survive the round-trips before the wait

`HPEXPIRE ... 20` → `500` in all four files. 500 ms is longer than any round-trip the test
makes, which is the direct fix for the failure #411 reports.

- `tests-integration/ioredis/hash/field-persist.test.ts` (+ `HPTTL <= 20` → `<= 500`)
- `tests-integration/node-redis/hash/field-persist.test.ts`
- `tests-integration/ioredis/hash/field-expire.test.ts`
- `tests-integration/node-redis/hash/field-expire.test.ts`

### 2. The wait for expiry no longer needs a margin either

Raising the TTL moved the problem to the *other* side of the same wait. With `HPEXPIRE 500`
followed by `delay(600)`, the sleep only has to overshoot by 100 ms to be correct — and it
doesn't always: running the suite under 4x CPU contention produced a run where **both**
`field-persist` ("`'value3' !== null`") and `field-expire` ("`'2' !== null`") failed in the
opposite direction, with the field still alive after the sleep. Both processes failed in the
same run, which points at the machine (a scheduling or wall-clock excursion), not at either
test.

So the fixed sleeps are gone. A new `waitUntilGone(read, description)` helper in
`tests-integration/utils.ts` polls until the field is actually gone, which removes the guess
from both directions: it waits exactly as long as the expiry takes, and still fails — after
5 s, reporting how long the value survived — if the field never expires. A test that
regresses now says so instead of looking like one more flake. Applied to all five expiry
waits in the four files (including the two 5 ms-TTL waits, same pattern, smaller margin).

### 3. The audit criterion was margin, not magnitude

The first pass swept for sub-100 ms TTLs, which is the wrong filter: what matters is how far
an assertion sits from the boundary that makes it flip, not how large the TTL is. Re-audited
on margin, which caught four sites per client in the `TTL rounds to nearest second (#59)`
test — TTLs over 1000 ms with **less** slack than the 500 ms adopted above, because
`ttlSeconds` is `Math.round((expiresAt - now) / 1000)`:

- `SET k v PX 1900` + `TTL === 2` flips at `PTTL < 1500` — 400 ms of slack, so a 401 ms stall
  fails it. Now `PX 1999`, at the top of the band: the full ~500 ms the band can give.
  Same for `PX 2900` → `PX 2999` and `PEXPIRE 1900` → `1999`.
- `PTTL > 1000` after `PX 1500` had 495 ms. The raw-milliseconds check now gets its own key
  at `PX 4999`, where a seconds-rounded PTTL would answer 4000 or 5000 — both outside the
  band — so the assertion is no weaker and has ~1 s of slack.

Files: `tests-integration/{ioredis,node-redis}/key/expire.test.ts`. The `x999` values carry a
comment explaining the band, because "tidying" them back to round numbers reintroduces the bug.

Everything else in `tests-integration/` was re-checked on the same criterion: the remaining
short TTLs (`SET ... PX 1`, `HPEXPIRE 0`, `PEXPIRE missing 10 XX`, the `PX 50`
keyspace-notification waits) assert that something *has* expired after a longer wait, and the
TTL-band assertions elsewhere (`move`/`copy` at 1000 s, `hsetex`/`hgetex` at 100 s,
`EXPIRETIME` against absolute timestamps) all have seconds of slack.

No `src/` changes.

Closes #411

## Test plan

- [x] **Baseline reproduced** — pre-fix files, full mock suite 3x under 4x parallel CPU
      contention: 1/3 runs failed exactly where #411 says, `node-redis/hash/field-expire.test.ts`
      "hash writes clear or preserve field TTLs like Redis", `HINCRBYFLOAT` → `'1'` instead of
      `'2.5'` because `float` had expired and was recreated from 0.
- [x] **Deterministic margin demo** — inserting a 450 ms stall before every TTL read in the
      `#59` rounding test fails on the old values (`expected: 2, actual: 1`) and passes on the
      new ones.
- [x] **Under load** — full `test:integration:mock` 5x under 4x parallel CPU contention:
      0/5 failed (1181 pass each). The six affected files 15x under 8x contention: 15/15 pass.
      (An earlier round on the fixed-sleep version produced both the 1-in-4 reverse failure
      described above and, separately, an unrelated `ioredis/zset/zmpop-bzmpop.test.ts`
      failure with `Too many Cluster redirections / Connection is closed` — a different
      load-induced flake class, left for its own issue.)
- [x] **The polling helper is not vacuous** — a value that never disappears fails it
      (`stuck field still present after 307ms`), a value that disappears late passes.
- [x] **Ground-truthed against real Redis 8.0.6** (`redis-cli`): `HPEXPIRE 500` → `HTTL` 1,
      `HPTTL` 486, gone after 600 ms; `HSET` 0 / `HINCRBY` 2 / `HINCRBYFLOAT` 2.5, then
      `replace` kept and `counter`/`float` gone. `PX 1999` → `TTL 2` (and still 2 after a
      450 ms stall, 1 after 550 ms), `PX 2999` → `TTL 3`, `PEXPIRE 1999` → `TTL 2`,
      `PX 4999` → `PTTL 4984`.
- [x] `tsc --noEmit` and `npm run lint` clean.

Note: the `TEST_BACKEND=real` cluster backend could not be exercised — the shared local test
cluster on :30000 is `cluster_state:fail` (two masters down) from unrelated load on this
machine, and restarting shared infrastructure would disrupt other agents. Real-Redis behavior
was verified directly against the standalone instance instead, as above. The change is
test-only and adds no command surface.


tests-integration/ioredis/hash/field-persist.test.ts
```diff
@@ -43,7 +43,7 @@ describe(`Hash Commands Integration (${testRunner.getBackendName()})`, () => {
         [1],
       )
       assert.deepStrictEqual(
-        await directClient.hpexpire(key, '20', 'FIELDS', '1', 'soon'),
+        await directClient.hpexpire(key, '500', 'FIELDS', '1', 'soon'),
```
fatal10110, line 47: **Pre-existing, not introduced by this PR — flagging because `500` is the moment to pin it. Not a reason to hold the PR.**

Real Redis computes hash-field `HTTL` with **ceiling**, while key-level `TTL` uses **round-to-nearest**. Two different rules — ground-truthed on redis 8.0.6 via atomic Lua (no round-trip skew):

```
# HTTL after HPEXPIRE <ms>
  20 -> 1     400 -> 1     999 -> 1    1000 -> 1
1001 -> 2    1400 -> 2    1500 -> 2    2400 -> 3     <- ceil(ms/1000)

# TTL after SET ... PX <ms>
 400 -> 0     500 -> 1    1400 -> 1
1500 -> 2    1900 -> 2    2400 -> 2    2900 -> 3     <- round(ms/1000)
```

This mock applies the *key-level* rule to both. `src/commands/hashes.ts:492` calls `ttlSeconds()`, which is `Math.round((expiresAt - Date.now()) / 1000)` at `src/commands/helpers.ts:81`. The helper's own comment — "Redis rounds remaining time to the nearest second ((ms+500)/1000), not ceil/floor" — is correct for key TTL and wrong for HTTL.

Measured divergence (mock `ttlSeconds` vs real `HTTL`):

```
  20ms -> mock 0 | real 1
 500ms -> mock 0 | real 1
1400ms -> mock 1 | real 2
2400ms -> mock 2 | real 3
```

**Why this matters for the value chosen here:** the unchanged assertion on line 64, `assert.ok(seconds[2] >= 0 && seconds[2] <= 1)`, accepts both the mock's `0` and real's `1`. That was already true at `20` (mock `0`, real `1`), so **this PR neither introduces nor worsens the gap** — the assertion was equally blind before. But it does pass up a free chance to close it: `1200` with `assert.strictEqual(seconds[2], 2)` would pin real Redis's ceiling and fail the mock today.

Minor, same line: `500` sits exactly on `Math.round`'s half-way boundary, so `seconds[2]` is `0` or `1` depending on whether `Date.now()` has ticked since the write. Deterministic at `20`; nondeterministic now. No failure today, but anyone tightening this assertion later inherits an instant flake. `600` or `750` stays unambiguous.

Same on the node-redis twin: `tests-integration/node-redis/hash/field-persist.test.ts:46`, assertion at line 59.

Suggest a follow-up issue for `ttlSeconds` rather than growing this PR.

tests-integration/ioredis/hash/field-expire.test.ts
```diff
@@ -255,7 +255,7 @@ describe(`Hash Commands Integration (${testRunner.getBackendName()})`, () => {
         '2.5',
       )
 
-      await delay(50)
+      await delay(600)
```
fatal10110: **Audit completeness (low, not a blocker).** The PR audited `tests-integration/` for sub-100ms TTLs and concluded the rest are safe. The filter was on TTL **magnitude**, but #411's failure mode is really about **margin** — an assertion on an exact positive TTL with less slack than the intervening round-trips can consume. A magnitude filter misses larger TTLs with tight margins.

`tests-integration/ioredis/key/expire.test.ts` has three such sites, all sitting on key-TTL's round-to-nearest boundary:

```
393  await redisClient!.set(upKey, 'v', 'PX', 1900)
394  assert.strictEqual(await redisClient!.ttl(upKey), 2)    // needs remaining > 1500ms -> 400ms slack
397  await redisClient!.set(up2Key, 'v', 'PX', 2900)
398  assert.strictEqual(await redisClient!.ttl(up2Key), 3)   // needs remaining > 2500ms -> 400ms slack
409  await redisClient!.pexpire(listKey, 1900)
410  assert.strictEqual(await redisClient!.ttl(listKey), 2)  // 400ms slack
```

Each tolerates only ~400ms between the write and the read — **tighter than the 500ms this PR just adopted as the safe floor.** Under the same 4x CPU contention that reproduced #411, `TTL` flips to `1` / `2` and these fail. Same bug class as the 20ms hash TTLs, just invisible to a sub-100ms grep.

Twin: `tests-integration/node-redis/key/expire.test.ts` lines 415, 421, 431 (same `PX 1900` / `PX 2900` / `pExpire 1900` pattern).

Note these are deliberately testing the rounding boundary, so the fix isn't simply "raise the TTL" — it's to assert a range, or re-read `PTTL` and derive the expected second, so the assertion stays meaningful when the machine is slow. Worth folding into the #411 audit or filing alongside it.

tests-integration/utils.ts
```diff
@@ -31,6 +31,47 @@ export function randomKey(): string {
...
+ * It is not a weaker assertion than `sleep(); assert.strictEqual(x, null)`: the
+ * value must still be gone, just within a generous deadline instead of at one
+ * arbitrary instant. A value that never expires still fails, and the message
+ * reports how long it survived, so a future regression says so directly
+ * instead of looking like one more flake.
+ */
+export async function waitUntilGone(
+  read: () => Promise<unknown>,
+  description: string,
+  { timeoutMs = 5000, intervalMs = 20 } = {},
```
fatal10110, line 54: **Low, not a blocker: at the 5ms sites, a 5000ms deadline doesn't catch a TTL 1000x too long.**

This is fine for the two 500ms-TTL call sites. But four call sites wait on a **5ms** field TTL and previously used `delay(40)` + `assert null`:

- `ioredis/hash/field-expire.test.ts:74` and `:119`
- `node-redis/hash/field-expire.test.ts:54` and `:95`

With the default `timeoutMs = 5000`, they now accept expiry anywhere up to ~5s. The loop reads, then checks elapsed, then sleeps 20ms, so after the last check under the deadline there's always one more read at about `start + 5000..5020ms`. A field that expires at `T_set + 5000` is normally gone by that read, and the helper returns.

I checked this on real Redis 8.0.6 with this exact loop, using `HEXPIRE h 5 FIELDS 1 f`. That's the TTL you'd get if HPEXPIRE's milliseconds were read as seconds:

```
HEXPIRE 5 (5 seconds): PASS elapsed 4634..4800   (10/10 concurrent)
  sequential:          PASS 5045, PASS 5027, PASS 4866, PASS 5018
no TTL at all:         FAIL elapsed 5009 / 5010
```

So a TTL that is ignored outright still fails, as the docblock says. But a ms→seconds mix-up, or expiry delayed by up to ~5s (e.g. lazy expiry on read broken, active expiry only), now passes at these sites, and the old `delay(40)` caught it. The ms/s case is still caught elsewhere: `field-persist` checks `HPTTL <= 500`. Delayed expiry isn't caught anywhere else.

Suggested fix, one line per site. Pass a deadline proportional to the TTL at the 5ms sites:

```ts
await waitUntilGone(() => directClient!.hget(key, 'soon'), "hash field 'soon' (5ms TTL)", { timeoutMs: 1000 })
```

That still leaves about 1s for load and rejects a 5s TTL every time. Otherwise, soften the docblock's "It is not a weaker assertion" so it doesn't over-promise.

The rest of the helper checks out: no timer is left pending on failure (`assert.fail` throws before the next `setTimeout`), a rejected `read()` propagates, both clients return `null` for a missing field, and a 20ms interval is sensible.

fatal10110:
## Re-review: `2627a31` → `a08831c` (3 commits)

**Verdict: SHIP.** One new low-severity note, posted inline, and it doesn't block the merge.

**Earlier findings**
- **Margin audit, `key/expire.test.ts` in both clients: fixed.** `PX 1999` / `PX 2999` / `PEXPIRE 1999` sit at the top of the window where round-to-nearest and `Math.floor` give different answers, `[N*1000-500, N*1000)`, which leaves ~499ms of slack. On real Redis, `PX 1999` reads `TTL 2` after a 452ms stall (`PTTL 1546`) and `TTL 1` after 551ms (`PTTL 1448`). `PX 2999` behaves the same way around `3`. The new `PX 4999` key with `pttl > 4000 && pttl <= 4999` rejects a seconds-rounded PTTL (4000 or 5000) throughout its ~999ms window: `stall 950 → 4046 pass`, `stall 1050 → 3945 fail`. The comments describing the windows are accurate, and both clients match.
- **HTTL ceiling divergence:** tracked in #432, so not raised again.

**New in this round**
- **`waitUntilGone`:** replacing the fixed sleeps with polling correctly fixes the expiry-side slack of the 600ms sleep. It adds no new assertion weakness at the 500ms sites. A TTL that is ignored outright still fails at about 5s, and the test "hash writes clear or preserve field TTLs" still fails if HSET stops clearing `replace`'s TTL, because one HPEXPIRE gives all three fields the same expiry time (checked on 8.0.6). **Low:** at the four 5ms sites, the 5000ms deadline also passes a TTL 1000× too long; see the inline comment on `utils.ts:54`.
- No `delay(` helpers are left over, and there are no unused imports.

**Verification**
- Affected files on the mock backend at `a08831c`: **33/33 pass** (6 files, 1.9s).
- CI is green on every job, including `ioredis-real` and `node-redis-real`.
- Checked against real Redis 8.0.6 (field TTLs) and 7.2.1 (key TTLs) on reviewer-owned instances. Both have been shut down.

Issue #417, opened 2026-09-22
SORT cluster guard runs on a parsed plan: wrong option reported, syntax error preempts denial, fires at MULTI queue time


fatal10110:
Found across the review and re-review of PR #407. Three separate divergences share one root cause, and none is fixable where the code currently sits — hence a single issue rather than three.

## Root cause
`ClusterPolicy`'s SORT guard runs in `beforeExecute`, against a **fully parsed** `CommandPlan`. By then the original argument order is gone and any syntax error has already thrown. Real Redis applies these rules inside `sortCommand()`'s own left-to-right argument scan, which is why it can report on order and can reach the cluster check before a later syntax error.

## The three symptoms

**1. `BY` is always reported before `GET`, regardless of argument order.**
```
real 8.0.6 : SORT l GET 'n_*' BY 'w_*'  ->  the GET error
mock       : the BY error
```
Applies to both the short (pre-7.4) and long (7.4+) wordings.

**2. A trailing syntax error preempts the cluster denial.**
```
real 8.0.6 : SORT l BY 'w_*' BADARG  ->  BY denied     (real aborts at the offending token)
mock       : ERR syntax error                          (parses the whole command first)
```
Real does give `ERR syntax error` when the junk *precedes* `BY`, so this is genuinely about scan order, not about which error wins.

**3. Inside `MULTI`, the guard fires at queue time.** Real Redis replies `+QUEUED` and surfaces the error inside `EXEC`, so a transaction real Redis executes is aborted wholesale here.

## Why the obvious fixes do not work
- **Cannot scan `plan.rawArgs` for token order.** Option values are opaque and can look like option names — `SORT k GET BY BY GET` is legal, and real Redis returns three nils for it. Any token-matching approach mis-parses it.
- **Symptom 2 is unreachable from the policy layer entirely**, and from `execute` too: `parseSort` throws during `plan()`, so `beforeExecute` never runs.

## Proposal
Move the cluster check into SORT's own argument scan — i.e. have `parseSort` preserve option order (or evaluate the cluster rule as it scans), and have SORT raise the cluster error itself rather than having `ClusterPolicy` inspect a finished plan.

That single change fixes all three: the error reported is whichever option the scan reaches first, the cluster error is raised before a later token fails to parse, and because the check becomes part of command execution rather than a pre-execution policy, `MULTI` queues normally and the error surfaces in `EXEC`.

[...]

Issue #41, opened 2026-06-09
Bug: Sorted-set lex tie-breaking uses localeCompare instead of raw byte order


artur-ma:
## Description

When sorted-set members have equal scores, the mock breaks ties using `String.prototype.localeCompare` instead of `memcmp` byte order. For bytes ≥ 0x80 (any non-ASCII byte), Unicode collation order differs from byte order, causing incorrect member ordering in all sorted-set range commands.

## Root Cause

`getSortedMembers` in `src/commands/zsets.ts` (or a helper) does:
```js
a.member.toString().localeCompare(b.member.toString())
```

Real Redis uses `memcmp` (raw byte comparison), equivalent to `Buffer.compare`.

## Example of divergence

Member `'é'` (UTF-8: `0xc3 0xa9`) ranks **after** `'z'` (0x7a) by byte order, but **before** it by most locale collations.

## Fix

Replace with `Buffer.compare(a.member, b.member)`. Both `a.member` and `b.member` are `Buffer` instances in the current data model.

PR #425, opened 2026-09-22
fix: read a zset SORT source in rank order, and honor DESC on the dontsort path (#418, #426)

### LINKED ISSUE #418: SORT with a constant BY returns a zset in insertion order, not rank order
Found while re-reviewing PR #407, reproduced against real Redis 8.0.6 and the mock. **Pre-existing on `main`**, not introduced by that PR.

## Problem
When `SORT` skips sorting (`BY nosort`, or any constant `BY` pattern) and the source key is a **sorted set**, the mock returns members in insertion order rather than rank order.

```
ZADD z 1 a 3 c 2 b       insertion order: a, c, b     rank order by score: a, b, c

real 8.0.6 : SORT z BY nosort  ->  a, b, c      (rank order, matching ZRANGE)
mock       : SORT z BY nosort  ->  a, c, b      (insertion order)
```

Real Redis reads a zset in rank order for the `dontsort` path (`sortCommand()` iterates the skiplist); the mock's `readSortSource` zset branch returns the underlying insertion sequence. `ZRANGE` on the same key is correct, so the bug is isolated to `readSortSource`.

## Why it surfaces now
PR #407 added a `forceDeterministicSetOrder` comment asserting that "lists and zsets already have a defined order" — which is the correct rationale for that code, but is only true for zsets once this is fixed. The comment is right about intent and wrong about the current implementation.

## Proposal
Have `readSortSource`'s zset branch go through `getSortedMembers` (or whatever the skiplist-ordered accessor is) rather than the raw member collection. Expected to be a one-line change.

Worth checking the same path for `SORT z BY nosort LIMIT`/`GET`/`STORE` while fixing, since they all consume the same source ordering.

## Acceptance
- `SORT z BY nosort` returns rank order, matching real Redis, for a zset whose insertion order differs from its score order.
- A regression test covering a zset where the two orders differ (a same-score case is not sufficient — it would pass by accident).

### LINKED ISSUE #426: SORT ... BY nosort DESC ignores DESC for lists and zsets
Spotted while fixing #418 (PR #425); pre-existing on `main` and left out of that PR to keep its diff surgical for #417.

## Problem

On the `dontsort` path (`BY nosort`, or any constant `BY` pattern) real Redis does not just skip sorting — it iterates the source **backwards** when `DESC` is given. `sortCommand()` has a dedicated branch for this: a list is read with `listTypeInitIterator(..., desc ? LIST_HEAD : LIST_TAIL)` and a zset is walked from `zsl->tail` backwards, with `LIMIT` applied directly in that same branch. Sets have no such branch, so `DESC` is genuinely a no-op for them.

The mock's `sortElements()` in `src/commands/keys.ts` returns `[...elements]` unchanged for a constant `BY` and never consults `args.desc`, so `DESC` is dropped for every source type.

## Reproduce

```
RPUSH l a c b
ZADD  z 1 a 3 c 2 b
SADD  s a c b
```

| command | real 8.0.6 | mock |
| --- | --- | --- |
| `SORT l BY nosort DESC` | `b, c, a` | `a, c, b` |
| `SORT z BY nosort DESC` | `c, b, a` | `a, b, c` (after #418) |
| `SORT s BY nosort DESC` | `a, c, b` (unordered, DESC ignored) | `a, c, b` |

`LIMIT` composes with it, e.g. `SORT z BY nosort DESC LIMIT 0 2` -> `c, b` on real Redis.

## Expected

`DESC` reverses the source order on the `dontsort` path for lists and zsets, and stays a no-op for sets. `LIMIT`/`GET`/`STORE` then consume that reversed order.

## Fix sketch

In `sortElements()`, when `args.by` is a constant pattern, reverse the element array for `DESC` unless the source key is a set. Touches the same `src/commands/keys.ts` region as #417, so it is probably cheapest to land after that restructure.

fatal10110:
## Summary

Two bugs on `SORT`'s `dontsort` path — the path taken by `BY nosort` or any constant `BY` pattern with no `*`.

**1. Source order (#418).** `readSortSource()` returned a sorted set's raw member map, i.e. insertion order. Nothing re-orders the source on this path, so that reached the client and disagreed with `ZRANGE` on the same key. Real Redis walks the skiplist, so the zset branch now goes through `getSortedMembers()` — the same rank-ordered accessor `ZRANGE` uses, memcmp tie-break included.

**2. `DESC` (#426).** `dontsort` does not mean *ignore DESC*: for `DESC`, `sortCommand()` walks a list and a zset's skiplist **from the tail toward the head**, and applies `LIMIT` to that reversed walk. `sortElements()` never read `args.desc` on this branch. That was invisible while zsets came out in insertion order, and fixing #418 turned it into a visible regression for the common leaderboard shape (populated highest-first, where insertion order *was* the DESC answer):

```
ZADD zr 3 c 2 b 1 a
                            real 8.0.6   main mock   #418 alone   this PR
SORT zr BY nosort           a,b,c        c,b,a       a,b,c        a,b,c
SORT zr BY nosort DESC      c,b,a        c,b,a       a,b,c  <-!   c,b,a
```

So this PR lands the `DESC` fix too rather than deferring it — **#417's scope can shrink accordingly.** A set has no such branch in real Redis, so `DESC` stays a genuine no-op there; a set under `STORE`/script is still force-sorted ALPHA first and only then reversed, which already worked.

**3. One type lookup per SORT.** `readSortSource()` returns `{ type, elements }` and the type is passed to `forceDeterministicSetOrder()` and `sortElements()`, instead of each calling `getType()` again. That removes two extra lookups (each a fresh clock read) and the window where the key could expire between them.

Closes #418
Closes #426

## Test plan
- [x] `SORT with a constant BY reads a zset in rank order` — rank order `c,a,b` differs from both insertion and ALPHA order (both `a,b,c`), so the STORE and script assertions are load-bearing; covers `BY nosort`, a constant `BY` glob, `SORT_RO`, `LIMIT`, `GET`, `STORE`, and that a zset is *not* force-sorted ALPHA in a script
- [x] `SORT with a constant BY keeps a zset rank tie-break by raw bytes` — `ZADD z 1 b 1 a 1 c 1 A 1 B` -> `A B a b c`; main returned `b a c A B`
- [x] `SORT with a constant BY reads the source backwards for DESC` — list, zset and set; `SORT_RO`; `DESC`+`LIMIT` at two offsets; `DESC`+`STORE` for all three types. Red on mock with the reverse disabled, green with it
- [x] Real-backend correctness: `ioredis-real` and `node-redis-real` green on CI (isolated Redis 8.0 containers). Local command-by-command probes were against a private redis-server — Redis **7.2.1** in round 2, and a private `redis:8.0` (8.0.6) container on 6511 for the tie-order check in round 3
- [x] `tsc --noEmit`, `npm run build`, `npm run lint`, `npx prettier --check`, `npm test` (378 unit), full mock integration suite, SORT + `sort-cluster-gates` suites on mock

## Known difference, deliberately untested
`SORT <zset> BY <glob> ALPHA` when every element compares equal (e.g. all weights missing): the mock's stable sort keeps the load order, which for a zset is now rank order. Real Redis loads a zset from its **dict**, whose hash seed is random per process — for `ZADD zr 3 c 2 b 1 a` the same 8.0.6 build returned `b c a`, `c a b` and `a c b` across three restarts. That order is undefined, so no test pins it; the note lives in the ALPHA branch of `sortElements()` where the order is produced.


fatal10110:
Follow-up from the last section of the description is now tracked as #426.

src/commands/keys.ts
```diff
@@ -687,7 +688,9 @@ function readSortSource(db: RedisDatabase, key: Buffer): Buffer[] {
   if (type === 'list') return db.getList(key)!.values
   if (type === 'set') return Array.from(db.getSet(key)!.members.values())
   if (type === 'zset') {
-    return Array.from(db.getSortedSet(key)!.members.values(), m => m.member)
+    // sortCommand() walks the skiplist, so a zset source is read in rank
+    // order — the same order ZRANGE reports, not insertion order (#418).
+    return getSortedMembers(db.getSortedSet(key)!).map(entry => entry.member)
```
fatal10110: **medium — reading the zset in rank order makes `SORT <zset> BY nosort DESC` unconditionally wrong, where it used to be right for the common "inserted highest-score-first" shape.**

The fix itself is correct and I verified it (incl. the equal-score tie-break — see the test comment). But `sortElements()` returns `[...elements]` for a constant `BY` and never consults `args.desc`, so the `DESC` reply is now exactly the *reverse* of real Redis for **every** zset, rather than "sometimes wrong depending on insertion order".

Reproduced against real Redis 8.0.6 and against both mock builds (`ZADD zr 3 c 2 b 1 a`):

```
# real redis 8.0.6 (redis_version:8.0.6)
127.0.0.1:6399> SORT zr BY nosort DESC
1) "c"  2) "b"  3) "a"
127.0.0.1:6399> SORT z2 BY nosort DESC LIMIT 0 2      # ZADD z2 3 three 1 one 2 two 10 ten
1) "ten"  2) "three"
```

```
# mock @ origin/main (7393647)        # mock @ pr-425 (c9620e9)
BY nosort      : [ 'c', 'b', 'a' ]    BY nosort      : [ 'a', 'b', 'c' ]   <- fixed, good
BY nosort DESC : [ 'c', 'b', 'a' ]    BY nosort DESC : [ 'a', 'b', 'c' ]   <- was right, now reversed
nosort DESC L02: [ 'three', 'one' ]   nosort DESC L02: [ 'one', 'two' ]    <- real: ten, three
```

`sortCommand()` doesn't just skip sorting on the `dontsort` path — it walks the skiplist from `zsl->tail` backwards when `DESC` is given (and applies `LIMIT` in that same branch). Reading the source in rank order is only half of that branch.

I understand the scope call — #426 is filed and the DESC fix belongs in `sortElements()`, which #417 restructures. Two things would make that defensible rather than a silent regression:

1. Extend this comment so the gap keeps an in-code marker. Before this PR the region carried an explicit `#418` pointer to its known gap; after it, the remaining gap has none, and #417 will rewrite `sortElements()` with no signal that `DESC` is unhandled on this path. Suggest: `... not insertion order (#418). DESC is not yet honoured on this path (#426).`
2. Or just land it here — it's `if (args.desc) return [...elements].reverse()` in the `isConstantSortPattern` early return at `keys.ts:711`, gated to non-set sources. Real Redis genuinely ignores `DESC` for sets (verified: `SORT s BY nosort DESC` -> `a c b`, unordered), so the set exemption is required.

Note the top-N leaderboard form (`SORT z BY nosort DESC LIMIT 0 N`) is the most likely real-world shape to hit this, and `grep -rn 'nosort' tests tests-integration` on the PR tree shows no `BY nosort DESC` case at all for any source type.

tests-integration/ioredis/sort-integration.test.ts
```diff
@@ -139,6 +139,63 @@ describe(`SORT / SORT_RO (${testRunner.getBackendName()})`, () => {
     })
   })
 
+  test('SORT with a constant BY reads a zset in rank order', async () => {
+    await withOps(async (c, k) => {
+      // sortCommand()'s dontsort path walks the skiplist, so the reply is in
+      // rank order — insertion order (a, c, b) is deliberately different.
+      await c.zadd(k('z'), '1', 'a', '3', 'c', '2', 'b')
```
fatal10110: **medium — the fixture can't discriminate rank order from an ALPHA sort, which makes the two most interesting assertions in this test (STORE and EVAL) vacuous.**

Members are `a, b, c` with distinct scores `1, 2, 3`, so **rank order and lexicographic order are the same sequence**. Every assertion here proves only "not insertion order" — none of them can tell rank order apart from a force-sorted ALPHA result.

That matters most for lines 181 and 189, whose own comment says *"a zset is already ordered, so it is never force-sorted ALPHA the way an unordered set is"*. That's the property the test is advertising, and it's the one property this data cannot test: if someone added `|| db.getType(args.key) === 'zset'` to `forceDeterministicSetOrder()` (`keys.ts:767`), `SORT z BY nosort` inside a script would be force-sorted ALPHA and still return `['a','b','c']` — plain, `sort_ro`, constant-BY, LIMIT, GET, STORE and EVAL all stay green and the regression ships.

Second gap in the same fixture: all three scores are distinct, so `getSortedMembers()`' equal-score tie-break (`src/commands/zsets/helpers.ts:14-19`) is never reached from the SORT path. That comparator is shared with ZRANGE; if it regressed to `localeCompare` (the exact bug #41 guards) or dropped to score-only, this test stays green while same-score zsets — the most common SORT-on-zset shape — silently fall back to insertion order.

I ground-truthed the behavior you *want* pinned here; the code is correct today, the test just doesn't pin it:

```
# real redis 8.0.6 — equal scores, inserted in non-lex order
127.0.0.1:6399> ZADD z 1 b 1 a 1 c 1 A 1 B
127.0.0.1:6399> SORT z BY nosort        127.0.0.1:6399> ZRANGE z 0 -1
1) "A" 2) "B" 3) "a" 4) "b" 5) "c"      1) "A" 2) "B" 3) "a" 4) "b" 5) "c"
```
```
mock @ pr-425 : SORT -> [ 'A', 'B', 'a', 'b', 'c' ]   (matches real; ZRANGE agrees)
mock @ main   : SORT -> [ 'b', 'a', 'c', 'A', 'B' ]   (insertion order)
```

One fixture change closes both gaps: pick scores so rank order *disagrees* with lexicographic order, and include an equal-score pair — e.g. `ZADD z 1 c 2 b 3 a` (rank `c, b, a`; ALPHA would give `a, b, c`), plus a `1 A`-style same-score member. Every existing assertion then discriminates both the accessor bug and an accidental force-sort.

Same applies to the node-redis twin, `tests-integration/node-redis/sort-integration.test.ts:189` (comment at :233).

(For what it's worth, the rest of the test is solid — genuinely red on `main` for all eight assertions, `node:test` + `assert` only, `withOps` mints a fresh hash tag so there's no key leakage or ordering dependence, and the ioredis/node-redis twins are argument-for-argument equivalent.)

src/commands/keys.ts
```diff
@@ -687,7 +688,9 @@ function readSortSource(db: RedisDatabase, key: Buffer): Buffer[] {
   if (type === 'list') return db.getList(key)!.values
   if (type === 'set') return Array.from(db.getSet(key)!.members.values())
   if (type === 'zset') {
-    return Array.from(db.getSortedSet(key)!.members.values(), m => m.member)
+    // sortCommand() walks the skiplist, so a zset source is read in rank
```
fatal10110, line 701: **low — removed-behavior audit: this also changes a second, unmentioned path — `SORT <zset> BY <glob> ALPHA` with tied weights.**

The PR describes itself as touching only the constant-`BY` (`dontsort`) path, and the comment here only talks about that path. But `readSortSource()`'s output order is also the tie-break-by-stability input for the ALPHA branch:

```js
// keys.ts:715-721 — ALPHA branch, no fallback on the element itself
[...elements].sort((a, b) =>
  Buffer.compare(sortByValue(a, args, db), sortByValue(b, args, db)),
)
```

Unlike the numeric branch (`keys.ts:730`, which does tie-break on the element), this comparator returns `0` for two members whose `BY` weight is the same, so `Array.prototype.sort`'s stability decides — and that input order was the `members` Map insertion order before this PR and is rank order after it. So `SORT zkey BY w_* ALPHA` changes its reply for tied weights, on a code path the description never mentions.

This is very likely an *improvement* — real Redis' `qsort`/`pqsort` is unstable there so the order is unspecified anyway, and deriving it from rank order is at least deterministic. Not asking for a behavior change; just worth a line in the PR description, since Map-insertion-order was the only other thing the deleted line was still buying and the audit shouldn't be silent about it.

Blast radius is otherwise clean: `readSortSource()` has exactly one caller (`runSort`, `keys.ts:737`), and the other two `sortElements()` paths are order-insensitive (numeric ties break on the element; ALPHA-without-BY compares the elements themselves; zsets can't hold duplicates).

src/commands/keys.ts
```diff
@@ -747,9 +750,8 @@ function applySortLimit(
  * means "do not sort", but an unordered SET source whose output has to be
  * reproducible — it is written by STORE, or returned to a script — is
  * force-sorted ALPHA with the `BY` dropped. Lists have a defined order, so
- * only sets are overridden. (Real Redis leaves zsets alone for the same
- * reason; `readSortSource` reads them in insertion rather than rank order
- * here, which is tracked separately in #418.)
+ * only sets are overridden, and zsets are left alone because `readSortSource`
```
fatal10110: **low — the reworded comment is factually correct, but it drops the real-Redis grounding and leaves a circular local justification.**

First, the substance: I ground-truthed the claim and it holds on 8.0.6. `sortCommand()`'s determinism override really is `OBJ_SET`-gated, so leaving zsets out is right, and the #407 comment this replaces is properly retired:

```
EVAL "return redis.call('SORT', KEYS[1],'BY','nosort')" 1 z   -> one two three ten   (zset: not force-sorted, rank order kept)
EVAL "return redis.call('SORT', KEYS[1],'BY','nosort')" 1 s   -> a b c d             (set:  force-sorted ALPHA; SMEMBERS is d b c a)
EVAL "return redis.call('SORT', KEYS[1],'BY','nosort')" 1 l   -> c a b               (list: not force-sorted)
SORT s BY nosort STORE dst ; LRANGE dst 0 -1                  -> a b c d             (STORE force-sorts a set)
SORT z BY nosort STORE dst ; LRANGE dst 0 -1                  -> one two three ten   (STORE does not force-sort a zset)
SORT_RO s BY nosort                                           -> d b c a             (plain call: no override)
MULTI / SORT s BY nosort / EXEC                               -> d b c a             (MULTI is not a trigger)
```

The nit is the wording. `zsets are left alone because readSortSource reads them in rank order` states the reason as a property of *our* implementation, which is circular — the real reason is that a zset has an intrinsic order, which is why `sortCommand()` doesn't override it either. Every other comment in this region anchors on the upstream function it mirrors (`:699-700`, `:727-728`, `:828-831`, `:866-869`), and the deleted text did too ("Real Redis leaves zsets alone for the same reason"). As written, if `readSortSource` ever changes the stated justification evaporates and a reader has nothing upstream to check against.

Suggested: `... and zsets are left alone because they already have a defined order — sortCommand() does not override them either, and readSortSource now reads them in rank order.`

fatal10110:
Review addressed in 766955b — all four points.

**1 (MEDIUM, DESC regression).** Took the preferred fix rather than the marker: `sortElements()` now reverses the source for `DESC` on the constant-`BY` branch, gated to non-set sources. Confirmed against real 8.0.6 that the gate is right — a set ignores `DESC` on its plain reply, while a set under `STORE`/script is force-sorted ALPHA *and then* reversed (`c,b,a`), which the existing ALPHA branch already handled. New test `SORT with a constant BY reads the source backwards for DESC` covers list + zset + set, `SORT_RO`, `DESC`+`LIMIT` at two offsets, and `DESC`+`STORE` for all three types; verified red on mock with the reverse disabled, green with it. **This closes #426, so #417's scope can shrink.** PR description updated to say so.

**2 (MEDIUM, weak fixture).** Rebuilt it so the three candidate orders are all distinct — insertion `a,b,c` / ALPHA `a,b,c` / rank `c,a,b` — which makes the STORE and EVAL assertions load-bearing instead of vacuous. Added the equal-score case you specified as its own test: `ZADD z 1 b 1 a 1 c 1 A 1 B` -> `A B a b c`.

**3 (LOW, ALPHA glob tie-break).** Noted rather than covered, deliberately. With every weight missing all elements compare equal, so the reply is just the load order: real Redis iterates the zset's **dict** (`c,b,a` for `ZADD zr 3 c 2 b 1 a`), the mock now yields rank order. A test pinning real's answer would be red against the mock by design, and pinning the mock's would assert an implementation detail real Redis leaves undefined — so it is documented in a comment at `readSortSource()` with the reasoning.

**4 (LOW, doc comment).** Restored the real-Redis grounding: "Real Redis leaves zsets alone for the same reason: `sortCommand()` walks the skiplist, so a zset source already arrives in rank order — which `readSortSource` now reproduces (#418)."

Every assertion was ground-truthed command-by-command against a private redis-server on **6499**; the shared fixtures were not flushed or bounced. Real-backend runs were CI-only this round since the shared cluster is `cluster_state:fail` locally — `ioredis-real` and `node-redis-real` both pass on CI, and all 15 checks are green (`raw-real` needed one rerun for the recurring `redis-cluster` container exited(1) infra flake, unrelated to this diff).

src/commands/keys.ts
```diff
@@ -687,7 +688,13 @@ function readSortSource(db: RedisDatabase, key: Buffer): Buffer[] {
   if (type === 'list') return db.getList(key)!.values
   if (type === 'set') return Array.from(db.getSet(key)!.members.values())
   if (type === 'zset') {
-    return Array.from(db.getSortedSet(key)!.members.values(), m => m.member)
+    // sortCommand() walks the skiplist, so a zset source is read in rank
+    // order — the same order ZRANGE reports, not insertion order (#418).
+    // This is also the base order an ALPHA `BY` glob falls back to when every
+    // weight is missing and all elements compare equal. Real Redis iterates
+    // the zset's dict there, an order it deliberately leaves undefined, so
```
fatal10110: **low (comment), plus a pre-existing bug worth its own issue.** This new ALPHA-tie note has the right conclusion (don't pin the zset case), but three of its claims are wrong, and it sits in the wrong function.

Ground-truthed on redis 8.0.6. Every `BY` weight is missing, so all elements compare equal:

```
# real 8.0.6
RPUSH lt c a b e d
SORT lt BY nw_* ALPHA        -> c a b e d
SORT lt BY nw_* ALPHA DESC   -> c a b e d      <- ties keep load order under DESC too
ZADD zr 3 c 2 b 1 a
SORT zr BY nw_* ALPHA        -> b c a          <- not c,b,a as the PR body says

# mock @ 766955b                 # mock @ origin/main
lt ALPHA      -> c a b e d       lt ALPHA      -> c a b e d
lt ALPHA DESC -> d e b a c       lt ALPHA DESC -> d e b a c
zr ALPHA      -> a b c           zr ALPHA      -> c b a
```

1. **The DESC case is not "rank order".** `sortCompare()` returns `-cmp`, and `-0` is still 0, so real Redis leaves tied elements in load order under DESC as well. The mock's ALPHA branch reverses the whole sorted array (`if (args.desc) sorted.reverse()`), which reverses the ties too. On a **list** this is deterministic and observable (`c a b e d` vs `d e b a c`, and also with `LIMIT` and `STORE`). The same output on `origin/main` shows the bug predates this PR, so it doesn't block here. It still deserves its own issue. The fix is a stable sort with a negated comparator instead of sort-then-reverse; the numeric branch already works this way because its element tie-break is negated too.
2. **"deliberately leaves undefined" / "rather than a mismatch".** `sort.c` has no tie-break for ALPHA+BY, so this is an accepted, documented difference from real Redis, not a match. The zset case really is unpinnable: real Redis iterates the dict, and siphash is seeded per process.
3. **Placement.** The behavior belongs to `sortElements()`' ALPHA comparator, and it affects lists and sets as well as zsets. Someone editing that comparator won't see a note in the zset loader. Suggest moving it to the ALPHA branch and phrasing it as a known difference, e.g. *"ALPHA+BY has no element tie-break, so ties keep load order (Array.prototype.sort is stable). Real Redis also keeps load order for ties, under DESC too. For a zset, real load order is dict order and can't be reproduced."*

Also, real Redis's SORT converts a listpack zset to skiplist as a side effect (`OBJECT ENCODING` `listpack` -> `skiplist` after `SORT zc BY nosort`). That predates this PR too; I'm noting it only so the follow-up issue can cover it.

src/commands/keys.ts
```diff
@@ -706,7 +713,13 @@ function sortElements(
   db: RedisDatabase,
 ): Buffer[] {
   if (args.by && isConstantSortPattern(args.by)) {
-    return [...elements]
+    // A constant BY does not simply skip the sort: sortCommand()'s dontsort
+    // branch reads the source *backwards* for DESC — a list from its head, a
```
fatal10110: **low (comment only): the list doesn't start at its head for DESC.** It starts at the tail and walks toward the head.

```
# real 8.0.6
RPUSH l a c b
SORT l BY nosort DESC           -> b c a     (b is the tail)
SORT l BY nosort DESC LIMIT 1 1 -> c
```

In `sort.c` the iterator starts at index `len - start - 1`. `LIST_HEAD` is the direction it moves, not where it starts:

```c
li = listTypeInitIterator(sortval,
        desc ? (long)(listTypeLength(sortval) - start - 1) : start,
        desc ? LIST_HEAD : LIST_TAIL);
```

The code is right. The comment is the problem, and it contradicts itself: "reads the source *backwards* ... a list from its head". If someone later writes a list-specific LIMIT path to match `sort.c` and follows this comment, `SORT l BY nosort DESC LIMIT 0 1` returns `a` instead of `b`.

Suggest: *"a list from its tail toward the head, a zset from the skiplist tail"*. The same wording appears in the test comments at `tests-integration/ioredis/sort-integration.test.ts:227` and `tests-integration/node-redis/sort-integration.test.ts:278`, and in the PR body.

src/commands/keys.ts
```diff
@@ -706,7 +713,13 @@ function sortElements(
...
+    // zset from the skiplist tail — and applies LIMIT to that reversed walk.
+    // A set has no such branch, so there DESC really is a no-op (#426).
+    const source = [...elements]
+    if (args.desc && db.getType(args.key) !== 'set') source.reverse()
```
fatal10110: **low (design, non-blocking): `sortElements()` now looks up the source key's type again instead of reusing the type `readSortSource()` already found.**

A SORT now does three live lookups of the same key:

```
runSort
  -> readSortSource        db.getType(key)       (keys.ts:~686)
  -> forceDeterministicSetOrder  db.getType(args.key)  (pre-existing)
  -> sortElements          db.getType(args.key)  (new, here)
```

Every `getType()` goes through `getLiveEntry()` -> `evictIfExpired(entry, Date.now())` (`src/state/database.ts:78, 406, 419`), each with a **fresh** `Date.now()`. Real Redis freezes the command time, so a key can't expire partway through a command there. In the mock, a SET with a PX TTL that crosses its deadline between the first and third lookups is read as a set, then reported as `null` here, so `!== 'set'` is true and the set gets reversed. The lookup that sees it expired also evicts it and emits an `evict` -> `expired` keyspace notification between reading the source and writing a `STORE` destination.

Wire impact is essentially nil. The window is about 1ms, and a plain set reply has no defined order anyway. `forceDeterministicSetOrder` already had half of this race on `main`. I checked: the missing-key case just reverses `[]`, which is harmless. Nothing else hangs off `getType`: no hit/miss stats, LRU/LFU or tracking. So this isn't a bug worth blocking on.

It's still the wrong shape for a sort helper. `sortElements()` should order what it's given, not reach back into the db for facts the caller already had. Suggest `readSortSource()` return `{ type, elements }` and pass `type` into both `forceDeterministicSetOrder()` and `sortElements()`. That drops two lookups, gives one consistent view of the key, and removes `db` from the dontsort branch. It also fixes a small naming nit: `const source` here reuses the name of `runSort`'s `source`, which is a different thing (`const ordered` would read better).

tests-integration/ioredis/sort-integration.test.ts
```diff
@@ -139,6 +139,151 @@ describe(`SORT / SORT_RO (${testRunner.getBackendName()})`, () => {
     })
   })
 
+  test('SORT with a constant BY reads a zset in rank order', async () => {
+    await withOps(async (c, k) => {
+      // sortCommand()'s dontsort path walks the skiplist, so the reply is in
+      // rank order. The fixture keeps all three candidate orders apart, so
+      // this cannot pass by accident:
+      //   insertion a, b, c   ALPHA a, b, c   rank c, a, b
```
fatal10110: **low (comment only): the fixture doesn't keep all three orders apart.** Insertion order and ALPHA order are the same here: both are `a, b, c`.

This doesn't weaken the test. Each assertion only needs rank order to differ from insertion order (the #418 bug) and from ALPHA order (an accidental script/STORE force-sort), and rank `c, a, b` differs from both. The DESC fixture below is fully discriminating too: reverse-rank `b, a, c` differs from reverse-insertion `c, b, a`, insertion and rank. I also mutation-tested the new code. With `reverse()` disabled, or with the `!== 'set'` guard removed, the DESC test goes red in both suites (68/70).

The problem is the comment. It overstates what the fixture proves, and someone who reuses this fixture for a test that must tell insertion order from ALPHA order would get a test that passes either way. Suggest: *"rank c, a, b differs from both insertion and ALPHA order (which coincide here: a, b, c)"*. The same comment is in the node-redis twin at `tests-integration/node-redis/sort-integration.test.ts:188-190` and in the PR test plan.

fatal10110:
## Re-review at `766955b`: **ship**

I re-reviewed the one new commit on top of `c9620e9`, looking at Redis wire compatibility and at code design. Everything was checked against my own **redis 8.0.6** and a mock built from this head. The DESC fix is correct and complete, and **this PR can close #426**. All 4 earlier findings are fixed. The 4 new findings are low-severity comment or design issues. None of them blocks the merge.

### Earlier findings

|---|---|---|---|
| 1 | medium | `BY nosort DESC` ignored DESC for zsets and lists | **Fixed.** Matches real 8.0.6 in every case below |
| 2 | medium | Fixture couldn't tell rank order from ALPHA; no equal-score test | **Fixed.** Rank `c,a,b` differs from ALPHA; equal-score test `A B a b c` added (fixture comment is off, see new finding 4) |
| 3 | low | ALPHA+glob-BY tie order changed on a path the PR didn't mention | **Addressed.** Not pinning it is the right call, but the new note is partly wrong (new finding 1) |
| 4 | low | Doc comment lost its real-Redis grounding | **Fixed.** Now matches `sort.c`'s determinism override |

### DESC fix on real 8.0.6 vs the mock at `766955b`: all match

| Case | Real 8.0.6 | Mock |
|---|---|---|
| list `RPUSH l a c b`, `BY nosort DESC` / `LIMIT 1 5` | `b c a` / `c a` | same |
| zset `ZADD z 2 a 3 b 1 c`, `DESC` / `SORT_RO DESC` | `b a c` / `b a c` | same |
| zset DESC `LIMIT 0 2` / `1 5` (also `2 1`, `5 2`, `-1 2`, `0 -1`, `1 -1`, `0 0`) | `b a` / `a c` / … | same |
| set: plain / `DESC` | `a c b` / `a c b` (DESC ignored) | same behavior |
| set `DESC STORE` / in `EVAL` with DESC / `SORT_RO` in `EVAL` | `c b a` (ALPHA, then reversed) | same |
| DESC + STORE: list / zset / set | replies `3`/`3`/`4`, `TYPE` is `list` | same |
| DESC + `GET #` / `GET w_*` / missing weights (RESP2 and RESP3) | interleaved, nil kept in its slot | same |
| missing key `DESC STORE dst` | `0`, existing `dst` deleted | same |
| constant `BY konstant` / `BY konst->field` with DESC; ALPHA + constant BY + DESC | reversed; ALPHA ignored | same |
| 200-member skiplist zset and 300-element list, ASC and DESC, 9 LIMIT specs | 0 differences | same |

The `!== 'set'` carve-out is complete. A set under STORE or in a script never reaches the dontsort branch, because `forceDeterministicSetOrder()` drops `by`, sets ALPHA and keeps `desc`. A plain set really does ignore DESC.

| # | Sev | Where | Finding |
|---|---|---|---|
| 1 | low | `keys.ts:695` | New ALPHA-tie note: real zset order is `b c a`, not `c,b,a`; it's an accepted difference, not "no mismatch"; and it belongs at the ALPHA comparator. It also exposes an **older bug**: real Redis keeps tied elements in load order under DESC, but the mock reverses them (list: real `c a b e d`, mock `d e b a c`, and `origin/main` gives the same) |
| 2 | low | `keys.ts:717` | DESC doesn't read "a list from its head". It starts at the tail and walks toward the head (`len - start - 1`). Same wording in both test files |
| 3 | low | `keys.ts:721` | Design: `sortElements()` looks up `args.key`'s type again, making 3 lookups per SORT, each with a fresh `Date.now()`. That opens a ~1ms expiry race; the effect on the reply is negligible. Suggest `readSortSource()` return `{ type, elements }` |
| 4 | low | ioredis test `:147` | The fixture comment says the three orders are all different, but insertion and ALPHA are both `a, b, c`. The test still discriminates; only the comment is wrong |

### Tests
- `sort-integration` (ioredis and node-redis) on the mock: **70/70 pass**.
- Mutation check: disabling `reverse()`, or removing the `!== 'set'` guard, makes the new DESC test fail in both suites (68/70), so both lines are covered.
- The ioredis and node-redis versions are equivalent (`DIRECTION: 'DESC'`, `LIMIT {offset,count}` and `sortStore` send the same tokens).
- Not covered, all low risk because they share the path that is covered: DESC+GET, a set in `EVAL` with DESC, list DESC+LIMIT, `SORT_RO` on a list.

### Older problems for follow-up issues (not from this PR, not blocking)
All confirmed on real 8.0.6:
- **ALPHA + glob BY + DESC reverses tied elements** (new finding 1). Real keeps load order. The fix is a stable sort with a negated comparator instead of sort-then-reverse.
- **More than one `BY`**: real Redis switches to no-sort the first time it sees a constant `BY` and stays that way. `SORT l BY nosort BY w_*` gives real `a c b`, while the mock uses the last `BY` and sorts, giving `b a c`. This sits in #417's parse code.
- **Intset set order**: `SADD si 3 1 2; SORT si BY nosort` gives real `1 2 3`, mock `3 1 2` (the mock iterates sets in insertion order).
- **SORT changes a zset's encoding** on real Redis: `OBJECT ENCODING` goes from `listpack` to `skiplist`. The mock doesn't model this.

**Verdict: ship.** I'd fix new finding 2 ("from its head") before merging because the comment is simply wrong and it's a one-word fix. The others can go into the follow-ups.

fatal10110:
Follow-ups addressed in 08f84eb; all 15 checks green without reruns.

1. **DESC direction.** The comment in `sortElements()` and both test files now say DESC walks a list and the skiplist from the tail toward the head. The PR description had the same wording, so I fixed that too.
2. **ALPHA-tie note.** Moved it to the ALPHA branch of `sortElements()`, where the order is produced. It now says plainly that this is a known difference from real Redis that is left undefined and untested on purpose. On the order itself: `b c a` is correct but not stable. I ran my own `redis:8.0` (8.0.6) container and got `b c a`, `c a b` and `a c b` across three restarts, because a zset's dict hash seed is random per process. The note gives all three rather than one. My earlier `c,b,a` came from a local **7.2.1**, and I wrongly labelled that probe 8.0.6. The description is corrected.
3. **One type lookup.** `readSortSource()` returns `{ type, elements }`, and the type is passed to `forceDeterministicSetOrder()` and `sortElements()`. SORT now calls `getType()` once.
4. **Fixture comment.** Both test files now say rank order `c,a,b` differs from insertion and ALPHA order, which are both `a,b,c`. The description had the same wrong claim and is fixed.

I left the four items reserved for the #417 follow-up alone.

fatal10110:
## Re-review (round 3) at `08f84eb`: **ship**, clean

All 4 round-2 findings are fixed. No new findings, so there are no new inline comments.

| Round-2 finding | Status |
|---|---|
| 1. ALPHA-tie note (`keys.ts:695`) | **Fixed.** Moved to the ALPHA branch of `sortElements()`, now called a known difference, left untested on purpose. Checked on my own `redis:8.0.6` (fresh container, 4 starts) with `ZADD zr 3 c 2 b 1 a; SORT zr BY missing_* ALPHA`: `c a b`, `b a c`, `c b a`, `c b a`. It changes per process, as the note says. `BY nosort` / `DESC` were the same every time (`a b c` / `c b a`) |
| 2. DESC direction comment (`keys.ts:717`) | **Fixed.** "from the tail toward the head" in `keys.ts`, both test files and the PR body |
| 3. Type looked up again (`keys.ts:721`) | **Fixed.** `readSortSource()` returns `{ type, elements }`, and `forceDeterministicSetOrder()` and `sortElements()` both use that `type`. The only `getType` left in SORT runs once on the source key; the other one (`readSortPattern`) looks up the BY/GET pattern keys. Missing key and all three types behave the same as before |
| 4. Fixture comment (ioredis `:147`) | **Fixed.** "rank `c, a, b` differs from both insertion and ALPHA order (both `a, b, c`)", in both test files and the PR body |

**Tests at `08f84eb`:**
- `tsc --noEmit` is clean. `sort-integration` (ioredis and node-redis) on the mock passes **70/70**.
- Mutation check: turning off the DESC `reverse()` fails 2 tests; removing the `!== 'set'` guard fails 2; turning off the set force-sort, which now takes the passed-in `type`, fails 6. All three lines are covered.

Two small notes that don't block this PR:
- The ALPHA note's "keeps the load order" is true for ASC only. Under DESC the mock reverses tied elements. That's the ALPHA+DESC tie issue being filed separately, and the note can be updated in that fix.
- `readSortSource()` still reads the key twice (`getType`, then `getList`/`getSet`/`getSortedSet`). If the key expired in the ~1ms between those reads, the `!` would hit `null`. That code was the same before this PR, and the new comment only claims that *the rest of* SORT never looks the key up again, which is accurate.

**Verdict: ship.** Closes #418 and #426.

PR #427, opened 2026-09-22
fix: decode map replies and zset scores by RESP version (#414)

### LINKED ISSUE #414: node-redis mock: map replies decode as objects at RESP2, zset scores as numbers; #385 not fully closed
Found while reviewing PR #408, which fixes the `flat-pairs` half of #385. Two sibling divergences in the same decoder remain, both the same bug class, both one `case` arm away from the one that was fixed.

## 1. `map` / `map-pairs` decode to an object regardless of protocol
`src/core/decode-redis-value.ts` (the `map`/`map-pairs` arm).

Real node-redis returns a **flat array** at RESP2 on the raw `sendCommand` path, and an object at RESP3. The mock returns an object at both. Verified (node-redis@6 `sendCommand` vs the mock, Redis 8.0.6):

| | real RESP2 | mock RESP2 | real RESP3 | mock RESP3 |
|---|---|---|---|---|
| `HGETALL` | `["f1","v1","f2","v2"]` | `{f1:...}` ✗ | `{f1:...}` | `{f1:...}` ✓ |
| `CONFIG GET maxmemory` | `["maxmemory","0"]` | `{maxmemory:"0"}` ✗ | `{maxmemory:"0"}` | ✓ |
| `XREAD` | array of pairs | object ✗ | object | ✓ |

This means #385 is **not fully closed** by PR #408: that PR's description states `CONFIG GET` "is a map reply" and leaves it, which is true on the RESP3 axis but misses that the divergence is at RESP2.

**Caveat for whoever fixes this:** node-redis' *curated* methods (`hGetAll()`, `configGet()`) return objects at **both** protocols — only the raw `sendCommand` path is protocol-dependent. A `mapShape` knob must not reach `CommandRunner.hGetAll()`.

Two comments currently assert the opposite of the table above and should be corrected at the same time: `src/client-mocks/node-redis-mock.ts:106` ("mirrors node-redis RESP2 defaults") and `:241` ("decode() turns a RESP2 map reply into a plain object already").

## 2. Sorted-set scores decode to numbers at RESP2
`ZSCORE` / `WITHSCORES` decode to JS numbers on both protocols; real node-redis returns **strings** at RESP2 and numbers at RESP3. PR #408 made RESP3 byte-for-byte correct and left RESP2 unchanged.

Worth noting that PR #408's new test (`['a', 1, 'b', 2]`) and the paragraph it added to `docs/TESTING.md` both write a RESP2 value real node-redis never produces, so this divergence is now documented as intended behaviour. Fixing it means correcting those too.

## Suggested approach
PR #408 added `flatPairsShape` as a fourth/fifth boolean-ish knob plus a `flatPairsShapeFor(respVersion)` helper. Fixing these two the same way adds `mapShape` + `mapShapeFor()` and `doubleShape` + `doubleShapeFor()`.

`encodeRedisValue` already takes `{ version: RespVersion }`. Putting `version?: RespVersion` on `DecodeRedisValueOptions` carries the same bit, lets all three shapes derive from it, and deletes the per-shape helpers. Recommended before adding a third knob.

## Acceptance
- `HGETALL`, `CONFIG GET` and `XREAD` return the real node-redis shape at RESP2 and RESP3, via the raw path, with the curated methods still returning objects at both.
- `ZSCORE`/`WITHSCORES` return strings at RESP2, numbers at RESP3.
- The two misleading comments and the `docs/TESTING.md` paragraph are corrected.

fatal10110:
## Summary

Two sibling divergences in `decodeRedisValue`, one `case` arm either side of the `flat-pairs` one #408 fixed. Both socketless clients read a `map` / `map-pairs` reply as a plain object and a `double` as a JS number on *every* connection, so a RESP2 consumer got RESP3 shapes. Real node-redis on the raw paths returns what the wire carries. Review found the same divergence in `boolean` and `big-number`, which are fixed the same way.

`version` on `DecodeRedisValueOptions` already carried the connection's protocol for `flat-pairs`. Rather than add `mapShape` / `doubleShape` knobs with their own helpers, every protocol-dependent kind now derives from that one bit. That is six kinds: `map`, `map-pairs`, `flat-pairs`, `double`, `big-number` and `boolean`.

### The curated-method caveat

node-redis' *curated* methods are protocol-independent exactly where its own `transformReply` is. `hGetAll()` and `configGet()` build their object from the RESP2 flat array as readily as from the RESP3 map, so they return an object on **both** protocols. Letting the version switch reach `CommandRunner.hGetAll()` would break the curated API, so it decodes through a dedicated `decodeMapReply` helper instead. An error reply (a wrong-type key) still throws, as it does from real node-redis, and any other non-map reply fails loudly instead of resolving to a string typed as an object.

The facade's raw paths follow the protocol, as in real node-redis: `sendCommand()`, `eval()` (real node-redis gives `EVAL` no `transformReply`), and `multi().addCommand(…).exec()`.

### Real-client ground truth

`node-redis@6.0.0` against real Redis, via `sendCommand` so no per-command `transformReply` is in the way:

| command | `RESP: 2` | `RESP: 3` |
| :-- | :-- | :-- |
| `HGETALL h` | `["f1","v1","f2","v2"]` | `{f1:"v1",f2:"v2"}` |
| `HGETALL missing` | `[]` | `{}` |
| `CONFIG GET maxmemory` | `["maxmemory","0"]` | `{maxmemory:"0"}` |
| `XREAD … STREAMS s 0` | `[["s",[…]]]` | `{s:[…]}` |
| `ZSCORE z b` | `"2.5"` | `2.5` |
| `ZSCORE z missing` | `null` | `null` |
| `ZINCRBY z 1.5 a` | `"2.5"` | `2.5` |
| `ZRANGE z 0 -1 WITHSCORES` | `["a","1","b","2.5"]` | `[["a",1],["b",2.5]]` |
| `ZPOPMIN z 1` | `["a","2.5"]` | `[["a",2.5]]` |
| `HELLO 3` / `HELLO 2` | object / flat array: the reply is already in the *new* protocol | same |
| `EVAL "redis.setresp(3); return true" 0` | `1` | `true` |
| `EVAL "redis.setresp(3); return false" 0` | `0` | `false` |
| `EVAL "return {big_number=\"12345678901234567890\"}" 0` | `"12345678901234567890"` | `12345678901234567890n` |
| `EVAL "redis.setresp(3); return {map={[{double=1/0}]=1}}" 0` | `["inf",1]` | `{Infinity:1}` |

The curated methods at both protocols: `hGetAll()` → `{f1:"v1",f2:"v2"}`, `configGet()` → `{maxmemory:"0"}`, `zScore()` → `2.5`; `hGetAll()` on a string key throws `WRONGTYPE`.

`map` and `map-pairs` differ only at RESP2, matching `encodeRedisValue`: a `map` is flattened into one array, and a `map-pairs` (XREAD) is written as an array of two-element arrays. A double used as a RESP3 map key keeps JavaScript's spelling (`Infinity`), because node-redis parses it to a number and keys the object with `String()`.

Double text formatting is unchanged from main; the d2string/profile-aware rewrite is tracked in #451 (branch feat/redis-double-format-d2string).

Nothing is compatibility-gated: the shape split is protocol-dependent, not Redis-version-dependent. No command implementation changes.

### Corrections carried in the same PR

- `src/client-mocks/node-redis-mock.ts`: the reply-decoding header comment, the `hGetAll` comment and the `respVersion` getter doc now state the protocol split and the curated-method exception. `decodeScalarReply`'s doc now tells the next implementer to ask what node-redis' `transformReply` produces, not to pass the live version.
- `docs/TESTING.md`: the paragraph #408 added documented `['a', 1, 'b', 2]`, a RESP2 value real node-redis never produces. It is replaced with a per-reply table and the curated-method note. The Lua note separates Redis's rule (booleans need `redis.setresp(3)`) from a gap in the mock's Lua engine.
- `tests/flat-pairs-resp3.test.ts`: RESP2 expectations move from `['a', 1, 'b', 2]` to `['a', '1', 'b', '2']`.
- `CHANGELOG.md`: a **BREAKING** `Changed` entry for the socketless clients' visible shapes.

Closes #414

## Test plan

- [x] New `tests/resp2-map-and-double-shapes.test.ts` (18 tests). It covers HGETALL / `CONFIG GET` / XREAD / ZSCORE / ZINCRBY / `ZRANGE WITHSCORES` / ZPOPMIN on both socketless clients at RESP2 and after `HELLO 3`, plus empty keys, `inf` / `-inf`, `HELLO`'s own reply shape, and MULTI replays. On the curated side it covers `hGetAll()` at both protocols, including a thrown `WRONGTYPE` and a loud failure on a non-map reply. It also covers `redis.setresp(3)` booleans and big numbers through `command()`, `sendCommand()`, `eval()` and `multi().exec()`. Red before the fix: 7 of the first 11 cases failed. The cases added in later review rounds each target a specific fix.
- [x] `tests/core/decode-redis-value.test.ts`: option-level assertions for each of the six kinds at each version, for *both* clients' option constants, plus a double map key pinned to the real shape at both protocols.
- [x] `npm test`: 403 pass, 0 fail.
- [x] `npm run test:integration:mock`: 1181 pass, 0 fail.
- [x] `npm run build`, `npm run lint`, `prettier --check`: clean.
- [x] Encoder byte-identity against `main` for doubles: 20,000 values (every bit pattern, the ±2^62 seam, powers of ten, small fractions, specials) encoded at RESP2 and RESP3 by `origin/main`'s encoder and this branch's, **0 differences**; the RESP2 decode reads back the exact wire text for each. As a negative control, the same check flags 14,100 differences against the parked d2string encoder.
- [ ] Real-backend integration suites (`TEST_BACKEND=real`) at this head: **not run locally**. They drive real clients over TCP, never touch the socketless decode path, and the node-redis suites FLUSHALL the shared cluster. The real-backend signal is CI. The behavioural checks against real Redis are the probes above, all against private servers (Redis 8.0.6 and 6.2 containers, and a local 7.2.1).


src/client-mocks/node-redis-mock.ts
```diff
@@ -1193,13 +1200,35 @@ function decodeReply(
...
+ * *both* protocols, the way node-redis' `transformReply` does — see #414.
+ * Entries still decode under the connection's own version, so only the
+ * container shape is pinned. A non-map reply (nothing currently produces one
+ * here) yields `{}` rather than a surprise scalar.
+ */
+function decodeMapReply(
+  value: RedisValue,
+  respVersion: RespVersion,
+): { [key: string]: NodeRedisReply } {
+  if (value.kind !== 'map' && value.kind !== 'map-pairs') {
```
fatal10110: **[high · Redis-compat + architecture — both lenses]** `decodeMapReply()`'s `{}` fallback swallows **error** replies, so `hGetAll()` on a wrong-type key now resolves to `{}` instead of throwing.

The guard runs *before* `decodeRedisValue`, which is the only thing that throws on `kind: 'error'`. `run()` does not throw — `CommandExecutor` returns errors as `RedisValue`s — so a `WRONGTYPE` reaches here as `{ kind: 'error' }` and falls straight into `return {}`.

The doc comment directly above ("A non-map reply (nothing currently produces one here) yields `{}` rather than a surprise scalar") is factually wrong: `HGETALL` against a string/list key produces exactly such a reply, every time.

Worse, it is *indistinguishable* from the legitimate empty result — `hGetAll('missing')` also yields `{}` — so the new test at `tests/resp2-map-and-double-shapes.test.ts:227` passes either way and nothing catches this.

Reproduced three ways (same script, same keys):

```
===== real node-redis@6, RESP2 and RESP3 =====
curated hGetAll(string key) THREW :: WRONGTYPE Operation against a key holding the wrong kind of value
curated hGetAll(list key)   THREW :: WRONGTYPE Operation against a key holding the wrong kind of value

===== origin/main (pre-PR) facade =====
hGetAll(list key) THREW :: ErrorReply :: WRONGTYPE Operation against a key holding the wrong kind of value

===== this PR (0aed867) facade, RESP2 and RESP3 =====
hGetAll(list key)   RETURNED :: {}
sendCommand HGETALL THREW   :: ErrorReply :: WRONGTYPE Operation against a key holding the wrong kind of value
```

So the raw path still throws and the curated path no longer does — a straight regression against both `main` and the real client, at both protocols. `NodeRedisMockCluster` inherits `hGetAll` from `CommandRunner`, so a `-MOVED`/`-CROSSSLOT` error value is swallowed the same way (read from code; not executed against a real cluster).

Fix: only special-case `map`/`map-pairs`, and let everything else fall through to `decodeRedisValue` so the throw survives — e.g.

```ts
if (value.kind !== 'map' && value.kind !== 'map-pairs') {
  decodeRedisValue(value, { ...NODE_REDIS_DECODE_OPTIONS, version: respVersion })
  return {}
}
```

and please add `await assert.rejects(() => c.hGetAll(listKey), /WRONGTYPE/)` — that is the red-before-green test this behaviour is missing.

src/core/decode-redis-value.ts
```diff
@@ -109,8 +123,16 @@ export function decodeRedisValue(
...
+      // ZSCORE is `"2.5"` there, not `2.5`.
+      if (options.version === 3) {
+        return value.value
+      }
+      const text = formatRedisDouble(value.value)
+      return options.returnBuffers ? Buffer.from(text) : text
+    }
     case 'boolean':
       return value.value
     case 'big-number':
```
fatal10110: **[medium · Redis-compat]** `big-number` and `boolean` have exactly the RESP2 divergence this PR just fixed for `double`, and are left alone — while the new doc block at line 28 asserts the protocol-dependent set is complete.

`encodeResp2` writes a `big-number` as a **bulk string** and a `boolean` as `:1`/`:0`, so a RESP2 client reads back a string / a number. This switch returns a JS `bigint` / a JS `boolean` at *both* versions, so encode and decode disagree for these two kinds in the same `switch` the PR is correcting.

Ground truth, `node-redis@6` raw `sendCommand` against Redis 8.0.6:

```
===== RESP2 =====
EVAL {big_number=...}      :: string :: "123456789012345678901234567890"
EVAL return true (boolean) :: number :: 1
===== RESP3 =====
EVAL {big_number=...}      :: bigint :: 123456789012345678901234567890n

$ redis-cli -p 6399 --no-raw eval "return {big_number='123456789012345678901234567890'}" 0
"123456789012345678901234567890"     # RESP2: bulk string
$ redis-cli -p 6399 -3 --no-raw eval "return {big_number='123456789012345678901234567890'}" 0
Unknown reply type: 13                # RESP3: '(' big-number frame
```

Reachable in principle via `src/core/lua-runtime.ts` (`RedisValue.bigNumber` / `RedisValue.boolean`); latent today only because the Lua `{big_number=…}` table conversion separately returns `[]` in the mock. The day that is fixed, RESP2 `EVAL` big-numbers regress.

Not asking you to grow the PR — but either fix the two arms alongside `double`, or narrow the new doc comment so it does not claim the decoder is now complete on the protocol axis. As written, the comment will stop the next reader from looking.

src/client-mocks/node-redis-mock.ts
```diff
@@ -1193,13 +1200,35 @@ function decodeReply(
 
 /**
  * Decode a reply a curated method immediately narrows to a scalar (or a flat
- * string array). No protocol-dependent kind can reach these, so the version
- * passed is unobservable.
+ * string array). None of the commands behind those methods replies with a
+ * protocol-dependent kind — no `map`, `double` or pair reaches here — so the
+ * version passed is unobservable. A curated method for a command that *does*
+ * (a `zScore`, say) needs the connection's real version, not this.
```
fatal10110: **[medium · Redis-compat + architecture]** This doc comment gives the **wrong recipe** for the next curated method, and would reintroduce the exact bug this PR fixes.

> A curated method for a command that *does* (a `zScore`, say) needs the connection's real version, not this.

Real node-redis' `zScore()` is protocol-**in**dependent — a `number` at both protocols — because its `transformReply` coerces. Threading "the connection's real version" into it would hand back `'1'` at RESP2. Ground truth, `node-redis@6` vs Redis 8.0.6:

```
CURATED zScore int                 | RESP2: 1 (number)        | RESP3: 1 (number)
CURATED zIncrBy                    | RESP2: 4 (number)        | RESP3: 4 (number)
CURATED zRangeWithScores           | RESP2: [{"value":"int","score":1},…] | RESP3: same
CURATED zPopMin                    | RESP2: {"value":"int","score":1}     | RESP3: same
CURATED hRandFieldCountWithValues  | RESP2: [{"field":"f1","value":"v1"},…] | RESP3: same
raw     ZSCORE via sendCommand     | RESP2: "1" (string)      | RESP3: 1 (number)
```

`zScore` is the same case as `hGetAll`, not its opposite: it needs a pinned shape, not the live version.

Second, related point on the same helper: the pin `version: 2` plus `asNumber()`'s throw-on-non-number has quietly become a landmine. Since the `double` arm now returns a **string** at v2, the moment anyone writes `asNumber(await this.run(['ZSCORE', …]))` it throws `expected an integer reply, got string` where pre-PR it returned `2.5`.

I traced every current call site of `asNumber`/`asString`/`asStringOrNull`/`asStringArray` (DEL, EXISTS, INCR, EXPIRE, TTL, HSET, LPUSH, RPUSH, SADD, ZADD-without-INCR, ZUNIONSTORE, PUBLISH, GET, SET, HGET, MSET, LRANGE, SMEMBERS, ZRANGE-without-WITHSCORES) plus `NodeRedisMockMulti` and `NodeRedisMockCluster` — **none** reaches a `double`/`map`/`map-pairs`/`flat-pairs` today, so the claim holds currently and `hGetAll` really is the only curated method that needed `decodeMapReply`. No live miss here; it is the guidance that is wrong.

Suggest rewording to: *a curated method for a `double`-replying command must pin the number shape (as `hGetAll` pins the object shape), not read the connection's version* — and consider having `asNumber` accept a numeric string so the pin stops being load-bearing.

src/core/decode-redis-value.ts
```diff
@@ -145,6 +171,27 @@ export function decodeRedisValue(
...
+ * return an object on both protocols while the raw `sendCommand` path follows
+ * the protocol. See #414.
+ */
+export function decodeRedisMapEntries(
+  entries: readonly [RedisValue, RedisValue][],
+  options: DecodeRedisValueOptions,
+): { [key: string]: NativeRedisReply } {
+  const out: { [key: string]: NativeRedisReply } = {}
+  for (const [key, value] of entries) {
+    out[decodeRedisKey(key)] = decodeRedisValue(value, options)
```
fatal10110, line 209: **[low · Redis-compat]** `decodeRedisKey` is the other half of this function, and it does **not** share `formatRedisDouble` — so the "encode and decode cannot drift" invariant the PR adds to `resp-encoder.ts` does not hold for map keys.

At `src/core/decode-redis-value.ts:205-207` a `double` key still goes through `String(value.value)`. Same key, two spellings depending on protocol: at RESP3 it becomes the property name `"Infinity"`, while at RESP2 the same key flattens through `decode()` → `formatRedisDouble` → `'inf'`, which is what the wire says.

Real Redis 8.0.6 double spelling (via Lua `{double=…}`), RESP3 then RESP2:

```
{double=1/0}   RESP3=(double) inf    RESP2="inf"
{double=-1/0}  RESP3=(double) -inf   RESP2="-inf"
{double=0/0}   RESP3=(double) nan    RESP2="nan"
{double=-0.0}  RESP3=(double) -0     RESP2="-0"

JS String(): Infinity | -Infinity | NaN | 0   — all four diverge.
```

(As an aside, this run also confirms the `'-0'` branch in `formatRedisDouble` is *correct* — real Redis does write `-0` for a genuine `-0.0`. It is only unreachable through `ZADD`, which normalises.)

No command emits a double map key today, so this is latent — but the fix is one line (`case 'double': return formatRedisDouble(value.value)`), and leaving it makes the new shared-formatter comment untrue at the moment it is written.

docs/TESTING.md
```diff
@@ -256,15 +256,23 @@ Three flavours, depending on which client you want to look like:
...
-object where RESP2 puts a flat array on the wire — both tracked in
-[#414](https://github.com/fatal10110/js-redis-server/issues/414).
+start on RESP2 and follow a `HELLO 3` the way a real connection does, so every
+reply whose shape the protocol decides changes with it — RESP2 has no map,
+double or pair type, and these clients hand back what a real client reads off
+the wire at each version:
+
+| reply                                 | RESP2                     | RESP3                     |
+| :------------------------------------ | :------------------------ | :------------------------ |
+| `ZRANGE … WITHSCORES`, `HRANDFIELD … WITHVALUES` | `['a', '1', 'b', '2']` | `[['a', 1], ['b', 2]]` |
```
fatal10110: **[low · Redis-compat, docs]** This row puts `ZRANGE … WITHSCORES` and `HRANDFIELD … WITHVALUES` on one line, but their RESP3 shapes are not the same: a `WITHSCORES` score is a `double` and comes back as a **number**, while a `WITHVALUES` value is a hash value and stays a **string**.

Ground truth, `node-redis@6` raw `sendCommand` against Redis 8.0.6:

```
RESP2 HRANDFIELD WITHVALUES    ["a","1","b","2"]
RESP2 ZRANGE WITHSCORES        ["a","1","b","2"]
RESP3 HRANDFIELD WITHVALUES    [["a","1"],["b","2"]]     <-- strings
RESP3 ZRANGE WITHSCORES        [["a",1],["b",2]]        <-- numbers
```

The table's RESP3 cell (`[['a', 1], ['b', 2]]`) is right for `WITHSCORES` and wrong for `WITHVALUES`. The RESP2 cell happens to be right for both, which is what makes the row look consistent.

The code is correct — `tests/flat-pairs-resp3.test.ts` pins `HRANDFIELD … WITHVALUES` to `['f1','v1',…]` / `[['f1','v1'],…]` and this PR correctly left those alone. It is only the new doc table that conflates them. Suggest splitting the row, or dropping `HRANDFIELD … WITHVALUES` from it and noting separately that its values stay strings at both protocols.

src/core/decode-redis-value.ts
```diff
@@ -24,15 +24,29 @@ export type NativeRedisReply =
  */
 export type DecodeRedisValueOptions = ClientDecodeOptions & {
   /**
-   * RESP version the connection served this reply under. Some replies are
-   * shaped by the protocol rather than by the client: `flat-pairs`
-   * (WITHSCORES / WITHVALUES) is a flat `[k, v, k, v, …]` array on RESP2 and
-   * `[[k, v], …]` tuples on RESP3, so a RESP3 consumer iterating
-   * `for (const [field, value] of reply)` must not be handed a flat array.
+   * RESP version the connection served this reply under. RESP2 has no map,
+   * double or pair type, so three reply kinds are shaped by the protocol
```
fatal10110: **[low · docs]** "three reply kinds … and all three derive from this one bit" is followed by **four** bullets: `map`, `map-pairs`, `flat-pairs`, `double`.

Worth fixing because this block is now the canonical description of the protocol-dependent decode contract, and "all four kinds derive from the existing `version` bit" is the PR's own headline claim — so the prose contradicts the framing it is there to support.

(See also the separate note on `big-number` / `boolean`: if those stay unfixed, this sentence should say the list is the set handled *today*, not the set that exists.)

tests/resp2-map-and-double-shapes.test.ts
```diff
@@ -0,0 +1,285 @@
...
+      )
+    })
+
+    test('curated hGetAll() stays an object on both protocols', async () => {
+      // node-redis' own `hGetAll` transformReply builds the object itself, so
+      // the curated method is protocol-independent where `sendCommand` is not.
+      // Letting the protocol switch reach it would break the curated API.
+      const resp2 = await seeded(2)
+      assert.deepStrictEqual(await resp2.hGetAll('h'), { f1: 'v1', f2: 'v2' })
+      assert.deepStrictEqual(await resp2.hGetAll('missing'), {})
```
fatal10110, line 239: **[low · test quality]** These two `{}` assertions cannot fail for the reason they are written.

`hGetAll('missing')` legitimately yields `{}` *and* the new `decodeMapReply` fallback returns `{}` for literally any non-map reply — including an `error`. So this test is satisfied by the correct path and by the swallow path alike, which is why the WRONGTYPE regression (see my comment on `node-redis-mock.ts`) ships undetected.

`HGETALL` on a missing key is `RedisValue.map([])` (`src/commands/hashes.ts`), so the empty case is real; what is missing is the case that separates them:

```ts
await c.sendCommand(['RPUSH', 'l', 'x'])
await assert.rejects(() => c.hGetAll('l'), /WRONGTYPE/)
```

Verified against `node-redis@6` + Redis 8.0.6: the real curated `hGetAll()` throws `WRONGTYPE Operation against a key holding the wrong kind of value` at **both** protocols, and so did this facade on `origin/main`.

tests/resp2-map-and-double-shapes.test.ts
```diff
@@ -0,0 +1,285 @@
...
+
+    test('HELLO itself answers in the shape of the version it lands on', async () => {
+      // HELLO 3's own reply is already RESP3, so it is a map; HELLO 2's is the
+      // RESP2 flat array. Both clients read the protocol off the session after
+      // the command has run, so this falls out of the same switch.
+      const c = await seeded(2)
+
+      const upgraded = await c.command('HELLO', 3)
+      assert.ok(
+        upgraded !== null && !Array.isArray(upgraded),
```
fatal10110: **[low · test quality]** `!Array.isArray(upgraded)` is weaker than the property the test name claims to pin — a string, a number or a Buffer all satisfy it, so the assertion would stay green if `HELLO`'s reply kind regressed away from a map.

The `HELLO 2` half is genuinely red-before-green, so the test does earn its place; it is only the first half that is loose. Something like `assert.ok(Object.keys(upgraded as object).includes('proto'))` pins the actual shape.

For reference, the real thing (`node-redis@6`, Redis 8.0.6) — the reply is indeed in the *new* protocol in both directions, so the behaviour under test is correct:

```
real, on a RESP2 conn:  HELLO 3 -> object {"server":"redis",…,"proto":3,…}
real, on a RESP3 conn:  HELLO 2 -> flat array ["server","redis",…,"proto",2,…]
```

src/core/decode-redis-value.ts
```diff
@@ -109,8 +123,16 @@ export function decodeRedisValue(
         return Number(value.value)
       }
       return value.value
-    case 'double':
-      return value.value
+    case 'double': {
```
fatal10110, line 131: **[low · release hygiene]** No `CHANGELOG.md` entry for what is a user-visible behaviour change on the **published root facade**.

`createInMemoryClient()` defaults to RESP2, and `InMemoryRedisClient.decode()` passes `version: this.session.protocolVersion`, so after this PR:

- `client.command('HGETALL', 'h')` returns `['f', 'v']` where it returned `{ f: 'v' }`
- `client.command('ZSCORE', 'z', 'b')` returns `'2.5'` where it returned `2.5`

The PR's own test churn is the evidence — `tests/in-memory-client.test.ts` and ~14 flipped assertions in `tests/flat-pairs-resp3.test.ts`. `CHANGELOG.md`'s preamble singles this entry point out ("**`js-redis-server`** — the curated consumer facade (`src/index.ts`). Changes here affect everyone"), and a downstream consumer upgrading has nothing to point at.

Flagging rather than blocking: sibling #408 landed the same class of change without an entry, so this is a repo-wide gap as much as a PR one — your call whether to fix it here or separately. (The new `formatRedisDouble` / `decodeRedisMapEntries` exports are *not* on the `/core` surface — `src/internal.ts` does not re-export `decode-redis-value` — so there is no export-surface note needed.)

fatal10110:
Review fixes pushed in 4d31d76. All nine findings addressed; nothing deferred.

**1 (HIGH, blocker) — `decodeMapReply` swallowed errors.** Confirmed and fixed. The `{}` fallback ran ahead of `decodeRedisValue`, so an `error` reply never reached the throw arm: `hGetAll()` on a wrong-type key returned `{}`, indistinguishable from the missing-key `{}`. A non-map reply now falls through to the ordinary decoder.

Ground truth, node-redis@6 against a real server, at **both** protocols:

```
hGetAll('str')    -> THREW SimpleError: WRONGTYPE Operation against a key holding the wrong kind of value
hGetAll('nohash') -> {}
```

**7 (why it shipped) — fixed too.** The curated test now has a wrong-type case at RESP2 and RESP3 asserting a `WRONGTYPE` rejection, which is the assertion that tells the correct path from the swallow path. The empty-hash `{}` assertions stay, and the comment says what each one is for.

**2 — `boolean` and `big-number`.** Fixed the same way rather than narrowing the doc: both are reachable (Lua's `redis.setresp(3)`, `luaReplyToRedisValue`), and our own encoder already writes `:1`/`:0` and a bulk string for them at RESP2.

```
EVAL 'return {big_number="12345678901234567890"}' 0
  RESP2 -> "12345678901234567890"      RESP3 -> 12345678901234567890n
```

`boolean` has no live real-Redis producer to probe (Lua's plain `true` converts to integer 1 before it reaches the wire), so it is pinned to what `encodeRedisValue` writes — `1` / `0` at RESP2 — which is the invariant that matters here. Both now have unit coverage, and the option doc lists all six kinds plus "every other kind decodes the same way at both versions".

**3 — `decodeScalarReply` guidance.** Rewritten. It no longer says to pass the live version; it says to ask what node-redis' `transformReply` produces, notes that a real `zScore()` is a number at both protocols and a real `hGetAll()` an object at both, and points at `decodeMapReply` as the pattern.

**4 — `decodeRedisKey`.** `double` keys now go through `formatRedisDouble`, so a key and a value holding the same double agree: `inf`, not `Infinity`. Unit test added.

**5 — TESTING.md table.** Split into separate `ZRANGE … WITHSCORES` and `HRANDFIELD … WITHVALUES` rows, with a sentence saying only the pair *shape* is shared: a score is a double (string at RESP2), a hash value is a bulk string at both.

**6 — "three reply kinds ... all three".** Rewritten to enumerate six without a count that can drift.

**8 — HELLO assertion.** Now pins `proto === 3` on the object and, on the flat array, that `proto` sits at an even index with `2` after it — so the map's entries-in-order property is what is being checked, not just "not an array".

**9 — CHANGELOG.** Added a `### Changed` entry marked **BREAKING** listing the six visible shape changes on the published `createInMemoryClient()` / `createNodeRedisMock().sendCommand()` surface, with the two ways forward (`HELLO 3`, or the curated `hGetAll()`), and a pointer to #408.

I did not add the exports to `/core` — `decode-redis-value` is not re-exported from `src/internal.ts` at all, so `decodeRedisMapEntries` and `formatRedisDouble` stay module-internal and need no changelog line.

**Verification:** `npm test` 397/397, `npm run test:integration:mock` 1181/1181, build + lint + prettier clean. Real-Redis ground truth above was probed on my own server, not the shared fixtures.

src/core/decode-redis-value.ts
```diff
@@ -154,8 +216,10 @@ export function decodeRedisKey(value: RedisValue): string {
       return value.value === null ? '' : value.value.toString('utf8')
     case 'verbatim':
       return value.value.toString('utf8')
-    case 'integer':
     case 'double':
+      // The Redis spelling, not JavaScript's: `inf`, not `Infinity`.
+      return formatRedisDouble(value.value)
```
fatal10110: **[medium · regression · Redis-compat] My round-1 finding 4 was wrong, and this fix for it introduces a divergence.** Please revert this arm to `String(value.value)`.

I asked for double **map keys** to use the wire spelling, on the theory that a key and a value holding the same double should agree. Real node-redis doesn't work that way. At RESP3 it parses `,inf` to a JS number first and then uses that number as the object key, so the key is `String(Infinity)` = `"Infinity"`. `decodeRedisKey` only builds object keys (via `decodeRedisMapEntries`), and a RESP2 map is flattened through `decode()` rather than through here. So on `origin/main`, `String(value.value)` already matched the real client.

Ground truth: node-redis@6 `sendCommand` against a real redis-server 7.2.1:

```
RESP2 redis.setresp(3); return {map={[{double=1/0}]=1}}    ["inf",1]
RESP2 redis.setresp(3); return {map={[{double=-1/0}]=1}}   ["-inf",1]
RESP2 redis.setresp(3); return {map={[{double=0/0}]=1}}    ["nan",1]
RESP3 redis.setresp(3); return {map={[{double=1/0}]=1}}    {"Infinity":1}
RESP3 redis.setresp(3); return {map={[{double=-1/0}]=1}}   {"-Infinity":1}
RESP3 redis.setresp(3); return {map={[{double=0/0}]=1}}    {"NaN":1}
RESP3 redis.setresp(3); return {map={[{double=2.5}]=1}}    {"2.5":1}
```

At RESP3 the PR mock now returns `{inf:1}` where real node-redis returns `{Infinity:1}`. (The RESP2 side is correct and still is: those are array values going through the `double` value arm, not keys.) `-0` is fine either way, because `String(-0)` is `"0"`, which is what real node-redis produces too.

What to change:
- Restore `case 'double':` into the `String(value.value)` group, next to `integer` / `big-number`.
- Replace the test at `tests/core/decode-redis-value.test.ts:347-360`. Right now it locks in the wrong value, and its comment ("a key and a value holding the same double never disagree") is false for real node-redis: the wire shows `inf`, the RESP3 object key is `Infinity`. Pin `decodeRedisKey({ kind: 'double', value: Infinity }) === 'Infinity'` instead, citing the probe above.

Apologies for the churn. This one was on the review, not the implementation.

src/core/resp-encoder.ts
```diff
@@ -195,7 +195,13 @@ function formatError(value: Extract<RedisValue, { kind: 'error' }>): string {
   return value.code ? `${sanitizeErrorText(value.code)} ${message}` : message
 }
 
-function formatNumber(value: number): string {
+/**
+ * Render a `double` the way Redis spells it — `inf` / `-inf` / `nan`, and `-0`
```
fatal10110: **[low · Redis-compat, pre-existing but newly surfaced]** "Render a `double` the way Redis spells it" doesn't hold for large whole-number doubles. Real Redis switches to exponent form there, while `value.toString()` prints every digit. The encoder bug predates this PR. What's new is that RESP2 *decoding* now shows this text to callers, so `ZSCORE` on the socketless clients diverges from real node-redis.

Real redis-server 7.2.1, `ZADD zb <v> m` then `ZSCORE zb m` (RESP2 bulk string), next to JS `String(v)`:

```
1e18      real=1000000000000000000      js=1000000000000000000     match
3e18      real=3000000000000000000      js=3000000000000000000     match
5e18      real=5e+18                    js=5000000000000000000     DIVERGES
9.3e18    real=9.3e+18                  js=9300000000000000000     DIVERGES
1e19      real=1e+19                    js=10000000000000000000    DIVERGES
1e20      real=1e+20                    js=100000000000000000000   DIVERGES
9.99e20   real=9.99e+20                 js=999000000000000000000   DIVERGES
1e21      real=1e+21                    js=1e+21                   match
```

So the diverging band runs from somewhere between 3e18 and 5e18 up to 1e21, where JS switches to exponent form as well. (Round 1 checked only 1e17 and 1e21, which fall either side of this band. That's my miss.) At RESP3 both sides are numbers, so only the RESP2 text is affected.

Not blocking, since values this large are rare in practice. But either handle this range (e.g. use exponent form for integral `|v| >= 2^62`-ish, after finding Redis' exact cutoff) or soften the doc comment so it doesn't claim exactness. A follow-up issue is fine.

docs/TESTING.md
```diff
@@ -256,15 +256,30 @@ Three flavours, depending on which client you want to look like:
...
+| :------------------------- | :----------------------- | :------------------------- |
+| `ZRANGE … WITHSCORES`      | `['a', '1', 'b', '2']`   | `[['a', 1], ['b', 2]]`     |
+| `HRANDFIELD … WITHVALUES`  | `['f1', 'v1']`           | `[['f1', 'v1']]`           |
+| `HGETALL`, `CONFIG GET`    | `['f1', 'v1']`           | `{ f1: 'v1' }`             |
+| `XREAD`                    | `[['s', […]]]`           | `{ s: […] }`               |
+| `ZSCORE`, `ZINCRBY`        | `'2.5'`                  | `2.5`                      |
+
+Only the pair *shape* is shared between the first two rows: a `WITHSCORES`
+score is a double, so it is a string at RESP2 and a number at RESP3, while a
+`WITHVALUES` hash value is a bulk string at both. Lua under `redis.setresp(3)`
```
fatal10110: **[low · Redis-compat, docs]** "Lua under `redis.setresp(3)` can also return a big number" reads as though `setresp(3)` is required. For big numbers (and `{double=…}`) it isn't: real Redis converts the table without it.

```
$ redis-cli --no-raw eval "return {big_number='12345678901234567890'}" 0
"12345678901234567890"                 # RESP2, no setresp needed
$ redis-cli -3 --no-raw eval "return {big_number='12345678901234567890'}" 0
Unknown reply type: 13                 # RESP3 '(' big-number, no setresp needed
```

The mock returns `[]` for a `{big_number=…}` / `{double=…}` / `{map=…}` table **without** `setresp(3)`. That's a gap in the mock's Lua engine, not a Redis rule, so the sentence currently documents the gap as though it were Redis behaviour.

The **boolean** half is right: `setresp(3)` *is* required there. Plain `return true` is `:1` even to a RESP3 client, and plain `return false` is null; only `redis.setresp(3); return true/false` gives `#t`/`#f` at RESP3 and `:1`/`:0` at RESP2. Suggest splitting the sentence so the qualifier sits only on boolean.

CHANGELOG.md
```diff
@@ -118,6 +118,36 @@ so the PR body is not a durable home for a breaking-change note.
   identical. `RespEncodeOptions` itself is **kept** — it is still the options
   parameter of `encodeRedisValue` / `encodeRedisResult`.
 
+### Changed
+
+- **BREAKING** The two socketless clients — `createInMemoryClient()` and
```
fatal10110: **[low · docs/release notes]** Four small accuracy problems in this entry:

1. **Scope is under-stated.** It names only `createNodeRedisMock()`'s raw `sendCommand()`. But `eval()` (`node-redis-mock.ts:326`) and `multi()…exec()` replay (`:770` / `:781`) decode through the same version-following `decodeReply`, so they change shape too. The Lua big-number / boolean rows below are mostly reached through `eval()`. `docs/TESTING.md` uses the same "only the raw `sendCommand` path" wording.
2. **"instead of always handing back the RESP3 shape"** contradicts the table's own `ZRANGE … WITHSCORES` row: its *before* value `['a', 1]` is flat, not the RESP3 tuple shape, because #408 already made the pair shape protocol-dependent. Suggest something like "…instead of the RESP3 object/number shapes".
3. **Line 148 points readers at `DecodeRedisValueOptions`**, which neither published entry point exports: `src/index.ts` and `src/internal.ts` don't re-export `decode-redis-value`. That's fine for an internal note, but in a consumer-facing CHANGELOG it names a type nobody can import.
4. **Boolean row (line 135):** add "under `redis.setresp(3)`". Verified against real Redis: without it, Lua `true` is `:1` at *both* protocols and `false` is null, so "true / false → 1 / 0" is only the before/after for scripts that opt in. The before/after values themselves are correct.

src/client-mocks/node-redis-mock.ts
```diff
@@ -1148,11 +1151,15 @@ function notify(
...
 //
-// node-redis leaves most RESP2 replies untransformed, and the shared
-// RedisValue → native JS decoder already matches those defaults: bulk-string →
-// utf8 string, integer → number, map → object, array/set → array. So the
-// curated methods are thin coercions over that decoder rather than per-command
-// reply tables — uncommon commands get the same native shapes via sendCommand.
+// node-redis leaves most replies untransformed, and the shared RedisValue →
+// native JS decoder already matches what it hands back: bulk-string → utf8
+// string, integer → number, array/set → array, plus the protocol-dependent
+// kinds (map, double, pairs) in whichever shape the negotiated RESP version
```
fatal10110: **[low · stale comment]** This header still lists the protocol-dependent kinds as "(map, double, pairs)". The fix added `boolean` and `big-number` to that set, so it's now five kinds (six counting `map-pairs` separately). The same staleness appears in the `respVersion` getter doc (around line 170, "what the `flat-pairs` shape keys off") and in `src/in-memory-client.ts` around line 234. A count-free wording like the one now in `DecodeRedisValueOptions` would stop this from drifting again.

src/core/decode-redis-value.ts
```diff
@@ -109,12 +128,30 @@ export function decodeRedisValue(
...
+      return options.returnBuffers ? Buffer.from(text) : text
+    }
     case 'boolean':
-      return value.value
-    case 'big-number':
-      return value.value
+      // RESP2 has no boolean: the encoder writes the `:1` / `:0` integer, so
+      // that is the number a client reads back. Only RESP3's `#t` / `#f` is a
+      // JS boolean. (Reachable through Lua's `redis.setresp(3)`.)
+      return options.version === 3 ? value.value : value.value ? 1 : 0
```
fatal10110: **[low · style]** A nested ternary (`options.version === 3 ? value.value : value.value ? 1 : 0`) sits right above the `big-number` arm, which uses the early-return `if` shape that AGENTS.md asks for. Suggest matching it:

```ts
case 'boolean':
  if (options.version === 3) {
    return value.value
  }
  return value.value ? 1 : 0
```

The behaviour is right. I verified it against real Redis: `redis.setresp(3); return true/false` gives `:1`/`:0` at RESP2 and `#t`/`#f` at RESP3, and the mock matches on both EVAL and FCALL.

tests/core/decode-redis-value.test.ts
```diff
@@ -180,19 +181,155 @@ describe('decode option divergences between the two clients', () => {
...
+        decodeRedisValue(
+          { kind: 'double', value: 2.5 },
+          { ...client, version: 3 },
+        ),
+        2.5,
+      )
+    }
+  })
+
+  test('version: big-number is the digit string on RESP2, a bigint on RESP3', () => {
```
fatal10110, line 252: **[low · test coverage]** The new boolean and big-number behaviour has only option-level unit tests (`decodeRedisValue` called directly). The CHANGELOG calls this a user-visible break reached through Lua, but no client-level test drives it end to end. A short case in `tests/resp2-map-and-double-shapes.test.ts` would pin the real path through `luaReplyToRedisValue` → session → decoder, and would catch a regression in the Lua conversion that these unit tests can't see. For example, on both socketless clients at RESP2 and after `HELLO 3`:
- `EVAL "redis.setresp(3); return true" 0` → `1` / `true`
- `EVAL "redis.setresp(3); return {big_number='12345678901234567890'}" 0` → `'12345678901234567890'` / `12345678901234567890n`

Expected values from real redis-server 7.2.1 + node-redis@6 `sendCommand`: boolean `1`→`true`, `0`→`false`; big number `"12345678901234567890"`→`12345678901234567890n`. The current mock already matches these.

fatal10110:
## Re-review of `4d31d76` (previous review: `0aed867`)

**Verdict: not yet.** There's one blocker, and my own round-1 review caused it (new finding N1). There's also one PR-description fix to make before merge. Everything else from round 1 is fixed. Once N1 is reverted and the PR body is corrected, this is shippable.

### Round-1 findings

| # | Round-1 finding | Status | Evidence |
|---|---|---|---|
| 1 | high: `hGetAll()` swallowed WRONGTYPE as `{}` | ✅ **fixed** | Non-map replies now fall through to `decodeRedisValue`. The PR mock matches real node-redis@6 at RESP2 and RESP3: `hGetAll(stringKey)` throws `WRONGTYPE …`, `hGetAll(missing)` → `{}`, `hGetAll(h)` → object. The new test is red on `0aed867` and green on `4d31d76`. |
| 2 | medium: `big-number` / `boolean` not version-dependent | ✅ **fixed** | Matches real Redis. `redis.setresp(3); return true/false` → `:1`/`:0` at RESP2, `#t`/`#f` at RESP3. Big number → digit string at RESP2, bigint at RESP3. The mock matched real on 4 EVAL and 2 FCALL cases. **Correction to the PR's framing:** a real boolean producer *does* exist (any script calling `redis.setresp(3)`, via EVAL or FCALL). Plain `return true` is `:1` even on RESP3, and plain `return false` is null. No native Redis 7.2 command replies with a RESP3 boolean. |
| 3 | medium: `decodeScalarReply` doc pointed at the wrong recipe | ✅ **fixed** | The new text is accurate: `zScore()` is a number and `hGetAll()` an object at both protocols, and it points at `decodeMapReply` as the pattern. |
| 4 | low: `decodeRedisKey` spelled double keys via `String()` | ⚠️ **fixed as asked, but my finding was wrong** | Real node-redis at RESP3 keys by the parsed number: `{"Infinity":1}`, not `{"inf":1}`. `origin/main` was already correct. See **N1**: please revert. |
| 5 | low: TESTING.md row conflated WITHSCORES / WITHVALUES | ✅ **fixed** | Every row matches real node-redis `sendCommand`. |
| 6 | low: "three kinds … all three" vs four bullets | ✅ **fixed** | Wording is now count-free. |
| 7 | low: no test separated the correct `{}` from the swallow path | ✅ **fixed** | Wrong-type case at RESP2 and RESP3. Red before, green after. |
| 8 | low: weak HELLO assertion | ✅ **fixed** | Pins `proto === 3` on the object, and `proto` at an even index followed by `2` on the flat array. |
| 9 | low: no CHANGELOG entry | ✅ **fixed** | The "no `/core` *Added* line" call is correct. tsup builds only `index.ts` and `internal.ts`, `package.json` exports only `"."` and `"./core"`, and neither re-exports `decode-redis-value` or `formatRedisDouble`. The before/after values in the table match real Redis and `origin/main`. Wording nits are in **N4**. |

### New findings (inline)

|---|---|---|---|
| N1 | **medium, blocker** | `src/core/decode-redis-value.ts:221` | Regression from finding 4's fix: RESP3 double map keys become `'inf'`/`'-inf'`/`'nan'`, where real node-redis gives `'Infinity'`/`'-Infinity'`/`'NaN'`. Revert to `String()`. The test at `tests/core/decode-redis-value.test.ts:347-360` locks in the wrong value. |
| N2 | low | `src/core/resp-encoder.ts:199` | `formatRedisDouble` prints every digit of whole-number doubles between ~3e18–5e18 and 1e21; real Redis uses exponent form (`1e20` → `"1e+20"`). The bug predates this PR, but RESP2 decoding now surfaces it. Fine as a follow-up. |
| N3 | low | `docs/TESTING.md:274` | The `redis.setresp(3)` qualifier implies it's needed for big numbers. It's needed only for booleans. The mock returning `[]` without it is an engine gap. |
| N4 | low | `CHANGELOG.md:123` | Scope names only `sendCommand()`, but `eval()` and `multi().exec()` change too. "Always the RESP3 shape" contradicts the WITHSCORES row. It cites `DecodeRedisValueOptions`, which isn't exported. The boolean row needs "under `redis.setresp(3)`". |
| N5 | low | `src/client-mocks/node-redis-mock.ts:1157` | The header still lists "(map, double, pairs)". The same is stale at `:170` and `in-memory-client.ts:234`. |
| N6 | low | `src/core/decode-redis-value.ts:145` | The nested ternary breaks the early-return style used by the `big-number` arm right below it. |
| N7 | low | `tests/core/decode-redis-value.test.ts:252` | Boolean and big-number are tested only at option level; add a client-level `EVAL "redis.setresp(3); …"` case. |

### PR description (no inline anchor, so reported here). Please fix before merge.

- **Real-backend coverage.** The test plan still has a ticked *"Targeted real-Redis run (`TEST_BACKEND=real`, node-redis hash + zset + config suites) — 163 pass, 0 fail"*. By the implementer's own account, the real-backend run for this round ended with `# cancelled 9` because the cluster fixture refused connections, and only the standalone real suites passed (6/6). The body should say what was actually run on this head. The real-Redis evidence for this PR comes from direct node-redis probes, which is fine, but the body should say that rather than imply suite coverage.
- Smaller stale lines: *"All four protocol-dependent kinds"* (there are now six arms: `map`, `map-pairs`, `flat-pairs`, `double`, `big-number`, `boolean`); *"Red before the fix (7 of 11 failing)"* (the file has grown since); *"The wire encoder … untouched"* (`resp-encoder.ts` is in the diff; `formatNumber` was renamed and exported, with no behaviour change).
- `npm test — 397 pass` is **accurate** for this head. I re-ran it.

### Verified on this head
- `tsc --noEmit`, `eslint`, `prettier --check`: clean.
- `npm test`: 397/397. `npm run test:integration:mock`: 1181/1181.
- The new tests fail on `0aed867` (4 red: big-number, boolean, double key, wrong-type `hGetAll`) and pass on `4d31d76`.
- Real-Redis evidence comes from node-redis@6 `sendCommand` and `redis-cli` against private redis-server **7.2.1** instances (the only version installed). **Not verified on Redis 8.x.**

### Pre-existing and out of scope (suggest follow-up issues, not blocking this PR)
- The mock's Lua engine returns `[]` for `{double=…}` / `{big_number=…}` / `{map=…}` tables when the script *hasn't* called `redis.setresp(3)`. Real Redis converts them regardless.
- `redis.setresp(3); return redis.call('HGETALL', 'h')` on a RESP3 connection: real → object, mock → flat array. Same with `ZSCORE`: real → number `2.5`, mock → string `"2.5"`. That's the RESP3→Lua→RESP3 conversion path in `lua-runtime`.
- Error class name: real node-redis throws `SimpleError`, the facade `ErrorReply`. This affects every error, not just this PR.

src/core/resp-encoder.ts
```diff
@@ -195,7 +195,35 @@ function formatError(value: Extract<RedisValue, { kind: 'error' }>): string {
...
+/**
+ * Largest magnitude Redis prints through its integer path: `double2ll()` in
+ * `util.c` rejects anything past `LLONG_MAX / 2` as a double, which is 2^62.
+ */
+const REDIS_DOUBLE_INTEGER_LIMIT = 2 ** 62
+
+/**
+ * Render a `double` the way Redis 7.2+ spells it (`d2string()` in `util.c`),
+ * which is the text of both a RESP3 `,` double and the RESP2 bulk string a
+ * client reads back — `decodeRedisValue` shares this so the two cannot drift.
```
fatal10110: **[medium · Redis-compat, incomplete change]** "shares this so the two cannot drift" misses a third copy. `src/core/lua-runtime.ts:368` still has a private `formatNumber`, a verbatim copy of the *old* formatter. `redisValueToLuaReply` (`:307-308`) uses it to hand a `double` reply from `redis.call` to a script. So after this PR the mock spells the **same score two ways**, depending on whether it's read directly or through Lua. Real Redis spells it one way.

Real Redis 8.0.6 (private container), `ZADD z 1e20 big 0.0000123 small 5e18 five`:
```
real 8.0.6 big    direct=1e+20      via-Lua=1e+20
real 8.0.6 small  direct=1.23e-5    via-Lua=1.23e-5
real 8.0.6 five   direct=5e+18      via-Lua=5e+18
```
This PR at `423adb8`, `createInMemoryClient()`, where via-Lua is `EVAL "return redis.call('ZSCORE', KEYS[1], ARGV[1])" 1 z <m>`:
```
mock big    direct=1e+20      via-Lua=100000000000000000000
mock small  direct=1.23e-5    via-Lua=0.0000123
mock five   direct=5e+18      via-Lua=5000000000000000000
```
Before this PR both paths used `toString()`: consistently wrong, but consistent. This change introduces the internal disagreement. The EVAL result goes out over TCP as a bulk string, so it's a wire difference too, and it contradicts the CHANGELOG `Fixed` entry's "a double on the wire … from the server and every client mock".

Fix: delete `formatNumber` from `lua-runtime.ts` and import `formatRedisDouble`. Pin it with a test comparing direct against `redis.call` for `1e20`.

src/core/resp-encoder.ts
```diff
@@ -195,7 +195,35 @@ function formatError(value: Extract<RedisValue, { kind: 'error' }>): string {
...
+ *    Plain digits while the trailing zeros number fewer than 8 (`5e18` is
+ *    `5e+18`, `1.2345678901234568e22` stays plain); a plain decimal while the
+ *    last digit sits fewer than 7 places after the point, or the magnitude is
+ *    below 10^4 (`0.000123`, but `1.23e-5`); exponent form otherwise.
+ *
+ * fpconv is Grisu2-based, which can in rare cases emit a longer digit string
+ * than the shortest one JS picks; those values are not matched here. Nor is
+ * the `redis-6.2` profile: Redis 6.2 prints `%.17g` (`0.1` is
+ * `0.10000000000000001` there), and the encoder is not profile-aware.
+ * (`redis-7.0` has not been checked.)
```
fatal10110: **[low · Redis-compat, disclosure]** I've now checked `redis-7.0`: it prints `%.17g` like 6.2. On both of those profiles the new formatter is slightly *worse* than the old one, not merely "still not matched". The comment, the PR body ("That was already true of the old formatter") and the CHANGELOG should say so.

My fuzz: 35,904 edge-weighted values, `ZADD` then `ZSCORE` over a raw socket at RESP2 and RESP3, on private containers:

| server | old formatter mismatches | new formatter mismatches | old matched, new doesn't |
|---|---|---|---|
| redis 8.0.6 | 4,442 | **112** | **0** |
| valkey 8.0.0 | 4,442 | 112 | 0 |
| valkey 9.0.0 | 4,442 | 112 | 0 |
| redis 7.0.15 | 23,766 | 23,929 | **456** |
| redis 6.2.14 | 23,765 | 23,928 | **456** |

Every one of the 456 is a 17-significant-digit value that `%.17g` and JS both print plain, while fpconv's layout moves it to exponent form:
```
value                   7.0.15 / 6.2.14 (real)   old (JS)                 new
489116.60955712036      489116.60955712036       489116.60955712036       4.8911660955712036e+5
0.00019999999999999998  0.00019999999999999998   0.00019999999999999998   1.9999999999999998e-4
```
(8.0.6 really does print `4.8911660955712036e+5`, so the new rule is right for 7.2+.)

On 7.2+ and both Valkey profiles, the 112 remaining mismatches are all digit choices that round-trip. Zero are layout errors, and RESP2 and RESP3 text is identical on every value. That confirms your 75/30,010 result independently.

On 6.2 and 7.0 both formatters are wrong for about two thirds of values (`0.1` → real `0.10000000000000001`), so neither has ever matched those profiles, and a profile-aware encoder is fine as a follow-up. But the disclosure should say "6.2 and 7.0 print `%.17g`; this formatter matches neither, and on those profiles it is marginally worse than before" rather than implying nothing changed there. Please also drop "(`redis-7.0` has not been checked.)".

CHANGELOG.md
```diff
@@ -126,6 +155,25 @@ so the PR body is not a durable home for a breaking-change note.
 
 ### Fixed
 
+- A double on the wire — a sorted-set score in `ZSCORE`, `ZINCRBY`,
```
fatal10110: **[scope] I recommend splitting the double-spelling change into its own PR.** The formatter itself checks out; this is about the change riding in #414's PR.

- **Different blast radius.** #414 is a decode-shape bug in the two *socketless clients*. This entry changes the bytes the **networked server** puts on the wire for every double at both protocols, which also reaches `createIoredisMock`, the RESP3 `,` frame and every TCP client. Neither the PR title nor #414 describes that. It's attributed to `[#414]` here, which isn't the issue it fixes.
- **Still incomplete** (see my comment on `resp-encoder.ts:207`: the Lua path keeps the old formatter), and it has a profile story (6.2 / 7.0) that deserves its own entry. It'll get a better review, a clean bisect point and a clean revert on its own. Sibling #430 is also working in `resp-encoder.ts`.
- **The split is clean.** Without it, the decoder's RESP2 double arm just reads back the same `toString()` text the encoder writes. That's self-consistent, and it's exactly the pre-existing state I flagged in round 2 as "fine as a follow-up". Move the `formatRedisDouble` body change, the `REDIS_DOUBLE_INTEGER_LIMIT`/`formatFpconv` helpers, the 36 encoder tests and this `Fixed` entry to a new PR (plus the Lua one-liner). Keep the rename/export here.

For the follow-up PR, my verification: 35,904-value fuzz, 0 layout errors against Redis 8.0.6, Valkey 8.0.0 and Valkey 9.0.0; ±2^62, 2^62 + 1ulp, ±2^63 and 2^64 all exact (`4611686018427387904`, `4611686018427389000`, `9223372036854776000`, `18446744073709552000`). That work is good and I'd approve it there with the Lua fix. If the maintainer prefers to keep it here, the minimum to ship is the Lua fix plus corrected 6.2/7.0 disclosure.

fatal10110:
## Re-review of `423adb8` (previous review: `4d31d76`)

**Verdict: not yet. Split the double-spelling change out; the decode half is ready.** All round-2 findings are fixed, including the blocker. The one thing holding this back is item 3's formatter rewrite. It's correct against Redis 7.2+ and Valkey, but it's a wire-wide server change riding in a client-decode PR, and it's incomplete (the Lua path still uses the old formatter). Move it to its own PR, or at minimum fix the Lua path and correct the 6.2/7.0 disclosure here.

### Round-2 findings

| # | Round-2 finding | Status | Evidence |
|---|---|---|---|
| N1 | **blocker:** RESP3 double map keys `'inf'` instead of `'Infinity'` | ✅ **fixed** | `decodeRedisKey` uses `String()` again. The replacement test pins real values at both protocols: RESP2 `['inf',1]`, RESP3 `{Infinity:1}` / `{'-Infinity':1}` / `{NaN:1}`, matching my 7.2.1 probe from last round. The false "never disagree" comment is gone. |
| N2 | `formatRedisDouble` wrong for large whole numbers | ✅ **fixed, but grew scope** | See below and the three new inline comments. |
| N3 | TESTING.md `setresp(3)` qualifier | ✅ **fixed** | Redis's rule (booleans only) is now separate from the mock's Lua gap. |
| N4 | CHANGELOG scope and wording | ✅ **fixed** | Names `eval()` and `multi().addCommand(…).exec()`. "Always RESP3" is gone, the WITHSCORES row is noted as already flat, there's no `DecodeRedisValueOptions` reference, and the boolean row is qualified. |
| N5 | stale comments | ✅ **fixed** | Header, `respVersion` getter and `in-memory-client.ts` now list all six kinds. |
| N6 | nested ternary | ✅ **fixed** | Early-return `if`. |
| N7 | no client-level Lua tests | ✅ **fixed** | `setresp(3)` `true` / `false` / big number at RESP2 and RESP3 on both clients, via `command()`, `sendCommand()`, `eval()` and MULTI replay. These pass on `4d31d76` as well, which is expected: they pin round-2 behaviour. |
| PR body | stale real-run claim and numbers | ✅ **fixed** | Real-run box unticked and reworded. "Six kinds" ✓. "18 tests" ✓ (counted). "36 values" ✓ (counted). "440" ✓ (re-ran: 440/440). "7 of the first 11" is historical, consistent with round 1. The encoder section replaces "untouched". The one inaccuracy left is the 6.2 sentence (inline). |
| (round-2 PR-body note) | `decodeMapReply` cast lie on non-map replies | ✅ **fixed** | Now throws `expected a map reply, got <kind>`; `+QUEUED` is tested. **It can't fire on a legitimate `hGetAll`:** `HGETALL` always produces a `map` value (`RedisValue.map([])` for a missing key, `hashes.ts:1052`), and the RESP2 flat array exists only at decode time. Wrong-type still throws WRONGTYPE. |

### Item 3: the formatter rewrite, checked independently

- **Fuzz** on private containers, not the shared fixtures: 35,904 edge-weighted values (±2^52/2^53/2^62/2^63/2^64 neighbourhoods, every power of ten from 1e-324 to 1e308 with assorted mantissas, random bit patterns, big integers, small fractions), `ZADD` then `ZSCORE` over a raw socket at RESP2 and RESP3:

  | server | old mismatches | new mismatches | layout errors | old ok → new wrong |
  |---|---|---|---|---|
  | Redis 8.0.6 | 4,442 | 112 | **0** | **0** |
  | Valkey 8.0.0 | 4,442 | 112 | 0 | 0 |
  | Valkey 9.0.0 | 4,442 | 112 | 0 | 0 |
  | Redis 7.0.15 | 23,766 | 23,929 | n/a (`%.17g`) | **456** |
  | Redis 6.2.14 | 23,765 | 23,928 | n/a (`%.17g`) | **456** |

  On 7.2+ and Valkey, every one of the 112 remaining mismatches round-trips and is a Grisu2 digit choice (e.g. `5e+22` → real `49999999999999996000000`). RESP2 and RESP3 carry identical text on every value. This independently confirms the implementer's 75/30,010. Redis 7.4 was not probed.
- **Boundaries** (8.0.6, exact): `±2^62` → `±4611686018427387904`; `2^62 + 1ulp` → `4611686018427389000`; `±2^63` → `±9223372036854776000`; `2^64` → `18446744073709552000`. `inf` / `-inf` / `nan` / `-0` are unchanged.
- **Callers:** `encodeResp2` (bulk), `encodeResp3` (`,`) and the decoder's `double` arm. **Missed:** `lua-runtime.ts:368` keeps a private copy of the old formatter for `redis.call` replies (inline, medium).
- **No existing test was quietly changed.** The PR touches no `tests-integration/` file, and no raw-TCP or integration test pinned a spelling in the changed ranges. The changed test files are all this PR's own.
- **Profiles:** 5 of 7 (`redis-7.2`, `redis-7.4` presumed, `redis-8.0`, `valkey-8.0`, `valkey-9.0`) strictly improve. On `redis-6.2` **and `redis-7.0`** (checked: 7.0.15 also prints `%.17g`), both formatters are about two-thirds wrong, and the new one is marginally worse (+163 net, 456 values that used to match). That's acceptable as a follow-up for a profile-aware encoder, but the disclosure should say so (inline).
- **Split or keep?** Split. My reasoning is inline on the CHANGELOG `Fixed` entry: it changes server wire bytes, not client decoding; it's attributed to #414, which it doesn't fix; it's incomplete; and it has its own profile story. The split is clean, because without it the RESP2 decode reads back the same `toString()` text the encoder writes.

### New findings (inline)

| Sev | Location | Finding |
|---|---|---|
| medium | `src/core/resp-encoder.ts:207` | `lua-runtime.ts` keeps the old formatter: the mock now returns `1e+20` directly but `100000000000000000000` via `redis.call`, where real 8.0.6 returns `1e+20` for both. This drift didn't exist before the PR. |
| low | `src/core/resp-encoder.ts:224` | 6.2 / 7.0 disclosure understates: 7.0 is `%.17g` too, and the new formatter regresses 456 values there. |
| scope | `CHANGELOG.md:158` | Recommend moving the double-spelling change to its own PR. |

### Verified on `423adb8`
`tsc --noEmit`, `npm run lint`, `prettier --check`, `npm run build`: clean. `npm test`: 440/440. `npm run test:integration:mock`: 1181/1181 (the #411 flake didn't trigger). Red-before-green: the new encoder, key and loud-failure tests fail on `4d31d76` (16 cases) and pass here. No `TEST_BACKEND=real` suites were run, and I didn't touch the shared fixtures.

fatal10110:
## Re-review of `030a1a7` (previous review: `423adb8`)

**Verdict: ship**, after an update-branch (GitHub reports `BEHIND`; see the merge check below). The split is clean, nothing from rounds 1-3 regressed, and I found **no new issues**, so there are no new inline comments this round.

### The split

| Check | Result |
|---|---|
| `resp-encoder.ts` vs the base | Only the `formatNumber` → `formatRedisDouble` rename at its three uses, `export`, and a doc comment. The function body is **identical** to current `origin/main`'s `formatNumber` (diffed line by line). The doc comment says only `inf`/`-inf`/`nan`/`-0` match Redis, gives the `1e+20` / `1.23e-5` / `%.17g` examples (all verified in round 3), and points at #451. |
| d2string tests | Gone. `tests/core-resp-encoder.test.ts` is identical to both the base and current `origin/main`. |
| CHANGELOG | The `Fixed` double-spelling entry is removed, and that is the only CHANGELOG change since `423adb8`. The BREAKING `Changed` entry is intact. |
| Leftovers | No remaining `d2string`/`fpconv`/`2^62`/`Grisu` references in the PR diff apart from the intentional #451 pointer. |
| Parked work | `feat/redis-double-format-d2string` @ `a86aaea` is `main` plus exactly `resp-encoder.ts` and the 36 tests. The #451 hand-off comment includes the Lua `redis.call` copy inconsistency, so that finding travels with the work. |

**My byte-identity check** (my own harness, not the implementer's): 48,001 doubles, made up of 30,000 random 64-bit patterns, 5,000 NaN/Inf payload patterns with random sign and mantissa, every power of ten from 1e-324 to 1e308 with six mantissas each, ±2^53/2^62/2^63 ±4096, `MIN_VALUE`/`MAX_VALUE`/`EPSILON`/`±0` and 5,000 small fractions. Each was encoded by `origin/main`'s encoder and by this branch's at RESP2 and RESP3:
```
{"values":48001,"nanCount":5020,"resp2ByteDiffs":0,"resp3ByteDiffs":0,"resp2DecodeNotWireText":0,"formatterDiffs":0}
```
As a control, the same harness against the parked `a86aaea` encoder reports `3919` byte differences, so it does detect a formatter change.

### Rounds 1-3: still in place
`src/core/decode-redis-value.ts`, `src/client-mocks/node-redis-mock.ts`, `src/in-memory-client.ts`, `docs/TESTING.md` and all four decode test files are **byte-identical to `423adb8`**, the head I verified in round 3. That covers:
- `Infinity`/`NaN` double map keys via `String()` (real node-redis shape at RESP2 and RESP3)
- boolean and big-number version handling
- the `redis.setresp(3)` client-level tests
- `decodeMapReply` throwing WRONGTYPE and failing loudly on non-map replies (can't fire on a legitimate `HGETALL`)
- the stale-comment fixes

The CHANGELOG BREAKING entry still names `eval()` and `multi().addCommand(…).exec()`. The PR body's real-backend box is still unticked, with the CI caveat.

### Verified on `030a1a7`
- `tsc --noEmit`, `npm run lint`, `prettier --check`, `npm run build`: clean.
- `npm test`: **403/403** (440 minus the 37 removed formatter tests). `npm run test:integration:mock`: **1181/1181**.
- CI: 15/15 pass (unit, lint, build, raw/ioredis/node-redis × mock/real, compatibility-profiles, both packages, CodeQL). `deploy` was skipped.

### Siblings (#430, #431, both still open)
Trial merge onto current `origin/main` (`b1995b8`, with #428 and #429): `#427 → #430 → #431`. #427 merges cleanly. #430 and #431 conflict only in `CHANGELOG.md`, which is textual. In the combined tree, `resp-encoder.ts` has #430's byte-based `encodeError` alongside `formatRedisDouble`, with no stray `formatNumber` left. `tsc` is clean, `npm test` passes 403/403, and `test:integration:mock` passes 1,240 of 1,242 with 0 failures (the other 2 are #430's own intentional `# SKIP redis-6.2 only` cases). **No semantic conflict.**

### Optional nit (PR body, no anchor)
The byte-identity bullet says "every bit pattern". It's 20,000 *random* bit patterns, since "every" would be 2^64. Worth rewording, but it doesn't block anything.

PR #428, opened 2026-09-22
fix(ci): gate redis-cluster on real cluster formation and capture logs (#398)

### LINKED ISSUE #411: Hash field-TTL tests flake under load: 20ms TTLs lose a race with their own round-trips

fatal10110:
## Summary

Two hash field-TTL tests (and their node-redis twins) set a **20 ms** field TTL and then
made more client round-trips than 20 ms survives on a loaded machine, so the field expired
mid-test and the assertions failed — the long-standing `test:integration:mock` flake in #411.

The tests are about TTL semantics, not about racing the clock. Nothing here loosens an
assertion: no tolerance for `-2`, no tolerance for a recreated counter, no widened equality.
The fix is in three parts.

### 1. The field must survive the round-trips before the wait

`HPEXPIRE ... 20` → `500` in all four files. 500 ms is longer than any round-trip the test
makes, which is the direct fix for the failure #411 reports.

- `tests-integration/ioredis/hash/field-persist.test.ts` (+ `HPTTL <= 20` → `<= 500`)
- `tests-integration/node-redis/hash/field-persist.test.ts`
- `tests-integration/ioredis/hash/field-expire.test.ts`
- `tests-integration/node-redis/hash/field-expire.test.ts`

### 2. The wait for expiry no longer needs a margin either

Raising the TTL moved the problem to the *other* side of the same wait. With `HPEXPIRE 500`
followed by `delay(600)`, the sleep only has to overshoot by 100 ms to be correct — and it
doesn't always: running the suite under 4x CPU contention produced a run where **both**
`field-persist` ("`'value3' !== null`") and `field-expire` ("`'2' !== null`") failed in the
opposite direction, with the field still alive after the sleep. Both processes failed in the
same run, which points at the machine (a scheduling or wall-clock excursion), not at either
test.

So the fixed sleeps are gone. A new `waitUntilGone(read, description)` helper in
`tests-integration/utils.ts` polls until the field is actually gone, which removes the guess
from both directions: it waits exactly as long as the expiry takes, and still fails — after
5 s, reporting how long the value survived — if the field never expires. A test that
regresses now says so instead of looking like one more flake. Applied to all five expiry
waits in the four files (including the two 5 ms-TTL waits, same pattern, smaller margin).

### 3. The audit criterion was margin, not magnitude

The first pass swept for sub-100 ms TTLs, which is the wrong filter: what matters is how far
an assertion sits from the boundary that makes it flip, not how large the TTL is. Re-audited
on margin, which caught four sites per client in the `TTL rounds to nearest second (#59)`
test — TTLs over 1000 ms with **less** slack than the 500 ms adopted above, because
`ttlSeconds` is `Math.round((expiresAt - now) / 1000)`:

- `SET k v PX 1900` + `TTL === 2` flips at `PTTL < 1500` — 400 ms of slack, so a 401 ms stall
  fails it. Now `PX 1999`, at the top of the band: the full ~500 ms the band can give.
  Same for `PX 2900` → `PX 2999` and `PEXPIRE 1900` → `1999`.
- `PTTL > 1000` after `PX 1500` had 495 ms. The raw-milliseconds check now gets its own key
  at `PX 4999`, where a seconds-rounded PTTL would answer 4000 or 5000 — both outside the
  band — so the assertion is no weaker and has ~1 s of slack.

Files: `tests-integration/{ioredis,node-redis}/key/expire.test.ts`. The `x999` values carry a
comment explaining the band, because "tidying" them back to round numbers reintroduces the bug.

Everything else in `tests-integration/` was re-checked on the same criterion: the remaining
short TTLs (`SET ... PX 1`, `HPEXPIRE 0`, `PEXPIRE missing 10 XX`, the `PX 50`
keyspace-notification waits) assert that something *has* expired after a longer wait, and the
TTL-band assertions elsewhere (`move`/`copy` at 1000 s, `hsetex`/`hgetex` at 100 s,
`EXPIRETIME` against absolute timestamps) all have seconds of slack.

No `src/` changes.

Closes #411

## Test plan

- [x] **Baseline reproduced** — pre-fix files, full mock suite 3x under 4x parallel CPU
      contention: 1/3 runs failed exactly where #411 says, `node-redis/hash/field-expire.test.ts`
      "hash writes clear or preserve field TTLs like Redis", `HINCRBYFLOAT` → `'1'` instead of
      `'2.5'` because `float` had expired and was recreated from 0.
- [x] **Deterministic margin demo** — inserting a 450 ms stall before every TTL read in the
      `#59` rounding test fails on the old values (`expected: 2, actual: 1`) and passes on the
      new ones.
- [x] **Under load** — RESULTS_PLACEHOLDER
- [x] **The polling helper is not vacuous** — a value that never disappears fails it
      (`stuck field still present after 307ms`), a value that disappears late passes.
- [x] **Ground-truthed against real Redis 8.0.6** (`redis-cli`): `HPEXPIRE 500` → `HTTL` 1,
      `HPTTL` 486, gone after 600 ms; `HSET` 0 / `HINCRBY` 2 / `HINCRBYFLOAT` 2.5, then
      `replace` kept and `counter`/`float` gone. `PX 1999` → `TTL 2` (and still 2 after a
      450 ms stall, 1 after 550 ms), `PX 2999` → `TTL 3`, `PEXPIRE 1999` → `TTL 2`,
      `PX 4999` → `PTTL 4984`.
- [x] `tsc --noEmit` and `npm run lint` clean.

Note: the `TEST_BACKEND=real` cluster backend could not be exercised — the shared local test
cluster on :30000 is `cluster_state:fail` (two masters down) from unrelated load on this
machine, and restarting shared infrastructure would disrupt other agents. Real-Redis behavior
was verified directly against the standalone instance instead, as above. The change is
test-only and adds no command surface.

---

## Review round 2 — changes

**Blocking**

1. **Timeout budgets reconciled.** The retry budget was 968s under a 240s `--wait-timeout`, so attempts 2-5 *and* the FATAL diagnostics dump were unreachable in CI — a hang still ended with no cluster state and no node logs, which is exactly what this PR exists to prevent. The three budgets are now explicitly ordered, tightest first, and each file documents its role:

   | Layer | Budget | Role |
   | --- | --- | --- |
   | `redis-cluster-init.sh` | ~315s (30s node gate + 3 x (60s create + 30s settle) + backoff) | **authoritative** — meant to lose first, exits non-zero after dumping diagnostics |
   | `--wait-timeout` | 420s | outer backstop only |
   | healthcheck `start_period` | 330s | covers the whole script budget so a slow-but-healthy boot cannot exhaust retries |

   Attempts dropped 5 -> 3 and per-phase timeouts tightened to make the script fit inside a sane CI step while keeping every retry and the dump reachable.

2. **Topology gate hardened.** `grep -c 'master'` matched `master,fail`, `slave,fail` and `noaddr` as healthy, so — since a dead *replica* leaves `cluster_state:ok` — the ready marker could be written over a degraded cluster. Flags are now split and matched as whole tokens, excluding `fail`/`fail?`/`handshake`/`noaddr`. Worth noting a naive `/fail/` regex would have been wrong in the other direction: it also matches the perfectly healthy `nofailover`.

**Also fixed**

3. The `--cluster-node-timeout` 5000 -> 15000 comment claimed pure upside; it now states the cost (15s is also the pfail -> fail window) and scopes it honestly — `cluster_formed()` probes all six nodes directly, so a node that is simply dead fails the gate regardless; the widened window only covers nodes that still answer us but that the cluster considers failed.
4. Stale `nodes-*.conf` in the `/data` volume is cleared on boot and `--dir` is passed explicitly. The ready marker moved `/run` -> `/dev/shm` (a per-start tmpfs), making a stale marker structurally impossible rather than a ~5s race.
5. Dropped the `|| return 1` that read awk's exit status instead of redis-cli's.
6. The healthcheck arithmetic that ignored the probe timeout is gone, replaced by the `start_period` design above.

**Also:** the healthcheck now calls the init script's own `check` mode instead of an inlined one-liner, so the probe and the startup gate cannot drift apart.

## Review round 2 — verification

- [x] **Dead-replica detection, end to end.** SIGKILL one replica in a formed cluster (pid from `INFO server`, never `SHUTDOWN`): `cluster_state` stayed `ok` for the entire observation window, and the container went **unhealthy at t+15s**. That is the blind spot finding 2 described, now closed — and it closes faster than the pfail window because the gate probes every node directly.
- [x] **Flag matcher unit-tested** against 10 `CLUSTER NODES` fixtures, asserting on the shipped function: healthy 3+3, `slave,fail`, `master,fail`, `master,fail?`, `handshake`, `noaddr`, the `nofailover` false-positive trap, the transient 5-master/1-replica shape, an address carrying a hostname, and empty input. All pass.
- [x] **Stale-config cleanup.** Confirmed `/data` is a real volume that survives `docker restart` (sentinel file persisted), and that with the cleanup the restart forms on **attempt 1/3**. Negative control: running `--cluster create` against already-clustered nodes reproduces `[ERR] Node 127.0.0.1:30000 is not empty`, the failure a stale config would cause.
- [x] **12 further up/down cycles under CPU load** (8 busy-loop containers + `linux/amd64` QEMU): **0 failures**, 38-57s to healthy against a 330s start period.
- [x] CI green on all three `-real` jobs; `raw-real` 49s end-to-end including startup, so the stricter 12-call probe costs nothing measurable.

## Compose binary

Verified locally with the standalone **`docker-compose` v2.26.1**; the `docker compose` CLI plugin is not installed on this machine. No portability issue in the shipped artifacts: neither the init script nor the compose file invokes docker at all, and the workflow's `docker compose` is available on `ubuntu-latest` and has now passed there. Both spellings are compose v2 with identical features (`--wait`, `--wait-timeout` need only v2.17+); `docs/TEST-INTEGRATION.md` now says so explicitly.


.github/workflows/ci.yml
```diff
@@ -121,7 +121,9 @@ jobs:
 
       - name: Start Redis (cluster + standalone)
         if: matrix.needsRedis
-        run: docker compose -f docker-compose.test.yml up -d --wait
+        # --wait-timeout bounds the wait: a cluster that never forms fails this
+        # step (with the log dump below) instead of hanging on the job clock.
+        run: docker compose -f docker-compose.test.yml up -d --wait --wait-timeout 240
```
fatal10110: **[high]** The retry budget and the compose wait budget are ~4x apart, which makes `CREATE_ATTEMPTS` and the `dump_diagnostics()` FATAL branch unreachable in CI — so the hang path still produces no diagnostics, which is the thing #398 asked for.

Arithmetic from the values in this PR:

```
NODE_READY_TIMEOUT            60s
per attempt: CREATE_TIMEOUT 120s + CLUSTER_READY_TIMEOUT 60s (+ sleep 2)
script worst case to reach the FATAL dump:  60 + 5*180 + 4*2  = 968s
clock after attempt 1 alone:                60 + 180          = 240s
docker compose --wait-timeout:                                  240s
```

Attempt 1's worst case *exactly equals* the compose budget. Attempts 2-5 can never run in CI, and `redis-cluster-init.sh:170-174` — the only place that dumps `CLUSTER INFO`, `CLUSTER NODES` and the six node logs — is unreachable on the slow/hung path.

**Failure scenario (the #398 one).** `--cluster create` wedges in "Waiting for the cluster to join" on a noisy runner. At t=240s compose gives up with a wait-timeout. The new `failure()` step runs and dumps `docker compose logs redis-cluster`, which contains exactly:

```
[cluster-init] started 6 redis-server processes (logs in /var/log/redis-cluster)
[cluster-init] all 6 nodes are up and cluster-enabled
[cluster-init] forming cluster (attempt 1/5)
```

No cluster state, no node logs — the same blind sighting, with a nicer message. (The `wait_for_nodes` FATAL path at line 103-105 *is* reachable at 60s and does dump correctly; it's only the create path that is starved.)

Two independent fixes, worth doing both:
1. Call `dump_diagnostics` after **every** failed attempt, not only after the last one. Then even a truncated run is diagnosable.
2. Make the budgets coherent — either bound the script's total below 240s (e.g. `CREATE_TIMEOUT=45`, `CLUSTER_READY_TIMEOUT=30`, `CREATE_ATTEMPTS=3` → 60+3*75+4 = 289s, so also raise `--wait-timeout`), or raise `--wait-timeout` past the script's real worst case. As written the two numbers look coordinated but aren't.

docker/redis-cluster-init.sh
```diff
@@ -0,0 +1,193 @@
...
+    echo "$info" | grep -q 'cluster_state:ok' || return 1
+    echo "$info" | grep -q 'cluster_slots_assigned:16384' || return 1
+    [ "$(echo "$info" | grep 'cluster_known_nodes:' | tr -dc '0-9')" = '6' ] || return 1
+
+    # Roles settle by gossip a beat after `--cluster create` returns: for a
+    # short window a node can report cluster_state:ok while still listing a
+    # freshly-assigned replica as a slotless master. Wait that out too, or the
+    # first client in sees a topology that is about to change under it.
+    roles=$(redis-cli -p "$port" cluster nodes 2>/dev/null | awk '{print $3}') || return 1
+    [ "$(echo "$roles" | grep -c 'master')" = '3' ] || return 1
```
fatal10110: **[high]** The "settled 3-master/3-replica" gate counts **failed** nodes as healthy, so it can write the ready marker on a degraded cluster — the exact class of bug this gate exists to close.

`awk '{print $3}'` yields the CLUSTER NODES *flags* field, and `grep -c 'master'` / `grep -c 'slave'` match it as a substring. Flags are comma-joined and carry `fail`, `fail?`, `noaddr` and `handshake` alongside the role. Literal output from a real `redis:8.0.6` cluster (the repo's own compose stack, currently degraded):

```
$ redis-cli -p 30000 cluster nodes
aaa96c9e... 127.0.0.1:30005@40005 slave 634da04b... 0 1790075040000 1 connected
f4bdb4d0... 127.0.0.1:30001@40001 master - 0 1790075041000 2 connected 5461-10922
2f038f82... :0@0                  master,fail,noaddr - 1790072553929 ... disconnected 10923-16383
2e3933cb... 127.0.0.1:30004@40004 master,fail        - 1790072461255 ... disconnected
842bc36b... 127.0.0.1:30003@40003 slave f4bdb4d0... 0 1790075041855 2 connected
634da04b... 127.0.0.1:30000@40000 myself,master      - 0 0 1 connected 0-5460
```

**Failure scenario.** One *replica* gets marked `fail` during or just after formation. A dead replica does not affect slot coverage, so `cluster_state` stays `ok` on every node, `cluster_slots_assigned:16384` holds, `cluster_known_nodes:6` holds — and the role counts still read 3/3 because `slave,fail` contains `slave`:

```
$ roles="myself,master
master
master
slave
slave
slave,fail"
$ echo "$roles" | grep -c master   # 3
$ echo "$roles" | grep -c slave    # 3
   -> all four conditions pass; touch /run/redis-cluster-ready on a 3-master + 2-live-replica cluster
```

The compose healthcheck does not catch it either — it only greps `cluster_state:ok`, which is `ok` in this state.

To be fair to the PR: the 5-masters + 1-replica case it was written for **is** caught (`grep -c master` = 5 ≠ 3). Verified. It's specifically `fail`/`noaddr` that leak through.

Suggested fix — match the role as a whole flag word and reject unhealthy flags outright:

```sh
roles=$(redis-cli -p "$port" cluster nodes 2>/dev/null \
  | awk '$3 ~ /fail|noaddr|handshake/ { print "BAD"; next } { print $3 }')
echo "$roles" | grep -q 'BAD' && return 1
[ "$(echo "$roles" | grep -cE '(^|,)master$')" = '3' ] || return 1
[ "$(echo "$roles" | grep -cE '(^|,)slave$')"  = '3' ] || return 1
```

Worth mirroring in the healthcheck too, otherwise a replica dying mid-run never flips the container unhealthy — which the compose comment claims it does.

docker/redis-cluster-init.sh
```diff
@@ -0,0 +1,193 @@
...
+NODE_READY_TIMEOUT="${NODE_READY_TIMEOUT:-60}"       # seconds, all nodes bound
+CLUSTER_READY_TIMEOUT="${CLUSTER_READY_TIMEOUT:-60}" # seconds, per create attempt
+CREATE_ATTEMPTS="${CREATE_ATTEMPTS:-5}"
+# `redis-cli --cluster create` can sit in "Waiting for the cluster to join"
+# indefinitely; without a timeout that hang is unrecoverable, with one it just
+# becomes another retryable attempt.
+CREATE_TIMEOUT="${CREATE_TIMEOUT:-120}"
+# 5000ms flaps under CI load and is what makes the join check give up; 15s is
+# still far below any test timeout and nothing here exercises failover.
+CLUSTER_NODE_TIMEOUT="${CLUSTER_NODE_TIMEOUT:-15000}"
```
fatal10110, line 109: **[medium]** The `5000 -> 15000` bump is safe for test expectations, but it triples the window in which the new gate above it will call a cluster with a dead node "ok".

The safety half of the claim checks out. Nothing in the suite exercises failover — the only matches in `tests-integration/` are standalone `INFO` assertions, not cluster failover:

```
tests-integration/ioredis/info-standalone.test.ts:32:    assert.match(info, /^master_failover_state:no-failover$/m)
tests-integration/node-redis/info-standalone.test.ts:32: assert.match(info, /^master_failover_state:no-failover$/m)
```

No test kills a node, calls `CLUSTER FAILOVER`, or asserts on `cluster_state:fail`. So "nothing here exercises failover" is accurate and no expectation changes.

The part the comment doesn't say: `cluster-node-timeout` is precisely what gates how long it takes for an unreachable node to go `pfail -> fail` and for `cluster_state` to flip. Tripling it triples the interval during which a node that has actually died still reads as healthy — both to `cluster_formed()` here and to the compose healthcheck's `cluster_state:ok` probe. Combined with the flag-substring issue on line 128, that widens the window for writing the ready marker on a degraded cluster from ~5s to ~15s.

Not a blocker, and the tradeoff is probably still right (5s bus links genuinely do flap under CI load). But the comment reads as pure upside; it's worth noting the cost, e.g. "15s also means a genuinely dead node takes up to 15s to fail the gate."

Minor related note, flagged as reasoned rather than measured: Redis pings any peer whose `pong_received` is older than `cluster-node-timeout / 2`, so this also moves that staleness threshold 2.5s -> 7.5s. I did not reproduce a measurable slowdown in role convergence (the 100ms `clusterCron` random-peer ping dominates across only 6 nodes), so I don't think it meaningfully eats the 60s `CLUSTER_READY_TIMEOUT` — but it is a mild force in the opposite direction to the one intended.

docker/redis-cluster-init.sh
```diff
@@ -0,0 +1,193 @@
...
+      exit 1
+    fi
+
+    attempt=$((attempt + 1))
+    reset_nodes
+    sleep 2
+  done
+}
+
+rm -f "$READY_MARKER"
```
fatal10110, line 376: **[medium]** The marker cannot go stale, but only by a ~5s race rather than by construction — and the sibling state in `/data` genuinely does survive.

I checked this one specifically because it's the obvious way a readiness gate fails open. Two facts about the container:

```
$ docker image inspect redis:8.0 --format '{{json .Config.Volumes}} WORKDIR={{.Config.WorkingDir}}'
{"/data":{}} WORKDIR=/data

$ docker inspect --format '{{json .HostConfig.Tmpfs}} mounts={{range .Mounts}}{{.Destination}} {{end}}' <redis-cluster>
null mounts=/data
```

So `/run` is **not** a tmpfs — it's the container's writable layer, and `/run/redis-cluster-ready` is durable across a restart. The only thing that clears it is this `rm -f`.

That does hold, because `rm -f` runs within milliseconds of container start while Docker's first probe lands at `interval` (5s). Worth knowing that the margin matters: a *passing* probe during `start_period` marks the container healthy **immediately**, and on a restart the nodes reload their old `nodes-<port>.conf` and can reach `cluster_state:ok` fast — so if that `rm -f` ever moved below `start_nodes`, the healthcheck would sign off on the previous cluster. Adding

```yaml
    tmpfs:
      - /run
```

to the service makes it structural instead of ordering-dependent, and is a one-liner.

The part that *is* live today: `--cluster-config-file nodes-$port.conf` is relative, WORKDIR is `/data`, and `/data` is a declared volume — so the node configs survive a restart even though the marker doesn't. On a `docker compose up` that reuses a stopped container, the six nodes come back already knowing the old cluster, attempt 1 fails on "not empty", and only then does `reset_nodes` clean up and attempt 2 succeed. It self-heals, but by luck of the retry rather than by design, and it burns one of the attempts that (per my comment on `ci.yml:126`) you cannot actually afford. CI hides this because the teardown is `down -v`; a developer running `docker compose up` twice locally hits it every time.

Cheap fix: call `reset_nodes` once before the loop, or `rm -f "$LOG_DIR"/../nodes-*.conf` (i.e. `/data/nodes-*.conf`) in `start_nodes` before launching. Then attempt 1 starts from a known state in every scenario.

docker/redis-cluster-init.sh
```diff
@@ -0,0 +1,193 @@
...
+    info=$(redis-cli -p "$port" cluster info 2>/dev/null) || return 1
+    echo "$info" | grep -q 'cluster_state:ok' || return 1
+    echo "$info" | grep -q 'cluster_slots_assigned:16384' || return 1
+    [ "$(echo "$info" | grep 'cluster_known_nodes:' | tr -dc '0-9')" = '6' ] || return 1
+
+    # Roles settle by gossip a beat after `--cluster create` returns: for a
+    # short window a node can report cluster_state:ok while still listing a
+    # freshly-assigned replica as a slotless master. Wait that out too, or the
+    # first client in sees a topology that is about to change under it.
+    roles=$(redis-cli -p "$port" cluster nodes 2>/dev/null | awk '{print $3}') || return 1
```
fatal10110: **[low]** This `|| return 1` is decorative — it reads `awk`'s exit status, not `redis-cli`'s, so it never fires on the case it looks like it's guarding.

In a pipeline the exit status is the *last* command's, and `awk` happily exits 0 on empty input. So if node `$port` is unreachable, `redis-cli` exits 1, `awk` exits 0, `roles` is set to the empty string, and the `|| return 1` is skipped. The function does still return 1 (via the `grep -c` count check on the next line reading 0), so there's no live bug — but the guard is misleading, particularly sitting nine lines below line 118 where the identical idiom **is** load-bearing:

```sh
info=$(redis-cli -p "$port" cluster info 2>/dev/null) || return 1   # real guard (no pipe)
roles=$(redis-cli ... | awk '{print $3}') || return 1                # no-op (pipe)
```

Either drop the `|| return 1` here, or make it real by capturing `CLUSTER NODES` first and piping second:

```sh
nodes_out=$(redis-cli -p "$port" cluster nodes 2>/dev/null) || return 1
roles=$(echo "$nodes_out" | awk '{print $3}')
```

While I'm on this function — I went looking for the classic `set -e` early-exit in the retry loop and there isn't one. Worth recording since it's the non-obvious part of the review:

```
$ sh -c 'set -eu; r="a"; z=$(echo "$r" | grep -c zzz); echo ALIVE'
$ echo $?
1                       # grep -c exits 1 on zero matches and kills the script
$ sh -c 'set -eu; r="a"; [ "$(echo "$r" | grep -c zzz)" = "3" ] || echo took-branch; echo ALIVE'
took-branch
ALIVE                   # inside [ ], the substitution status does not propagate
```

The script only ever uses the second form (lines 121, 128, 129), and every other command-substitution assignment is `|| return 1`-guarded, so `set -eu` is safe throughout. `attempt=$((attempt + 1))` is also fine (plain assignment, always 0 — it's bash's `((i++))` that bites). `sh -n` is clean.

docker-compose.test.yml
```diff
@@ -11,44 +11,32 @@ services:
...
           'CMD-SHELL',
-          'redis-cli -p 30000 cluster info | grep -q cluster_state:ok',
+          'test -f /run/redis-cluster-ready && for p in 30000 30001 30002 30003 30004 30005; do redis-cli -p $$p cluster info 2>/dev/null | grep -q cluster_state:ok || exit 1; done',
         ]
       interval: 5s
-      timeout: 5s
-      retries: 30
-      start_period: 20s
+      timeout: 10s
+      # 30s grace + 40 x 5s = 230s, just inside the workflow's --wait-timeout,
```
fatal10110: **[low]** This arithmetic only holds if every probe returns instantly, so the two numbers aren't actually coordinated the way the comment says.

`30s + 40 x 5s = 230s` assumes the probe costs nothing. But `timeout: 10s` (raised from 5s in this PR) and the probe now shells out to **six** `redis-cli` invocations instead of one. Docker's interval is the gap *between* probes, so a probe that stalls costs `interval + timeout`:

```
instant probes:          30 + 40*5        = 230s   <- the comment's number
probes hitting timeout:  30 + 40*(5 + 10) = 630s
docker compose --wait-timeout                240s
```

So in the slow case — which is exactly the loaded-runner case this PR is about — the container reaches "unhealthy" long after compose has already given up. The real failure mechanism in CI is always the `--wait-timeout`, never the retry count; the `retries: 40` budget can only win the race when probes are fast, and when probes are fast the cluster is healthy anyway.

Not harmful (both paths fail the step, which is what matters), but the comment asserts a coordination that doesn't exist. Either drop the "just inside the workflow's --wait-timeout" clause, or make it true by sizing `retries` against `interval + timeout`.

Verified separately that the probe's exit status is what you'd want, since `test -f ... && for ...; done` is easy to get wrong:

```
$ sh -c 'touch /tmp/mk; test -f /tmp/mk && for p in 1 2 3; do echo cluster_state:ok | grep -q cluster_state:ok || exit 1; done'; echo $?
0
$ sh -c 'touch /tmp/mk; test -f /tmp/mk && for p in 1 2 3; do echo ok | grep -q cluster_state:ok || exit 1; done'; echo $?
1
$ sh -c 'rm -f /tmp/mk2; test -f /tmp/mk2 && for p in 1 2 3; do ... done'; echo $?
1
```

All three behave correctly.

fatal10110:
## Review summary — PR #428

Reviewed through both lenses (Redis cluster semantics + architecture/correctness). Six inline comments posted. Redis-behaviour claims below are ground-truthed against a real `redis:8.0.6` cluster and disposable local `redis-server` instances; literal output is quoted in the comments.

### Findings

| 1 | high | `ci.yml:126` | Script's retry budget (968s) vs `--wait-timeout` (240s): attempts 2-5 and the FATAL `dump_diagnostics()` are unreachable in CI, so the hang path still yields no diagnostics — the #398 symptom recurs |
| 2 | high | `redis-cluster-init.sh:128` | Topology gate counts `master,fail` / `slave,fail` / `noaddr` as healthy; a dead **replica** keeps `cluster_state:ok`, so all four conditions pass and the ready marker is written on a degraded cluster |
| 3 | medium | `redis-cluster-init.sh:43` | `--cluster-node-timeout 15000` is safe for test expectations (verified: nothing exercises failover) but triples the window in which finding #2 can fire |
| 4 | medium | `redis-cluster-init.sh:182` | `/run` is the writable layer, not tmpfs — marker safety is a ~5s race, not structural; `/data` node configs genuinely do survive a restart and burn attempt 1 |
| 5 | low | `redis-cluster-init.sh:127` | `\|\| return 1` reads `awk`'s status, not `redis-cli`'s — dead guard next to an identical live one |
| 6 | low | `docker-compose.test.yml:36` | `30s + 40x5s = 230s` ignores the 10s probe timeout; real worst case is 630s, so the two budgets aren't coordinated as claimed |

### Claims that were checked and hold

The root-cause analysis is **correct**, not hand-waved. `--daemonize yes` really does return success before the bind:

```
$ redis-server --port 6391 ... --daemonize yes   # port already in use
$ echo $?
0
$ tail -2 /tmp/n6391.log
# Warning: Could not create server TCP listening socket *:6391: bind: Address already in use
# Failed listening on port 6391 (tcp), aborting.
```

Also verified:
- `node_ready()` is right about `cluster_enabled`. `INFO cluster` → `cluster_enabled:1`; `CLUSTER INFO` does **not** contain the field. Confirmed it doesn't false-positive against `cluster_enabled:0` on a non-cluster server.
- **No `set -e` early-exit bug.** Every command-substitution assignment is either `|| return 1`-guarded or inside `[ ]`, where the substitution's status doesn't propagate. The killer form (`z=$(… | grep -c x)`) is demonstrated in comment #5 and is not used. `attempt=$((attempt + 1))` is safe. `sh -n` clean.
- The 5-masters + 1-replica case the PR was written for **is** caught by the new gate (`grep -c master` = 5 ≠ 3). The fix does address the measured bug.
- `cluster_state:ok` / `cluster_slots_assigned:16384` / `cluster_known_nodes:` are exact field names; the `tr -dc '0-9'` parse yields `6`, and no other `CLUSTER INFO` line matches that grep.
- `CLUSTER NODES` field 3 is the flags field and Redis 8.0.6 still emits `slave`, not `replica` — the gate's most load-bearing assumption holds.
- `--cluster-yes` is a real flag (`redis-cli --cluster help`), equivalent to the old `echo "yes" |`.
- `FLUSHALL` then `CLUSTER RESET HARD` is the right order, and `CLUSTER RESET`'s "containing keys" guard applies only to masters, so swallowing a `-READONLY` FLUSHALL on a replica doesn't break `reset_nodes`.
- `if: failure() && matrix.needsRedis` **does** fire for a failed *startup* step (`failure()` is job-scoped, not step-scoped) and skips cleanly on the four mock matrix entries. That part of the fix is sound.
- Healthcheck `CMD-SHELL` exit status is correct in all three cases (marker missing / a node down / all healthy).

### Verdict

**Not yet — close #1 and #2, the rest are optional.** The direction is right and the diagnosis is genuinely correct rather than a retry-shaped guess. But the two high findings undercut the PR's own two goals: #2 means the gate can still sign off on a degraded cluster (the thing it exists to prevent), and #1 means the CI hang path still produces no diagnostics (the thing #398 asked for). Both are small, local fixes.

Not reproduced: no end-to-end compose bring-up — `docker compose` is unavailable on this machine and the Docker VM is memory-constrained with a shared cluster in use. The 36-cycle soak in the test plan is taken at face value; findings here are from code trace plus real-Redis ground truth.

docker/redis-cluster-init.sh
```diff
@@ -0,0 +1,277 @@
...
+  for port in $PORTS; do
+    log "--- redis-server $port log (last 50 lines) ---"
+    tail -n 50 "$LOG_DIR/$port.log" 2>&1 || log "(no log file for $port)"
+  done
+  log '===== end diagnostics ====='
+}
+
+start_nodes() {
+  mkdir -p "$LOG_DIR" "$CONFIG_DIR"
+  rm -f "$CONFIG_DIR"/nodes-*.conf
```
fatal10110: **[low] new** This clears `nodes-*.conf` but not `dump.rdb`, and a persisted `dump.rdb` in the same volume brings back the "not empty" failure this change was meant to remove.

`--save ''` does not keep `/data/dump.rdb` empty. On a replica, full sync writes the received RDB to disk (`repl-diskless-load` defaults to `disabled`). All six nodes share `--dir /data` and the default `dbfilename`, so they all write, and load, **one** `dump.rdb`. On a fresh private stack of this head, the file already existed right after the first formation:

```
-rw------- 1 root root 171 18:33:21 /data/dump.rdb
```

When that file holds keys, e.g. after a replica full-resyncs mid-run (I forced it here with `SAVE` on 30000 after writing 200 keys), `docker restart` gives:

```
[cluster-init] forming cluster (attempt 1/3)
[ERR] Node 127.0.0.1:30000 is not empty. Either the node already knows other nodes (check with CLUSTER NODES) or contains some key in database 0.
[cluster-init] redis-cli --cluster create failed (or timed out after 60s) on attempt 1
[cluster-init] forming cluster (attempt 2/3)
[cluster-init] cluster_state:ok on all 6 nodes, 16384/16384 slots, 3 masters + 3 replicas
```

and in the node logs every node claims the dump's slots:

```
* I have keys for unassigned slot 4775. Taking responsibility for it.
```

`reset_nodes` recovers it on attempt 2, and the create fails fast rather than eating the 60s timeout. CI never hits this because its teardown is `down -v`. So this is low severity. It is still the same failure class, and it silently uses up one of the three attempts. Clearing the whole directory fixes it:

```sh
rm -f "$CONFIG_DIR"/nodes-*.conf "$CONFIG_DIR"/*.rdb
```

(Optionally also give each node `--dbfilename dump-$port.rdb` so six processes stop sharing one file.)

docker/redis-cluster-init.sh
```diff
@@ -0,0 +1,277 @@
...
+CREATE_ATTEMPTS="${CREATE_ATTEMPTS:-3}"
+# `redis-cli --cluster create` can sit in "Waiting for the cluster to join"
+# indefinitely; without a timeout that hang is unrecoverable, with one it just
+# becomes another retryable attempt.
+CREATE_TIMEOUT="${CREATE_TIMEOUT:-60}"
+# 5000ms is what made bus links flap under CI load, and a flapping link is
+# exactly what makes redis-cli's join consistency check give up. 15s buys that
+# stability at a real cost, not for free: cluster-node-timeout is also the
+# pfail -> fail promotion window, so a node the cluster has lost stays merely
+# "pfail" — invisible to the flag check in cluster_formed() — for up to 15s
```
fatal10110: **[low] new** The comment gets the mechanism backwards. A pfail node is **not** invisible to the flag check. The blind window is the period *before* pfail, when the node has no flag at all.

Pfail is shown as the `fail?` flag, and `healthy_role_counts` excludes `fail?` (checked against the image's mawk: `master,fail?` gives `2 3` and fails the gate). I ran a real timeline: 3-node cluster, `cluster-node-timeout 3000`, one node SIGSTOPped, its flags polled from a peer every 0.5s:

```
t=0.5s .. 4.0s  flags=master          cluster_state:ok    <- no flag at all: invisible to the gate
t=4.5s          flags=master,fail?    cluster_state:ok    <- pfail: CAUGHT by the flag check
t=5.0s          flags=master,fail     cluster_state:fail
```

So the conclusion holds: the blind window scales with `cluster-node-timeout`, roughly 5s → 15s. Two sentences need rewording:

- "stays merely 'pfail' — invisible to the flag check" → something like: *carries no flag at all until cluster-node-timeout expires, and until then it is invisible to the flag check.*
- "only covers nodes that still answer us but that the cluster considers failed" → the flag check does catch nodes the cluster considers failed. What it misses is nodes that still answer us but *that peers have not flagged yet*, e.g. a bus-link partition inside the first node-timeout.

The other claim here does hold: a node that has actually crashed fails the direct `redis-cli` probe at once. On a private stack of this head, SIGKILL of replica 30003 flipped the container to `unhealthy` at t+13s while `cluster_state` stayed `ok`.

docker/redis-cluster-init.sh
```diff
@@ -0,0 +1,277 @@
...
+      --daemonize yes
+  done
+  log "started 6 redis-server processes (logs in $LOG_DIR)"
+}
+
+# A node counts as ready only once it answers PING *and* reports that it came
+# up in cluster mode — PING alone can succeed before the cluster subsystem is
+# initialised, and `--daemonize yes` returns 0 even when the bind fails.
+node_ready() {
+  [ "$(redis-cli -p "$1" ping 2>/dev/null)" = 'PONG' ] || return 1
```
fatal10110: **[low] new** "Designed to lose first" only holds for nodes that crash, not for nodes that hang. Every `redis-cli` call in the startup path is unbounded, so a wedged node stalls the script before it reaches its own deadline check.

Neither `node_ready`, `cluster_formed` (lines 197/202) nor `dump_diagnostics` sets a timeout, and `redis-cli` has no default one. Against a SIGSTOPped `redis-server` (the TCP connect succeeds but no reply ever comes):

```
$ kill -STOP <pid>; redis-cli -p 7104 ping &   # still blocked after 6s, only returned after SIGCONT
redis-cli still blocked after 6s against a stopped node (no default timeout)
```

If one node wedges during boot (swap storm, stuck `fork`, a QEMU stall), `wait_for_nodes` / `wait_for_cluster` never get back to their `now() >= deadline` check. The 420s `--wait-timeout` backstop then fires with no diagnostics. If it wedges during a failed attempt, `dump_diagnostics` itself hangs on that node. That is the #398 failure mode again, in a narrower case.

This is an edge case and not a regression: the first revision had the same gap. But the budget comment now promises the script is authoritative. The image has coreutils `timeout`, so a one-line wrapper covers every call site:

```sh
rcli() { timeout 5 redis-cli "$@"; }
```

The healthcheck path is fine as it is, because Docker's `timeout: 15s` bounds `check` mode.

fatal10110:
## Re-review: PR #428 at `c910e82` (previously `8f6ea46`)

Checked against a private stack of this head (own compose project, no host ports, `docker-compose` v2.26.1), a local 3-node cluster for pfail timing, and the image's own `mawk`. CI is 14/14 green, including `raw-real`, `ioredis-real` and `node-redis-real`.

### Prior findings

| # | Sev | Finding | Status | How I checked |
|---|-----|---------|--------|---------------|
| 1 | high | Budget inversion (968s script vs 240s wait) | **Fixed** | 30 + 3×(60+30) + 2×2 = **304s** + dump, below `start_period` 330s, below `--wait-timeout` 420s. Every retry and the FATAL dump can now run. The probe timeout doesn't matter during boot, because failures inside `start_period` don't count. Steady state is 3×(5+15) = 60s, matching the comment. Hung-node caveat in new finding C. |
| 2 | high | Flag substring match let `fail` nodes through | **Fixed** | Tested whole-token awk under mawk with 8 fixtures: clean→`3 3`, `slave,fail`→`3 2`, `master,fail?`→`2 3`, `nofailover`→`3 3` (the trap is handled), `handshake`/`noaddr`→`1 3`, 5m/1s→`5 1`, empty→`0 0`. Live: SIGKILL of replica 30003 → **unhealthy at t+13s** while `cluster_state:ok`. |
| 3 | medium | node-timeout 15000 widens the window | **Addressed, wording wrong** | The tradeoff is now stated honestly. But the comment says pfail is invisible to the flag check, and it isn't: `fail?` is excluded. The blind window is the pre-pfail period when the node has no flag. See new finding B. |
| 4 | medium | Marker race and stale `/data` | **Mostly fixed** | After `docker restart`: `/dev/shm` sentinel gone, `/run` sentinel still there. So the move was necessary and makes staleness impossible by construction. Plain restart forms on **attempt 1/3**. The gap left is `dump.rdb` (new finding A). |
| 5 | low | Dead awk-status guard | **Fixed** | Line 202 captures `CLUSTER NODES` with no pipe, so `\|\| return 1` is a real guard now. |
| 6 | low | Healthcheck arithmetic didn't add up | **Fixed** | The probe runs the script's `check` mode, so gate and probe can't drift. `start_period` covers the full boot budget; `retries: 3` is sized for steady state. `check` mode does nothing destructive before its early exit, and it never removes the marker. |

| | Sev | Location | Finding |
| A | low | `redis-cluster-init.sh:116` | Boot clears `nodes-*.conf` but not the shared `/data/dump.rdb`, which replicas write during full sync. Reproduced `[ERR] Node 127.0.0.1:30000 is not empty` on attempt 1 after a restart; recovers on attempt 2. CI is unaffected (`down -v`). |
| B | low | `redis-cluster-init.sh:78` | node-timeout comment gets the mechanism backwards. Real timeline: no flag for about node-timeout, then `fail?` (caught), then `fail`. |
| C | low | `redis-cluster-init.sh:141` | Unbounded `redis-cli` calls: a wedged (SIGSTOPped) node blocks the script before it reaches its deadline check, so the 420s backstop fires with no dump. Suggest `timeout 5 redis-cli`. |

### Compose spelling
I accept this. Nothing in the script or compose file calls docker. The standalone `docker-compose` v2 binary and the `docker compose` plugin are the same implementation, and ubuntu-latest has the plugin. I did my own private bring-up with `docker-compose` 2.26.1: healthy in 15s on attempt 1/3.

### Verdict: **ship**
Both high findings and all four others from the first round are fixed, and I checked each fix at runtime rather than trusting the claims. The three new findings are low-severity edge cases (a local restart after a replica resync, comment wording, a wedged-not-crashed node). None of them can make the gate report ready on a degraded cluster, and none affects CI as configured. They're fine as a follow-up or a quick amend.

docs/TEST-INTEGRATION.md
```diff
@@ -166,7 +166,18 @@ The `docker-compose.test.yml` file defines three services, all on the official `
 - Ports: 30000-30005 (bus ports 40000-4000x stay container-internal).
 - All six `redis-server` processes run in **one** container so they reach each other over 127.0.0.1 and the host port map is 1:1.
 - Mac-compatible networking via `--cluster-announce-ip 127.0.0.1`, so MOVED/ASK redirects resolve from the host.
-- Healthcheck waits for `cluster_state:ok` (every slot assigned) before the cluster is marked healthy.
+- Startup is driven by [`docker/redis-cluster-init.sh`](../docker/redis-cluster-init.sh), mounted read-only into the container. It wipes each node's data directory (below), waits for **all six** nodes to answer `PING` and report `cluster_enabled:1`, then retries `redis-cli --cluster create` (up to 3 times, resetting the nodes between attempts) until the readiness gate below passes.
```
fatal10110, line 169: **[nit] new** Docs drift: the script no longer sends `PING`. `node_ready()` now sends `INFO cluster` alone, and its comment says so ("A reply at all proves the node is bound and serving (so no separate PING)"). Suggest: *waits for **all six** nodes to answer `INFO cluster` with `cluster_enabled:1`*. Not blocking.

fatal10110:
## Re-review, round 3: PR #428 at `62577df` (previously `c910e82`)

I checked this on a private stack of this head (own compose project, no host ports), using the image's own `dash`/`mawk`. CI has no failures, and all three -real jobs pass.

### Round-2 follow-ups

| | Finding | Status | Evidence |
|---|---------|--------|----------|
| A | Shared `/data/dump.rdb` poisons restart | **Fixed** | `/data/<port>` per node, and `redis-server` is launched from one place only, so every node gets its own `--dir`. `rm -rf` of each node's whole directory also removes anything AOF would create there. Live: 30 keys, `SAVE` on all six (dump sizes 259–342 bytes, one per dir), `docker restart` → **formed on attempt 1/3**, `dbsize` 0, healthy in 13s. |
| B | pfail comment backwards | **Fixed** | The three-stage sequence (no flag → `fail?` → `fail`) matches my round-2 SIGSTOP timeline. Only the first stage is blind, and its length is `cluster-node-timeout`. |
| C | Unbounded `redis-cli` on a hung node | **Fixed** | See below. |

### C in detail (the probe logic changed a lot)

- **Hung node during formation, exact CI command.** I SIGSTOPped 30005 right after "all 6 nodes are up". `up -d --wait --wait-timeout 420` returned `exited (1)` after **157s**. The log shows attempts 1/3, 2/3, 3/3, the FATAL line, a dump that completed, and `(no CLUSTER INFO reply from node 30005 within 5s: down or hung)`. Every `[cluster-init]` line starts on its own line, so the trailing-dots fix works.
- **Hung node on a formed cluster.** `check` exited 1 in **5.2s** and the container went unhealthy. Every Docker probe ran 5–6s.
- **No orphans.** I sampled `/proc` four times over 12s. Every `timeout`/`redis-cli` process was under 5s old and a child of the probe currently running. Nothing piled up.
- **`set -e` + parallel rounds.** dash's bare `wait` returns 0 even after failing and timed-out background jobs (`(exit 3) & timeout 1 sleep 3 & wait` → status 0, script alive). So `reset_nodes`/`dump_diagnostics` can't kill the retry loop.
- **Nothing relies on exit status or an empty reply.** `node_ready` greps the reply. `judge_views` reads content and walks the *expected* port list in `END`. The only remaining exit-status check is `--cluster create`, which is cluster mode rather than stdin mode, and `wait_for_cluster` re-checks its result by reading replies anyway. 14 adversarial fixtures in mawk all behave as they should: jemalloc noise on every file → `ok`; noise-only (a timeout-killed call), empty file, refused connection, killed between INFO and NODES, **missing file** (awk errors, verdict empty ≠ `ok`), `slave,fail`, `cluster_state:fail`, 16383 slots, 7 known nodes, 5m/1s and a duplicate node line all **fail**. `nofailover` and a trailing error line → `ok`. A call killed by `timeout` cannot pass as success.
- **Budget.** 35 + 3×(45+30+5) + 2×(5+2) + 5 = **294s**, which matches the code. The per-round bound holds because `query_all` runs one call per node in parallel. The only extra time is sleep granularity and awk/mktemp startup (under 2s per attempt), which is well inside the 36s margin to `start_period` 330s. The measured 157s for a hung node matches 3×45 + 2×7 + 5 plus boot.

### New
One nit, posted inline: `docs/TEST-INTEGRATION.md:169` still says the script waits for nodes to answer `PING`. It sends `INFO cluster` alone now.

### Verdict: **ship**
All three round-2 follow-ups are fixed. I checked every claim myself and got the same results the implementer reported. The rewritten probe fails closed in every case I could construct. The docs nit doesn't block merging.

Issue #267, opened 2026-06-19
Flaky test: SCAN MATCH interleaved-batches iteration lower-bound is nondeterministic on real backend


fatal10110:
## Problem
`tests-integration/ioredis/scan-integration.test.ts:183` (`SCAN MATCH handles interleaved matching and non-matching batches`) intermittently fails on the **real** Redis backend in CI.

Observed in CI run [27838605393](https://github.com/fatal10110/js-redis-server/actions/runs/27838605393) (job 82392326433):

```
AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
  assert.ok(result.iterations > Math.ceil(expected.length / 2))
  at TestContext.<anonymous> (tests-integration/ioredis/scan-integration.test.ts:183:14)
  actual: false, expected: true
```

A re-run passed without code changes — confirming flakiness, not a regression.

## Root cause
The assertion `result.iterations > Math.ceil(expected.length / 2)` assumes top-level `SCAN ... COUNT 2` takes "many" iterations. But the test's own comment notes top-level SCAN sweeps the **whole node keyspace**, and the suite runs files concurrently, so the number of keys on the node — and therefore the iteration count — is not deterministic. When the node keyspace is small at scan time, real Redis returns the matches in fewer iterations and the lower bound fails.

The `values` assertion (line 175) is exact and reliable because `MATCH` filters; only the `iterations` lower-bound is environment-dependent.

## Fix options
- Drop the iteration lower-bound assertion (keep the exact `values` check).
- Or make the test deterministic: seed enough non-matching keys on the same node under a unique tag and scope the lower bound to a stable per-key-count, rather than relying on shared keyspace size.

## Notes
- `values` correctness is unaffected — this is purely about the iteration-count heuristic.
- Surfaced while working PR #266 (unrelated cluster-policy refactor); failure is in SCAN code, not that change.

PR #429, opened 2026-09-22
fix: isolate integration keys per run and make clean:redis fail loudly (#420, #395)

### LINKED ISSUE #395: clean:redis is a silent no-op in CI (redis-cli not installed, failure swallowed) — likely root cause of the SCAN flakes
Found while working on PR #394. This is a plausible **root cause for the long-running SCAN flakes**, including #267.

## Problem
`package.json`'s cleanup script:
```
"clean:redis": "redis-cli --cluster call 127.0.0.1:30000 FLUSHALL || redis-cli FLUSHALL || echo 'Redis cleanup attempted'"
```
The trailing `|| echo` swallows every failure. In CI, `redis-cli` is **not installed** — the Redis services run as docker-compose containers, and no step installs the client on the runner. So both branches fail, `echo` succeeds, the script exits 0, and **the keyspace is never flushed**.

It has therefore never cleaned anything in CI, silently, while reporting success. Locally it works only if the developer happens to have `redis-cli` on PATH.

## Consequence
`test:integration:real` runs `clean:redis` before the suite specifically to start from an empty keyspace. It doesn't. Every real-backend run inherits whatever the previous run left behind.

That is a strong candidate explanation for the recurring SCAN flakes:
- **#267** — `SCAN MATCH handles interleaved matching and non-matching batches`, `tests-integration/ioredis/scan/keys-scan.test.ts:161`
- The same assertion class at `tests-integration/node-redis/scan/keys-scan.test.ts:126`, seen failing once during #394 and passing on a plain re-run

Both assert `result.iterations > Math.ceil(expected.length / 2)` — i.e. they assume a known keyspace size. Leftover keys from a previous run change the iteration count, so the assertion's premise ("extra keys only raise iterations") does not hold against a polluted real Redis. Note this is *also* a questionable assertion independent of pollution, since real Redis makes no guarantee about `SCAN` batch sizes — worth revisiting alongside.

## Proposal
1. Make the cleanup actually run in CI: either install `redis-cli` on the runner, or flush via `docker compose exec` against the running containers, or add a `FLUSHALL` through one of the existing Node clients as a test-harness pretest step.
2. **Remove the unconditional `|| echo`.** A cleanup step that cannot fail is a cleanup step nobody can trust. If a best-effort fallback is genuinely wanted locally, make it explicit and loud.
3. Consider per-suite isolation instead of global flush — a unique key prefix per run would make the suites robust to leftovers regardless.
4. Revisit the `iterations` assertions so they do not depend on total keyspace size.

## Acceptance
- `clean:redis` fails loudly when it cannot reach Redis.
- A real-backend CI run demonstrably starts from an empty keyspace (assert `DBSIZE 0` at suite start).
- #267 and its node-redis twin stop recurring, or are rewritten not to depend on keyspace size.

### LINKED ISSUE #420: String integration suites use fixed key names on an unflushed shared server and collide across runs
Found while working on PR #409. Third distinct flake class in the integration suites, after #395 (SCAN, keyspace pollution) and #411 (hash field TTLs racing their own round-trips).

## Problem
Several string integration suites use **fixed, shared key names** against a real backend that is never flushed. When many real-backend files run in one process, they inherit each other's data — and data from previous runs.

Observed directly against node 30000:
```
string-integrationappendkey  ->  "hello worldhellohellohello"
```
so the `APPEND` assertion fails with `26 !== 5`. Run individually the same files pass, because the key happens to be absent or short enough.

This is not the same bug as #395. #395 is about `clean:redis` silently failing in CI, which means the keyspace is never flushed. This is about tests that **would still collide even if it worked**, because two files can use the same key name within a single run, and a test that appends is not idempotent against its own previous execution.

## Why it matters
- It produces failures that look like real regressions on an unrelated PR. During PR #409's work this misdirected effort twice.
- It is load-dependent and ordering-dependent, so it reproduces on CI and not locally, which is the worst possible signature.
- Combined with #411 and #395, the integration suites currently have three independent reasons to fail for no code reason. Each one individually teaches people to re-run; together they make a red run nearly uninformative.

## Proposal
Use unique keys per test run. The repo already has a `randomKey()` helper — PR #409's own new suites use it and were unaffected. Options, in order of preference:

1. Sweep the string suites (and any other suite using fixed literal key names) onto `randomKey()`.
2. Or give each test file a unique key prefix derived from its path plus a per-run nonce, which also makes leftover keys attributable when debugging.

Either removes the dependence on a flush entirely, which is more robust than fixing the flush — and is the "per-suite isolation instead of global flush" option already suggested in #395.

## Acceptance
- No integration test depends on a key being absent at start.
- Running the full real-backend suite twice in a row, without any flush in between, passes both times.

fatal10110:
## Summary

Two related integration-test infrastructure bugs that together made a red
real-backend run nearly uninformative. No `src/` changes.

Closes #420
Closes #395

### #395 — `clean:redis` claimed flushes that never happened

The old script ended in `|| echo 'Redis cleanup attempted'`, so it exited 0
whatever happened. CI runs Redis as docker-compose services and never installs
`redis-cli`, so both `redis-cli` branches failed, the `echo` succeeded, and the
keyspace was never flushed — silently, while reporting success.

It is replaced by `scripts/flush-redis.ts` (ioredis, already a devDependency:
no `redis-cli` on PATH and no `docker compose exec`, which does not exist on a
Docker CLI without the Compose plugin). Its rule is that **"flushed" is only
ever claimed on evidence, never inferred from a command not erroring**:

- **The whole cluster, not just the listed ports.** `REDIS_CLUSTER_PORTS` holds
  *seeds*: the harness's cluster clients discover every node from any one of
  them, so a run pointed at a single port still uses every master. The script
  therefore reads `CLUSTER NODES` from each seed and flushes the entire
  topology it finds. A listed-but-unreachable node, or one with no usable
  address, is a failure — never a silent omission.
- **Masters** are flushed with `FLUSHALL`, then asserted empty with `DBSIZE`.
- **Replicas** refuse `FLUSHALL` with `-READONLY`, so they are verified rather
  than trusted: masters are flushed first, then each replica must belong to a
  master this run actually flushed *and* reach `DBSIZE 0` itself. A replica
  whose master link is down keeps a full stale keyspace while happily refusing
  the write.
- **Cluster health.** `FLUSHALL`/`DBSIZE` are keyless and answer normally on a
  cluster that is down, so every cluster node must also report
  `cluster_state:ok` — otherwise the suite's next keyed command gets
  `-CLUSTERDOWN` against a keyspace just declared clean.
- **Standalones** (`REDIS_STANDALONE_PORT`, `REDIS_STANDALONE_AUTH_PORT`) are
  flushed when set — the same condition under which the harness uses them.
- **Exit 1 on any unverified endpoint**, with a hint that fits the failure:
  "start the backends" only when something is actually unreachable; separate
  advice for an unhealthy cluster, a stale replica, a non-cluster seed, or keys
  surviving `FLUSHALL`.

All env parsing lives in `tests-integration/redis-endpoints.ts`, shared by the
harness and the script so they cannot disagree about which endpoints exist.
Every port — cluster list and both standalones — is range-checked, and a
malformed entry throws instead of being dropped (`"30000,30001-30005"` used to
collapse to one port). `test:integration:raw:real` now runs `clean:redis` too.

### #420 — suites collided on fixed literal key names

Several suites used fixed keys against a real backend that is never flushed
between files or runs, so they inherited each other's data and their own
previous run's — e.g. `APPEND` returning `26 !== 5` because
`string-integrationappendkey` still held `"hello worldhellohellohello"`.

Every such suite, in both clients, is swept onto the existing `randomKey()`
helper, following the convention PR #409's suites already use: a file-level
`const RUN = randomKey()` namespacing plain keys (`` `appendkey:${RUN}` ``) and
hash tags (`` `{same:${RUN}}mget1` ``, with `RUN` inside the braces so
same-slot grouping is preserved). CROSSSLOT probes either keep fixed literal
tags or build the second key with `keyInAnotherSlot(first, …)`, so the two keys
are guaranteed to hash apart rather than colliding ~1 run in 16384.

Assertions that silently depended on the size of the shared keyspace are
rewritten:

- **SCAN paging** (`scan/keys-scan.test.ts`, #267 and its node-redis twin):
  `iterations > Math.ceil(expected.length / 2)` became `iterations > 1`. It
  still proves the backend paginates top-level SCAN — a backend returning
  everything with cursor 0 fails — without assuming how many keys the node
  holds. The full traversal must also yield exactly the matching keys.
- **`DBSIZE - baseline` deltas** (`key/commands`, `key/workflow`): total DBSIZE
  counts every other suite's keys, including ones whose TTL expires mid-test,
  so the delta drifted (`28 !== 29` reproduced on a freshly flushed cluster).
  `assertKeyCount` / `assertNodeRedisKeyCount` instead assert that exactly the
  expected keys exist *and* that `KEYS <suite tag>:*` on the tag's slot owner
  finds nothing extra — so a command creating an unexpected key still fails.
  Exact DBSIZE is still asserted (`== 0` after a flush) by the
  flush-async-sync suites.
- An unawaited `set()` that an `EVAL` then read back, in both `scripts`
  suites, is awaited; with per-run keys it had become load-bearing.

## Decision: reviewer item 9b considered, deliberately not implemented

**"No suite asserts `DBSIZE 0` at start, which is #395's literal acceptance
criterion."** Considered and intentionally declined. This is a resolved
decision, not outstanding work.

The two acceptance criteria cannot both hold at test level:

- #395 asks a suite to assert `DBSIZE 0` at start.
- #420 asks the full real-backend suite to pass **twice in a row with no flush
  in between** — so the second run legitimately starts on the first run's
  leftovers. Any suite asserting `DBSIZE 0` at start would fail that second run
  by construction.

Implementing 9b would therefore have broken the other issue this same PR
closes, and re-introduced exactly the coupling #420 removes ("no integration
test depends on a key being absent at start"). The intent — *demonstrate that
a real-backend run starts from an empty keyspace* — is met in the cleanup step
instead, more strongly than a test-level check could: `clean:redis` asserts
`DBSIZE 0` on **every** node of the discovered topology and both standalones,
plus `cluster_state:ok`, rather than on whichever single node one suite
happens to touch.

## Test plan

- [x] `npm run lint`, `npx tsc --noEmit`, `npx prettier --check` clean;
      `npm test` — 378 unit tests pass.
- [x] `TEST_BACKEND=mock` full integration suite passes — 1181 tests. One
      early run this round showed 4 nondeterministic failures that I did not
      capture, so I cannot name them; three immediate reruns at load average
      72–95 were all 1181/1181. This round's test changes cannot fail
      nondeterministically (the watch keys either work or don't; the
      standalone-port parsing only runs on the real backend).
- [x] CI green, including `ioredis-real`, `node-redis-real` and `raw-real`.
      Their logs show `clean:redis` flushing every master, verifying every
      replica, and asserting `cluster_state:ok` on all six nodes — in the
      environment (no `redis-cli`) where the old script silently no-opped.
- [x] **`clean:redis` behaviour**, each case reproduced on a private cluster.
      The first now succeeds where it used to lie; every other case exits 1
      with the hint that matches it:
      - `REDIS_CLUSTER_PORTS=<one port>` (the partial-list case): the old
        script flushed the one listed master, left **43 keys on the two
        unlisted masters, and exited 0**. The new one discovers the other five
        nodes, flushes all three masters, verifies all three replicas, and
        leaves every node at `DBSIZE 0`. Seeding from a lone *replica* port
        works the same way.
      - An unlisted node down while `cluster_state` still reads `ok` → reported
        as unreachable with the cluster-unhealthy hint.
      - Replica with `master_link_status:down` holding 3 stale keys (`FLUSHALL`
        → `READONLY`) → `still holds 3 key(s) after its master was flushed`.
      - Cluster in `cluster_state:fail` (master + replica downed) → every
        reachable node reported, cluster-unhealthy hint.
      - Unreachable standalone → "start the backends" hint.
      - Non-cluster node given as a cluster seed → its own hint.
      - Malformed ports (`30000,30001-30005`, `30000,abc,30002`, `99999`, and
        malformed standalone ports) → rejected before connecting, by the
        script and by the harness alike.
- [x] **#420 acceptance** — full real suite twice in a row with no flush
      between, on a private cluster. Met; see *Acceptance runs* below.
- [x] **This round's changes on the real backend, single seed**
      (`REDIS_CLUSTER_PORTS=31100`): the 39 suites they touch — both watch
      suites plus every suite that goes through the standalone-port path —
      twice in a row with no flush between: **236 tests, 0 failures, both
      passes**. `clean:redis` discovered all six nodes from the one seed.
- [x] `string/bitmap.test.ts` was excluded from the local runs only:
      `SETBIT <key> 4294967295 1` allocates a 512MB string, which OOM-killed
      nodes on this machine's 1.9GB Docker VM. It is green in CI.

### Acceptance runs (#420)

Full real suite, twice in a row, no flush between, on a private cluster
(`string/bitmap.test.ts` excluded locally, see above). It was run on the
round-1 and on the round-2 code:

| code    | pass 1           | pass 2 (on pass 1's leftovers) |
| ------- | ---------------- | ------------------------------ |
| round 1 | 1135, **1 fail** | 1135, 0 fail                   |
| round 2 | 1135, 0 fail     | 1135, **4 fail**               |

**The acceptance criterion is met: every one of those 5 failures is #411.**
They are all the same two tests — `hash writes clear or preserve field TTLs
like Redis` and `HPERSIST, HTTL and HPTTL report and clear hash field TTLs`,
in `hash/field-expire.test.ts` / `hash/field-persist.test.ts`, both clients —
in files this PR does not touch, and nothing else failed in any of the four
passes. The failure moving between pass 1 and pass 2 across the two runs is
itself the signature of a timing flake, not keyspace state.

It was also confirmed directly: re-running exactly those four suites against
the *same unflushed* keyspace, without full-suite contention, gives **19/19
pass**. A state-dependent failure would have reproduced there. The machine was
at load average 17–20 during those passes.

## Review history

- **Round 1** found two ways `clean:redis` could still report a flush that
  never happened (replicas trusted on `-READONLY`; keyless commands passing on
  a `cluster_state:fail` cluster), the port parser dropping malformed entries,
  the SCAN paging bound removed rather than fixed, a vacuous DBSIZE check, an
  unawaited `set()`, missing `clean:redis` on `raw:real`, and random-vs-random
  CROSSSLOT probes. All fixed; 9b settled as above.
- **Round 2** found that a partial `REDIS_CLUSTER_PORTS` list still produced a
  partial flush with exit 0 (fixed by topology discovery), that the harness
  still parsed standalone ports with a bare `Number(...)` (now shared and
  validated), that every failure advised starting the backends (now
  failure-specific), and two remaining fixed `watch:{multi}` keys (now
  run-namespaced in both clients). This description was rewritten to match the
  current code.


scripts/flush-redis.ts
```diff
@@ -0,0 +1,174 @@
...
+  client.on('error', () => {})
+
+  try {
+    await client.connect()
+
+    try {
+      await client.flushall()
+    } catch (err) {
+      if (endpoint.allowReadonly && isReadonlyReplicaError(err)) {
+        return `${endpoint.label}: replica, flushed via its master`
```
fatal10110: **HIGH — the `-READONLY` branch reports success without ever verifying the node is empty.**

This returns `"replica, flushed via its master"` and **returns early**, so the `DBSIZE === 0` check on line 128 never runs for any node that refuses the write. The assumption "it refused, therefore its master's FLUSHALL covers it" is not sound: a replica whose master link is down still refuses writes while holding a complete, stale keyspace.

Reproduced live on a real replica:

```
127.0.0.1:6502> INFO replication
role:slave
master_link_status:down
127.0.0.1:6502> FLUSHALL
(error) READONLY You can't write against a read only replica.
127.0.0.1:6502> DBSIZE
(integer) 31
127.0.0.1:6502> KEYS *
1) "s1"
2) "s3"
3) "s2"
```

`flush()` returns a fulfilled promise, `main()` prints `replica, flushed via its master`, and `clean:redis` exits 0 — with 31 keys still there. That is precisely the #395 failure mode this script exists to eliminate: a cleanup step that reports success without cleaning.

Compounding it: `Promise.allSettled(targets.map(flush))` (line 141) starts every endpoint concurrently with no ordering, so a replica can be probed *before* its master has been flushed, in which case "its master's flush covers it" is not even true at the moment the claim is made.

Suggested fix: don't return early. After the READONLY refusal, still poll `DBSIZE` on that node (short retry loop, since replication is async) and fail if it doesn't reach 0. That keeps the replica-is-fine case working and closes the stale-replica and ordering holes in one go.

scripts/flush-redis.ts
```diff
@@ -0,0 +1,174 @@
...
+      }
+      throw err
+    }
+
+    const remaining = await client.dbsize()
```
fatal10110: **HIGH — `DBSIZE === 0` is a vacuous success signal on a broken cluster.**

`FLUSHALL` and `DBSIZE` are keyless, so they bypass Redis's cluster-state gate entirely. On a cluster in `fail` state (slots unassigned, a master down, resharding in flight) every node happily answers both, the script prints six `flushed` lines and exits 0 — while the suite that follows gets `-CLUSTERDOWN` on its first key-bearing command.

Reproduced against the cluster this repo's `docker-compose.test.yml` publishes:

```
$ redis-cli -p 30000 cluster info | head -1
cluster_state:fail
$ redis-cli -p 30000 dbsize
(integer) 0
$ redis-cli -p 30000 get probe
(error) CLUSTERDOWN Hash slot not served
```

So #395's acceptance criterion ("a real-backend CI run demonstrably starts from an empty keyspace") is satisfied vacuously: the keyspace is empty because nothing can be stored in it. The run then fails deep in an unrelated test file with a confusing error instead of failing here with a clear one.

Cheap fix: for endpoints created from `clusterPorts()`, also `await client.cluster('INFO')` and require `cluster_state:ok` before reporting the node flushed. Two extra lines, and it turns a whole class of "the tests are red for no code reason" runs into one actionable message from `clean:redis`.

scripts/flush-redis.ts
```diff
@@ -0,0 +1,174 @@
...
+function clusterPorts(): number[] {
+  const configured = process.env.REDIS_CLUSTER_PORTS
+  if (!configured) {
+    return DEFAULT_CLUSTER_PORTS
+  }
+
+  const ports = configured
+    .split(',')
+    .map(part => Number(part.trim()))
+    .filter(port => Number.isInteger(port) && port > 0)
```
fatal10110: **MEDIUM-HIGH — this `.filter()` silently discards malformed entries, so a typo yields a partial flush that still exits 0.**

The guard only fires when *every* entry is unparsable. A value with one bad element is silently truncated to the parts that happened to parse:

```
"30000-30005"        -> THROWS  (good)
"30000,30001-30005"  -> [30000]                     <-- 5 nodes silently skipped
"30000,abc,30002"    -> [30000,30002]               <-- 1 node silently skipped
"99999,30000"        -> [99999,30000]               <-- 99999 is not a valid port
```

(run against this PR's `clusterPorts()` verbatim.)

The mixed case is the realistic one, because `CONTRIBUTING.md` in this same PR prints the default as the *range* `30000-30005`. A developer who copies that shape and edits part of it gets `[30000]`: one node flushed, five left dirty, `clean:redis` exits 0, and the suite then fails in whichever file happens to land on a polluted node. That is the same silent-partial-cleanup class #395 is about — just moved from "no `redis-cli`" to "bad env var".

Suggested fix: reject the whole value if *any* element fails to parse, and bound the range (`> 0 && <= 65535`), e.g. map each part and throw naming the offending token. Same change is needed in the copy of this logic at `tests-integration/test-config.ts:412`.

tests-integration/test-config.ts
```diff
@@ -394,8 +394,30 @@ export class TestRunner {
...
-    return [30000, 30001, 30002, 30003, 30004, 30005]
+    const configured = process.env.REDIS_CLUSTER_PORTS
+    if (!configured) {
+      return [30000, 30001, 30002, 30003, 30004, 30005]
+    }
+
+    const ports = configured
+      .split(',')
+      .map(part => Number(part.trim()))
+      .filter(port => Number.isInteger(port) && port > 0)
```
fatal10110: **MEDIUM-HIGH — duplicate of the parser in `scripts/flush-redis.ts:40-58`, with the same silent-drop bug.**

These ~15 lines are byte-for-byte the logic in the new flush script, so both the harness and the cleanup step silently drop malformed entries in `REDIS_CLUSTER_PORTS` (see the detail on `scripts/flush-redis.ts:49`). Two copies also means the next fix has to land twice, and the failure mode if they ever drift is "the suite and the flush target different nodes", which is very hard to debug.

The default `[30000, 30001, 30002, 30003, 30004, 30005]` is confirmed unchanged from `main`, and the env var is read consistently by both readers today — this is about keeping it that way. Suggest exporting one `parseClusterPorts()` (from `tests-integration/test-config.ts`, which the script can import as it already lives in the repo, or from a small shared module) and calling it from both.

tests-integration/ioredis/scan/keys-scan.test.ts
```diff
@@ -150,15 +150,24 @@ describe(`Scan Commands Integration (${testRunner.getBackendName()})`, () => {
...
-      // Lower bound holds on both backends — extra keys only raise iterations.
-      assert.ok(result.iterations > Math.ceil(expected.length / 2))
+      // Cross-check the multi-step cursor traversal against a single-shot
+      // KEYS over the same live keyspace — both must agree.
+      assert.deepStrictEqual(
+        (await directClient.keys(`${tag}:hit:*`)).sort(),
+        expected,
+      )
+
+      // Deliberately no assertion on `result.iterations`. COUNT is only a
```
fatal10110: **MEDIUM — the stated reason for deleting the `iterations` bound is not what real Redis does, and this removes the only coverage that top-level SCAN paginates at all.**

The comment says the old bound "assumed a known keyspace size". Measured against real Redis 8.0 with this test's exact shape (`MATCH <tag>:hit:* COUNT 2`, 6 matching keys):

```
empty db, 10 keys total      -> iterations = 4      (bound: 4 > 3   holds)
db with 5010 keys total      -> iterations = 2175   (bound: 2175 > 3 holds)
```

The bound holds at both extremes and only gets *safer* as the keyspace grows — extra keys can only raise the iteration count, never lower it. Keyspace pollution was not what made #267 flake here.

What the deletion actually costs: this is now the only SCAN test in the file with no statement about the traversal, so a backend that ignores `COUNT` and returns the whole matching set in one round-trip with cursor `0` passes it silently.

The replacement `KEYS` cross-check does not recover that. `assert.deepStrictEqual(result.values, expected)` two lines above already pins the exact set, and `KEYS matches Redis glob pattern semantics` (line 219) already covers KEYS/SCAN glob agreement — so the new assertion can only fail in a scenario the file already tests.

Note also that this PR keeps a structurally identical assertion 50 lines above at line 114, `assert.ok(result.iterations > result.values.length)` (and again at line 212), which with `values.length === 1` is exactly `iterations > 1`. That is the keyspace-size-independent bound this test wanted. Suggest restoring it here as `assert.ok(result.iterations > 1)` — it drops the `Math.ceil(expected.length / 2)` premise you're right to reject, while keeping the paging coverage.

tests-integration/node-redis/scan/keys-scan.test.ts
```diff
@@ -151,15 +151,24 @@ describe(`Scan Commands Integration (node-redis, ${testRunner.getBackendName()})
...
-      // Lower bound holds on both backends — extra keys only raise iterations.
-      assert.ok(result.iterations > Math.ceil(expected.length / 2))
+      // Cross-check the multi-step cursor traversal against a single-shot
+      // KEYS over the same live keyspace — both must agree.
+      assert.deepStrictEqual(
+        (await directClient.keys(`${tag}:hit:*`)).sort(),
+        expected,
+      )
+
+      // Deliberately no assertion on `result.iterations`. COUNT is only a
```
fatal10110: **MEDIUM — same as the ioredis twin at `tests-integration/ioredis/scan/keys-scan.test.ts:165`.**

Deleting the `iterations` bound leaves this test with no assertion about the cursor traversal, so a SCAN that ignores `COUNT` and returns everything in one batch with cursor `0` passes. The `KEYS` cross-check is near-tautological against the `deepStrictEqual(result.values, expected)` above it.

Measured on real Redis 8.0 with this test's shape: `iterations = 4` on a 10-key db, `2175` on a 5010-key db — the old bound (`> 3`) holds at both ends, so keyspace size was not the flake cause. `assert.ok(result.iterations > 1)` restores the coverage without the keyspace-size premise, and matches what lines 114 / 212 of this same file already assert.

tests-integration/utils.ts
```diff
@@ -42,13 +42,37 @@ export async function getTotalDbSize(redisClient: Cluster): Promise<number> {
...
+export async function assertKeyCount(
+  redisClient: Cluster,
+  keys: readonly string[],
+  expected: number,
 ): Promise<void> {
-  const size = await getTotalDbSize(redisClient)
-  assert.strictEqual(size - baseline, expectedDelta)
+  assert.strictEqual(await countExistingKeys(redisClient, keys), expected)
+  assert.ok(
+    (await getTotalDbSize(redisClient)) >= expected,
```
fatal10110: **MEDIUM — this second assertion cannot catch anything, and it drops the one thing the old delta did catch.**

*Vacuous:* if line 71 passed, those `expected` keys demonstrably exist on the masters, so the sum of the masters' `DBSIZE` is necessarily `>= expected`. There is no server-side bug this can detect. Its only realistic failure mode is a false one: `getTotalDbSize` sums `redisClient.nodes('master')`, so while ioredis's slot map is refreshing (failover, slot migration, a node briefly dropping out) that list can be incomplete and the sum comes back too low — a spurious failure with the message "DBSIZE must count at least the keys this suite created", in a PR whose whole purpose is removing flakes. It also costs an extra round-trip per master on every call site.

It also sits oddly against the paragraph this PR adds to `AGENTS.md:120` ("never assert on total `DBSIZE`").

*Lost coverage:* `assertDbSizeDelta(baseline, 35)` failed if the commands under test created a **36th, unexpected** key. `assertKeyCount(keys, 35)` only counts names the test already knows about, so a bug that writes a wrongly-named key — `COPY`/`RENAME` landing on the wrong destination, a leaked temp key, an off-by-one in a key builder — now passes. That is a real regression class, and it is cheap to keep: every call site already namespaces its keys under one hash tag, so

```ts
assert.strictEqual((await redisClient.keys(`${tag}:*`)).length, expected)
```

is exact, single-slot, deterministic, and immune to other suites' keys — which is exactly the property the delta lacked. Suggest replacing the `>=` DBSIZE check with that (and the node-redis equivalent at line ~344).

tests-integration/ioredis/scripts.test.ts
```diff
@@ -48,7 +52,7 @@ describe(`Redis scripts (ioredis) ${testRunner.getBackendName()}`, () => {
     })
 
     test('gets utf8 value correctly', async () => {
-      redisClient?.set('myKey', 'фвфв')
+      redisClient?.set(`myKey:${RUN}`, 'фвфв')
```
fatal10110: **MEDIUM — missing `await`, and this PR is what makes it load-bearing.**

`redisClient?.set(...)` here is not awaited, while the `EVAL` on line 66 reads the key back and the test asserts it saw `'фвфв'`.

Before this PR the key was the fixed literal `myKey`, which the shared, never-flushed cluster kept populated from earlier tests and earlier runs — so the assertion passed regardless of whether this SET had landed. The leftover data was masking the missing `await`. Now the key is unique per run, and the test's correctness depends entirely on this unawaited write reaching the node before the `EVAL` does.

It happens to hold today only because both commands hash to the same slot and ioredis writes them FIFO on one node connection. Nothing in the test states or enforces that, and if the SET rejects (MOVED during a slot move, connection reset) it becomes a floating unhandled rejection rather than a test failure.

This is the good kind of consequence of #420 — removing the pollution exposes tests that were quietly relying on it. Just add `await`.

tests-integration/node-redis/scripts.test.ts
```diff
@@ -48,7 +52,7 @@ describe(`Redis commands (${testRunner.getBackendName()})`, () => {
     })
 
     test('gets utf8 value correctly', async () => {
-      redisClient?.set('myKey', 'фвфв')
+      redisClient?.set(`myKey:${RUN}`, 'фвфв')
```
fatal10110: **MEDIUM — same missing `await` as `tests-integration/ioredis/scripts.test.ts:55`.**

The `EVAL` on line 66 reads this key back and asserts the value. Unawaited, the test's correctness now rests on ioredis's connection-level ordering rather than on anything the test states — and before this PR the fixed `myKey` left over in the shared keyspace was hiding that. Add `await`.

scripts/flush-redis.ts
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  return ports
+}
+
+function endpoints(): Endpoint[] {
+  const targets: Endpoint[] = clusterPorts().map(port => ({
+    label: `cluster node 127.0.0.1:${port}`,
+    port,
+    allowReadonly: true,
```
fatal10110: **LOW — `allowReadonly: true` is applied unconditionally to every configured cluster port.**

There's no check that the refusing node's master is actually in the port list. Combined with the new `REDIS_CLUSTER_PORTS` override, that makes a degenerate case reachable: point it at the three replica ports (or hit it after a failover that made the listed nodes replicas) and the script prints a success line per endpoint having flushed nothing, then exits 0.

Verifying `DBSIZE` reaches 0 on the READONLY path (see the comment on line 123) also fixes this one, since a replica whose master was never flushed would then fail loudly. Alternatively, require at least one endpoint to have actually performed a flush before reporting success.

scripts/flush-redis.ts
```diff
@@ -0,0 +1,174 @@
...
+    })
+  }
+
+  const authPort = process.env.REDIS_STANDALONE_AUTH_PORT
+  if (authPort) {
+    targets.push({
+      label: `standalone (auth) 127.0.0.1:${authPort}`,
+      port: Number(authPort),
+      password:
+        process.env.REDIS_STANDALONE_AUTH_PASSWORD ?? DEFAULT_AUTH_PASSWORD,
```
fatal10110: **LOW — this env var is read by nothing else, so setting it makes the flush and the harness disagree.**

`tests-integration/test-config.ts:22` exports `STANDALONE_AUTH_PASSWORD = 'testpass'` and uses it unconditionally for both the docker-compose path and the spawned-child path; it never consults an env var. This script re-declares the same literal as `DEFAULT_AUTH_PASSWORD` (line 28) and adds `REDIS_STANDALONE_AUTH_PASSWORD` on top. Set that variable and the flush authenticates with one password while every test authenticates with another.

Two copies of a credential literal also drift silently if `docker-compose.test.yml`'s `--requirepass` ever changes.

Also worth improving while here: the failure message is unhelpful for this case. Verified against the running `redis-standalone-auth` container:

```
$ REDIS_STANDALONE_AUTH_PORT=6400 REDIS_STANDALONE_AUTH_PASSWORD=wrong node --import tsx ./scripts/flush-redis.ts
clean:redis failed — could not flush 1 of 2 endpoint(s):
  standalone (auth) 127.0.0.1:6400: Connection is closed.
```

"Connection is closed." is what ioredis reports for a wrong password *and* for a refused connection, so the suggested remedy ("Start the test backends first") sends you the wrong way. Surfacing `err.code` / the original `WRONGPASS` would fix that.

Suggest importing `STANDALONE_AUTH_PASSWORD` and dropping both the duplicate literal and the env var.

package.json
```diff
@@ -78,7 +78,7 @@
     "test:package": "npm run build && node --enable-source-maps --import tsx --no-warnings --test \"./tests-package/**/*.test.ts\"",
     "test:demo": "node --enable-source-maps --import tsx --no-warnings --test \"./examples/browser-demo/*.test.ts\"",
     "test:all": "npm run test && npm run test:package && npm run test:demo && npm run test:integration:mock && npm run test:integration:real",
-    "clean:redis": "redis-cli --cluster call 127.0.0.1:30000 FLUSHALL || redis-cli FLUSHALL || echo 'Redis cleanup attempted'",
+    "clean:redis": "node --import tsx --no-warnings ./scripts/flush-redis.ts",
```
fatal10110, line 81: **LOW — `clean:redis` is still not wired into every real-backend entry point, and nothing asserts an empty keyspace at suite start.**

`test:integration:real`, `:real:ioredis` and `:real:node-redis` all run `clean:redis` first — but `test:integration:raw:real` (line 77) does not, so the raw-tcp real job starts from whatever the previous job left behind. Since this PR's premise is that the cluster is never flushed between files, that job is the one place still exposed to the original #420 symptom.

Separately, #395's second acceptance criterion — "a real-backend CI run demonstrably starts from an empty keyspace (assert `DBSIZE 0` at suite start)" — isn't implemented by any suite. The only `DBSIZE === 0` assertions in the tree are in `flush-async-sync.test.ts`, which flush a standalone themselves and so don't say anything about the shared cluster's starting state. (The docstring added at `tests-integration/utils.ts:63` cites those suites as if they covered it.) Worth either adding the check or adjusting the issue's acceptance notes.

AGENTS.md
```diff
@@ -117,6 +117,10 @@ The integration test suite supports two backends via `TEST_BACKEND` (see [tests-
 
 Integration tests live in [tests-integration/](tests-integration/) with subdirectories for `ioredis/` and `node-redis/` clients.
 
+**Key isolation is mandatory.** The real backend is a long-lived, shared Redis cluster that is *not* flushed between test files, so every key a test touches must be unique per run — derive it from `randomKey()` ([tests-integration/utils.ts](tests-integration/utils.ts)), either per test (`const tag = \`{feature:${randomKey()}}\``) or via a file-level `const RUN = randomKey()`. Never assert on a fixed literal key name, never assume a key is absent at start, and never assert on total `DBSIZE` (other suites' keys — including ones expiring mid-test — make it drift); count the suite's own keys with `assertKeyCount` / `assertNodeRedisKeyCount` instead. The suite must pass twice in a row with no flush in between (#420).
```
fatal10110, line 120: **LOW — two residual key-isolation gaps that this new rule makes worth closing.**

The rule as written is "never assert on a fixed literal key name" and the sweep is otherwise thorough (an exhaustive hash-tag check across all 38 changed files found the only multi-slot call sites are the 5 intentional CROSSSLOT probes, and the `{same:${RUN}}` / `{test:${RUN}}` regrouping keeps every same-slot group on one slot). Two leftovers:

**1. The randomised CROSSSLOT probes can collide.** Four sites build both sides of a "must be different slots" probe from independent `randomKey()` calls, e.g. `tests-integration/ioredis/key/commands.test.ts:60-61`:

```ts
const crossSlotA = `{touch-cross-a:${randomKey()}}:key`
const crossSlotB = `{touch-cross-b:${randomKey()}}:key`
```

Both tags are random, so there is a ~1/16384 chance per run that they hash to the same slot and `assert.rejects(... CROSSSLOT ...)` fails. Monte-Carlo over 200k draws of the actual templates: 0.00006 and 0.00005 collision rates. Also at `tests-integration/node-redis/key/commands.test.ts:93`, `tests-integration/ioredis/set/setops.test.ts:104`, `tests-integration/node-redis/set/setops.test.ts:98`. Pre-existing, but this PR is the natural place to pin them — the sibling probes that use fixed literals (`{mget-slot-a}key` / `{mget-slot-b}key`, `watchkey3` / `watchkey4`) are the right pattern and verified to hash apart (15092/2711 and 7186/11509).

**2. A sweep miss.** `tests-integration/node-redis/watch.test.ts:74-75` still uses the fixed literals `watch:{multi}:3` / `watch:{multi}:4` while `watchkey` / `watchkey2` in the same file were namespaced. Not a live flake — the test writes both values itself — but it's the file the new rule was applied to, so the inconsistency will read as "the rule is optional".

scripts/flush-redis.ts
```diff
@@ -0,0 +1,319 @@
...
+    }
+  })
+
+  try {
+    // Masters first: a replica can only be verified once whatever feeds it has
+    // been emptied, so these two steps cannot overlap.
+    const masters = targets.filter(t => t.replication.role === 'master')
+    await forEachTarget(masters, flushMaster)
+
+    const flushedPorts = new Set(
```
fatal10110, line 440: **MEDIUM (new, round 2) — a well-formed *subset* of a healthy cluster still gets a partial flush that exits 0.**

The new parser correctly rejects malformed lists, but nothing checks that the configured ports cover the whole cluster. The harness only uses `REDIS_CLUSTER_PORTS` as **seed nodes**: ioredis `new Cluster(...)` (`test-config.ts:99`) and node-redis `rootNodes` (`:146`) both discover the full topology from a single seed. The flush script, though, only flushes the masters in the list. So with one seed, the suite runs against every master while `clean:redis` emptied one of them.

Reproduced on a private, healthy 6-node cluster (native redis 7.2.1, ports 7101-7106):

```
before: 7101=12 7102=10 7103=18
$ REDIS_CLUSTER_PORTS=7101 node --import tsx ./scripts/flush-redis.ts
  cluster node 127.0.0.1:7101: flushed, cluster_state:ok
EXIT=0
after:  7101=0 7102=10 7103=18
```

28 keys are left on two masters the harness will use, and the script reports success. That is the #395 failure class again, reached through a well-formed value this time. It's also realistic: listing a single seed is normal usage for cluster clients, and the harness accepts it without complaint. It also contradicts the `AGENTS.md` claim this round adds ("shared with the harness so the two cannot disagree about which nodes exist"). Today they agree on the *list*, but not on the *topology* that list resolves to.

Suggested fix: after connecting, read `CLUSTER NODES` (or `CLUSTER SHARDS`) from any reachable cluster endpoint and fail loudly, naming the missing ports, if any master in the topology isn't in `flushedPorts`. The alternative is to flush the masters discovered from the topology. Either way it's about ten lines, and it makes "every endpoint is demonstrably empty" true for the cluster the suite will actually talk to.

tests-integration/redis-endpoints.ts
```diff
@@ -0,0 +1,66 @@
...
+function invalidPort(name: string, entry: string, raw: string): Error {
+  return new Error(
+    `${name} entry "${entry}" is not a TCP port between ${MIN_PORT} and ${MAX_PORT} ` +
+      `(got "${raw}"). Pass a comma-separated list of individual ports — ranges ` +
+      `such as "30000-30005" are not supported, spell every port out.`,
+  )
+}
+
+/** Parse one port-valued env var, rejecting anything that is not a TCP port. */
+export function parsePort(name: string, raw: string): number {
```
fatal10110, line 27: **LOW (new, round 2) — the dedupe covers cluster ports but not the standalone ports.**

`scripts/flush-redis.ts` now validates `REDIS_STANDALONE_PORT` / `REDIS_STANDALONE_AUTH_PORT` with `parsePort`, but the harness still reads them with a bare `Number(configuredPort)` in `startRealStandalone()` / `startRealStandaloneAuth()` (`test-config.ts`, outside this diff). For a value like `6399x` the flush throws a clear error while the harness would connect to `NaN`. Both fail, so this isn't a silent-success hole. But the module docstring's promise ("so the two can never disagree about which endpoints exist") only holds for the cluster half.

Suggest exporting a `realStandalonePort()` / `realStandaloneAuthPort()` pair from here and using it in both places, the same way `realClusterPorts()` is shared.

fatal10110:
## Re-review (round 2) of `678d994`

**Verdict: not yet, but close.** One medium finding stands between this and ship (**N1**, a small fix). Both HIGHs from round 1 are fixed, and I reproduced the fixes live rather than taking the description's word for them. Everything else is either fixed or non-blocking.

### How I verified it
Everything ran on my own native Redis 7.2.1 instances, on ports I checked were free first: a private 6-node cluster on 7101-7106, plus a master/replica pair on 7201/7202. Nothing touched the shared 6399 / 6400 / 30000-30005 backends. I ran the new `scripts/flush-redis.ts` from an isolated worktree.

- **Healthy cluster.** All 6 nodes report `flushed` / `replica of …, empty` / `cluster_state:ok`, exit 0.
- **Stale replica.** Replica with `master_link_status:down` holding 3 keys: `replica of 127.0.0.1:7201 still holds 3 key(s) after its master was flushed (master_link_status:down) — its keyspace is stale, not clean`, exit 1.
- **Replicas-only list** (`7104,7105,7106`): all three fail with `replica of …, which this run did not flush`, exit 1.
- **Cluster down.** I downed master 7103 and its replica 7106. The two dead nodes are reported unreachable and every surviving node fails with `cluster_state:fail — …would get -CLUSTERDOWN`, exit 1. Same result when only the surviving nodes are listed.
- **Parser.** `"30000,30001-30005"`, `"30000,abc,30002"`, `"99999"`, `"30000-30005"` and `"30000,,30001"` all throw now.
- **Tests.**
  - Mock backend: 125/125 across the 14 affected suites.
  - Real backend (my private cluster), 10 affected suites, **run twice with no flush in between**: 87/90 both times, with an identical failure set. The 3 failures are version artifacts of my 7.2.1 server, not this PR: `HSCAN NOVALUES` returns `ERR syntax error` (NOVALUES arrived in 7.4), and `scan COUNT errors match Redis` hits COUNT-0 handling that differs by version. The restored `iterations > 1` test and every `assertKeyCount` call site pass on real Redis in both runs, which also confirms the `${PREFIX}${tag}:*` KEYS pattern matches ioredis's `keyPrefix` on the wire.
  - `assertKeyCount` is non-vacuous. Injecting one extra `${tag}:leaked_extra` key fails with `expected exactly 2 key(s) matching key-integration{monitor:hpsmipkn}:*`.

### Round-1 findings
|---|-----|---------|--------|
| 1 | HIGH | `-READONLY` branch returned success without verifying | ✅ **Fixed.** Roles are read via `INFO replication`, masters are flushed first, and each replica must belong to a flushed master *and* reach DBSIZE 0. Verified live (stale-replica and replicas-only cases above) |
| 2 | HIGH | Keyless FLUSHALL/DBSIZE pass on a `fail` cluster | ✅ **Fixed.** `cluster_state:ok` is required on every cluster endpoint. Verified live |
| 3 | MED-HIGH | Parser silently dropped bad entries | ✅ **Fixed.** Every entry is validated and range-checked. *Residual:* a well-formed **subset** still flushes partially, see **N1** |
| 4 | MED-HIGH | Parser duplicated in harness | ✅ **Fixed** via `redis-endpoints.ts`. Partial for standalone ports, see **N2** |
| 5/6 | MED | SCAN paging coverage lost | ✅ **Fixed** in both twins. `iterations > 1` is restored and the tautological KEYS check is removed. Passes on real Redis |
| 7 | MED | `DBSIZE >= expected` vacuous; extra keys undetected | ✅ **Fixed.** Now an exact KEYS count over the suite's own tag, and the dead `*TotalDbSize` helpers are gone. Verified by injection |
| 8/9 | MED | Unawaited `set()` before EVAL | ✅ **Fixed** in both |
| 10 | LOW | `allowReadonly` unconditional | ✅ **Fixed.** Role detection subsumes it |
| 11 | LOW | Invented `REDIS_STANDALONE_AUTH_PASSWORD` | ✅ **Fixed.** The env var is gone and the constant is shared. *Nit:* every failure still ends with "Start the test backends first: docker compose … up", which is the wrong remedy for the new stale-replica and `cluster_state` failures, where the backends *are* up |
| 12 | LOW | `raw:real` skipped `clean:redis`; no DBSIZE-0-at-start check | ✅ `raw:real` **fixed**. DBSIZE-0-at-start: **accepted as a deliberate decision** (below). *Nit:* the `utils.ts:78` docstring still cites the flush-async-sync suites as exact-DBSIZE coverage. They flush a standalone they own, so the citation doesn't say what it implies |
| 13 | LOW | Random CROSSSLOT probes (1/16384) + sweep miss | ✅ Probes **fixed** via `keyInAnotherSlot` at all 4 sites (all route through un-prefixed direct connections, so the slot comparison is on the actual wire keys). ⚠️ `node-redis/watch.test.ts:74-75` `watch:{multi}:3/4` sweep miss **not addressed**. Non-blocking |

### Adjudication: 9b (DBSIZE 0 at suite start)
**Accept.** The argument holds: #420 requires the suite to pass twice with no flush in between, so run 2 starts non-empty by design, and a suite-level `DBSIZE 0` assertion would fail it by construction. It would also reintroduce the "depends on a key being absent at start" coupling #420 removes. Checking in `clean:redis` is the right place, and it's stronger: every endpoint, replicas included, plus `cluster_state:ok`. One caveat ties back to N1: "every endpoint" currently means every *listed* endpoint. Once N1 is fixed, the claim covers the whole topology the suite uses.

| N1 | MED | `scripts/flush-redis.ts:271` | `REDIS_CLUSTER_PORTS=7101` on a healthy cluster: 1 master flushed, **28 keys left on the other two, exit 0**. The harness uses the list as *seeds* and discovers every master, so a single seed (normal cluster-client usage) gives a silent partial flush, which is the #395 class. Fix: compare `CLUSTER NODES` masters against `flushedPorts` and fail on any gap |
| N2 | LOW | `tests-integration/redis-endpoints.ts:31` | `parsePort` validates standalone ports in the script, but the harness still uses bare `Number(...)`, so the dedupe only covers the cluster half |

### PR description
- **The 4 pass-2 failures are framed clearly enough** in the round-2 section. "Read on before treating 'pass 2: 4 failures' as the criterion failing", the #411 attribution, and the 19/19 low-load rerun on the same unflushed keyspace make the case. I could not independently rerun those four suites: HEXPIRE/HPERSIST need Redis ≥7.4 and my private cluster is 7.2.1. So that part is the author's evidence, not mine. The rest of my twice-without-flush run is consistent with it.
- **The top half is now stale and contradicts the round-2 section**, and a reader who stops early will be misled:
  - "replicas are recognised by their `-READONLY` refusal and skipped" is no longer true.
  - The iterations assertions "are gone … cross-check … single-shot `KEYS`" has been reversed.
  - The original *Test plan* says "pass 1: 1 failure / pass 2: 0 failures", while round 2 says "pass 1: 0 / pass 2: 4". Both are consistent with #411 being load-dependent flakiness, but presented side by side without comment they read as a contradiction. It's worth one sentence saying #411 moves between passes, which is itself the evidence that it's timing and not keyspace state.

  Suggest rewriting the Summary and Test plan to describe the current code, and keeping round 2 as a changelog.

scripts/flush-redis.ts
```diff
@@ -0,0 +1,497 @@
...
+      }
+    } catch (err) {
+      seed.failure = {
+        message: `cannot read CLUSTER NODES: ${describe(err)}`,
+        hint: HINT.notCluster,
+      }
+    }
+  }
+
+  const unlisted = [...topology.values()].filter(node => !seedIds.has(node.id))
```
fatal10110, line 303: **LOW (new, round 3) — a transient `handshake` entry fails `clean:redis` on a healthy cluster, and whether it fails depends on which seed you list.**

`unlisted` is every entry any seed reports in `CLUSTER NODES`, including nodes in `handshake`. Those aren't cluster members yet: they own no slots and don't appear in `CLUSTER SLOTS`, so the harness's clients never route to them. If the handshake target isn't reachable, the entry is still opened, fails, and turns the run red.

Reproduced on a private, healthy 6-node cluster (native 7.2.1, ports 7321-7326). I ran `CLUSTER MEET 127.0.0.1 7399` on 7321, with nothing listening on 7399:

```
7321's view:  14a920e4…  127.0.0.1:7399@17399 handshake

$ REDIS_CLUSTER_PORTS=7321 ./scripts/flush-redis.ts
clean:redis failed — 1 of 7 endpoint(s) are not verifiably empty:
  cluster node 127.0.0.1:7399 (discovered): Connection is closed.
  - The cluster itself is unhealthy: …
EXIT=1

$ REDIS_CLUSTER_PORTS=7322 ./scripts/flush-redis.ts     # same cluster, same moment
EXIT=0
```

This is loud rather than silent, so it isn't the #395 class. It's a false positive, and a seed-dependent one, which makes a red `clean:redis` harder to trust. A stray `CLUSTER MEET` or a node that is mid-join is enough to trigger it.

Suggested fix: skip entries flagged `handshake`. More generally, derive the must-flush set from the topology the harness actually uses: every master that owns slots per `CLUSTER SHARDS` / `CLUSTER SLOTS`, plus their replicas. Keep failing hard if one of *those* is unreachable. The noaddr/port-0 handling below could follow the same rule: fail only when the node owns slots or is a replica of a flushed master.

scripts/flush-redis.ts
```diff
@@ -0,0 +1,497 @@
...
+  flushedPorts: ReadonlySet<number>,
+): Promise<string> {
+  const replication = target.replication
+  if (replication.role !== 'replica') {
+    throw new Error('not a replica')
+  }
+
+  if (!flushedPorts.has(replication.masterPort)) {
+    throw new FlushFailure(
+      `replica of port ${replication.masterPort}, which this run did not flush — ` +
```
fatal10110, line 386: **LOW (new, round 3) — when a master fails its post-FLUSHALL check, its replicas are blamed with the wrong message and the wrong hint.**

`flushedPorts` only holds masters that passed `flushMaster`. If a master's `FLUSHALL` ran but `DBSIZE` then found keys, the master is excluded, and every replica of it gets this message plus the stale-replica hint. The message says the run "did not flush" a master it did flush, and the hint ("wait for it to resync, or restart it") points away from the actual cause.

Reproduced with a writer pipelining `SET`s into a slot owned by 7321 while the script ran:

```
cluster node 127.0.0.1:7321: FLUSHALL left 400 key(s) behind
cluster node 127.0.0.1:7324 (discovered): replica of port 7321, which this run did not flush — nothing emptied it, and it refuses FLUSHALL itself
What to do:
  - Keys survived FLUSHALL: something is writing to this node concurrently — …
  - A replica is not following a flushed master: wait for it to resync, or restart it …
```

The `leftKeys` line itself works as intended; that's the one hint the description says wasn't tested, and it fires correctly. Suggest distinguishing "master was never flushed / not in topology" from "master was flushed but failed verification", e.g. by keeping a second set of *attempted* ports, and giving the latter a message like `replica of port 7321, whose master failed its own flush check` with no extra hint. The master's line already carries the right one.

fatal10110:
## Re-review (round 3) of `1e17eed`

**Verdict: ship.** N1 (the silent partial flush) is fixed and holds up under every edge case I could build. N2 and the round-2 nits are fixed too. The two new findings are both LOW. Each produces a loud, misleading failure; neither produces a silent success. They're fine as follow-ups.

### How I verified it
I used a private native Redis 7.2.1 cluster on 7321-7326, plus a standalone on 7421, all on ports I checked were free first. I ran everything from my own worktree. I didn't touch the shared fixtures (6399 / 6400 / 30000-30005); they are still `PONG` / `cluster_state:ok`. I stopped all the private servers afterwards.

### Round-2 findings
| # | Finding | Status |
|---|---------|--------|
| N1 | One-port `REDIS_CLUSTER_PORTS` → partial flush, exit 0 | ✅ **Fixed.** The script now discovers the whole topology. See the edge-case table |
| N2 | Standalone ports parsed with a bare `Number()` in the harness | ✅ **Fixed.** `realStandalonePort()` / `realStandaloneAuthPort()` are shared by the harness and the script. `abc`, `99999`, `7811x` and `-1` all throw; blank means unset |
| nit | "Start the backends" printed for every failure | ✅ **Fixed.** Hints are failure-specific and printed once. I triggered each one: unreachable, unhealthy cluster, stale replica, non-cluster seed, and `leftKeys` (see the note on line 386) |
| nit | `watch:{multi}:3/4` fixed keys | ✅ **Fixed** in both clients, with `RUN` inside the braces |
| nit | "flush-async-sync suites" cited as exact-DBSIZE coverage | ⚪ Still in the `utils.ts` docstring and the description. Cosmetic |

### N1 edge cases, all on the private cluster
| Case | Result |
|---|---|
| Single **master** seed (`7321`) | Discovers the other 5 nodes, flushes 3 masters, verifies 3 replicas, every node ends at `DBSIZE 0`. Exit 0 ✅ |
| Single **replica** seed (`7326`) | Whole cluster flushed and verified. Exit 0 ✅ |
| Replica seed whose **master is down** (failover disabled) | Master reported unreachable, replica reported "did not flush", 4 nodes `cluster_state:fail`. Exit 1 ✅ |
| **Unlisted replica down**, `cluster_state` still `ok` | Reported unreachable. Exit 1 ✅ (conservative; I agree with failing here) |
| **Hostname announce** (7322 with `cluster-announce-hostname localhost`) and **`0.0.0.0` announce** (7323) | The `,hostname` suffix is parsed correctly. 7323 is opened **once** when it's also a seed (dedupe by node id), and it's reachable at `0.0.0.0` when discovered. Exit 0 ✅ |
| **Mixed list** (cluster `7321` + standalone `7421`) | Whole cluster flushed; the standalone fails with the "non-cluster seed" hint and is **not** flushed. Exit 1 ✅ |
| **Gossip not converged**: a `handshake` entry visible to one seed only | ⚠️ Seed 7321 fails (exit 1), seed 7322 passes (exit 0), on the same healthy cluster. False positive, loud not silent. **New finding, line 303** |
| **Concurrent writer** during flush | `FLUSHALL left 400 key(s) behind` with the correct `leftKeys` hint. Exit 1 ✅. The replica cascade message is wrong, though. **New finding, line 386** |
| `noaddr` | Not reproducible on demand. By inspection, `noaddr` or a port ≤ 0 is reported as a failure, not skipped. It's the same over-strictness as the handshake case; the note on line 303 covers both |

### Tests
- **Mock, full suite, twice:** 1181/1181 both times at load average ~27.
- **Real backend:** all 34 touched test files (39 suites, 272 tests) against the private cluster through a **single seed**, twice with no flush in between. Both passes gave 265/272 with an **identical** failure set. All 7 failures come from my test server being 7.2.1 rather than the repo's 8.0 target:
  - `HSCAN NOVALUES` returns `ERR syntax error`; NOVALUES arrived in 7.4.
  - `scan COUNT errors match Redis` hits COUNT-0 handling that differs by version.
  - `INCRBYFLOAT` / `HINCRBYFLOAT` return `3.79999999999999982` instead of `3.8`: 7.2's long-double formatting, reproduced directly with `redis-cli`. The PR never touches those assertions.

  The rewritten watch tests and every `assertKeyCount` site pass on real Redis. Afterwards, `clean:redis` via the single seed took the leftovers to `DBSIZE 0` on all six nodes.

### The unexplained intermittent mock failures
**Acceptable.** The PR has no `src/` changes, so the mock backend's behaviour is identical to `main`. Round 3's only mock-visible change is the two watch-test key strings, which are deterministic. The flush script and the standalone-port parsing never run under `TEST_BACKEND=mock`. The reruns support that: three at load 72-95, plus my two at ~27, were all 1181/1181. At that load, 4 failures that came and went point to the suite's existing timing-sensitive tests (the #411 class), not to this PR. The description is honest that the failures weren't captured, and it doesn't overclaim.

### PR description
It now reads as **one consistent account of the current code**, and I found nothing in it that contradicts the code:
- The topology-discovery, replica, cluster-health and hint claims all match `scripts/flush-redis.ts` and my repros.
- The round-1/round-2 table reconciles the old contradiction (1+0 vs 0+4 failures, all 5 traced to the two #411 tests), and "the failure moving between passes is itself the signature of a timing flake" frames it correctly.
- The 9b decision and the 19/19 evidence are kept. I still accept 9b, and with N1 fixed, "DBSIZE 0 on **every** node of the discovered topology" is now literally true.

The only stale phrase left is the flush-async-sync citation, which is cosmetic. I couldn't rerun the #411 hash-TTL suites myself (they need Redis ≥7.4), so that part is the author's evidence, which is consistent with everything I did observe.

### New findings (inline)
| R3-1 | LOW | `scripts/flush-redis.ts:303` | `handshake` (and, by inspection, `noaddr`) entries are treated as must-flush nodes. On a healthy cluster, a transient handshake fails the run, and the outcome depends on which seed is listed. Suggest deriving the must-flush set from slot-owning masters (`CLUSTER SHARDS`/`SLOTS`) plus their replicas |
| R3-2 | LOW | `scripts/flush-redis.ts:386` | A master that fails its own post-flush check makes its replicas report "which this run did not flush" with the stale-replica hint. Both the message and the hint point away from the real cause |

Neither blocks. Both are cases where `clean:redis` fails loudly with a confusing message, which is the right failure direction for this script.

Issue #419, opened 2026-09-22
CONFIG SET: gate multi-pair arity on the 7.0 boundary, then add duplicate-parameter detection


fatal10110:
Deferred from PR #409 with verified reasoning. Two layers, and they must land in this order.

## Layer 1 — multi-pair `CONFIG SET` is not arity-gated
Accepting more than one parameter pair is part of the **7.0 rewrite**. Real 6.2 rejects the whole form:

```
CONFIG SET proto-max-bulk-len 2mb maxmemory 100mb
  REAL 6.2.14  ERR Unknown subcommand or wrong number of arguments for 'SET'. Try CONFIG HELP.
  REAL 7.0.15  OK
  MOCK (redis-6.2 profile)  OK        <- diverges
```

So under the `redis-6.2` profile this server accepts a form real 6.2 refuses. Pre-existing and ungated.

## Layer 2 — no duplicate-parameter detection
```
CONFIG SET proto-max-bulk-len 2mb proto-max-bulk-len 3mb
  REAL 7.2 / 8.0  ERR ... - duplicate parameter
  MOCK            OK
```

## Why the order matters
Adding duplicate detection first would emit a 7.0-only message under the `redis-6.2` profile — the exact class of ungated divergence that PR #409's review was about. The duplicate error cannot exist on a profile where the multi-pair form itself does not exist.

So: gate `CONFIG SET`'s arity on the 7.0 boundary first (the `config.set.failure-message` FeatureId added in PR #409 covers the same boundary and may be reusable, or a sibling `config.set.multi-pair` may be cleaner), then layer duplicate detection on top of the 7.0+ path only.

## Acceptance
- Multi-pair `CONFIG SET` is rejected on the `redis-6.2` profile with real 6.2's wording, and accepted on 7.0+.
- Duplicate parameters produce real Redis' `duplicate parameter` error on 7.0+ profiles only.
- Both pinned by raw-TCP tests, verified against real 6.2 and a 7.x/8.x server.

PR #430, opened 2026-09-22
refactor: one profile-aware unknown-subcommand error helper (#413)

### LINKED ISSUE #413: refactor: 15 hand-rolled copies of the unknown-subcommand error; three wordings live, none version-gated
Found while reviewing PR #410. Closely related to #384 — most of the mechanism that issue needs is this refactor.

## Problem
`git grep "nknown subcommand" -- src/commands/` returns **15 hand-rolled copies** of the same error template, across `command.ts`, `config.ts`, `connection.ts`, `pubsub.ts`, `scripts.ts`, `streams/xgroup.ts`, `streams/xinfo.ts`, plus two error classes in `core/redis-error.ts`.

Because each site is independent, **three different wordings are live at once**, and each has drifted separately from real Redis. Verified against real servers:

| | real | mock (PR #410 head) |
|---|---|---|
| `CONFIG BOGUS` on 7.0+ | `ERR unknown subcommand 'BOGUS'. Try CONFIG HELP.` | matches (fixed by #410) |
| `CONFIG BOGUS` on 6.2 | `ERR Unknown subcommand or wrong number of arguments for 'BOGUS'. Try CONFIG HELP.` | **wrong** — emits the 7.0+ text |
| `XGROUP BOGUS` on 7.2/8.0 | `ERR unknown subcommand 'BOGUS'. Try XGROUP HELP.` | **wrong** — emits the pre-7.0 text |

Three separate defects, one cause:

1. **The wording is version-specific and ungated.** Real Redis changed the template at **7.0.0** (confirmed: 6.2.24 uses the capital-U form, 7.0.15 already uses the new one). This repo ships a `redis-6.2` profile and sweeps it in CI via `test:integration:compatibility:mock`, so every one of these 15 sites is wrong under that profile. The gate precedent already exists — `client.setinfo.unknown-subcommand-error: { redis: '7.0.0', valkey: '7.2.0' }`, consumed by `clientSetInfoUnavailableError` in `src/commands/connection.ts:716-726`.
2. **No `%.128s` truncation.** Real Redis truncates the echoed subcommand at 128 bytes. Boundary verified exact on 7.2.1 and 8.0.6: at length 128 mock and real agree, at 129 they diverge (real echoes 128 `X`, mock echoes all 129). With a 300-char name the mock reflects all 300.
3. **Non-UTF-8 bytes are mangled.** The schema decodes the subcommand token with `toString()` (UTF-8), so raw bytes become U+FFFD:
   ```
   real 8.0.6 : -ERR unknown subcommand '\xff\xfe\xfd'. Try CONFIG HELP.
   mock       : -ERR unknown subcommand '\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd'. Try CONFIG HELP.
   ```

## Proposal
One profile-aware helper — `unknownSubcommandError(container, subcommand, profile)` — that owns the template, the version gate, the 128-byte truncation and binary-safe echoing, replacing all 15 sites. Keep the subcommand as a `Buffer` through to formatting so (3) falls out for free.

## Relationship to #384
#384 part 2 is the same echo/truncation/binary-byte model, filed against `UnknownRedisCommandError`. #384 part 1 (`plan.definition.name` cannot express `config|get` in the subscribed-mode error) is a different, policy-layer problem — but once this helper exists, the formatting half of #384 is already solved. Worth doing this first, then #384.

## Acceptance
- One helper; `git grep "nknown subcommand" -- src/commands/` returns a single definition site.
- `CONFIG BOGUS` and `XGROUP BOGUS` match real Redis byte-for-byte on both the `redis-6.2` and `redis-8.0` profiles, pinned by raw-TCP tests (where exact wire text belongs per AGENTS.md).
- A 129-byte subcommand and a non-UTF-8 subcommand are both pinned.

fatal10110:
## Summary

`git grep "nknown subcommand" -- src/commands/` returned 15 hand-rolled copies of one error template across `command.ts`, `config.ts`, `connection.ts`, `pubsub.ts`, `scripts.ts`, `streams/xgroup.ts`, `streams/xinfo.ts` and two error classes in `core/redis-error.ts`. Three wordings were live at once and each had drifted separately. All of them now route through **one helper**, and the three defects the duplication was hiding are fixed.

Re-verified against real `redis-server` 6.2.24, 7.0.15 and 8.0.6 before touching any code — the issue's tables hold up:

| | 6.2.24 | 7.0.15 / 8.0.6 |
|---|---|---|
| `CONFIG BOGUS` | `ERR Unknown subcommand or wrong number of arguments for 'BOGUS'. Try CONFIG HELP.` | `ERR unknown subcommand 'BOGUS'. Try CONFIG HELP.` |
| `PUBSUB CHANNELS a b` | `ERR Unknown subcommand or wrong number of arguments for 'CHANNELS'. Try PUBSUB HELP.` | `ERR unknown subcommand or wrong number of arguments for 'CHANNELS'. Try PUBSUB HELP.` |
| 300-byte name | echoed whole | cut to 128 bytes |
| `CONFIG \xff\xfe\xfd` | those three bytes | those three bytes |

### One helper

`unknownSubcommandError(container, subcommand, profile)` in `src/commands/helpers.ts` owns the template, the version gate, the 128-byte cut and the binary-safe echo. All 15 sites call it; `git grep "nknown subcommand" -- src/commands/` now returns only `helpers.ts`.

The sweep turned up a **second** real template that the grep also covered, so it is handled here rather than left as a fourth stray copy: real Redis' `addReplySubcommandSyntaxError`, raised by a container itself when a *known* subcommand gets arguments it cannot use (`PUBSUB CHANNELS a b`). It keeps the `or wrong number of arguments` clause and is **not** truncated, but it took the same 7.0 case flip. It is the sibling `subcommandSyntaxError(...)`, sharing one private formatter — on 6.2 the two templates are literally the same string, because before container commands entered the command table every one of these replies came out of `addReplySubcommandSyntaxError`.

### 1. Version gate

Gated on the existing `error.unknown-subcommand-wording` FeatureId (`{ redis: '7.0.0', valkey: '7.2.0' }`), following the `client.setinfo.unknown-subcommand-error` precedent. That precedent gate is now **removed**: its versions were identical to the general one and its only consumer, `clientSetInfoUnavailableError`, is gone.

Every profile-gated subcommand (`COMMAND DOCS`/`GETKEYSANDFLAGS`, `CLIENT SETINFO`/`NO-EVICT`, `ACL DRYRUN`, `PUBSUB SHARDCHANNELS`/`SHARDNUMSUB`) now goes through the same helper, so "this server does not have that subcommand yet" and "no such subcommand" produce the reply that real Redis of that vintage produces — including the case the old code got right only by accident, `CLIENT SETINFO` on `redis-7.0`, where the feature gate (7.2) and the wording gate (7.0) disagree.

### 2. `%.128s` truncation

The cut is taken on the Buffer, so it is 128 **bytes**, not characters. `'A'.repeat(127) + 'é'` is 129 bytes and real 8.0.6 replies with 128 of them — the last a bare `0xc3`, the lead byte of a sequence whose continuation byte was cut away. 6.2 does not truncate at all.

### 3. Binary-safe echo

The schemas decoded the subcommand token with `toString()`, so raw bytes became U+FFFD — three bytes on the wire where real Redis sends one. Every container's `subcommand` field is now `t.bulk()`, and the raw Buffer reaches the formatter untouched (this also fixes XGROUP/XINFO echoing an upper-cased name instead of the client's casing).

Getting bytes onto the wire needed a path that did not exist — `RedisCommandError` carries a `string` and `resp-encoder.ts` formatted strings. The smallest one that works:

- `RedisCommandError` accepts a `Buffer` message and keeps it in `readonly messageBytes?: Buffer` alongside the lossy `Error.message`, read through a new `errorReplyBody(error)`.
- `RedisValue`'s error variant gains an optional `messageBytes?: Buffer`; `message` stays a `string`, so every existing reader (`redisErrorText`, `decodeRedisValue`, the client mocks, the unit tests) is untouched.
- `resp-encoder.ts` builds error frames as bytes (`encodeError` + a byte-wise `sanitizeErrorBytes`, the 1:1 `\r`/`\n` map from #388 applied to bytes rather than characters).
- The four sites that turn a caught `RedisCommandError` into a reply call `errorReplyBody(err)`.
- Lua sees the raw bytes too — `redis.call('CONFIG', <raw bytes>)` now hands the script the same `err` a client would get.

This is deliberately *not* #364: `redis-error.ts` gains two lines on the base class and loses the two unknown-subcommand subclasses, nothing else.

### Breaking change

`UnknownScriptSubcommandError` and `UnknownClusterSubcommandError` are removed from `src/core/redis-error.ts` and from the root export in `src/index.ts`. They hard-coded the 7.0+ wording with no profile in reach, which is precisely the bug; keeping them would perpetuate it. No doc referenced them (`docs/API.md`'s exported-error example names only `RedisCommandError`).

### Out of scope, confirmed untouched

- **#384.** The `UnknownRedisCommandError` (*command*-name) echo is unchanged; this PR only supplies the helper #384 part 2 can reuse.
- **`XGROUP CREATE` / `XINFO STREAM` arity naming.** Real 7.0+ says `wrong number of arguments for 'xgroup|create' command`; this server says `'xgroup'`, because the check lives in a schema parser that only knows `ctx.commandName`. That is #384 part 1; the raw-TCP test says so in a comment and asserts only `CONFIG GET`, which already matches.
- **#419** (CONFIG SET arity gating) — `config.ts` keeps its own `configSetFailed`; the helper's signature is `(container, subcommand, profile)` and stable.
- `src/commands/keys.ts` and `src/commands/bitmaps.ts` are not touched.

Closes #413

## Test plan

- [x] **`tests-integration/raw-tcp/command-errors-subcommand.test.ts`** (new, 8 tests) — exact reply bytes on the default `redis-8.0` profile: one shared template across all ten containers, casing preserved, trailing args ignored, truncation at 1/127/128/129/300, the 128-byte cut landing inside a multi-byte character, a non-UTF-8 name echoed byte for byte, the PUBSUB syntax-error wording, and `CONFIG GET`'s arity reply as a guard that the unknown path did not widen over it. **Red on mock before the fix: 6 of 8 failed.** Green against real 8.0 from the start.
- [x] **`tests-integration/compatibility/profile-gates.test.ts`** — the two CONFIG-only wording tests are generalized to five containers and joined by two new ones (the syntax-error wording, and the non-UTF-8 echo on every profile). Red on `redis-6.2` (3 fail) and on `redis-8.0` (3 fail) before the fix. This is where `CONFIG BOGUS` / `XGROUP BOGUS` get pinned byte-for-byte on **both** the `redis-6.2` and `redis-8.0` profiles.
- [x] `tests-integration/raw-tcp/helpers.ts` / `tests/shared-test-helpers.ts` — `commandFrame` and `expectReply` accept Buffers, so a test can put bytes on the wire that are not valid UTF-8 and diff the reply by byte.
- [x] Unit-test expectations corrected, not accommodated: `tests/compatibility/behavior.test.ts` expected the 7.0+ SLOWLOG wording from a `redis-6.2` session — that expectation was the bug. The neighbouring 6.2 assertions for `COMMAND DOCS`, `PUBSUB SHARDCHANNELS`, `CLIENT SETINFO`/`NO-EVICT` and `ACL DRYRUN` were already correct and still pass unchanged.
- [x] `npm test` — 378 pass, 0 fail.
- [x] `npm run test:integration:mock` — 1191 pass, 0 fail.
- [x] `npm run test:integration:compatibility:mock` — all 7 profiles, 18 pass / 0 fail each.
- [x] `npm run test:integration:real` — 1069 pass, 8 fail, 15 cancelled on a real backend shared with ~9 other agents. Every failure re-ran green in isolation (`node-redis` set/scan/keys/proto-max-bulk-len, `list/blocking`, `ioredis` string/watch/zset, cluster `READONLY` replica routing — the last is the known flake called out in #410), confirming contention rather than regression. The change-adjacent suites (all of `raw-tcp/`, both `command-integration`, both `cluster-integration`, `ioredis/scripts`) run 177 pass / 0 fail against real Redis.
- [x] `npm run build`, `npm run lint` — clean.
- [x] Docs: `docs/API.md` gains a "Current profile gates" row for the unknown-subcommand wording. `README.md`, `docs/COMMANDS.md` and `docs/ARCHITECTURE.md` checked and unaffected — none quotes these error strings or names the removed classes, and the byte path is internal to the existing encode pipeline. `docs/COMPATIBILITY-PROFILES-PLAN.md` still names the removed FeatureId but reads as a superseded design record (its example gate table lists 8 of the current 24 entries), so it was left alone.


src/commands/helpers.ts
```diff
@@ -164,34 +159,89 @@ const SUBCOMMAND_ECHO_LIMIT = 128
...
+ * three bytes, where a UTF-8 decode would turn each into U+FFFD.
  */
-export function unknownSubcommandError(
+function subcommandError(
   container: string,
   subcommand: Buffer | string,
   profile: CompatibilityProfile,
+  kind: 'unknown' | 'syntax',
 ): RedisCommandError {
   const raw = Buffer.isBuffer(subcommand) ? subcommand : Buffer.from(subcommand)
```
fatal10110, line 209: **HIGH — the echoed subcommand is not cut at the first NUL byte.**

Real Redis formats the echo with `%s` on a C string, so it stops at the first NUL — and that cut happens *before* the `%.128s` cut. This helper emits the NUL and everything after it, raw, inside a `-ERR ...\r\n` simple-error frame: a body real Redis cannot produce.

Ground-truthed on real 8.0.6 vs this PR's mock over raw RESP sockets:

```
CONFIG AA\x00BB
  real 8.0.6 : -ERR unknown subcommand 'AA'. Try CONFIG HELP.
  PR mock    : -ERR unknown subcommand 'AA\x00BB'. Try CONFIG HELP.
```

Ordering proof (the NUL cut precedes the 128-byte cut):

```
CONFIG ('A'*100 + NUL + 'A'*100)   real echo_len=100   mock echo_len=128
CONFIG (NUL + 'A'*10)              real echo=''        mock echo_len=11
```

`sanitizeErrorBytes` in `resp-encoder.ts` maps CR/LF (verified correct against real Redis) but leaves `0x00` untouched, so nothing downstream catches this either.

Fix, in `subcommandError` before the `subarray`:

```ts
const nul = raw.indexOf(0)
const cut = nul === -1 ? raw : raw.subarray(0, nul)
```

Apply it on the **6.2 branch too** (line 211) — real 6.2.14 NUL-truncates even though it does not length-truncate.

src/core/lua-runtime.ts
```diff
@@ -339,7 +339,7 @@ function redisValueToLuaReply(value: RedisValue): ReplyValue {
       return null
     case 'error':
       return {
-        err: Buffer.from(value.message),
+        err: value.messageBytes ?? Buffer.from(value.message),
```
fatal10110, line 357: **MEDIUM — the byte-exact error body is re-lossified at the Lua boundary, so the defect this PR exists to fix survives on the Lua path.**

This line correctly carries `messageBytes` *into* the script, but the two return directions in this same file still go through the lossy string:

- `luaReplyToRedisValue` at **line 218**: `RedisValue.error(value.err.toString(), value.code?.toString())` — `value.err` is already a Buffer and gets UTF-8 decoded.
- `redisErrorToLuaReply` at **line 363**: `Buffer.from(err.message)`, never consulting `messageBytes`.

So within one file bytes go in and U+FFFD comes out:

```
EVAL "return redis.pcall('CONFIG', ARGV[1])" 0 "\xff\xfe\xfd"
  wire reply echoes 9 bytes (3x EF BF BD) instead of the 3 the client sent
```

That is exactly the divergence #413 was filed to close — the PR description's claim that "Lua sees the raw bytes too" holds only for the inbound direction.

Second affected path: a parse-time XGROUP/XINFO error reaches Lua via `ctx.executor.plan()` throwing (`scripts.ts:80`, caught at `:86`) and is converted by `redisErrorToLuaReply` at `:363`, losing the bytes before the engine ever sees them.

A third site is unconverted for the same reason: `scripts.ts:492` uses `RedisResult.error(err.message, err.code)` where the other four call sites use `errorReplyBody(err)`.

Neither Lua path has a test — the new raw-TCP suite only exercises direct dispatch. Worth a case there, since this is the headline fix.

Fixes are one-liners: `RedisValue.error(value.err, value.code?.toString())` at 218, `errorReplyBody(err)` at 363 and at `scripts.ts:492`.

src/commands/helpers.ts
```diff
@@ -164,34 +159,89 @@ const SUBCOMMAND_ECHO_LIMIT = 128
  * 8.0.6   CONFIG BOGUS -> unknown subcommand 'BOGUS'. Try CONFIG HELP.
  * ```
  *
- * The subcommand is echoed with the casing the client sent. `container` is the
- * upper-case parent name as it appears in the `Try ... HELP.` suffix.
+ * `container` is the upper-case parent name as it appears in the
+ * `Try ... HELP.` suffix; `subcommand` is echoed with the bytes the client
+ * sent, casing and all.
+ */
+export function unknownSubcommandError(
```
fatal10110, line 166: **MEDIUM — the gate models the 7.0 wording change but not the dispatch-timing change that caused it.**

Redis 7.0 moved container subcommands into the command table. That is one change with two observable effects: the new wording (which this helper now gates correctly) **and** the fact that an unknown container subcommand is rejected at *queue* time inside `MULTI` rather than at `EXEC`. This PR takes the first and leaves the second on the 6.2 model for every profile.

Ground-truthed:

```
MULTI / CONFIG BOGUS / EXEC
  real 7.0.15, 8.0.6 : +OK / -ERR unknown subcommand 'BOGUS'... / -EXECABORT ...
  PR mock 7.0 & 8.0  : +OK / +QUEUED                          / *1 -ERR unknown subcommand 'BOGUS'...
```

A client that treats `EXECABORT` as "nothing ran" sees a one-element array of errors instead, on the profiles that are supposed to be the accurate ones.

Affects every container that dispatches in `execute()`: CONFIG (`config.ts:350`), CLIENT, ACL, SLOWLOG, COMMAND, PUBSUB, SCRIPT, FUNCTION.

Not necessarily this PR's job to fix — but the gate comment in `feature-gates.ts` now describes the 7.0 command-table move as if wording were its only consequence, which will mislead the next reader. Either handle the timing under the same gate or record the known gap here and in a follow-up issue.

See also the mirror-image problem on XGROUP/XINFO, commented separately.

src/commands/streams/xgroup.ts
```diff
@@ -133,9 +136,7 @@ function createXgroupSchema() {
         }
       }
 
-      throw new RedisCommandError(
-        `Unknown subcommand or wrong number of arguments for '${subcommand}'. Try XGROUP HELP.`,
-      )
+      throw unknownSubcommandError('XGROUP', rawSubcommand, ctx.profile)
```
fatal10110, line 139: **MEDIUM — throwing from the schema parser aborts the transaction on `redis-6.2`, where real 6.2 queues.**

This is the mirror image of the CONFIG/CLIENT timing gap. XGROUP and XINFO resolve their subcommand inside `createXgroupSchema()` / `createXinfoSchema()`, which run at **queue** time, so the error lands before `EXEC` on every profile — including `redis-6.2`, where real 6.2 had not yet moved container subcommands into the command table and therefore queues them.

Ground-truthed:

```
MULTI / XGROUP BOGUS / EXEC
  real 6.2.14 : +OK / +QUEUED                                    / *1 -ERR Unknown subcommand or wrong number of arguments for 'BOGUS'. Try XGROUP HELP.
  PR mock 6.2 : +OK / -ERR Unknown subcommand or wrong number... / -EXECABORT Transaction discarded because of previous errors.
```

Same for `xinfo.ts:103`.

So the two families now diverge from real Redis in opposite directions: CONFIG-style containers queue when real 7.0+ aborts, and XGROUP/XINFO abort when real 6.2 queues. Moving the subcommand check out of the schema parser into `execute()` would fix this one and is also the prerequisite for the `xgroup|create` arity naming in #384 part 1.

src/commands/helpers.ts
```diff
@@ -164,34 +159,89 @@ const SUBCOMMAND_ECHO_LIMIT = 128
...
- * Note this covers the *unknown subcommand* path only. On 6.2 the same template
- * also served wrong-arity replies for a known subcommand, where 7.0+ says
- * `wrong number of arguments for 'config|get' command`; that arity path is a
- * separate divergence and is not handled here (#413).
+ * ```
+ * 6.2.24  PUBSUB CHANNELS a b -> Unknown subcommand or wrong number of arguments for 'CHANNELS'. Try PUBSUB HELP.
+ * 8.0.6   PUBSUB CHANNELS a b -> unknown subcommand or wrong number of arguments for 'CHANNELS'. Try PUBSUB HELP.
+ * ```
+ */
+export function subcommandSyntaxError(
```
fatal10110, line 186: **MEDIUM — `subcommandSyntaxError` is wired into PUBSUB only (1 of 11 containers), so the 6.2 profile still emits a wording that did not exist until 7.0.**

Folding `addReplySubcommandSyntaxError` into this helper is the right call, but every other container still answers a known-subcommand-with-bad-arity via `WrongNumberOfArgumentsError`, which produces the 7.0-era `wrong number of arguments for 'config|resetstat' command` form on **all** profiles. Real 6.2 answers every one of these with the subcommand-syntax template instead.

Ground-truthed on real 6.2.14 vs the PR mock on `redis-6.2`:

```
CONFIG RESETSTAT x
  real 6.2.14 : -ERR Unknown subcommand or wrong number of arguments for 'RESETSTAT'. Try CONFIG HELP.
  PR mock 6.2 : -ERR wrong number of arguments for 'config|resetstat' command
```

Same divergence for `SCRIPT LOAD`, `ACL WHOAMI x`, `SLOWLOG LEN x`, `COMMAND COUNT x`, `XINFO STREAM` — and for `PUBSUB NUMPAT x`, which sits three lines from a correct call site at `pubsub.ts:246` and still takes the arity path.

7.0 / 7.2 / 8.0 are correct; only the 6.2 side is wrong. Since `redis-6.2` is swept in CI by `test:integration:compatibility:mock`, this is the same class of defect #413 was filed about — one template applied inconsistently — just on the sibling wording. Either route the remaining containers' max-arg checks through `subcommandSyntaxError`, or scope the gap explicitly in the helper doc and a follow-up issue.

src/commands/streams/xinfo.ts
```diff
@@ -98,9 +100,7 @@ function createXinfoSchema() {
         }
       }
 
-      throw new RedisCommandError(
-        `unknown subcommand '${subcommand}'. Try XINFO HELP.`,
-      )
+      throw unknownSubcommandError('XINFO', rawSubcommand, ctx.profile)
```
fatal10110, line 103: **MEDIUM — XINFO arity errors name `xinfo` instead of `xinfo|<subcommand>`, and this is wrong on every profile.**

The PR description scopes the `container|subcommand` arity naming out as #384 part 1, and for `XGROUP CREATE` that is fair. But the XINFO arity sites (`:95`, and also `:41`, `:47`, `:65`, `:71`, `:83`) use `ctx.commandName` — which is `'xinfo'` — and there is no profile on which that is correct:

```
XINFO STREAM            (missing key)
  real 7.0.15 / 8.0.6 : -ERR wrong number of arguments for 'xinfo|stream' command
  real 6.2.14         : -ERR Unknown subcommand or wrong number of arguments for 'STREAM'. Try XINFO HELP.
  PR mock (all 5)     : -ERR wrong number of arguments for 'xinfo' command
```

So the mock is wrong on 6.2 (should be the subcommand-syntax template — see the separate comment on `helpers.ts:186`) *and* wrong on 7.0+ (should be `xinfo|stream`). Pre-existing rather than introduced here, but this PR is the one consolidating these templates, and the raw-TCP suite's comment says only `CONFIG GET` is asserted "because XGROUP/XINFO reach the same real reply" — which understates it: they do not reach the same reply on any profile.

Minor, same line: the container name is hard-coded as `'XINFO'` here even though `ParseContext` already carries `ctx.commandName`, used two lines up at `:95`. The 13 `execute()`-side literals genuinely have no alternative (`RedisExecutionContext` carries no command name), so the free-form `container: string` parameter is the right signature overall — but these two schema-parser sites could derive it.

src/index.ts
```diff
@@ -106,9 +106,7 @@ export {
   StringExceedsMaxSizeError,
   RedisSyntaxError,
   TransactionDiscardedError,
-  UnknownClusterSubcommandError,
   UnknownRedisCommandError,
```
fatal10110, line 109: **MEDIUM — breaking change to the root facade with no CHANGELOG entry.**

`UnknownScriptSubcommandError` and `UnknownClusterSubcommandError` are removed here, and therefore also from `/core` via `internal.ts:236`. The removal itself is safe — I grepped the whole repo (src, tests, tests-integration, docs, examples, scripts) and there are no dangling references, and the classes really did hard-code the 7.0+ wording with no profile in reach, so deleting them rather than gating them is the right call.

The problem is only the note: `CONTRIBUTING.md:107-112` makes a CHANGELOG entry mandatory **in the same PR** for a change like this, and the `Unreleased` / `Removed` section already carries entries of exactly this kind. The root facade is the curated public surface, so this is a stronger break than a `/core`-only removal and shouldn't land silently.

Add an `Unreleased` / `Removed` entry naming both classes and pointing at `unknownSubcommandError` as the replacement.

src/core/redis-error.ts
```diff
@@ -1,13 +1,33 @@
...
+      this.messageBytes = message
+    }
   }
 }
 
+/**
+ * The reply body to put on the wire for a {@link RedisCommandError} — its raw
+ * bytes when it carries any, otherwise its message.
+ */
+export function errorReplyBody(error: RedisCommandError): string | Buffer {
```
fatal10110, line 29: **LOW (design) — `errorReplyBody` is an opt-in convention with no type-level enforcement, and it has already been missed three times.**

`RedisResult.error(err.message, err.code)` still typechecks and still *looks* correct at a call site, so the lossy form stays reachable. The split is currently maintained by discipline alone:

- Converted: `command-executor.ts:159`, `:242`, `:319`, `session-adapter.ts:160`
- Not converted: `auth-policy.ts:38`, `scripts.ts:492`, `lua-runtime.ts:363`

Of those three, `scripts.ts:492` and `lua-runtime.ts:363` are live byte-loss bugs (see the `lua-runtime.ts` comment); `auth-policy.ts:38` is harmless today only because its messages are static strings.

A `RedisResult.fromError(err: RedisCommandError)` overload would make the lossy form unreachable for a caught `RedisCommandError` and would have caught all three mechanically, rather than leaving the invariant to be re-checked by hand on every future call site. Worth doing while the byte path is fresh — the same reasoning that motivated collapsing 15 hand-rolled templates into one helper applies to the 7 hand-rolled error conversions.

src/commands/scripts.ts
```diff
@@ -226,7 +231,9 @@ export const functionCommand = defineCommand<FunctionArgs>({
   name: 'function',
   since: { redis: '7.0.0', valkey: '7.2.0' },
   schema: t.object({
```
fatal10110, line 233: **LOW (pre-existing, adjacent) — the `since` gate hides FUNCTION on `redis-6.2`, but the resulting unknown-command error uses 7.0+ formatting on that profile.**

Not introduced by this PR, but it is the same class of defect on the line this diff touches, and this PR is the one establishing that gated-off things must also *sound* like the server of that vintage.

```
FUNCTION LIST   on the redis-6.2 profile
  real 6.2.14 : -ERR unknown command `FUNCTION`, with args beginning with: `LIST`,
  PR mock 6.2 : the 7.0+ unknown-command form
```

Real 6.2 uses backticks and a trailing comma-space per argument. This affects every command gated off on `redis-6.2`, not just FUNCTION, so it belongs in `UnknownRedisCommandError` — which is #384 part 2, the issue this PR explicitly sets up the helper for. Worth a line in that issue noting the wording is version-specific too, not only the echo/truncation model.

src/core/resp-encoder.ts
```diff
@@ -224,3 +235,24 @@ function formatNumber(value: number): string {
 function sanitizeErrorText(value: string): string {
   return value.replace(/[\r\n]/g, ' ')
 }
+
+/**
+ * Byte-wise counterpart of {@link sanitizeErrorText}. `\r` and `\n` are
+ * single-byte, so mapping them here is equivalent to mapping the characters —
+ * and it works on bodies that are not valid UTF-8 at all.
+ */
+function sanitizeErrorBytes(value: Buffer): Buffer {
```
fatal10110, line 241: **LOW — two near-identical sanitizers now coexist and can drift.**

After this change `sanitizeErrorText` survives only to sanitize `value.code` at `:201` — a 3-10 character ASCII token like `ERR`, `WRONGTYPE`, `MOVED`, `EXECABORT`. Every actual error *body* goes through `sanitizeErrorBytes`.

The byte-wise version is correct and its comment justifying the equivalence is right (I verified the CR/LF mapping against real Redis). But the pair is now a maintenance hazard: a future fix to the escaping rules — say the `\r\n` run-collapsing question from #388, or the NUL handling from the `helpers.ts:209` comment — has to be made in two places, and the string one is the one that *looks* canonical because it is the older name.

Simplest resolution: drop `sanitizeErrorText` and run the code through `sanitizeErrorBytes` as well (`Buffer.from(value.code)`), leaving one implementation. The code is ASCII by construction, so the byte-wise map is exactly equivalent.

src/commands/pubsub.ts
```diff
@@ -368,26 +378,17 @@ function expectArgCount(
   }
 }
 
+/**
+ * PUBSUB's variadic subcommands police their own argument count, so an excess
+ * argument is real Redis' `addReplySubcommandSyntaxError` rather than the
+ * dispatch-level unknown-subcommand reply or an arity error.
+ */
 function expectPubSubSubcommandMaxArgCount(
```
fatal10110, line 386: **LOW (readability) — the parameter order makes the call sites unreadable.**

`expectPubSubSubcommandMaxArgCount` previously took `(subcommand, args, count)`; it now takes the whole `PubSubArgs` object plus a trailing `ctx`, with the count wedged in the middle. The call sites read:

```ts
expectPubSubSubcommandMaxArgCount(args, 1, ctx)
```

Nothing at the call site hints that `1` is a maximum argument count — it reads like an index or a subcommand arity. The old form at least had `args.args` adjacent to the count it constrained.

`(args, ctx, { max: 1 })` or simply renaming to `expectPubSubSubcommandAtMost(args, ctx, 1)` would restore that. Minor, but this helper is now the only call site of `subcommandSyntaxError`, so it will be the template when the other containers are wired up (see the `helpers.ts:186` comment).

tests/core-cluster-policy.test.ts
```diff
@@ -126,7 +126,10 @@ describe('new cluster execution policy', () => {
     )
     assert.deepStrictEqual(
       await session.execute('cluster', [Buffer.from('nope')]),
-      RedisResult.error("unknown subcommand 'nope'. Try CLUSTER HELP.", 'ERR'),
+      RedisResult.error(
+        Buffer.from("unknown subcommand 'nope'. Try CLUSTER HELP."),
```
fatal10110, line 130: **LOW (test hygiene) — the expectation is now coupled to the internal `messageBytes` field rather than to observable behavior.**

Wrapping the expected message in `Buffer.from(...)` makes this `deepStrictEqual` pass only when the error carries `messageBytes`. A refactor that keeps the wire bytes byte-for-byte identical but happens not to set `messageBytes` — for instance routing this path through a `string` message because the subcommand was already known to be ASCII — fails this assertion for no user-visible reason.

Same pattern at `tests/commands-scripts.test.ts:324` and `:535`.

Byte-exactness is already properly pinned where it belongs, in `tests-integration/raw-tcp/command-errors-subcommand.test.ts`, which compares the actual reply frame. These unit tests would be more robust asserting on the decoded message text and leaving the byte question to the raw-TCP suite — otherwise they are testing the representation, not the behavior, which is what AGENTS.md means by putting exact wire text in raw-TCP tests.

fatal10110:
Review addressed in 4a8e426. Both blockers fixed, both extras done, four follow-ups filed.

## 1. NUL truncation (HIGH) — fixed

Re-verified on my own 6.2.24 and 8.0.6 containers before touching anything. Your ordering proof holds exactly:

| input | real 8.0.6 | real 6.2.24 |
|---|---|---|
| `AA\0BB` | echoes `AA` | echoes `AA` |
| `'A'*100 + \0 + 'A'*100` | 100 | 100 |
| `\0 + 'A'*10` | 0 | 0 |
| `'A'*200 + \0 + 'A'*200` | 128 | 200 |

The last row is the one that settles the ordering *and* shows why the cut cannot ride along inside the 7.0 branch: 6.2 NUL-truncates while not length-truncating. `asCString()` therefore runs in `subcommandError` before the profile branch, and the `%.128s` `subarray` is taken on its result. It covers `subcommandSyntaxError` too — same `%s` conversion — even though that path cannot currently carry a NUL, so the two templates cannot drift.

You were also right about the protocol hazard: the raw NUL was going out inside a `-ERR ...\r\n` simple-error frame. Pinned at both levels — exact bytes for all four cases on the default profile in `command-errors-subcommand.test.ts`, and the 200/`\0`/200 case across all seven profiles in `profile-gates.test.ts` (128 vs 200 by gate).

## 2. Lua boundary (MEDIUM) — fixed, with one correction to the repro

All three sites fixed: `luaReplyToRedisValue` passes `value.err` through (it is already a `Buffer`), `redisErrorToLuaReply` uses a new `errorReplyBytes(err)`, and `scripts.ts:492` uses `RedisResult.fromError(err)`.

One correction worth recording, because it changed how the test is written. Your repro `EVAL "return redis.pcall('CONFIG', ARGV[1])" 0 "\xff\xfe\xfd"` does not reach this template on real Redis at all. From 7.0, script command lookup resolves `container|subcommand`, so an unknown subcommand fails lookup and 8.0.6 answers `-ERR Unknown Redis command called from script` — verified for CONFIG, COMMAND, XINFO and XGROUP. Real 6.2.24 has no such lookup and does produce the container's error. So the byte loss you found is real and worth fixing (it is this server's behavior on every profile), but there is no real-Redis-green assertion for that exact input.

Two tests, split accordingly:

- **Real-safe, not skipped:** `EVAL "return redis.pcall('PUBSUB','CHANNELS','a','b')" 0` — a *known* subcommand with unusable arguments, which does cross the Lua boundary and return the container's own reply on both 6.2.24 and 8.0.6, byte-identical to direct dispatch. This is the round trip through `redisValueToLuaReply` → engine → `luaReplyToRedisValue`.
- **Mock-only, with the reason in the comment and an issue behind it:** the raw-byte version, covering both paths you named — `PUBSUB <raw bytes>` (throws from `execute()`, travels as a `RedisValue`) and `XINFO <raw bytes> k` (throws from the schema parser before a plan exists, travels as a `RedisCommandError` through `redisErrorToLuaReply`).

Both new tests are red on 15fed82: `not ok 7 - the echoed subcommand stops at the first NUL byte` and `not ok 10 - a nested error keeps raw bytes across the Lua boundary`, 9 pass / 2 fail.

## 3. CHANGELOG — added

`Unreleased / Removed` gets a **BREAKING** entry for `UnknownScriptSubcommandError` and `UnknownClusterSubcommandError` with the call-site mapping, and states plainly that there is no drop-in replacement class: `unknownSubcommandError` needs a profile, so it is not exported from either entry point, and catching still works through `RedisCommandError`. `Unreleased / Fixed` gets the behavior entry.

## 4. `RedisResult.fromError` — done, landed here

Worth doing exactly for the reason you gave, so it is in this PR rather than deferred. Every conversion from a caught `RedisCommandError` to a reply now goes through it (`command-executor` ×3, `session-adapter`, `auth-policy`, `scripts.ts`), and `git grep "RedisResult.error(err"` over `src/` returns only the `Resp2ParseError` case, which is a plain `Error` with no bytes to lose.

## Follow-ups filed

| | |
|---|---|
| #435 | MULTI dispatch timing — 7.0+ rejects at queue time with `-EXECABORT`; this server queues and errors at EXEC. Verified both directions on 6.2.24 and 8.0.6. |
| #436 | The mirror — XGROUP/XINFO throw from the schema parser at queue time on every profile, aborting a transaction that real 6.2 queues. |
| #437 | `subcommandSyntaxError` reaches PUBSUB only. Filed with the full verified table for all eight commands you listed, `PUBSUB NUMPAT x` included. |
| #439 | Script dispatch — the lookup behavior in §2 above. |
| #438 | XINFO/XGROUP arity errors name `'xinfo'` rather than `'xinfo|stream'`. |

The gate comment on `error.unknown-subcommand-wording` now opens with "This gates the *wording* only" and lists all four gaps with their issue numbers, so it no longer implies coverage it does not have.

## Verification

`npm test` 378/0 · `test:integration:mock` 1195/0 · compatibility sweep 18/0 on each of the seven profiles · build and lint clean.

Against real 8.0.6 on a container of my own (per the fixture note — the shared ports were left alone): the standalone raw-TCP suites run 40 pass / 0 fail / 1 skipped, the skip being the mock-only Lua test above. The shared cluster is `cluster_state:fail` right now, so the cluster-touching suites (`ioredis/scripts`, both `command-integration`, both `cluster-integration`) hang on discovery and could not be re-run; they were 177/0 against real Redis on 15fed82, and nothing in this commit touches cluster routing.

One unit expectation changed shape rather than behavior: a script's error reply now carries `messageBytes`, so `tests/commands-scripts.test.ts` builds its expected `RedisResult` from a `Buffer`. Same bytes on the wire.

src/core/lua-runtime.ts
```diff
@@ -215,7 +216,9 @@ export function luaReplyToRedisValue(value: ReplyValue): RedisValue {
   if ('err' in value) {
     // The engine supplies an explicit code (or none, for verbatim returned error
     // tables); the host does not infer one.
-    return RedisValue.error(value.err.toString(), value.code?.toString())
+    // `value.err` is already bytes; decoding it here would undo the
+    // byte-exact echo a command like `CONFIG <raw bytes>` produced.
+    return RedisValue.error(value.err, value.code?.toString())
```
fatal10110, line 221: **MEDIUM (new, re-review) — `redis.call` errors still lose their bytes: `renderScriptError` round-trips the body through a UTF-8 string.**

This line and `redisErrorToLuaReply` fix the `pcall` path. The `redis.call` path takes a different exit, though. When the error propagates as a Lua error, `renderScriptError` appends the ` script: <sha>, on @user_script:N.` suffix, and it does that on a string:

- `:295` `body = value.err.toString('utf8')`
- `:299` `return { err: Buffer.from(message, 'utf8'), code: value.code }`

Ground-truthed over raw sockets, `EVAL "return redis.call('PUBSUB', ARGV[1])" 0 "\xff\xfe\xfd"`:

```
real 6.2.24 : -ERR Error running script (call to f_7208...): @user_script:1: ERR Unknown subcommand or wrong number of arguments for '\xff\xfe\xfd'. Try PUBSUB HELP.
mock 6.2    : -ERR Unknown subcommand or wrong number of arguments for '\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd'. Try PUBSUB HELP. script: 7208..., on @user_script:1.
mock 8.0    : -ERR unknown subcommand '\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd'. Try PUBSUB HELP. script: 7208..., on @user_script:1.
```

`XGROUP` gives the same result. On the `redis-6.2` profile the right reply has the raw bytes, and the mock sends U+FFFD. The 6.2 decoration format (`Error running script (call to f_...)`) also differs, but that predates this PR and is out of scope here.

This contradicts the new CHANGELOG entry (`CHANGELOG.md:168-169`): "Error replies are now assembled as bytes end to end, including across the Lua boundary, so a nested `redis.call` error keeps its bytes too." The new Lua tests only use `redis.pcall`, so nothing catches it.

Fix: build the default case as bytes:

```ts
default:
  return {
    err: Buffer.concat([value.err, Buffer.from(` script: ${sha}, on @user_script:${line}.`)]),
    code: value.code,
  }
```

Also add a `redis.call` case next to the `pcall` ones. It can be real-verified on the `redis-6.2` profile, where the mock should match real 6.2's echoed bytes.

tests-integration/raw-tcp/command-errors-subcommand.test.ts
```diff
@@ -0,0 +1,335 @@
...
+  // (#439) and is not what this test is about.
+  //
+  // What it pins is the property that #413 is for: whatever reply a nested
+  // command produces, its bytes reach the client unchanged. Both Lua paths are
+  // covered — PUBSUB throws from `execute()` (the reply travels as a
+  // RedisValue), XINFO throws from its schema parser before the plan exists
+  // (the reply travels as a RedisCommandError through `redisErrorToLuaReply`).
+  test(
+    'a nested error keeps raw bytes across the Lua boundary',
+    { skip: testRunner.backend === 'real' && 'see the comment above' },
```
fatal10110: **LOW (new, re-review): this mock-only test expects a reply real 8.0 never sends. Real-correct alternatives exist.**

The #439 finding checks out. On real 7.2.1 and 8.0.6, `redis.pcall` with an unknown subcommand for CONFIG, COMMAND, PUBSUB, XINFO and XGROUP all return `-ERR Unknown Redis command called from script`. The real-safe `PUBSUB CHANNELS a b` test above is also byte-identical to real 8.0.6 (and to 6.2.24 with the capital `U`). It is ASCII, though, so it would have passed before this fix too.

This test runs on the default `redis-8.0` profile. There it pins `-ERR unknown subcommand '\xff\xfe\xfd'. Try PUBSUB HELP.` as the *expected* value, which is the #439 divergence, so whoever fixes #439 will have to rewrite it. Both byte paths can be pinned against output that real Redis actually sends:

- **`luaReplyToRedisValue` (`lua-runtime.ts:221`)**: `EVAL "return {err=ARGV[1]}" 0 "\xff\xfe\xfd"` returns `-\xff\xfe\xfd\r\n`, byte-identical on real 6.2.24, real 8.0.6 and both mock profiles. It can run **unskipped against real**.
- **`redisErrorToLuaReply` (`:366`) and the `execute()` path**: on the `redis-6.2` profile, `redis.pcall('XGROUP', ARGV[1])`, `redis.pcall('PUBSUB', ARGV[1])` and `redis.pcall('XINFO', ARGV[1])` (no key) all match real 6.2.24 byte for byte, for example `-ERR Unknown subcommand or wrong number of arguments for '\xff\xfe\xfd'. Try XGROUP HELP.` `profile-gates.test.ts` is already the per-profile home for this.

It is still mock-only, but the expectation would be real behavior for that profile instead of a known bug, and it would not break when #439 lands.

src/core/compatibility/feature-gates.ts
```diff
@@ -18,15 +18,30 @@ export const FEATURE_GATES: Record<FeatureId, VersionGate> = {
...
+  // src/commands/helpers.ts). Verified against redis-server 6.2.24, 7.0.15 and
+  // 8.0.6. Valkey forked at 7.2, so every Valkey profile has the newer wording.
+  //
+  // This gates the *wording* only. Three things the same 7.0 change brought are
+  // NOT gated, and are divergences on the older profiles rather than gaps in
+  // this entry:
+  //  - Dispatch timing. 7.0 rejects an unknown container subcommand at MULTI
+  //    queue time with -EXECABORT; this server queues it and errors at EXEC
+  //    (#435). XGROUP/XINFO are the mirror: they throw from a schema parser at
+  //    queue time even on `redis-6.2`, where real 6.2 queues them (#436).
```
fatal10110: **LOW (new, re-review): the gap list here leaves out one 6.2 divergence, and the lead-in sentence has the direction wrong.**

1. **Missing gap.** With trailing arguments, real 6.2 XINFO and XGROUP look up the key and check its type *before* they check the subcommand. 7.0+ does not. Ground-truthed, with `nokey` absent, `str` a string and `s` a stream:

```
                        real 6.2.24                                    real 7.2.1 / 8.0.6
XINFO BOGUS nokey    -> ERR no such key                                ERR unknown subcommand 'BOGUS'. Try XINFO HELP.
XINFO BOGUS str      -> WRONGTYPE Operation against a key holding ...  ERR unknown subcommand 'BOGUS'. ...
XINFO BOGUS s        -> ERR Unknown subcommand or wrong number of ...  ERR unknown subcommand 'BOGUS'. ...
XGROUP BOGUS nokey g -> ERR The XGROUP subcommand requires the key to exist. Note that for CREATE you may want to use the MKSTREAM option ...
XGROUP BOGUS str g   -> WRONGTYPE Operation against a key holding ...
```

On `redis-6.2` the mock answers every one of these with the unknown-subcommand template. The 8.0-profile test "trailing arguments do not change the unknown-subcommand reply" is right as written; the same claim is false on 6.2. This predates the PR, since XGROUP/XINFO never did the key check. The cause is the same as #436 (the subcommand check sits in the schema parser, ahead of any key access), so it probably belongs in #436 and in this list rather than in a new issue.

2. **Direction.** Line 32 says the listed items "are divergences on the older profiles". #435 and #439 are divergences on the **7.0+** profiles, and #436/#437 are 6.2-only. Maybe: "divergences on one side of this boundary or the other".

fatal10110:
## Re-review of `4a8e426` (previous review: `15fed82`)

Every claim below was checked against real servers (**redis 6.2.24, 7.2.1, 8.0.6**). I also ran this PR's mock on the `redis-6.2` and `redis-8.0` profiles and diffed its raw RESP replies byte for byte against the real servers.

### Status of the previous findings

|---|-----|---------|--------|
| 1 | HIGH | Echo not cut at first NUL | ✅ **Fixed.** `asCString` runs before the profile branch. All four NUL cases and `'A'*200 + \0 + 'A'*200` are byte-identical to real: 6.2 echoes 200, 7.2/8.0 echo 128, `\0A10` echoes `''`. |
| 2 | MED | Byte loss at the Lua boundary | ✅ **Fixed for the three named sites** (`lua-runtime.ts:221`, `:366`, `scripts.ts:492`). `return {err=ARGV[1]}` with raw bytes gives `-\xff\xfe\xfd` on real and mock, and so does `pcall` on 6.2. ⚠️ The `redis.call` path still loses the bytes; see new finding A. |
| 3 | MED | MULTI dispatch timing on 7.0+ | ⏸ Deferred to #435 and listed in the gate comment. **Fine to defer.** |
| 4 | MED | XGROUP/XINFO abort at queue time on 6.2 | ⏸ Deferred to #436. **Fine to defer.** New finding C adds a related gap to it. |
| 5 | MED | `subcommandSyntaxError` reaches PUBSUB only | ⏸ Deferred to #437, and the CHANGELOG says so. **Fine to defer.** |
| 6 | MED | `xinfo|stream` arity naming | ⏸ Deferred to #438. **Fine to defer.** |
| 7 | MED | No CHANGELOG entry for the breaking change | ✅ **Fixed.** The Removed entry is marked BREAKING and maps the old call sites to the new ones. |
| 8 | LOW | `errorReplyBody` not enforced | ✅ **Fixed.** `RedisResult.fromError` is used at all six sites. `git grep "RedisResult.error(err"` in `src/` now finds only the `Resp2ParseError` case, which is not a `RedisCommandError`. The `ExecCommandAbortError` site at `command-executor.ts:154` carries static text. |
| 9 | LOW | 6.2 unknown-command wording for gated-off commands | ➖ Not addressed. It predates this PR and belongs to #384. OK. |
| 10 | LOW | Two sanitizers | ❌ Not fixed. Non-blocking. |
| 11 | LOW | `(args, 1, ctx)` signature | ❌ Not fixed. Non-blocking. |
| 12 | LOW | Unit tests coupled to `messageBytes` | ❌ Not fixed. It also grew: `luaReplyToRedisValue` now puts `messageBytes` on *every* Lua error, so `commands-scripts.test.ts:637` needed another `Buffer.from`. Non-blocking. |

### The #439 claim (script dispatch)

**Verified**, with one correction. On real 7.2.1 and 8.0.6, `redis.pcall` with an unknown subcommand for CONFIG, COMMAND, PUBSUB, XINFO and XGROUP returns `-ERR Unknown Redis command called from script`. On 6.2.24 the container's own error comes back for PUBSUB, COMMAND, XGROUP and a key-less XINFO, but **not for CONFIG**, which 6.2 refuses from scripts (`-@user_script: 1: This Redis command is not allowed from scripts`). With a missing key, 6.2 XINFO returns `ERR no such key`. The real-safe `PUBSUB CHANNELS a b` test is byte-identical to real 8.0.6. Filing #439 separately was the right call.

| | Sev | Location | Finding |
| A | **MED** | `lua-runtime.ts:221` (fix at `:295`/`:299`) | `redis.call` errors still decode to U+FFFD: `renderScriptError` round-trips the body through a string to add the `script: <sha>` suffix. Real 6.2 sends `'\xff\xfe\xfd'` and the mock sends `'\xef\xbf\xbd'` × 3, on both profiles. This contradicts the new CHANGELOG line "a nested `redis.call` error keeps its bytes too". No test covers `redis.call`. |
| B | LOW | `command-errors-subcommand.test.ts:294` | The mock-only test expects the #439-divergent reply on `redis-8.0`. Better options, all real-correct: `return {err=ARGV[1]}` works unskipped against real, and on `redis-6.2`, `pcall` XGROUP/PUBSUB/XINFO match real 6.2 byte for byte. |
| C | LOW | `feature-gates.ts:37` | On 6.2 with trailing args, XINFO/XGROUP check the key before the subcommand (`ERR no such key`, `WRONGTYPE`, `requires the key to exist`). This predates the PR and has the same root cause as #436, but the gap list leaves it out. Also, line 32 says "older profiles", while #435/#439 are 7.0+ divergences. |

### Tests on `4a8e426`
- `tsc --noEmit`: clean
- Targeted unit and raw-TCP tests (`behavior`, `profile`, `core-cluster-policy`, `commands-scripts`, `command-errors-subcommand`, `command-errors-config`): **86/86**
- `profile-gates.test.ts` on `redis-6.2`, `redis-7.0`, `redis-8.0`, `valkey-8.0`: **14/14 each**
- Lua and script unit suites: **23/23**
- The `node-redis-real` failure at `keys-scan.test.ts:162` is the #395/#267 flake that #429 fixes. It is not counted against this PR.

### Verdict: **not yet. One small fix, then ship.**

Finding A is the only blocker. It is the byte loss #413 is about, on the Lua `redis.call` path, and the CHANGELOG currently says it is fixed. The fix is about three lines in `renderScriptError`, plus a `redis.call` test case. Findings B and C and the leftover LOWs are non-blocking, and the four deferrals are properly tracked (#435 to #439) and documented in the gate comment.

fatal10110:
Second round addressed in 6b93b49.

## A. `redis.call` bytes (MEDIUM) — fixed

`renderScriptError` now keeps the body as a `Buffer` and concatenates the ` script: <sha>, on @user_script:<line>.` suffix as bytes, so the CHANGELOG line about nested `redis.call` errors is now true.

Before choosing test inputs I re-probed real 6.2.24, 7.0.15 and 8.0.6 on my own containers. On 7.0 and 8.0, all twelve containers I tried answer `ERR Unknown Redis command called from script` for an unknown subcommand from a script. So a raw-byte echo cannot come back through `redis.call` on any 7.0+ profile. The tests are split to match:

- **`redis-8.0`, unskipped against real:** `EVAL "error(ARGV[1])" 0 "\xff\xfe\xfd"`. It goes through the same `renderScriptError` branch a failing `redis.call` takes, and real 8.0.6 sends `-ERR user_script:1: \xff\xfe\xfd script: <sha>, on @user_script:1.` (94 bytes). The mock sent 100 bytes before the fix.
- **`redis-6.2` profile sweep:** `redis.call('PUBSUB', ARGV[1])` with raw bytes. Real 6.2 wraps the error in the older `Error running script (call to f_<sha>): …` decoration. This server uses the 7.0 decoration on every profile, which is a separate, pre-existing divergence, now filed as **#442**. So this test checks only the body segment that real 6.2 also sends. It must contain the three raw bytes, not three U+FFFD.

Both are red on 4a8e426 and green now.

## B. Test inputs — replaced

The mock-only test is gone. In its place:

- `return {err=ARGV[1]}`: unskipped, identical on real 6.2.24 and 8.0.6.
- `redis.pcall` of `PUBSUB` / `XGROUP` / `XINFO` with raw bytes, on the `redis-6.2` profile only, byte-for-byte against real 6.2.24. `XINFO` is called without a key on purpose, because with one 6.2 checks the key first (C below).
- `PUBSUB CHANNELS`: I checked this path, and real Redis never echoes non-ASCII here. The echoed name is `argv[1]`, and on this path it has to be a subcommand Redis recognizes. `PUBSUB CHANNELSé a b` fails at dispatch instead. So the test now passes a raw-byte *argument* to pin that arguments are not echoed. It also now covers a failing `redis.call`, not just `pcall`, and matches 8.0.6 including the decoration. It is a shape guard, not a red test, and the comment says so.

The raw-TCP suite runs 12/12 against real 8.0.6 with nothing skipped (18/18 including the CONFIG suite).

## C. Gate comment — fixed

It no longer calls these "older profiles" divergences. Each gap now names the profiles where this server is wrong:

- #435 MULTI: 7.0+
- #436 XINFO/XGROUP: `redis-6.2`, plus the key-before-subcommand ordering
- #437 arity: `redis-6.2`
- #439 scripts: 7.0+

I verified the key-ordering claim on 6.2.24 before adding it: missing key → `no such key` / `requires the key to exist`, string key → `WRONGTYPE`, stream or no key → the subcommand error. 8.0.6 returns the subcommand error in every case. #436 now has a second section with the full table.

## D. #439 — corrected

On 6.2, `CONFIG`, `CLIENT`, `ACL` and `SCRIPT` are refused as `noscript`, and `FUNCTION` does not exist. `SLOWLOG`, `COMMAND`, `PUBSUB`, `XGROUP` and keyless `XINFO` do reach the container. You listed four of those; `SLOWLOG` is the fifth. On 7.0+ all twelve behave the same way. The acceptance criteria now use `PUBSUB` instead of `CONFIG`.

## Non-blocking

- **Two sanitizers:** done. The code is prefixed onto the body before one byte-wise map, and `sanitizeErrorText` is deleted.
- **`expectPubSubSubcommandMaxArgCount(args, 1, ctx)`:** left as is. Reordering it would be churn with no clear benefit.
- **Tests coupled to `messageBytes`:** left as is (4 sites build the expected `RedisResult` from a `Buffer`). A proper decoupling means an `assertRedisError(result, text, code)` helper across `tests/`, which is more than trivial. Happy to do it as a follow-up if you want it.

## Verification

- `npm test`: 378 pass, 0 fail
- Compatibility sweep, all 7 profiles, 0 fail: 21 pass on `redis-6.2`; 19 pass and 2 skipped on the others (the two `redis-6.2`-only tests)
- Build and lint: clean
- `test:integration:mock`: 1194 pass, 2 fail. Both failures were node-redis hash field-TTL tests (the known #411 flake), and `hash/field-expire`, `field-persist` and `basic` ran 19/19 on their own.

Real-backend checks used only my own servers (6.2 on 7662, 7.0 on 7670, 8.0 on 7680). Nothing touched the shared fixtures, and no node-redis real suite ran against the shared cluster.

src/commands/connection.ts
```diff
@@ -569,7 +570,9 @@ export const infoCommand = defineCommand({
 export const clientCommand = defineCommand({
   name: 'client',
   schema: t.object({
-    subcommand: t.string(),
+    // Raw bytes, not `t.string()`: the unknown-subcommand reply echoes the
+    // name the client sent, and a UTF-8 decode here would lose its bytes.
+    subcommand: t.bulk(),
     args: t.variadic(t.bulk()),
   }),
   flags: ['readonly', 'admin'],
```
fatal10110, line 578: **LOW (new, round 3; it predates this PR and doesn't block it): `CLIENT` has no `noscript` flag, so scripts can run it on every profile. #439's acceptance criterion assumes the opposite.**

This line hasn't changed in the PR, but the corrected #439 now says:

> `CONFIG`/`CLIENT`/`ACL`/`SCRIPT` from a script **keep** their `noscript` refusal on every profile

That holds for CONFIG, ACL and SCRIPT. It does not hold for CLIENT, which is flagged `['readonly', 'admin']` only. Results over raw sockets:

```
EVAL "return redis.pcall('CLIENT','GETNAME')" 0
  real 6.2.24 : -@user_script: 1: This Redis command is not allowed from scripts
  real 8.0.6  : -ERR This Redis command is not allowed from script
  mock 6.2    : (nil)   <- runs
  mock 8.0    : (nil)   <- runs

EVAL "return redis.pcall('CLIENT','BOGUS')" 0
  real 6.2.24 : -@user_script: 1: This Redis command is not allowed from scripts
  mock 6.2    : -ERR Unknown subcommand or wrong number of arguments for 'BOGUS'. Try CLIENT HELP.
```

For comparison, `CONFIG GET maxmemory` from a script is refused correctly on the mock.

Whoever works on #439 will take "keep" at face value, and nothing currently refuses CLIENT. Either add `'noscript'` here (real 7.0+ marks every `client|*` subcommand NOSCRIPT, and 6.2 marks the whole command) or change "keep" to "gain" for CLIENT in #439. Fine as a follow-up. It isn't a reason to hold this PR.

fatal10110:
## Re-review of `6b93b49` (previous review: `4a8e426`)

I ran real **redis 6.2.24** and **8.0.6** on my own ports, and this PR's mock from both `4a8e426` and `6b93b49` on the `redis-6.2` and `redis-8.0` profiles. I compared raw RESP replies byte for byte. The work was done in isolated worktrees; the shared fixtures were not touched.

### Round-2 findings

| | Finding | Status |
|---|---|---|
| **A** (blocker) | `redis.call` errors lost raw bytes | ✅ **Fixed.** `renderScriptError` now keeps the body as a Buffer and appends the suffix as bytes. The old-head vs new-head mock diff has exactly two changed lines, both this fix: `error(ARGV[1])` goes from 100 B to **94 B**, which is byte-identical to real 8.0.6 (`-ERR user_script:1: \xff\xfe\xfd script: <sha>, on @user_script:1.`). `redis.call('PUBSUB', ARGV[1])` on 6.2 goes from 157 B with U+FFFD to 151 B with the raw bytes. |
| A: tests | Do the new tests go red before the fix? | ✅ **Yes.** I ran the new test files against `4a8e426` code. Exactly the two claimed tests fail and nothing else: `a script-aborting error keeps raw bytes through the script decoration` (raw-TCP, 8.0) and `a nested unknown subcommand keeps raw bytes through redis.call` (profile-gates, 6.2). |
| A: `error(ARGV[1])` as a stand-in for `redis.call` | Same code path? | ✅ Both reach the `default:` branch of `renderScriptError`, which is the code that changed. The 8.0 test covers the decoration, and the 6.2 `redis.call` test covers the whole chain (command error → `redisErrorToLuaReply` → engine rethrow → decoration). Taken together, the coverage is complete. |
| A: the 6.2 `includes()` assertion | Would it pass without the fix? | ✅ **No.** The expected substring contains the raw `\xff\xfe\xfd`, so U+FFFD cannot match it, and it did fail on the old code. Asserting only the body segment is the right choice while #442 (6.2 wrapper wording) is open. |
| **B** | Test inputs | ✅ **Fixed.** The mock-only test that expected #439 behaviour is removed. `return {err=ARGV[1]}` gives `-\xff\xfe\xfd\r\n` on real 6.2.24, real 8.0.6 and both mock profiles. On 6.2, `pcall` PUBSUB/XGROUP/keyless XINFO match real 6.2.24 byte for byte. The `PUBSUB CHANNELS <raw> b` shape guard matches real 8.0.6 through both `pcall` and `call`, and real never echoes the argument. The new raw-TCP tests run **unskipped against a real 8.0.6: 18/18**. |
| **C** | Gate comment | ✅ **Fixed.** Each gap now names its profile side (#435/#439 on 7.0+, #436/#437 on 6.2), and 6.2's key-before-subcommand order is listed. The new #436 section's table matches my 6.2.24/8.0.6 captures row for row. *Nit:* #436's "Related" section cites **#433** (a MONITOR issue) where it means **#438** (xinfo arity). |
| **D** | #439 text | ✅ **Accurate.** On real 6.2.24, CONFIG/CLIENT/ACL/SCRIPT are refused from scripts, FUNCTION returns `Unknown Redis command called from Lua script`, and SLOWLOG/COMMAND/PUBSUB/XGROUP/keyless XINFO reach the container. On 8.0.6, all 11 return `Unknown Redis command called from script`. The acceptance criteria use PUBSUB. One correction: the criterion that CLIENT **keeps** its noscript refusal is wrong, because the mock never refuses CLIENT (inline comment on `connection.ts:578`). |
| Sanitizer merge (was #10) | Any drift? | ✅ **None.** Mapping CR/LF 1:1 distributes over concatenation, and the `code␠` separator contains no CR/LF, so sanitizing `code + ' ' + body` gives the same bytes as sanitizing each part. Checked on the wire: `CONFIG a\r\nb`, `CONFIG \r\n\r\n`, `{err='AB\rC x\ny'}` (CR inside the *code*) and `{err='\r\nX'}` are byte-identical to real on both profiles. `sanitizeErrorText` is fully removed. |
| #11 pubsub arg order | Deferred | 👍 **Fine to defer.** Readability only, and it has one caller. |
| #12 four tests tied to `messageBytes` | Deferred | 👍 **Fine to defer.** It's brittleness in unit tests only. Byte exactness is pinned where it belongs, in raw-TCP, and now also against real Redis. |

### New finding (inline)

| Sev | Location | Finding |
|---|---|---|
| LOW (predates this PR, non-blocking) | `src/commands/connection.ts:578` | `CLIENT` has no `noscript` flag. Real 6.2.24 and 8.0.6 refuse `redis.pcall('CLIENT','GETNAME')`, while the mock runs it and returns nil. #439's acceptance criterion says CLIENT "keeps" its refusal, which it doesn't have. Add `'noscript'`, or fix the wording in #439. |

Not filed as findings: two divergences that already existed on `4a8e426` and that this round did not change. `error_reply('E\r1 z')` gets an extra `ERR ` (CR inside the code token; the normal `error_reply('MYERR x')` case matches), and 6.2's noscript refusal wording differs (already mentioned in #439).

### Tests on `6b93b49`
- `tsc --noEmit` and `eslint`: clean
- Unit tests (`tests/**`): **378/378**
- Compatibility sweep: **0 failures on all 7 profiles**. redis-6.2 runs 21 tests; the other six run 19 each, because the two 6.2-only Lua tests are correctly skipped.
- Raw-TCP `command-errors-subcommand` + `command-errors-config`: **18/18 on mock** and **18/18 against my own real 8.0.6** (`REDIS_STANDALONE_PORT`, not the shared fixture)
- Red check: the new tests on the old code give **2 failures**, exactly the claimed pair.

### Verdict: **SHIP** ✅

Blocker A is fixed and pinned by tests that fail on the old code and pass now, one of which runs against real Redis. B, C and D are verified. The sanitizer merge is byte-equivalent, and both deferrals are acceptable. The single new finding predates this PR, is LOW, and can be a follow-up; the same goes for the #433 → #438 link in #436.

PR #424, opened 2026-09-22
refactor: back the in-memory transports with stream.duplexPair (#360)

fatal10110:
## Summary

Pure refactor, no behavior change. Three hand-written in-memory
`ConnectionTransport` implementations each carried their own reader queue,
waiter set, abort controller and listener map. They are replaced by one end of
a [`stream.duplexPair()`](https://nodejs.org/api/stream.html#streamduplexpairoptions)
wrapped by the production `SocketConnectionTransport`, so the socketless path
now exercises the same framing and teardown code as a real socket.

- **`SocketConnectionTransport`** takes any `Duplex` (was `net.Socket`). Its
  no-op ternary (`Buffer.isBuffer(chunk) ? Buffer.from(chunk) : Buffer.from(chunk)`)
  is gone, and `read()` swallows `ERR_STREAM_PREMATURE_CLOSE` (an ordinary
  disconnect) while letting every other error propagate.
- **`DuplexConnectionTransport`** (~130 lines) is deleted.
  `createVirtualConnection` now builds a `duplexPair` and bridges teardown
  between the two ends — `duplexPair` does **not** propagate `destroy()` the way
  a real socket pair does, so without the bridge a destroyed end leaves the
  other waiting forever. There is a regression test for each direction.
- **`InMemoryConnectionTransport`** is deleted and replaced by a duplexPair-backed
  test helper (`tests/in-memory-transport-test-helper.ts`), keeping the same
  `feed()` / `endRead()` / `getWrittenBuffer()` surface (and the same
  feed-after-close and single-reader guards) the four unit-test files used.
- **`ConnectionTransport.on()`** plus `ConnectionTransportEvent` /
  `…Listener` / `…Unsubscribe` and the three listener maps are deleted —
  `on('close' | 'drain' | 'error')` had zero subscribers anywhere.
- **`VirtualClientSocket`** becomes a type: it is a `duplexPair` end decorated
  with the `net.Socket` surface ioredis touches (the `setNoDelay` / `ref` /
  `unref` / … no-ops, `remoteAddress` / `remotePort`, and the `nextTick`
  `'connect'` emit), so it can no longer be constructed directly.
- **`engines.node`** moves to `>=22.6` (`stream.duplexPair` landed in 22.6);
  README updated, and CI now runs the unit suite on 22.6 as well as 24 so the
  floor is exercised rather than only declared.

Net **-329 lines** under `src/`.

### What `duplexPair` does not give you for free

Found across three review rounds; each has a regression test, and each test was
[...]

PR #431, opened 2026-09-22
fix: enforce proto-max-bulk-len at the protocol layer and derive the bit-offset ceiling from it (#415)

### LINKED ISSUE #415: proto-max-bulk-len is not enforced at the protocol layer, and SETBIT/BITFIELD hardcode their own ceiling
Found while reviewing PR #409, which makes `proto-max-bulk-len` a live setting for the first time (it was previously an inert entry in the CONFIG defaults map that nothing read). PR #409 enforces it in `APPEND` and `SETRANGE` only. Now that the setting actually does something, the places that ignore it are divergences.

## 1. The protocol layer does not enforce it — this is real Redis's *primary* enforcement point
`src/core/transports/resp2/decoder.ts`

In real Redis every oversized bulk argument is rejected at parse time, for every command, before any command handler runs. Verified with the limit lowered to 1 MB and a 2 MB `APPEND` value on a fresh key:

```
mock    -> :2000000  then STRLEN 2000000, connection alive
7.2.16  -> -ERR Protocol error: invalid bulk length   (connection closed)
6.2.24  -> -ERR Protocol error: invalid bulk length   (connection closed)
```

This single check is what covers `SET`, `GETSET`, `SETEX`, `MSET`, `LPUSH`, `HSET` and everything else in real Redis. All of them currently diverge in the mock. It also means PR #409's statement that "`APPEND` to a missing key is never size-checked" is correct at the command layer but wrong end-to-end — real Redis still rejects an oversized fresh `APPEND`, just earlier.

Note this changes connection lifecycle, not just the reply: real Redis closes the connection on a protocol error.

## 2. Bit-offset ceiling is hardcoded instead of derived
`src/commands/bitmaps.ts:28` — `MAX_BIT_OFFSET = 0x1_0000_0000`.

Real Redis derives the ceiling from the live setting: `(offset >> 3) >= server.proto_max_bulk_len`. The hardcoded constant was correct only while the config was inert. With the limit at 1048576:

```
SETBIT k 8388608 1  ->  mock: 0 (allocates)
                        real 6.2 and 7.2: ERR bit offset is not an integer or out of range
```

Affects `SETBIT`/`GETBIT` (`bitmaps.ts:55`) and `BITFIELD` (`bitmaps.ts:462`). They agree at the default limit, so this is newly reachable as a result of making the config live.

## Also
`docs/COMMANDS.md` (as updated by PR #409) should not imply the limit is generally enforced until (1) lands — today it is enforced by exactly two commands.

## Acceptance
- An oversized bulk argument is rejected at the protocol layer for every command, with real Redis's `Protocol error: invalid bulk length` wording and connection-close behaviour, pinned by a raw-TCP test.
- `SETBIT`/`GETBIT`/`BITFIELD` derive their offset ceiling from the live `proto-max-bulk-len`.
- Docs describe what is actually enforced.

fatal10110:
## Summary

PR #409 made `proto-max-bulk-len` a live setting and enforced it in `APPEND` /
`SETRANGE`. This closes the two places that still ignored it.

### 1. The protocol layer — Redis' *primary* enforcement point

`src/core/transports/resp2/decoder.ts` accepted a bulk argument of any length,
so the limit bound exactly two commands. Real Redis rejects an oversized bulk
while the multibulk frame is still being parsed, before the payload is read and
before any handler runs — one check that covers `SET`, `GETSET`, `SETEX`,
`MSET`, `LPUSH`, `HSET`, `ECHO`, an `APPEND` to a missing key, and the
command-name bulk itself. It is also a **connection-lifecycle** event, not just
a reply: the error is followed by a close.

The decoder now refuses a bulk header longer than the live limit with
`-ERR Protocol error: invalid bulk length`, and the adapter closes the
connection — after answering any valid commands that preceded the bad frame in
the same pipeline (the #79 behaviour, unchanged).

To make "live" actually true, `Resp2CommandDecoder` became pull-based:
`push(chunk)` buffers, `next()` yields one frame, and `Resp2SessionAdapter` runs
each command before parsing the next. A pipelined
`CONFIG SET proto-max-bulk-len` therefore applies to the frames behind it even
when they arrived in the same TCP read — which is what Redis does, and what the
old "decode the whole chunk, then execute" shape could not. The limit reaches
the decoder as a `maxBulkLength: () => bigint` getter off
`session.server.protoMaxBulkLen`, so the transport layer keeps no dependency on
server state.

### 2. The bit-offset ceiling

`src/commands/bitmaps.ts` hardcoded `MAX_BIT_OFFSET = 0x1_0000_0000`. Redis
derives it: `(offset >> 3) >= server.proto_max_bulk_len` (bitops.c,
`getBitOffsetFromArgument`). The two agree at the 512MB default — which is why
this only became reachable once the config went live — and diverge as soon as
the limit moves. `SETBIT`, `GETBIT`, `BITFIELD` and `BITFIELD_RO` now derive it,
including the `#<index>` form (multiplied by the type width first).

`ParseContext` carries only `commandName` + `profile`, with no access to server
state, so a ceiling that depends on live config cannot live in a command's
`schema`. Their offset / bit-value / op-list validation moved to `execute`,
which is also closer to Redis: every one of those is a runtime error there, so
inside `MULTI` they queue and fail at `EXEC` while only arity fails at queue
time. Error ordering is preserved and pinned — `SETBIT k <over-ceiling> 2`
reports the *offset* error, not the bit error, and a bad offset anywhere in a
`BITFIELD` op list rejects the whole command without touching the key.

Two smaller things fall out of the ceiling becoming movable: `>>> 3` truncates
to 32 bits, so byte indices are now `Math.floor(bitIndex / 8)` (offsets past
2^32 are reachable with a raised limit), and bitmap allocations go through the
same `RangeError` → `StringExceedsMaxSizeError` backstop `strings.ts` already
uses.

### 3. Docs

`docs/COMMANDS.md` said enforcement "is not yet general" and pointed at this
issue. It now describes the three places the limit is actually enforced, and the
Bitmap section notes the derived ceiling. `docs/ARCHITECTURE.md` gets a sentence
on the decoder's pull-based loop.

## Compatibility

No gate. Ground-truthed against **redis-server 6.2.24, 7.2.16 and 8.0.6** — all
three byte-identical on every case asserted here:

```
$1048576 with proto-max-bulk-len 1048576  -> accepted (waits for payload)
$1048577 with proto-max-bulk-len 1048576  -> -ERR Protocol error: invalid bulk length, close
SETBIT k 8388608 1  (limit 1MB)           -> ERR bit offset is not an integer or out of range
SETBIT k 8388607 1  (limit 1MB)           -> 0
SETBIT k 8388608 2  (limit 1MB)           -> ERR bit offset ...   (offset wins over bit)
BITFIELD k GET u8 8388608  (limit 1MB)    -> ERR bit offset ...
BITFIELD k GET i64 8388552 (limit 1MB)    -> [0]                  (no (bits-1)/8 term)
```

The limit is `>` and not `>=`: a bulk exactly the size of the limit is accepted.
(An earlier revision of this description said inline commands are exempt. That is
vacuous upstream: the smallest settable limit is 1MB, far above the 64KB cap Redis
puts on inline requests — a cap the mock does not model yet, tracked in #441.)

Closes #415

## Test plan

- [x] `tests-integration/raw-tcp/proto-max-bulk-len-protocol.test.ts` — new.
      Exact wire bytes + the close, per AGENTS.md. Eight commands each rejected
      at the protocol layer, the oversized command-name bulk, pipelined
      commands answered before the error, the command never running, inline
      commands exempt, and a mock-only block that lowers the limit to 1MB and
      pins the boundary from both sides plus the ceiling following the setting
      back up. Red before the fix: **14 of 15 failed**.
- [x] `tests-integration/{ioredis,node-redis}/string/bitmap-offset-limit.test.ts`
      — new twin pair. Default-limit ceiling, offset-before-bit-value ordering,
      BITFIELD / BITFIELD_RO / `#<index>`, no partial write on a rejected op
      list, and the mock-only lowered/raised-limit cases. Red before the fix:
      **6 of 12 failed** (the default-limit half was already correct — that is
      the point).
- [x] All new tests green against **real Redis** before the fix landed
      (18 pass, 6 mock-only skips), proving the divergence was ours.
- [x] `npm test` — 378/378
- [x] `npm run test:integration:mock` — 1208/1208
- [x] Real backend, targeted: the three new files plus `raw-tcp/protocol-errors`
      and `raw-tcp/proto-max-bulk-len` — 21 pass, 0 fail, 6 skipped
- [x] `npm run build`, `npm run lint`, `prettier --check` clean


src/commands/bitmaps.ts
```diff
@@ -52,12 +90,23 @@ function parseBitOffset(token: Buffer | undefined): number {
...
+  assertBitOffsetWithinLimit(value, protoMaxBulkLen)
   return value
 }
 
+/** @see parseBitOffset — the shared `(offset >> 3) >= limit` ceiling. */
+function assertBitOffsetWithinLimit(
+  offset: number,
+  protoMaxBulkLen: bigint,
+): void {
+  if (BigInt(byteIndexOf(offset)) >= protoMaxBulkLen) {
```
fatal10110: **[high] The derived ceiling drops `MAX_MATERIALISABLE_LENGTH`, so a raised `proto-max-bulk-len` lets SETBIT/BITFIELD allocate without bound — and the docs this PR edits still promise it can't.**

`strings.ts` deliberately does *not* use `proto-max-bulk-len` raw. `assertWithinProtoMaxBulkLen` (`src/commands/strings.ts:740-752`) caps the effective limit at `MAX_MATERIALISABLE_LENGTH` (`536870912n`, exported at `strings.ts:64`) with the comment "because `proto-max-bulk-len` is configurable far past anything this process can allocate". `assertBitOffsetWithinLimit` compares against `protoMaxBulkLen` unclamped, so the bitmap path loses that guard.

Reproduced against this branch (mock server, in-process):

```
CONFIG SET proto-max-bulk-len 629145600   -> OK
SETRANGE k1 629145599 x                   -> ERR string exceeds maximum allowed size (proto-max-bulk-len)
SETBIT   k2 5033164792 1                  -> 0   | STRLEN = 629145600 | RSS +1793 MB
```

Same server, same configured limit: `SETRANGE` refuses without allocating, `SETBIT` materialises 600MB.

`allocateBitmapBuffer` is not a substitute for the cap. On Node 22 `require('buffer').constants.MAX_LENGTH` is `9007199254740991`, so `Buffer.alloc` raises no `RangeError` for any size this ceiling admits — the process simply allocates until it is killed, and a V8 allocation failure is not a catchable `RangeError` anyway. The try/catch only fires past 2^53, which `Number.isSafeInteger` has already excluded.

This also contradicts the paragraph the PR keeps at `docs/COMMANDS.md:134`: "Raising it past **512MB**, Redis' own default, does not raise what the mock will allocate: beyond that the size error is returned rather than a buffer the test process may not survive producing." That is now false for `SETBIT`/`BITFIELD`.

Suggested fix: clamp in `assertBitOffsetWithinLimit` the same way `assertWithinProtoMaxBulkLen` does —

```ts
const limit =
  protoMaxBulkLen < MAX_MATERIALISABLE_LENGTH
    ? protoMaxBulkLen
    : MAX_MATERIALISABLE_LENGTH
if (BigInt(byteIndexOf(offset)) >= limit) {
  throw new BitOffsetError()
}
```

and keep `allocateBitmapBuffer` as the backstop it is described as, rather than the primary defence.

src/core/transports/resp2/session-adapter.ts
```diff
@@ -31,28 +31,47 @@ export class Resp2SessionAdapter {
...
-        const { frames, error } = this.decoder.push(chunk)
-        for (const frame of frames) {
+        this.decoder.push(chunk)
+
+        for (;;) {
+          let frame: Resp2CommandFrame | null
+          try {
+            // Pulled one at a time so each command has run — and any config it
+            // changed has landed — before the next frame is parsed.
+            frame = this.decoder.next()
```
fatal10110, line 54: **[high] The pull-based decoder rewrite — the most invasive change in this PR — is not pinned by a single test. Either add the test or drop the rewrite.**

I restored the old batch-drain shape here (collect every frame from the chunk, *then* execute them), keeping the new bulk-length check and the live `maxBulkLength` getter exactly as this PR has them. Result: **38/38 still green** across `raw-tcp/proto-max-bulk-len-protocol`, `raw-tcp/protocol-errors`, `raw-tcp/protocol`, and both `bitmap-offset-limit` twins.

The behaviour does differ, though. One TCP write containing `CONFIG SET proto-max-bulk-len 1048576` followed by `*3\r\n$6\r\nAPPEND\r\n$3\r\nfoo\r\n$2097152\r\n`:

```
pull-based (this PR):  "+OK\r\n-ERR Protocol error: invalid bulk length\r\n"   connection closed
batch-drain (old):     "+OK\r\n"                                              connection still open
```

So the rewrite does deliver what the description claims — but nothing in the suite would notice if it were reverted. Every new test's `setLimit` helper opens a *separate* connection and awaits `+OK` before the affected frame is written, so a config change and a frame behind it never share a read.

This is the one behaviour the rewrite exists for; please pin it with a raw-TCP test that writes both frames in a single `conn.write()`. Without it, an invasive change to the core read loop is carried on the strength of a comment.

src/commands/bitmaps.ts
```diff
@@ -423,7 +487,20 @@ type BitFieldOp =
       operand: bigint
       overflow: Overflow
     }
-type BitFieldArgs = { key: Buffer; ops: BitFieldOp[] }
+/**
+ * BITFIELD's operation list is kept as raw tokens and validated at execute
+ * time, because a field offset is bounded by the live `proto-max-bulk-len`.
+ *
+ * That also matches Redis, which parses and validates the whole operation list
+ * in one pass inside the command (bitops.c, `bitfieldGeneric`) before running
```
fatal10110: **[medium] This new docblock asserts an ordering rule that real Redis does not follow.**

"the first bad token in argument order wins" is not how `bitfieldGeneric` behaves. Real Redis matches an opcode only once it has confirmed there are enough remaining arguments, and checks the `BITFIELD_RO` restriction *after* parsing the op's type and offset. Ground-truthed on `redis-server 7.2.1` (identical on 6.2.14):

```
BITFIELD_RO b foo            -> -ERR syntax error
BITFIELD_RO b SET u8 0       -> -ERR syntax error
BITFIELD_RO b GET u8 0 foo   -> -ERR syntax error
BITFIELD_RO b SET x9 0 1     -> -ERR Invalid bitfield type. Use something like i16 u8. ...
BITFIELD_RO b INCRBY u8 0 1  -> -ERR BITFIELD_RO only supports the GET subcommand

BITFIELD b GET x9            -> -ERR syntax error
BITFIELD b SET x9 0          -> -ERR syntax error
```

This branch, same seven:

```
BITFIELD_RO b foo            -> -ERR BITFIELD_RO only supports the GET subcommand
BITFIELD_RO b SET u8 0       -> -ERR BITFIELD_RO only supports the GET subcommand
BITFIELD_RO b GET u8 0 foo   -> -ERR BITFIELD_RO only supports the GET subcommand
BITFIELD_RO b SET x9 0 1     -> -ERR BITFIELD_RO only supports the GET subcommand
BITFIELD b GET x9            -> -ERR Invalid bitfield type. ...
BITFIELD b SET x9 0          -> -ERR Invalid bitfield type. ...
```

The divergence is pre-existing on `origin/main` — the missing piece is a per-op "are there `N` remaining args?" check before the opcode is accepted, and hoisting the `readonly` check to after `parseFieldType`/`parseFieldOffset`. What is new is this comment, which now documents the behaviour as matching Redis. Please either fix the ordering or soften the claim to just the part that *is* verified (a rejected op anywhere in the list leaves the key untouched — that one reproduces exactly).

src/core/transports/resp2/decoder.ts
```diff
@@ -20,39 +24,55 @@ type ParseOutcome =
...
+    while (this.buffered.length > 0) {
+      const outcome = this.parseFrame(0)
+      if (outcome.kind === 'incomplete') {
+        return null
       }
-    } catch (err) {
-      if (!(err instanceof Resp2ParseError)) {
-        throw err
+
+      this.buffered = this.buffered.subarray(outcome.nextIndex)
```
fatal10110: **[medium] The decoder is left poisoned after `next()` throws, and it is a published export.**

`parseFrame` throws *before* this line runs, so on a protocol error the bad frame stays at the head of `buffered`. That is fine for `Resp2SessionAdapter`, which returns immediately — but `Resp2CommandDecoder` is re-exported from `src/core/transports/resp2/index.ts` and again from `src/internal.ts`, so it is part of the surface other code can drive.

An external consumer that catches `Resp2ParseError` and calls `next()` again gets the identical throw forever, and any further `push()` grows a buffer that can never drain. The old `push()` returned `{ frames, error }` and always advanced past what it had consumed, so this footgun is new.

Worth a `failed` flag so a second `next()`/`push()` fails loudly ("decoder closed after protocol error"), or at minimum a doc line saying the instance must be discarded once it throws.

Related, on the same surface: `push()` changed its return type from `Resp2DecodeResult` to `void` and `Resp2DecodeResult` was deleted outright. That is a breaking change to an exported API with no `CHANGELOG.md` entry in this diff (#424, touching the same layer, does update it).

tests-integration/raw-tcp/proto-max-bulk-len-protocol.test.ts
```diff
@@ -0,0 +1,312 @@
...
+
+    assert.strictEqual((await conn.readRawFrame()).toString(), '+PONG\r\n')
+  })
+})
+
+describe(
+  `Raw TCP proto-max-bulk-len protocol errors follow the live limit (${testRunner.getBackendName()})`,
+  {
+    skip:
+      testRunner.backend === 'real' &&
```
fatal10110, line 255: **[medium] Every load-bearing boundary assertion in this PR is mock-only.**

All three tests in this `describe` skip on the real backend, as do three of six in each `bitmap-offset-limit` twin. Between them that is the `>` vs `>=` boundary (the headline claim), the #415 lowered-limit repro, the ceiling following the setting back up, and the lowered `SETBIT`/`GETBIT`/BITFIELD ceilings — i.e. the claims this PR rests on are pinned only against the implementation under review.

I did ground-truth them by hand and **they are all correct** (`redis-server` 6.2.14 / 7.0.15 / valkey 7.2.14 / 7.2.1 / 8.0.6, byte-identical), so this is a durability concern, not a correctness one. The skip reason is fair for the *shared* standalone, but `setupRawStandalone()` already starts a server per suite, so a dedicated disposable one could carry these.

One related coverage gap worth closing cheaply: every BITFIELD assertion in both twins uses `u8`, where a `(bits-1)/8` term would be invisible. The PR's claim that real Redis has no such term is **correct** — verified on all five engines, `BITFIELD b GET i64 8388607` → `0` and `8388608` → error, exactly the `SETBIT` boundary — but a wide type is what would pin it, and `GET` never allocates, so `BITFIELD k GET i64 4294967288` works at the *default* limit with no memory cost and no skip.

src/commands/bitmaps.ts
```diff
@@ -114,34 +174,37 @@ export const setbitCommand = defineCommand({
...
+          value: input[index + 2]!,
+        },
         nextIndex: index + 3,
       }
     },
   ),
   flags: ['write', 'denyoom'],
   keys: args => [args.key],
   execute: (args, ctx) => {
+    const offset = parseBitOffset(args.offset, ctx.server.protoMaxBulkLen)
```
fatal10110: **[medium] This move changes MULTI *and* MONITOR semantics. It is the correct change — but nothing in the repo pins it.**

Ground-truthed against `redis-server 7.2.1`: inside `MULTI`, `SETBIT k -1 1`, `SETBIT k 1 5`, `GETBIT k -1`, `SETBIT k <over-ceiling>`, `BITFIELD k GET u8` and `BITFIELD_RO k SET u8 0 1` all reply `+QUEUED`; only `SETBIT k` replies `-ERR wrong number of arguments for 'setbit' command`, after which `EXEC` → `-EXECABORT Transaction discarded because of previous errors.` So the description's "only arity fails at queue time" is right, and moving these out of `schema` is a **fix**.

What is worth pinning is that it is a change at all:

- `CommandExecutor.plan()` runs the schema at `src/core/command-executor.ts:127`, **before** `executePlan()` reaches the transaction policy that queues.
- A plan failure goes through `rawCommandErrorResult` → `ctx.session.markTransactionDirty()` (`command-executor.ts:156`).

So on `origin/main` these errors surfaced at queue time and poisoned the transaction into `EXECABORT`; on this branch they queue and surface inside the `EXEC` array. Second-order, `publishMonitorEvent` only fires from `executePlan` (`command-executor.ts:359`), so a malformed `SETBIT`/`BITFIELD` is now newly visible to `MONITOR` — also correct, also unpinned.

`git grep` over `tests/` and `tests-integration/` finds **no** MULTI/EXEC coverage for `SETBIT`, `GETBIT` or `BITFIELD` at all. Given that the entire justification for restructuring three commands' argument handling is this queue-vs-EXEC distinction, it should have a test.

src/core/transports/resp2/decoder.ts
```diff
@@ -101,8 +121,14 @@ export class Resp2CommandDecoder {
...
 
+      // Redis' primary `proto-max-bulk-len` enforcement point: the header is
+      // judged before a single byte of the payload is read, so an oversized
+      // argument to *any* command is refused at parse time rather than by the
+      // command that would have received it (networking.c,
+      // `processMultibulkBuffer`). `>` and not `>=` — a bulk exactly the size
+      // of the limit is accepted. Verified on redis 6.2.24, 7.2.16 and 8.0.6.
       const length = parseLength(bulkHeader.line, 'bulk')
-      if (length < 0) {
+      if (length < 0 || BigInt(length) > this.maxBulkLength()) {
```
fatal10110, line 174: **[low/medium] This check bounds `$` headers only, and the adjacent multibulk bound is both missing *and* version-specific — so fixing it later will need a compatibility gate.**

Ground-truthed:

```
redis 6.2.14:        *1048577     -> -ERR Protocol error: invalid multibulk length, close
                     *2147483647  -> -ERR Protocol error: invalid multibulk length, close
redis 7.2.1 / 8.0.6: *1048577     -> (no reply, connection open)
                     *2147483647  -> (no reply, connection open)
                     *2147483648  -> -ERR Protocol error: invalid multibulk length, close
PR mock:             *2147483648  -> (no reply, connection open)
```

So 6.2 caps the element count at 1024*1024 and 7.x/8.x only reject past 2^31 — this is exactly the version-specific behaviour `src/core/compatibility/` exists for. Flagging it here because `docs/COMMANDS.md` now says the limit is "enforced in the three places Redis enforces it", which reads as completeness while the sibling bound at the same enforcement point is still absent. If it is out of scope for this PR, an issue referencing the split above would keep the gate from being missed later.

Two more pre-existing wording divergences in this same function, since they sit right next to the new check:

```
redis 7.2.1:  '*2\r\n$3\r\nGET\r\n$5\r\nabcdef\r\n'  -> $-1        connection stays open
PR mock:                                            -> -ERR Protocol error: bulk string not terminated, close

redis 7.2.1:  '*2\r\n$3\r\nGET\r\n+x\r\n'            -> -ERR Protocol error: expected '$', got '+'
PR mock:                                            -> -ERR Protocol error: expected bulk string
```

Real Redis never validates a bulk's trailing CRLF — it just advances past it — so `bulk string not terminated` is an invented error that closes a connection Redis keeps open.

tests-integration/raw-tcp/proto-max-bulk-len-protocol.test.ts
```diff
@@ -0,0 +1,312 @@
...
+    const doomed = await connect()
+    doomed.write(frameWithOversizedTail(OVER_DEFAULT_LIMIT, 'SET', key))
+    await expectThenClose(doomed, INVALID_BULK_LENGTH)
+
+    const survivor = await connect()
+    survivor.write(commandFrame('EXISTS', key))
+    assert.strictEqual((await survivor.readRawFrame()).toString(), ':0\r\n')
+  })
+
+  test('an inline command is not subject to the bulk limit', async () => {
```
fatal10110: **[low] The one test covering inline commands asserts nothing about the limit — and the exemption it names is unobservable on real Redis.**

`PING\r\n` is four bytes. This passes under any implementation, including one that applied the bulk limit to inline arguments, so it would not catch a regression.

More importantly, the exemption cannot be demonstrated on a real server at all. The minimum settable `proto-max-bulk-len` is 1048576:

```
CONFIG SET proto-max-bulk-len 1024
-> ERR CONFIG SET failed (possibly related to argument 'proto-max-bulk-len')
   - argument must be between 1048576 and 9223372036854775807 inclusive
```

and real Redis caps inline requests at `PROTO_INLINE_MAX_SIZE` (64KB), which is always the smaller of the two:

```
redis 7.2.1 / 8.0.6:  write('PING ' + 'a'*70000)  -> -ERR Protocol error: too big inline request, close
PR mock:              write('PING ' + 'a'*70000)  -> ""   connection open, buffers indefinitely
```

So "inline commands are exempt from proto-max-bulk-len" is vacuously true upstream, while the limit that actually applies to inline input is missing here — `parseInlineFrame`'s `readLine` scans for `\r\n` with no bound, so a client can pin memory and park a connection with no reply.

Suggest renaming this test to what it really covers (an inline command still works), and tracking the 64KB inline cap separately.

src/core/transports/resp2/decoder.ts
```diff
@@ -20,39 +24,55 @@ type ParseOutcome =
   | { kind: 'skip'; nextIndex: number }
   | { kind: 'incomplete' }
 
+/**
+ * Redis' compiled-in default for `proto-max-bulk-len`: 512MB. Duplicated from
+ * `RedisServerState` rather than imported, so the transport layer keeps no
+ * dependency on server state — the real value arrives through
+ * {@link Resp2CommandDecoderOptions.maxBulkLength}.
+ */
+const DEFAULT_PROTO_MAX_BULK_LEN = 536870912n
```
fatal10110: **[low] `DEFAULT_PROTO_MAX_BULK_LEN` is now duplicated, and the fallback it guards is unreachable in-repo.**

`src/state/server-state.ts:44` already has the identical `const DEFAULT_PROTO_MAX_BULK_LEN = 536870912n`. Two sources of truth for the same Redis default will drift.

The stated reason — "so the transport layer keeps no dependency on server state" — does not really apply: the dependency is already expressed as the injected `maxBulkLength` getter, and `src/core/transports/attach-session.ts:48` is the only place that constructs an adapter, always passing it. So this default is dead code for every in-repo path and only reachable by an external consumer who constructs the decoder bare.

Either export the constant from `server-state.ts` and import it, or make `maxBulkLength` a required option and delete the fallback.

fatal10110:
Pushed `bc5b559`. Walking the review in order.

### 1. HIGH — derived ceiling dropped `MAX_MATERIALISABLE_LENGTH` — **fixed**

You are right, and the measurement is the proof: I removed the only thing
bounding that allocation and then left the paragraph promising it could not
happen. `assertBitOffsetWithinLimit` now takes the lower of the live setting and
`MAX_MATERIALISABLE_LENGTH`, the same shape `assertWithinProtoMaxBulkLen` uses
for `APPEND`/`SETRANGE`. Your `Buffer.alloc` point is also right — `MAX_LENGTH`
is 2^53-1 on Node 22, so `allocateBitmapBuffer` never fired; it stays only as
the low-memory backstop `allocateStringBuffer` is, and its docblock now says so.

Pinned in both twins by `raising the limit past 512MB does not raise the ceiling
with it`. The **allocation-free** `BITFIELD GET` probe is deliberately the first
assertion, so that with the cap removed the test fails before reaching a `SETBIT`
that would really allocate half a gigabyte. Verified red with the clamp dropped.

### 2. HIGH — pull-based rewrite unpinned — **fixed**

Also correct, and the reason is now obvious: every assertion I wrote put the bad
frame in a write of its own, which both shapes handle identically.

New test: `a CONFIG SET earlier in the same TCP write is already in force` sends
`CONFIG SET proto-max-bulk-len 1048576` and an over-limit frame in **one** TCP
write. I re-applied your revert (batch-drain, new check kept) and confirmed it is
the only failing test — 15 pass, 1 fail — then restored the pull-based loop.

### 3. MEDIUM — BITFIELD op-ordering docblock — **fixed, and the code with it**

Re-ground-truthed; the claim was wrong about `BITFIELD_RO`, whose GET-only check
Redis applies in a *second* pass after everything parses:

```
BITFIELD_RO k SET u8 8388608 1 -> ERR bit offset is not an integer or out of range
BITFIELD_RO k SET u99 0 1      -> ERR Invalid bitfield type...
BITFIELD_RO k SET u8 0 notanint-> ERR value is not an integer or out of range
BITFIELD_RO k NOPE             -> ERR syntax error
BITFIELD_RO k SET u8 0 1 ...   -> ERR BITFIELD_RO only supports the GET subcommand
```

Identical on 6.2.24 and 7.2.16. The mock threw the RO error inline (pre-existing),
so rather than document a rule we did not follow I moved the check to a second
pass. Covered in the new raw-TCP suite. The docblock now states the verified rule
and names the exception.

### 4. MEDIUM — decoder poisoned after `next()` throws — **fixed**

Made explicitly terminal with a documented contract, which is what Redis does
(`CLIENT_CLOSE_AFTER_REPLY`; the stream cannot be resynchronised). The error is
stored, every later `next()` re-throws the same one, `push()` becomes a no-op,
and the buffer is dropped. Silent resumption mid-frame is no longer reachable.

### 6. MEDIUM — MULTI/MONITOR semantics uncovered — **fixed**

New `tests-integration/raw-tcp/bitmap-runtime-errors.test.ts`, on **both**
backends (default limit only, no server-wide mutation): 7 runtime errors that
queue and surface in `EXEC`, 3 arity cases that abort at queue time, a failed
element not stopping the rest of the transaction, and the `BITFIELD_RO` ordering
above. **9 of its 12 tests are red on `origin/main`** and all 12 pass against
real Redis. Raw TCP because `+QUEUED` is the observation, and both clients'
`multi()` are client-side pipelines that cannot show *when* the server rejected.

### 7. LOW/MED — invented strings and the multibulk bound — **partly fixed**

- `Protocol error: expected bulk string` → Redis' real
  `Protocol error: expected '$', got '%c'` (verified: `*2\r\n%3\r\nfoo\r\n`
  answers `got '%'`). Pure wording, no behaviour change, so it landed here.
- Multibulk bound: added the `INT_MAX` ceiling, which is the one every supported
  profile agrees on. The tighter pre-7.0 bound is genuinely version-specific —
  `*1048577` errors on 6.2.24 and is accepted on 7.2.16 — so it needs a
  `FeatureId`. **Deferred to #441.** No memory hazard meanwhile: nothing is
  allocated from the count.
- `Protocol error: bulk string not terminated` is also invented, and matching
  Redis means *deleting* the check — it reads `ll` bytes and skips two, so
  `*1\r\n$3\r\nfooXX` dispatches `foo`. That changes framing for malformed input,
  has no coverage today, and is unrelated to `proto-max-bulk-len`. **Deferred to
  #441**, marked in the code as a known divergence rather than left looking
  correct.

### 8. LOW — duplicated default / unreachable fallback — **fixed**

`maxBulkLength` is now required and the duplicated constant is gone.

### 5. MEDIUM — boundary assertions mock-only — **partly addressed, and I want to push back on the rest**

More now runs against real: the entire new MULTI/`BITFIELD_RO` suite, plus the
refusal side of the bulk boundary, which was already there (`$536870913` at the
default limit, refused by mock and real alike, with no payload sent).

The **accept** side I do not think can move, and I would rather say so than paper
over it. Redis pre-sizes its query buffer straight from the header —
`sdsMakeRoomForExact(querybuf, ll + 2)` for any bulk at or above
`PROTO_MBULK_BIG_ARG` — so asserting `$<exactly the limit>` at the 512MB default
costs the *server* a 512MB allocation, and proving it at a cheaper limit means
`CONFIG SET proto-max-bulk-len`, which is server-wide state on a shared backend.
Given the OOM-kills today, spending 512MB on the shared standalone to prove `>`
rather than `>=` is the wrong trade. It stays mock-only, ground-truthed by hand,
with the reasoning written into the suite header so this does not have to be
rediscovered. Happy to revisit if the harness ever gets a disposable real server.

### On #424

Thanks — that is a real interaction and my tests would not catch it, since they
are raw-TCP only and the flush-then-close guarantee holds on a real socket. I
have not touched `close()`; leaving the `end()`-before-`destroy()` fix to #424.

### Verification

`npm test` 378/378 · `test:integration:mock` 1222/1223 (the one failure is the
known #411 hash-TTL flake, green on a previous full run) · real backend,
targeted: 33 pass / 0 fail / 8 mock-only skips · build, lint, prettier clean.

src/core/transports/resp2/decoder.ts
```diff
@@ -15,44 +20,85 @@ export class Resp2ParseError extends Error {
...
+  private readonly maxBulkLength: () => bigint
+  /**
+   * Set once a protocol error is raised. A RESP stream cannot be resynchronised
+   * after one — Redis' own parser marks the client `CLIENT_CLOSE_AFTER_REPLY`
+   * and never reads another command from it — so the decoder is deliberately
+   * terminal rather than silently resuming mid-frame.
+   */
+  private fatalError: Resp2ParseError | null = null
+
+  constructor(options: Resp2CommandDecoderOptions) {
```
fatal10110, line 53: **[medium · new in bc5b559] Making this option required is a second breaking change to published `/core` API, and neither break has the CHANGELOG entry the repo requires.**

`CHANGELOG.md` is explicit: `js-redis-server/core` (`src/internal.ts`) "is smaller but still published API, so removals from it are marked **BREAKING (`/core`)**", and "Entries are added when a change lands on `main` ... the PR body is not a durable home for a breaking-change note." `src/internal.ts` re-exports `Resp2CommandDecoder`.

This PR now changes that exported class twice:

1. `push(chunk)` returns `void` instead of `{ frames, error }` (round 1), so every existing caller silently gets `undefined`.
2. `constructor(options)` is now required (this round). That's a type error for TS callers and a crash for JS callers. Against this head:

```
new Resp2CommandDecoder()
-> TypeError: Cannot read properties of undefined (reading 'maxBulkLength')
```

Removing the duplicated default was the right call; it just needs recording. Please add an `[Unreleased]` entry marking both changes **BREAKING (`/core`)**, e.g. "`Resp2CommandDecoder` is now pull-based (`push()` buffers, `next()` yields one frame and is terminal after a protocol error) and requires `{ maxBulkLength }`."

src/commands/bitmaps.ts
```diff
@@ -52,12 +91,37 @@ function parseBitOffset(token: Buffer | undefined): number {
...
+ * allocate unbounded memory inside the test process — `Buffer.alloc` is no
+ * backstop, since `buffer.constants.MAX_LENGTH` is 2^53-1 on Node 22 and so
+ * never throws for any size this ceiling would otherwise admit.
+ */
+function assertBitOffsetWithinLimit(
+  offset: number,
+  protoMaxBulkLen: bigint,
+): void {
+  const limit =
+    protoMaxBulkLen < MAX_MATERIALISABLE_LENGTH
```
fatal10110: **[low · new in bc5b559] The clamp also covers reads that never allocate, so the fix adds a divergence that doesn't make anything safer. The new test then pins it.**

`MAX_MATERIALISABLE_LENGTH` exists to stop the process from materialising a huge buffer, but this function also gates `GETBIT` and `BITFIELD[_RO] GET`. Neither allocates in the mock (`getBit` returns 0 past the end of the buffer) or in Redis. Ground-truthed with `proto-max-bulk-len 629145600` on private servers:

```
                                  6.2.24   7.2.1   8.0.6   this PR
GETBIT gk 4294967296              0        0       0       ERR bit offset is not an integer or out of range
BITFIELD_RO gk GET u8 4294967296  0        0       0       ERR bit offset ...
BITFIELD gk GET u8 4294967296     0        0       0       ERR bit offset ...
EXISTS gk / STRLEN gk             0 / 0    0 / 0   0 / 0   0 / 0
```

`strings.ts`, the precedent this mirrors, clamps only the writers (`APPEND`/`SETRANGE`, via `assertWithinProtoMaxBulkLen`), not `GETRANGE`. Suggest applying the clamp only where an allocation follows: `SETBIT`, and the SET/INCRBY offsets in BITFIELD (GET ops keep the raw limit). Then drop the `GETBIT` and BITFIELD GET assertions from `raising the limit past 512MB does not raise the ceiling with it`. The test can stay red-first and allocation-free by probing the write path with an invalid value, so which error wins decides the outcome and nothing is allocated either way:

```
SETBIT k 4294967296 2                   clamped   -> ERR bit offset is not an integer or out of range
                                        unclamped -> ERR bit is not an integer or out of range
BITFIELD k SET u8 4294967296 notanint   clamped   -> ERR bit offset ...
                                        unclamped -> ERR value is not an integer or out of range
```

(Offset-before-value ordering is already verified against real Redis in this PR.)

src/core/transports/resp2/decoder.ts
```diff
@@ -93,16 +145,27 @@ export class Resp2CommandDecoder {
       }
 
       if (prefix !== 0x24) {
-        throw new Resp2ParseError('Protocol error: expected bulk string')
+        // Redis' exact wording, which echoes the offending byte:
+        // `Protocol error: expected '$', got '%c'`. Verified on 6.2.24 and
+        // 7.2.16 — `*2\r\n%3\r\nfoo\r\n` answers `got '%'` and closes.
+        throw new Resp2ParseError(
+          `Protocol error: expected '$', got '${Buffer.from([prefix]).toString('latin1')}'`,
```
fatal10110: **[nit · new in bc5b559] Bytes ≥ 0x80 are echoed as 2-byte UTF-8, while Redis sends the raw byte.**

`toString('latin1')` gives a one-char JS string, but the reply is UTF-8 encoded on the way out:

```
'*1\r\n\xE9x\r\n'   6.2.24 / 7.2.1 / 8.0.6 -> -ERR Protocol error: expected '$', got '<0xE9>'       (1 byte)
                    this PR               -> -ERR Protocol error: expected '$', got '<0xC3 0xA9>'  (2 bytes)
```

Everything else here matches byte for byte on all three versions (`got '%'`, and CR/LF sanitised to `got ' '`). Only non-ASCII input is affected, and the connection closes right after, so a short known-divergence note alongside the #441 ones is enough.

fatal10110:
## Re-review of #431 at `bc5b559`

Previously reviewed: `bb2137a`. I re-derived every claim below myself rather than relying on commit messages. Checks were run against:

- my own `redis-server` **6.2.24 / 7.2.1 / 8.0.6** on private ports, with every `CONFIG SET` restored
- a mock built from this head
- mutation runs, where I reverted each fix and re-ran the suites to show the new tests actually catch its absence
- `#424`+`#431` merged in a scratch tree, for the cross-PR check

### Prior findings

| # | Sev | Finding | Status | Evidence |
|---|---|---|---|---|
| 1 | high | Unbounded SETBIT/BITFIELD allocation past 512MB | ✅ **Fixed** | Clamp removed → exactly the new test fails in each twin (12 pass / 2 fail), peak RSS 158 MB, so the red case really is allocation-free. Clamp now also covers reads that don't need it → new comment below. |
| 2 | high | Pull-based loop unpinned | ✅ **Fixed** | Restoring the batch-drain loop → 15 pass / **1 fail**, and the one failure is `a CONFIG SET earlier in the same TCP write is already in force`. |
| 3 | med | BITFIELD op-ordering docblock wrong | ◐ **Partly fixed** | BITFIELD_RO second pass fixed and confirmed. Test cases that separate a second pass from a per-op check after parsing match on 6.2.24 / 7.2.1 / 8.0.6 (`BITFIELD_RO k SET u8 0 1 GET u99 0` → type error; `… GET u8 4294967296` → offset error; `INCRBY u8 0 1 NOPE` → syntax error). **Not fixed:** the per-op arity half. `BITFIELD k GET x9`, `BITFIELD k SET x9 0`, `BITFIELD k GET u8 0 SET x9` and `BITFIELD_RO k GET x9` all return `ERR syntax error` on all three versions, but `ERR Invalid bitfield type…` here. Redis checks `remargs` before it reads the type. The rewritten docblock ("type, then offset, then value, per op") still claims the opposite. Pre-existing, non-blocking: add the remargs pre-check or soften the sentence. |
| 4 | med | Decoder poisoned after throw; `/core` API break with no CHANGELOG | ◐ **Partly fixed** | Terminal state implemented as described (`fatalError`, buffer cleared, `push` ignored, `next` re-throws). Pipelined `PING` + unbalanced-quotes inline still answers `+PONG` then the error, then closes. **CHANGELOG still missing, and there are now two breaks.** New comment on `decoder.ts:42`. |
| 5 | med | Load-bearing boundary assertions mock-only | ✅ **Pushback accepted** | Correction to my round-1 comment: `setupRawStandalone()` only spawns a private server locally. In CI it connects to the shared docker-compose standalone, and the runner has no `redis-server` binary, so a cheap accept-side test on the real backend isn't available. The claims were ground-truthed by hand (round 1, five engines). Minor leftover, non-blocking: still no wide-type (`i64`) BITFIELD assertion, although `BITFIELD k GET i64 4294967288` would run on real at the default limit with no allocation. |
| 6 | med | MULTI / MONITOR change unpinned | ✅ **Fixed** | New `raw-tcp/bitmap-runtime-errors`: **9/12 red on `origin/main`** (the 3 arity cases stay green, as they should), **12/12 green** on real 6.2.24 / 7.2.1 / 8.0.6 and in CI `raw-real`. The MONITOR visibility change is still unpinned (minor). |
| 7 | low/med | Multibulk bound + invented error wording | ✅ **Fixed / deferral acceptable** | `*2147483648` → `invalid multibulk length`, close; `*2147483647` → waits. Byte-identical to 7.2.1 / 8.0.6. `got '%'` matches, and CR/LF is sanitised to `got ' '` exactly like Redis. **#441 deferrals are acceptable:** 6.2.24 does reject `*1048577` while 7.2.1 / 8.0.6 accept it, so it truly needs a `FeatureId`. The CRLF check is a pre-existing framing change unrelated to #415. Both are marked in code. New nit on non-ASCII echo below. |
| 8 | low | Inline test asserts nothing; 64KB inline cap missing | ❌ **Not fixed** | The test still sends `PING\r\n`, and the 64KB `too big inline request` cap isn't tracked anywhere: #441 doesn't mention inline. Non-blocking. Please add it to #441. |
| 9 | low | Duplicate `DEFAULT_PROTO_MAX_BULK_LEN` | ✅ **Fixed** | Constant removed and option made required. That fix is what causes the second `/core` break in finding 4. |

### New findings (inline)

- **[medium]** `decoder.ts:42`: the now-required constructor option plus the `push()` return change are two breaks to `js-redis-server/core` with no **BREAKING (`/core`)** entry. `CHANGELOG.md` says one is required when the change lands. A bare `new Resp2CommandDecoder()` now throws `TypeError: Cannot read properties of undefined (reading 'maxBulkLength')`.
- **[low]** `bitmaps.ts:117`: the clamp also gates `GETBIT` / `BITFIELD GET`, which never allocate. With the limit at 600MB, real 6.2 / 7.2 / 8.0 answer `0` (EXISTS 0, STRLEN 0) where this PR errors, and the new test pins that divergence. The comment includes an allocation-free red probe for write-only clamping.
- **[nit]** `decoder.ts:152`: a byte ≥ 0x80 is echoed as 2-byte UTF-8 where Redis sends the raw byte.

### Cross-PR (#424, `c5d22ad`)

No semantic problem. Merged #424 + #431 in a scratch tree: typecheck clean, #424's transport tests 19/19, #431's protocol suites 30/30. Over the virtual wire, #431's write-then-close delivers `+PONG\r\n-ERR Protocol error: invalid bulk length\r\n` to every consumer style I tried: flowing `'data'`, async iterator, and a **paused client reading 50 ms after teardown**. #424's `end()` → `'finish'`/`setImmediate` → destroy doesn't lose the error.

### Tests / CI

- Mock, this head: 79/79 targeted integration, 277/277 unit.
- Real, this head (private servers): new suites 30 pass / 8 mock-only skips on each of 6.2.24, 7.2.1, 8.0.6.
- CI: all green, including `ioredis-real`, `node-redis-real`, `raw-real`.

### Verdict: **not yet, but one small change away**

Both highs are fixed and properly pinned, and the MULTI change is now proven red→green against three real versions. The remaining blocker is doc-only: add the **BREAKING (`/core`)** CHANGELOG entry for `Resp2CommandDecoder`, which is repo policy. Once that's in, this ships; it doesn't need another review round. The rest are non-blocking and can land here or in #441: read-path clamp, BITFIELD arity wording/docblock, inline test and 64KB cap, non-ASCII echo.

fatal10110:
Pushed **`a183734`**. Everything from the second review, in order.

### Blocking — `/core` break unrecorded — **fixed**

`CHANGELOG.md` → `[Unreleased]` → new `### Changed` with a **BREAKING (`/core`)**
entry for `Resp2CommandDecoder` covering both breaks — `push()` now returns
`void` (frames come from `next()`, which throws the `Resp2ParseError` and is
terminal after it), and the constructor requires `{ maxBulkLength }`, so a bare
`new Resp2CommandDecoder()` throws the `TypeError` you quoted. It has a
before/after migration snippet and notes that `Resp2SessionAdapter` /
`attachSession` keep their signatures. I also added **Fixed** entries for the
protocol-layer enforcement (flagging that the server now *closes the
connection*) and the bit-offset ceiling, since both are consumer-visible.

### `bitmaps.ts:117` — clamp on reads — **fixed**

Right: the cap now applies only where an allocation follows — `SETBIT`, and
BITFIELD `SET` / `INCRBY` offsets. `GETBIT` and BITFIELD `GET` keep Redis' own
ceiling, as `strings.ts` caps the writers but not `GETRANGE`. Re-verified your
table on my own 6.2.24 / 7.2.16 / 8.0.6 at `proto-max-bulk-len 629145600`: reads
answer `0` / `[0]`, and your two write probes answer the bit / value error
(unclamped). I used your probes as-is, plus the `INCRBY` variant — the mock now
answers the offset error on the write probes and `0` / `[0]` on the reads.
Nothing in the test allocates, capped or not.

### Finding 3 — BITFIELD per-op arity — **fixed in code, not just the comment**

Added the `remargs` pre-check before any argument is parsed (`GET` ≥ 2, `SET` /
`INCRBY` ≥ 3, `OVERFLOW` ≥ 1). All your cases answer `ERR syntax error` on
6.2.24, 7.2.16 and 8.0.6 and now on the mock, and `GET x9 0` / `SET x9 0 1` still
reach the type error. The docblock now says: argument count first, then type,
offset and value.

### Finding 8 — inline test / 64KB cap — **both**

The test is renamed to what it covers and now does something a regression could
break: an inline `PING` with a 60KB argument is served (byte-identical on all
three versions). The 64KB cap is added to **#441** with ground truth. One nuance
worth recording there: the cap bounds an *unterminated* inline buffer. `'ECHO ' +
70000*'a' + '\r\n'` in one write is served on all three versions, and only the
no-CRLF form trips `too big inline request`, so a test has to use that
deterministic form.

### `decoder.ts:152` — non-ASCII echo — **fixed**

`Resp2ParseError` has a new `messageBytes` field and the adapter pre-encodes the
reply from it, so `0xE9` goes out as one byte. CR/LF still map to a space. This
is additive: `new Resp2ParseError(message)` is unchanged. A new test compares
the bytes as hex, because a UTF-8 decode would hide exactly this difference.

### Also from your notes

- **i64 at the default limit** — `BITFIELD k GET i64 4294967288` → `[0]` and
  `… 4294967296` → error, in both twins. It runs on real with no allocation.
- **MONITOR visibility of malformed SETBIT/BITFIELD** — still unpinned. I left
  it: it is a second-order effect of the same move the MULTI suite already pins,
  and I would rather not grow this PR further. Happy to take it in a follow-up.

### Verification

Mutation-checked, one fix at a time: clamping reads, dropping the arity
pre-check, and writing the parse error as a string each fail **exactly their own
test** (both twins for the read clamp), and nothing else.

`npm test` 378/378 · `test:integration:mock` **1225/1225** · new suites against
my **private** servers, not the shared fixtures: 32 pass / 0 fail / 8 mock-only
skips on each of 6.2.24, 7.2.16 and 8.0.6 · build, lint, prettier clean. I did
not run any node-redis suite against the shared cluster.

docs/COMMANDS.md
```diff
@@ -107,21 +107,37 @@ surface.
...
+>      limit lowers the ceiling with it.
+>
 >   Raising it past **512MB**, Redis' own default, does not raise what the mock
->   will allocate: beyond that the size error is returned rather than a buffer
->   the test process may not survive producing. At or below the default the
+>   will allocate — for the bit-offset ceiling (3) just as for `APPEND` and
+>   `SETRANGE` (2). Beyond the default the mock answers an error rather than
+>   producing a buffer the test process may not survive: the size error for a
+>   string that would grow too large, and the ordinary bit-offset error for an
+>   offset the raised setting would otherwise admit. At or below the default the
```
fatal10110, line 139: **[low · new in a183734] This paragraph describes the round-2 clamp, not the one now in the code.**

Since this commit, only offsets that allocate are capped at 512MB. Reads keep Redis' own ceiling. But the paragraph still says that past the default "the mock answers ... the ordinary bit-offset error for an offset the raised setting would otherwise admit", and it groups the whole bit-offset ceiling (3) with `APPEND`/`SETRANGE` as allocation-capped.

Against this head, with `proto-max-bulk-len 629145600` (same result on 6.2.24 / 7.2.1 / 8.0.6):

```
GETBIT rk 4294967296                    -> :0                (was ERR in round 2)
BITFIELD_RO rk GET u8 4294967296        -> *1 :0
SETBIT rk 4294967296 2                  -> ERR bit offset ... (write: capped, documented divergence)
BITFIELD rk SET u8 4294967296 notanint  -> ERR bit offset ...
```

`CHANGELOG.md` already has the right wording ("capped at 512MB for operations that allocate (`SETBIT`, and `BITFIELD`'s `SET` / `INCRBY`) ... reads are not capped, because they never allocate"). Suggest mirroring that here, e.g. "... and the ordinary bit-offset error for a `SETBIT` or `BITFIELD SET`/`INCRBY` offset the raised setting would otherwise admit; `GETBIT` and `BITFIELD GET` keep Redis' ceiling, since they never allocate."

fatal10110:
## Round 3 re-review of #431 at `a183734`

Previously reviewed: `bc5b559`, one new commit. I re-derived every claim below myself rather than relying on the commit message or the reply. Checks were run against:

- my own `redis-server` **6.2.24 / 7.2.1 / 8.0.6** on private ports, with every `CONFIG SET` restored
- a mock built from this head, with its RSS tracked
- four mutation runs, where I undid each fix and re-ran the suites to see what breaks

### Round-2 findings

| Round-2 finding | Status | Evidence |
|---|---|---|
| **[medium · blocker]** `/core` breaking changes missing from CHANGELOG | ✅ **Fixed** | `[Unreleased] → Changed` has a **BREAKING (`/core`)** entry covering both breaks: `push()` → `void` with `next()` throwing and terminal, and the required `{ maxBulkLength }` with the exact `TypeError`. The migration snippet **type-checks verbatim** against `src/internal.ts`. The Fixed entries cover the protocol-layer check, including that the connection now closes, plus the derived ceiling and the MULTI/EXEC change. |
| **[low]** Clamp also gated allocation-free reads | ✅ **Fixed** | With `proto-max-bulk-len 629145600`, reads past the old 512MB clamp return the same bytes as 6.2.24 / 7.2.1 / 8.0.6: `GETBIT` at 4294967296 and at 5033164791, `BITFIELD_RO GET u8`, `BITFIELD GET i64`, and a mixed `SET u8 0 1 GET u8 4294967296` → `*2 :0 :0` with `STRLEN` 1. **Mock RSS unchanged, 99 → 99 MB.** Writes stay capped (documented). Mutations: clamping reads again, or removing the clamp entirely, each fail **only** `raising the limit past 512MB raises the read ceiling but not the write ceiling` in each twin. With the clamp removed, peak RSS was 198 MB against a 193 MB baseline, so the red case allocates nothing. The value-error probes also ran on real 64 MB containers without OOM. |
| **[nit]** Non-ASCII byte echoed as 2-byte UTF-8 | ✅ **Fixed** | `*1\r\n\xE9x\r\n` → `…got '\xE9'\r\n` is **hex-identical** on all four. CR and LF → `got ' '` and `+` → `got '+'` also match. Mutation (send it as a string again) fails only `a non-'$' element prefix is echoed as the raw byte`. `messageBytes` is additive; the only in-repo consumer of `Resp2ParseError` is the adapter, and `.message` is unchanged for every existing error. |
| **[prior #3]** BITFIELD per-op arity half | ✅ **Fixed** | A **35-case battery** matches byte for byte across 6.2.24 / 7.2.1 / 8.0.6 / mock. It covers short ops (`GET x9`, `SET x9 0`, `INCRBY x9 0`, `GET u8 0 SET x9`, `GET`, `OVERFLOW`), OVERFLOW (short, bad mode, bad mode before a later op, trailing), the BITFIELD_RO second pass (`SET u8 0 1 GET u99 0` → type error; `SET u8 0 1 GET u8` → syntax error; `INCRBY u8 0 1` → GET-only), and offset vs missing value (`SET u8 4294967296` → syntax; `SET u8 4294967296 1 GET x9` → offset). Mutation (remove the `remaining` gate) fails only `a BITFIELD op missing arguments is a syntax error before its type is read`. The docblock is now accurate. |
| **[prior #5 leftover]** No wide-type coverage | ✅ **Fixed** | `BITFIELD GET i64 4294967288` → `[0]` and `…4294967296` → offset error, at the default limit, in both twins. Identical on all three real versions. |
| **[prior #8]** Inline test asserted nothing | ✅ **Fixed** | The test now sends `PING` with a 60,000-byte argument and asserts the 60,010-byte bulk reply. Same bytes on all three versions. The 64KB cap is tracked in #441 (see note below). |

### New findings (inline)

- **[low]** `docs/COMMANDS.md:139`: the paragraph added in round 2 still says reads past a raised limit get the bit-offset error. Since this commit they return `0`, like Redis. CHANGELOG already has the right wording; mirror it here.

### Deferrals: acceptable

- **#441** tracks three items. (1) The pre-7.0 multibulk bound is truly version-specific (6.2.24 rejects `*1048577`; 7.2.1 / 8.0.6 accept it), so it needs a `FeatureId`. (2) The invented unterminated-bulk error: matching Redis means changing framing for malformed input, unrelated to #415. (3) The 64KB inline cap. All three are pre-existing and marked in code or in the issue.
- **Correction for #441 item 3:** "a >64KB inline line that arrives complete is served" doesn't hold in general. With the CRLF included and sent in one write, **100,000- and 300,000-byte inline lines are rejected by all three versions** (`-ERR Protocol error: too big inline request`, then close). Redis reads inline input 16KB at a time and checks the cap before the CRLF arrives, so the 70KB case is served only because of where that read boundary falls. The issue's recommended model ("no `\n` in the first 64KB buffered") and its deterministic-test advice are right. Only that sentence and the 70KB example should be qualified, so the eventual fix doesn't admit arbitrarily long complete lines.
- **MONITOR visibility** of malformed `SETBIT`/`BITFIELD` is still untested. The behaviour itself is correct (it runs inside `execute`, as in Redis). Minor; fine to leave.

### Tests / CI

- CI: 15 pass, 1 skip (deploy), including `ioredis-real`, `node-redis-real`, `raw-real`.
- Unit: **378/378** (`npm test`).
- Mock, the five relevant suites: 46/46.
- Real, the new test files on private servers: **32 pass / 0 fail / 8 mock-only skips** on each of 6.2.24, 7.2.1, 8.0.6, matching the claim exactly.
- Cross-PR (#424): the close/flush path is unchanged this round, and last round's merged-tree check still stands.

### Verdict: **ship** ✅

The blocker and every round-2 item are fixed, each fix is pinned by a test that fails without it (mutation-verified), and every wire claim reproduces byte for byte on 6.2.24 / 7.2.1 / 8.0.6. The one new item is a single doc sentence and doesn't block merge. It can go in this PR or a follow-up.

PR #462, opened 2026-09-23
fix: round HTTL hash-field TTL up to the next second (#432)

### LINKED ISSUE #432: HTTL/HPTTL: hash-field TTL rounds to nearest second; real Redis uses ceiling
Found by a reviewer ground-truthing PR #423 against real Redis 8.0.6. **Pre-existing on `main`**, out of that PR's diff.

## Problem

`ttlSeconds()` (`src/commands/helpers.ts:81`) computes `Math.round((expiresAt - Date.now()) / 1000)` and is shared by both key-level `TTL` and hash-field `HTTL`. Real Redis uses **two different rules**: round-to-nearest for key TTL, **ceiling** for hash-field TTL.

The helper's own comment ("Redis rounds remaining time to the nearest second, `(ms+500)/1000`") is correct for `TTL` and wrong for `HTTL`.

## Evidence

Measured on real redis 8.0.6 via atomic Lua, so there is zero round-trip skew:

```
HPEXPIRE k <ms> FIELDS 1 f ; HTTL ; HPTTL
  20ms   -> HTTL 1   HPTTL 20
  400ms  -> HTTL 1   HPTTL 400
  999ms  -> HTTL 1   HPTTL 999
  1000ms -> HTTL 1   HPTTL 1000
  1001ms -> HTTL 2   HPTTL 1001
  1400ms -> HTTL 2   HPTTL 1400
  2400ms -> HTTL 3   HPTTL 2400
=> CEILING
```

Same server, key-level TTL for contrast:

```
  PX 400  -> TTL 0   PTTL 400
  PX 500  -> TTL 1   PTTL 500
  PX 1400 -> TTL 1   PTTL 1400
  PX 1500 -> TTL 2   PTTL 1500
  PX 2400 -> TTL 2   PTTL 2400
=> round-to-nearest
```

Mock, identical sequence:

```
  20ms   -> HTTL [0]   <- real 1
  400ms  -> HTTL [0]   <- real 1
  500ms  -> HTTL [1]
  1001ms -> HTTL [1]   <- real 2
  1400ms -> HTTL [1]   <- real 2
  2400ms -> HTTL [2]   <- real 3
```

`HPTTL` matches throughout; only the seconds-granularity reply diverges.

## Failure scenario

Any consumer asserting `HTTL` on a sub-second or fractional-second hash-field TTL gets a different integer from the mock than from real Redis — 0 vs 1, 1 vs 2, 2 vs 3.

## Why it survived

The existing assertion in `tests-integration/ioredis/hash/field-persist.test.ts:64` is `assert.ok(seconds[2] >= 0 && seconds[2] <= 1)`, a range wide enough to swallow the divergence at every TTL the suite currently uses. Note this was equally true at the old 20ms TTL as at the 500ms one PR #423 adopts — the PR neither introduces nor worsens the gap.

## Proposal

Give `ttlSeconds` a ceiling variant and route `HTTL`/`HPTTL` (and `HGETEX`/`HGETDEL`-adjacent TTL replies, if any report seconds) through it. Keep `TTL`/`PTTL` on round-to-nearest. Correct the helper's comment to name which rule belongs to which command family.

## Acceptance

- `HTTL` matches real Redis at 20ms, 400ms, 999ms, 1000ms, 1001ms, 1400ms and 2400ms.
- Regression test pins it. A TTL of `1200` with `assert.strictEqual(seconds[2], 2)` fails the mock today and passes real Redis — that is the minimal discriminating case.
- Key-level `TTL` rounding is unchanged, pinned by the existing tests.

fatal10110:
## Summary
- `ttlSeconds()` in `src/commands/helpers.ts` (round-to-nearest, `(ms+500)/1000`) was shared by key-level `TTL` and hash-field `HTTL`. Real Redis uses **ceiling** (`(ms+999)/1000`) for hash-field TTL, so e.g. `HPEXPIRE ... 1200` → `HTTL` returned 1 on the mock vs 2 on real Redis.
- Added `hashFieldTtlSeconds()` (ceiling) and routed `HTTL` through it. `TTL` stays on `ttlSeconds()` (round-to-nearest); `HPTTL` is ms-granularity and unaffected. No other command reports hash-field TTL in seconds.
- Fixed the helper comments so each one names the command family its rounding rule applies to.
- Tightened the loose `seconds[2] >= 0 && seconds[2] <= 1` assertion in the `field-persist` tests (500ms field → `HTTL` is exactly 1) in both the ioredis and node-redis suites.
- Not version-specific, so no compatibility gate.

Closes #432

## Test plan
- [x] New test `HTTL rounds hash-field TTLs up to the next second, unlike key TTL` (ioredis + node-redis `hash/field-persist.test.ts`): `HPEXPIRE 1200`/`2400` → `HTTL [2, 3]`, and `SET PX 1200` → `TTL 1` as a check that key TTL is unchanged. Failed on mock before the fix (`[1, 2]`).
- [x] Passes against real Redis (docker-compose cluster), both clients
- [x] Passes on mock after the fix
- [x] `npm run build`, `npm run lint`, `npm test` (462/462), `npm run test:integration:mock` (1250 pass, 0 fail)
- [x] `tests-integration/ioredis/hash/*` against real: 33/33


tests-integration/ioredis/hash/field-persist.test.ts
```diff
@@ -127,6 +127,40 @@ describe(`Hash Commands Integration (${testRunner.getBackendName()})`, () => {
...
+    const hashKey = `${tag}:hash`
+    const stringKey = `${tag}:string`
+    let directClient: Redis | undefined
+
+    try {
+      assert.ok(redisClient)
+      directClient = await connectToSlotOwner(redisClient, hashKey)
+      await directClient.hset(hashKey, 'a', '1', 'b', '1')
+      assert.deepStrictEqual(
+        await directClient.hpexpire(hashKey, '1200', 'FIELDS', '1', 'a'),
```
fatal10110: [medium] The ceiling assertion only has 200ms of slack. With `HPEXPIRE 1200`, `HTTL` returns 2 only while more than 1000ms remain, so the test fails if 200ms or more pass between the HPEXPIRE and the HTTL read. That can happen on a loaded CI runner, during a GC pause, or on the real cluster with a slot redirect or reconnect. Choose TTLs that keep ceiling and round-to-nearest apart with more margin. For example, `1450`/`2450` gives ceil = 2/3 and round = 1/2 at t=0, and the ceiling reading stays correct for about 450ms. For the key-TTL check, keep a value such as `PX 1450` (round gives 1 while 500ms or more remain). Update the "Holds as long as under 200ms elapse" comment to match. Apply the same change to the node-redis twin at `tests-integration/node-redis/hash/field-persist.test.ts:118-133`.

Ground truth on redis 7.4.11 and 8.0.6 (pipelined, a few ms elapsed): `HPEXPIRE 1200/2400/500/1600` then `HTTL` gives `2 3 1 2`, and `HPTTL` gives `1196 2397 498 1598`. The expected values are right; only the timing margin is thin.

tests-integration/node-redis/hash/field-persist.test.ts
```diff
@@ -105,6 +105,41 @@ describe(`Hash Commands Integration (node-redis, ${testRunner.getBackendName()})
...
+    const tag = `{hash-field-ttl-ceil:${randomKey()}}`
+    const hashKey = `${tag}:hash`
+    const stringKey = `${tag}:string`
+    let directClient: RedisClientType | undefined
+
+    try {
+      directClient = await connectToNodeRedisSlotOwner(redisClient, hashKey)
+      await directClient.hSet(hashKey, { a: '1', b: '1' })
+      assert.deepStrictEqual(
+        await directClient.hpExpire(hashKey, 'a', 1200),
```
fatal10110: [medium] (twin of the ioredis comment) `hpExpire 1200` leaves only 200ms of slack before `hTTL` stops returning 2. Widen the TTLs, for example to 1450/2450, so the test does not become flaky under load.

src/commands/helpers.ts
```diff
@@ -77,12 +77,18 @@ export function typeName(type: RedisDataTypeName | null): string {
...
+// ((ms+500)/1000), not ceil/floor — matches EXPIRETIME and real TTL behavior.
 export function ttlSeconds(expiresAt: number): number {
-  // Redis rounds remaining time to the nearest second ((ms+500)/1000),
-  // not ceil/floor — matches EXPIRETIME and real TTL behavior.
   return Math.max(0, Math.round((expiresAt - Date.now()) / 1000))
 }
 
+// Hash-field TTL (HTTL): Redis rounds remaining time *up* to the next second
+// ((ms+999)/1000), so any sub-second remainder reports 1, not 0 (#432).
+export function hashFieldTtlSeconds(expiresAt: number): number {
```
fatal10110: [medium] No deterministic test pins the rounding edges. The only proof of the ceiling behavior is a wall-clock integration test, so boundary regressions go untested. Examples: `ceil` vs `(ms+999)/1000` at exactly 1000ms remaining should give 1 and 1001ms should give 2, and a sub-second remainder should give 1, never 0. Add a small `node:test` unit test that fixes `Date.now()` with `mock.timers.enable({ apis: ["Date"], now })` (or `mock.method(Date, "now", ...)`). It should assert `hashFieldTtlSeconds` at 1ms→1, 999ms→1, 1000ms→1, 1001ms→2 and 2400ms→3, and `ttlSeconds` at 499ms→0, 500ms→1 and 1499ms→1. That fixes the two rounding rules side by side.

Ground truth (redis 7.4.11 / 8.0.6): `HTTL` gives 1200ms→2, 500ms→1, 1600ms→2 (ceiling). `TTL` gives 1200ms→1, 1600ms→2 (nearest).

src/commands/helpers.ts
```diff
@@ -77,12 +77,18 @@ export function typeName(type: RedisDataTypeName | null): string {
...
+  return Math.max(0, Math.ceil((expiresAt - Date.now()) / 1000))
```
fatal10110: [low] This call reads `Date.now()` a second time. `hash.getField()` already checked expiry with its own clock read (`isHashFieldExpired`), then this helper reads `Date.now()` again. If the clock moves past `expiresAt` between the two reads, the result clamps to `0`. Real Redis never returns `HTTL 0` for a field: a live field with any sub-second remainder reports 1, and an expired one reports -2. The window is tiny, so this is a nit. You could take `now` as a parameter, or map a result of `<= 0` to `-2` in `hashFieldTtls`.

src/commands/helpers.ts
```diff
@@ -77,12 +77,18 @@ export function typeName(type: RedisDataTypeName | null): string {
   return type ?? 'none'
 }
 
+// Key-level TTL: Redis rounds remaining time to the nearest second
+// ((ms+500)/1000), not ceil/floor — matches EXPIRETIME and real TTL behavior.
 export function ttlSeconds(expiresAt: number): number {
```
fatal10110: [low] Naming nit: `ttlSeconds` now applies only to key TTL and EXPIRETIME, while its generic name sits next to `hashFieldTtlSeconds`. Renaming it to `keyTtlSeconds` would make the split obvious to the next caller, since it has only one call site (`src/commands/keys.ts:151`). The comment itself is correct: on redis 7.4.11/8.0.6, `PEXPIREAT k 4102444800300` then `EXPIRETIME` gives `4102444800`, and `...700` gives `4102444801` (nearest).

src/commands/hashes.ts
```diff
@@ -26,7 +26,7 @@ import {
   ok,
   parseIntegerToken,
   ttlMilliseconds,
-  ttlSeconds,
+  hashFieldTtlSeconds,
```
fatal10110: [low] Import order: `hashFieldTtlSeconds` was dropped into the slot where `ttlSeconds` was, so the named-import list is no longer alphabetical (`... parseIntegerToken, ttlMilliseconds, hashFieldTtlSeconds`). Move it after `bulk`.

fatal10110:
## Review of #462 (head `96b0752`): **NOT YET**

The fix is correct and matches real Redis. It is held back only by two test-quality findings, both medium.

**Checked against real Redis 7.4.11 and 8.0.6, same results on both:**
- `HTTL` rounds up. `HPEXPIRE` values of 1200 / 2400 / 500 / 1600 ms give `2 3 1 2`.
- `TTL` rounds to the nearest second. `PX` values of 1200 / 1600 give `1 2`.
- `EXPIRETIME` rounds to the nearest second, as the helper comment says. `PEXPIREAT …800300` gives `…800` and `…800700` gives `…801`.
- The rounding is the same on 7.4 and 8.0, so no compatibility gate is needed.

**Tests:**
- The two `field-persist` files pass on the mock backend (8/8).
- With `origin/main`'s `src`, the new and tightened tests fail on the mock, as they should.
- `npm test` passes 462/462 and `tsc --noEmit` is clean.

| Severity | Location | Finding |
|---|---|---|
| medium | `tests-integration/{ioredis,node-redis}/hash/field-persist.test.ts` (1200/2400 TTLs) | The ceiling test has only 200ms of slack. Use values such as 1450/2450. |
| medium | `src/commands/helpers.ts:88` | No deterministic unit test covers the rounding edges. Add one that mocks `Date.now()`. |
| low | `src/commands/helpers.ts:89` | A second `Date.now()` read after the expiry check can clamp the result to `HTTL 0`, which real Redis never returns. |
| low | `src/commands/helpers.ts:82` | Consider renaming `ttlSeconds` to `keyTtlSeconds`. |
| low | `src/commands/hashes.ts:29` | The import list is out of alphabetical order. |

**Out-of-scope follow-up:** the mock does not implement `HEXPIRETIME`/`HPEXPIRETIME`. Real Redis rounds `HEXPIRETIME` **up** too: `…800300` and `…800700` both give `…801`. That differs from `EXPIRETIME`'s nearest-second rounding, so a future implementation should use `hashFieldTtlSeconds`-style ceiling rounding.

tests-integration/ioredis/hash/field-persist.test.ts
fatal10110: Fixed in e7a9f1e. The TTLs are now 1450/2450ms. Ceiling gives [2, 3] and round-to-nearest would give [1, 2], so the test still tells the two rules apart, now with about 450ms of slack. The key-TTL check uses SET PX 1450 → TTL 1, with about 950ms of slack. I also updated the comment.

tests-integration/node-redis/hash/field-persist.test.ts
fatal10110: Fixed in e7a9f1e, same change as the ioredis twin: hpExpire 1450/2450 → hTTL [2, 3], SET PX 1450 → TTL 1, and the comment now says 450ms instead of 200ms.

src/commands/helpers.ts
fatal10110: Fixed in e7a9f1e. The new unit test tests/ttl-rounding.test.ts passes a fixed `now` to each helper. It pins hashFieldTtlSeconds at 1→1, 999→1, 1000→1, 1001→2, 1999→2, 2000→2, 2001→3, and 0/-1/-1500→0. It pins keyTtlSeconds at 1→0, 499→0, 500→1, 1499→1, 1500→2, and 0/negative→0. ttlMilliseconds is checked too, including the clamp at 0.

src/commands/helpers.ts
fatal10110: Fixed in e7a9f1e. All three helpers now take an optional `now = Date.now()`. hashFieldTtls reads the clock once, before the map and so before getField's own expiry check, and passes that value to hashFieldTtlSeconds and ttlMilliseconds. A field that is still live at getField's later check has expiresAt > now, so it can no longer report 0.

src/commands/helpers.ts
fatal10110: Done in e7a9f1e. The rename was small: ttlSeconds had one call site outside helpers.ts (TTL in src/commands/keys.ts), so it is now keyTtlSeconds, and its comment names TTL.

src/commands/hashes.ts
fatal10110: Fixed in e7a9f1e. The named imports in hashes.ts are back in alphabetical order, and I fixed the same order in keys.ts after the rename.

src/commands/helpers.ts
```diff
@@ -77,14 +77,23 @@ export function typeName(type: RedisDataTypeName | null): string {
...
-  return Math.max(0, Math.round((expiresAt - Date.now()) / 1000))
+// Key-level TTL (TTL): Redis rounds remaining time to the nearest second
+// ((ms+500)/1000), not ceil/floor — matches EXPIRETIME and real TTL behavior.
+export function keyTtlSeconds(expiresAt: number, now = Date.now()): number {
+  return Math.max(0, Math.round((expiresAt - now) / 1000))
 }
 
-export function ttlMilliseconds(expiresAt: number): number {
-  return Math.max(0, expiresAt - Date.now())
+// Hash-field TTL (HTTL): Redis rounds remaining time *up* to the next second
```
fatal10110: [low] Missing CHANGELOG entry. This PR changes behavior clients can see: `HTTL` now returns 2 instead of 1 for 1200ms remaining. `CHANGELOG.md` says entries are added "when a change lands on `main`", and recent fix PRs (#427, #431, #410) each added a `### Fixed` bullet under `[Unreleased]`. Add one here too, for example: "`HTTL` rounds hash-field TTLs up to the next second, as real Redis does, instead of to the nearest second ([#462], [#432])". Include the link references at the bottom of the file. The `ttlSeconds` to `keyTtlSeconds` rename needs no entry, because `src/commands/helpers.ts` is not exported from `src/index.ts` or `src/internal.ts`.

fatal10110:
## Re-review of #462 (head `e7a9f1e`, after `96b0752`): **SHIP**

All 6 earlier findings are fixed:

| Severity | Finding | Status |
|---|---|---|
| medium (x2: ioredis + node-redis) | Ceiling test had 200ms of slack | Fixed. It now uses 1450/2450 and `PX 1450` in both clients. The comments are correct: HTTL holds while under 450ms elapse, and TTL = 1 holds while under 950ms elapse. |
| medium | No deterministic test for the rounding edges | Fixed. `tests/ttl-rounding.test.ts` passes an explicit clock and covers 1/999/1000/1001/1999/2000/2001ms plus the ≤0 clamp for the ceiling helper, and 1/499/500/1499/1500ms for nearest-second. |
| low | Second `Date.now()` read could produce `HTTL 0` | Fixed. `hashFieldTtls` takes one `now` before `getField`. Any field still live at the later expiry check has `expiresAt > now`, so HTTL is at least 1 and HPTTL above 0. |
| low | `ttlSeconds` naming | Fixed. It is renamed to `keyTtlSeconds`. No stale references remain, and the helper is not part of the `/core` export surface. |
| low | Import order | Fixed in both `hashes.ts` and `keys.ts`. |

New finding: **low**, a missing `### Fixed` CHANGELOG entry (inline comment on `helpers.ts:86`). It does not block the merge.

Tests on `e7a9f1e`:
- `tests/ttl-rounding.test.ts`: 3/3 pass.
- Both `field-persist` integration files on the mock backend: 8/8 pass.
- TTL-related integration files (key/expire, string/set, hash/hgetex, hash/hsetex, key/workflow): 36/36 pass.
- `npm test`: 465/465 pass.
- `tsc --noEmit`, eslint and prettier are clean on the changed files.

The real-Redis results from the first round still apply, since the values are the same. On 7.4.11 and 8.0.6, HTTL rounds up and TTL/EXPIRETIME round to the nearest second.

Issue #437, opened 2026-09-22
The 6.2 subcommand syntax-error template reaches PUBSUB only; 10 other containers emit 7.0-era arity wording on redis-6.2


fatal10110:
Found while reviewing #430 (the #413 wording consolidation).

## Problem

Redis has **two** container-error templates, and #430 only routed one of them everywhere.

1. `unknown subcommand '%.128s'. Try %s HELP.` — dispatch-level, the subcommand is not in the command table. #430's `unknownSubcommandError`, wired into all 11 containers.
2. `addReplySubcommandSyntaxError` — a *known* subcommand given arguments it cannot use. #430 added `subcommandSyntaxError` for it, but wired it into **PUBSUB only**.

Both flipped case at Redis 7.0. Before 7.0 there was no command-table entry for a subcommand, so template 2 served the wrong-arity case for **every** container; from 7.0 a wrong-arity known subcommand is an ordinary arity error naming `'<container>|<subcommand>'`.

So on the `redis-6.2` profile this server emits the 7.0-era arity wording for everything except `PUBSUB CHANNELS`/`NUMSUB`/`SHARDCHANNELS`. Verified against real 6.2.24 and 8.0.6:

| command | real 6.2.24 | real 8.0.6 | this server, `redis-6.2` profile |
|---|---|---|---|
| `CONFIG RESETSTAT x` | `ERR Unknown subcommand or wrong number of arguments for 'RESETSTAT'. Try CONFIG HELP.` | `ERR wrong number of arguments for 'config\|resetstat' command` | the 8.0 form |
| `SCRIPT LOAD` | `... for 'LOAD'. Try SCRIPT HELP.` | `... for 'script\|load' command` | the 8.0 form |
| `ACL WHOAMI x` | `... for 'WHOAMI'. Try ACL HELP.` | `... for 'acl\|whoami' command` | the 8.0 form |
| `SLOWLOG LEN x` | `... for 'LEN'. Try SLOWLOG HELP.` | `... for 'slowlog\|len' command` | the 8.0 form |
| `COMMAND COUNT x` | `... for 'COUNT'. Try COMMAND HELP.` | `... for 'command\|count' command` | the 8.0 form |
| `CLIENT GETNAME x` | `... for 'GETNAME'. Try CLIENT HELP.` | `... for 'client\|getname' command` | the 8.0 form |
[...]

Issue #435, opened 2026-09-22
MULTI: an unknown container subcommand is queued instead of rejected at queue time on redis-7.0+ profiles


fatal10110:
Found while reviewing #430 (the #413 wording consolidation). #430 gated the unknown-subcommand *wording* on `error.unknown-subcommand-wording`; this is the **dispatch timing** the same Redis 7.0 change brought, which is still ungated.

## Problem

Redis 7.0 moved container commands into the command table, so from 7.0 an unknown subcommand fails **command lookup**, which happens before a command is queued. In a transaction it is therefore rejected at `MULTI` queue time and dirties the transaction; 6.2 had no such lookup and queued it like anything else.

Captured from real servers, one connection, `MULTI` → `CONFIG BOGUS` → `EXEC`:

| | `CONFIG BOGUS` reply | `EXEC` reply |
|---|---|---|
| real 6.2.24 | `+QUEUED` | `*1` then `-ERR Unknown subcommand or wrong number of arguments for 'BOGUS'. Try CONFIG HELP.` |
| real 8.0.6 | `-ERR unknown subcommand 'BOGUS'. Try CONFIG HELP.` | `-EXECABORT Transaction discarded because of previous errors.` |
| this server (every profile) | `+QUEUED` | `*1` then the error |

So the mock is right for `redis-6.2` and wrong for every profile from `redis-7.0` up — including the default `redis-8.0`. A client that uses `MULTI` to batch a typo'd container subcommand sees the whole transaction execute here and get discarded on real Redis.

Affects every container this server dispatches at execute time: `CONFIG`, `CLIENT`, `ACL`, `SLOWLOG`, `COMMAND`, `PUBSUB`, `SCRIPT`, `FUNCTION`.

## Why it is not in #430

#430 is a wording consolidation: it replaced 15 copies of a message with one profile-aware helper. Moving the *dispatch* of container subcommands from execute time to plan time is a pipeline change — it needs the registry to resolve `config|get`-style names, or an equivalent pre-execution check on `CommandPlan` — and would have doubled that PR.

## Sketch

[...]

Issue #436, opened 2026-09-22
XGROUP/XINFO abort a transaction at queue time on the redis-6.2 profile, where real 6.2 queues them


fatal10110:
Found while reviewing #430 (the #413 wording consolidation). This is the mirror of #435: there, container subcommands are dispatched too *late* for `redis-7.0+`; here, `XGROUP` and `XINFO` are dispatched too *early* for `redis-6.2`.

## Problem

`XGROUP` and `XINFO` resolve their subcommand inside a `t.custom(...)` schema parser (`createXgroupSchema` in `src/commands/streams/xgroup.ts`, `createXinfoSchema` in `src/commands/streams/xinfo.ts`). Schema parsing runs in `CommandExecutor.plan()`, which in a transaction runs at **queue** time. So an unknown `XGROUP`/`XINFO` subcommand dirties the transaction and `EXEC` answers `-EXECABORT` — on every profile, including `redis-6.2`, where real 6.2 queues it and reports the error inside the `EXEC` array.

Real 6.2.24, one connection:

```
MULTI            -> +OK
CONFIG BOGUS     -> +QUEUED
EXEC             -> *1 then -ERR Unknown subcommand or wrong number of arguments for 'BOGUS'. Try CONFIG HELP.
```

6.2 has no command-table entry for container subcommands at all, so nothing can fail before queueing. `XGROUP`/`XINFO` behave the same way there.

For `redis-7.0+` the current early dispatch happens to land on the right *timing* — but for the wrong reason (a parser, not a command-table lookup), and it is not gated, so it cannot be made correct for 6.2 without addressing where the dispatch lives.

## Second symptom: 6.2 checks the key before the subcommand

The same difference in *where* the subcommand is resolved shows up outside transactions too. Real 6.2 looks the key up first whenever one is present, and only then rejects the subcommand; 7.0+ rejects the subcommand at lookup and never touches the key. Captured on 6.2.24 and 8.0.6:

| command | real 6.2.24 | real 8.0.6 |
|---|---|---|
| `XINFO BOGUS <missing>` | `ERR no such key` | `ERR unknown subcommand 'BOGUS'. Try XINFO HELP.` |
| `XINFO BOGUS <string key>` | `WRONGTYPE Operation against a key holding the wrong kind of value` | same as above |
| `XINFO BOGUS <stream>` | `ERR Unknown subcommand or wrong number of arguments for 'BOGUS'. Try XINFO HELP.` | same as above |
[...]

PR #463, opened 2026-09-23
fix: name xinfo|<sub> and xgroup|<sub> in subcommand arity errors (#438)

### LINKED ISSUE #438: XINFO/XGROUP arity errors name 'xinfo' instead of 'xinfo|stream'
Found while reviewing #430 (the #413 wording consolidation).

## Problem

`XINFO`'s arity errors name the container instead of the dispatched subcommand, on every profile:

| | real 8.0.6 | this server |
|---|---|---|
| `XINFO STREAM` | `ERR wrong number of arguments for 'xinfo\|stream' command` | `ERR wrong number of arguments for 'xinfo' command` |

`XGROUP` has the same defect (`XGROUP CREATE` → `'xgroup'` against real `'xgroup|create'`), noted in a comment in `tests-integration/raw-tcp/command-errors-subcommand.test.ts` where the assertion is deliberately omitted for that reason.

## Cause

Both resolve their subcommand inside a `t.custom(...)` schema parser, and a parser only gets `ctx.commandName` — the container. Every `throw new WrongNumberOfArgumentsError(ctx.commandName)` in `src/commands/streams/xinfo.ts` (lines 41, 47, 65, 71, 83, 95) and the equivalents in `xgroup.ts` therefore name `'xinfo'` / `'xgroup'`.

This is the same root cause as #384 part 1: `plan.definition.name` cannot express a dispatched subcommand. Containers that dispatch in `execute()` (`CONFIG`, `CLIENT`, …) pass the literal `'config|get'` to `expectArgCount` and get it right by hand.

## Fix

The narrow version is mechanical — the parser knows which branch it is in, so pass `'xinfo|stream'` etc. explicitly, exactly as the execute-time containers already do. The broader version is #384's naming mechanism, which would remove the hand-written literals everywhere. Either is fine; the narrow one is a handful of lines and unblocks #437's 6.2 arity sweep, which needs a correct subcommand name to gate on.

## Acceptance

- `XINFO STREAM`, `XINFO GROUPS`, `XINFO CONSUMERS` and the `XGROUP` subcommands produce `wrong number of arguments for '<container>|<subcommand>' command` on `redis-7.0+`, matching real Redis byte-for-byte.
- The assertions currently commented out in `tests-integration/raw-tcp/command-errors-subcommand.test.ts` (`XGROUP CREATE`, `XINFO STREAM`) are enabled.

fatal10110:
## Summary
- `XINFO`/`XGROUP` resolve their subcommand inside a `t.custom(...)` schema parser, which only gets `ctx.commandName` (the container name). Because of that, every arity error said `'xinfo'`/`'xgroup'`, while real Redis says `'xinfo|stream'`, `'xgroup|create'`, and so on.
- Narrow fix from the issue: each branch of the parser now passes its own `'<container>|<subcommand>'` name, the same way the containers that dispatch in `execute()` (`CONFIG`, `CLIENT`, …) already do. A bare `XINFO`/`XGROUP` with no subcommand still names just the container, which matches real Redis.
- Some of the old `WrongNumberOfArgumentsError(ctx.commandName)` calls were not arity errors in real Redis. These are the option-list failures: a `COUNT`/`ENTRIESREAD` with no value, a stray option token, or extra arguments after `FULL COUNT n`. Real 8.0.6 answers all of them with `addReplySubcommandSyntaxError` (`unknown subcommand or wrong number of arguments for 'STREAM'. Try XINFO HELP.`). Renaming them to `'xinfo|stream'` would still have been wrong, so they now use the existing profile-aware `subcommandSyntaxError` helper. The two `RedisSyntaxError` calls for the same kind of input were changed the same way.
- The subcommand is still resolved in the schema parser; moving it out is #435/#436.

**Profiles:** the `<c>|<sub>` arity name is now emitted on every profile. That is the same form every other container already uses on `redis-6.2`. Real 6.2 uses a different template for all of them, which is the #437 sweep, and this change gives that sweep a correct subcommand name to gate on. The option-list syntax errors go through `subcommandSyntaxError`, which already applies the 6.2/7.0 wording gate. 6.2's `xinfoReplyWithStreamInfo` uses `addReplySubcommandSyntaxError` for these same inputs.

Closes #438

## Test plan
- [x] `tests-integration/raw-tcp/command-errors-subcommand.test.ts`: enabled the `XGROUP CREATE` / `XINFO STREAM` arity assertions that were previously left out, and added every `XGROUP`/`XINFO` subcommand in the too-few and too-many cases, plus lower-case input and the bare container. Also added a new test for the option-list syntax errors, run against a live stream and group with per-run keys. Expected strings were captured from real 8.0.6.
- [x] The new tests failed on mock before the fix and passed against real Redis 8.0.
- [x] After the fix they pass on mock and real.
- [x] `npm run build`, `npm run lint`, `npm test` (462/462), `npm run test:integration:mock` (1249 pass, 0 fail)
- [x] Real backend: `ioredis/stream`, `node-redis/stream`, `raw-tcp` (242/242)


tests-integration/raw-tcp/command-errors-subcommand.test.ts
```diff
@@ -342,19 +343,84 @@ describe(`Raw TCP unknown-subcommand errors (${testRunner.getBackendName()})`, (
...
+        args,
+        `-ERR wrong number of arguments for '${name}' command\r\n`,
+      )
+    }
+  })
 
+  // Past the arity table, an XINFO STREAM / XGROUP CREATE|SETID option list
+  // the container cannot use — a dangling `COUNT`/`ENTRIESREAD`, a stray token,
+  // trailing junk — is `addReplySubcommandSyntaxError`, not an arity error.
+  // Real Redis looks the key up first, so this runs against a live stream and
```
fatal10110: [medium] This comment gets the ordering wrong, and it hides a divergence.

- **XGROUP:** "Real Redis looks the key up first" is wrong here. XGROUP parses its options *before* it looks up the key (real 8.0.6):
  ```
  XGROUP CREATE <missing> g $ BOGUS    -> ERR unknown subcommand or wrong number of arguments for 'CREATE'. Try XGROUP HELP.
  XGROUP SETID <missing> g $ MKSTREAM  -> ... for 'SETID'. Try XGROUP HELP.
  XGROUP CREATE <string> g $ BOGUS     -> ... for 'CREATE' ... (no WRONGTYPE)
  ```
- **XINFO STREAM:** the statement is true for real Redis, but this server does **not** do it. The mock raises the syntax error in the schema parser, before any lookup. The test works around that by creating a live stream first, and the comment doesn't say so:
  ```
  real 8.0.6 / 7.0.15 / 6.2.24:  XINFO STREAM <missing> x -> ERR no such key
                                 XINFO STREAM <string> x  -> WRONGTYPE Operation against a key holding the wrong kind of value
  this PR:                       XINFO STREAM <missing> x -> ERR unknown subcommand or wrong number of arguments for 'STREAM'. Try XINFO HELP.
  ```
  A later maintainer could take this as the rule for both commands and move XGROUP's option check behind the key lookup, which would be wrong.

Suggested fix: say that XGROUP validates options before the key lookup and XINFO STREAM validates them after, and that the mock does not match XINFO yet (see the follow-up comment on xinfo.ts:63). Also, "an ... option list the container cannot use" should say *subcommand*, which is the wording the source comments use.

Lens: Redis wire-compat, confirmed against 8.0.6, 7.0.15 and 6.2.24. Also raised by the architecture reviewer.

src/commands/streams/xinfo.ts
```diff
@@ -41,34 +44,40 @@ function createXinfoSchema() {
...
-        if (!key) throw new WrongNumberOfArgumentsError(ctx.commandName)
+        if (!key) throw new WrongNumberOfArgumentsError('xinfo|stream')
         let cursor = index + 2
         let full = false
         let count: number | null = null
 
         if (cursor < input.length) {
           if (input[cursor].toString().toUpperCase() !== 'FULL') {
-            throw new RedisSyntaxError()
+            throw syntaxError()
```
fatal10110: [low] Out of scope and not blocking. The behavior predates this PR, but no open issue covers it, so it needs a new one.

The option-list errors are now worded correctly, but they are still raised in the schema parser. Real 7.0+ raises them at execute time: after the key lookup for XINFO, and at execute time for XGROUP too. That causes two differences from real Redis:

1. **Key checked after the options.** `XINFO STREAM <missing> x` (also `FULL COUNT`, `FULL COUNT 1 x`, `COUNT 5`) returns `ERR no such key` on real 8.0.6, and `WRONGTYPE` on a string key. The mock returns the STREAM syntax error.
2. **MULTI aborts.** In a transaction, both XINFO and XGROUP option-list errors abort at queue time here. Real 8.0.6 queues them:
   ```
   MULTI -> OK
   XINFO STREAM s x -> QUEUED
   XGROUP SETID s g $ MKSTREAM -> QUEUED
   EXEC -> 1) ERR unknown subcommand or wrong number of arguments for 'STREAM'. Try XINFO HELP.
           2) ERR unknown subcommand or wrong number of arguments for 'SETID'. Try XGROUP HELP.
   ```
   The mock gives an immediate error, then `EXECABORT`. Arity errors such as `XINFO STREAM` with no key *are* rejected at queue time on real Redis, so those are already right.

This was wrong before the PR as well (`-ERR syntax error`, also at parse time). #435 and #436 cover only the timing of *unknown-subcommand* errors, so neither covers this case.

Lens: Redis wire-compat (confirmed against 8.0.6). The architecture reviewer flagged the same thing as a follow-up.

tests-integration/raw-tcp/command-errors-subcommand.test.ts
```diff
@@ -342,19 +343,84 @@ describe(`Raw TCP unknown-subcommand errors (${testRunner.getBackendName()})`, (
...
     await expectReply(
       conn,
-      ['CONFIG', 'GET'],
-      "-ERR wrong number of arguments for 'config|get' command\r\n",
+      ['XGROUP', 'CREATE', key, 'g', '$', 'MKSTREAM'],
+      '+OK\r\n',
     )
+
+    const cases: [string[], string, string][] = [
+      [['XINFO', 'STREAM', key, 'x'], 'XINFO', 'STREAM'],
```
fatal10110: [low] Every case in the syntax-error table sends the subcommand in upper case, so nothing checks that the error echoes the subcommand exactly as the client sent it.

Real 8.0.6 echoes the client's casing:
```
xinfo stream s x               -> ... for 'stream'. Try XINFO HELP.
XINFO sTrEaM s x               -> ... for 'sTrEaM'. Try XINFO HELP.
xgroup setid s g $ mkstream    -> ... for 'setid'. Try XGROUP HELP.
```
The code is correct today, because it passes `rawSubcommand`. But a refactor that passed the upper-cased `subcommand` instead would still pass this test. Suggest adding a row such as `[['xinfo', 'stream', key, 'x'], 'XINFO', 'stream']`.

Lens: raised by both reviewers.

tests-integration/raw-tcp/command-errors-subcommand.test.ts
```diff
@@ -342,19 +343,84 @@ describe(`Raw TCP unknown-subcommand errors (${testRunner.getBackendName()})`, (
...
+      [['XINFO', 'STREAM', key, 'FULL', 'x'], 'XINFO', 'STREAM'],
+      [['XINFO', 'STREAM', key, 'FULL', 'COUNT'], 'XINFO', 'STREAM'],
+      [['XINFO', 'STREAM', key, 'FULL', 'BOGUS', '1'], 'XINFO', 'STREAM'],
```
fatal10110: [low] `FULL BOGUS 1` fails at the same branch as `FULL x` on line 404, the `!== 'COUNT'` check after FULL, so it adds no coverage.

`XINFO STREAM key COUNT 1` with no `FULL` is not tested, and it is a plausible client mistake. Real 8.0.6 answers it with `ERR unknown subcommand or wrong number of arguments for 'STREAM'. Try XINFO HELP.`, and the mock matches. Consider replacing this row with that case.

Lens: architecture, with the real-Redis reply checked by the wire-compat reviewer.

src/commands/streams/xgroup.ts
```diff
@@ -57,13 +61,21 @@ function createXgroupSchema() {
...
+        subcommandSyntaxError('XGROUP', rawSubcommand, ctx.profile)
 
       if (subcommand === 'CREATE' || subcommand === 'SETID') {
         const key = input[index + 1]
         const group = input[index + 2]
         const rawId = input[index + 3]?.toString()
         if (!key || !group || rawId === undefined) {
-          throw new WrongNumberOfArgumentsError(ctx.commandName)
+          throw new WrongNumberOfArgumentsError(
+            subcommand === 'CREATE' ? 'xgroup|create' : 'xgroup|setid',
```
fatal10110: [low] Nit. Each branch writes the subcommand-to-name mapping twice: here, and again for `value.subcommand` around line 105. The same happens at lines 133-137 and 140-143 for CREATECONSUMER/DELCONSUMER.

Hoisting it once removes the duplication and keeps the literal type, with no cast:
```ts
const name = subcommand === 'CREATE' ? 'create' : 'setid'
throw new WrongNumberOfArgumentsError(`xgroup|${name}`)
```
Optional, since other containers also use plain literals.

Lens: architecture.

src/commands/streams/xgroup.ts
```diff
@@ -57,13 +61,21 @@ function createXgroupSchema() {
         throw new WrongNumberOfArgumentsError(ctx.commandName)
       }
       const subcommand = rawSubcommand.toString().toUpperCase()
+      // A parser only knows the container (`ctx.commandName`), so arity errors
+      // for a dispatched subcommand spell out `xgroup|<sub>` themselves, as
+      // real Redis 7.0+ does (#438). An option list the subcommand cannot use
+      // is real Redis' `addReplySubcommandSyntaxError`, not an arity error.
+      const syntaxError = () =>
```
fatal10110: [low] Nit. The `syntaxError` closure, and its twin at xinfo.ts:51, is allocated on every parse. Only the CREATE/SETID branch here and the STREAM branch in XINFO use it. Moving it into those branches, or making it a module-level function that takes `(rawSubcommand, profile)`, would make that clear.

Lens: architecture.

tests-integration/raw-tcp/command-errors-subcommand.test.ts
```diff
@@ -342,19 +343,84 @@ describe(`Raw TCP unknown-subcommand errors (${testRunner.getBackendName()})`, (
...
+      [['XGROUP', 'DESTROY', key], 'xgroup|destroy'],
+      [['XGROUP', 'DESTROY', key, 'g', 'x'], 'xgroup|destroy'],
+      [['XGROUP', 'CREATECONSUMER', key, 'g'], 'xgroup|createconsumer'],
+      [
+        ['XGROUP', 'CREATECONSUMER', key, 'g', 'c', 'x'],
+        'xgroup|createconsumer',
+      ],
+      [['XGROUP', 'DELCONSUMER', key, 'g'], 'xgroup|delconsumer'],
+      [['XGROUP', 'DELCONSUMER', key, 'g', 'c', 'x'], 'xgroup|delconsumer'],
+      [['XINFO'], 'xinfo'],
```
fatal10110: [low] Out of scope and not blocking; noting it as a follow-up. XGROUP and XINFO have no `HELP` subcommand, so the HELP arity row that real Redis has can't be tested here:
```
real 8.0.6 / 7.0.15:  XGROUP HELP x -> ERR wrong number of arguments for 'xgroup|help' command
                      XINFO HELP x  -> ERR wrong number of arguments for 'xinfo|help' command
this server:          XGROUP HELP x -> ERR unknown subcommand 'HELP'. Try XGROUP HELP.
```
The mock's reply tells the client to run the very command it just rejected. Implementing `XGROUP HELP`/`XINFO HELP` is a new feature, as CONFIG, CLIENT and the others already have, so it belongs in its own issue.

Lens: Redis wire-compat.

fatal10110:
## Review of 1a61bce: NOT YET (one medium finding; every other finding is low)

This PR fixes what #438 asks for. The only thing blocking it is one misleading test comment, which is a quick fix.

**Checked against real Redis (8.0.6, 7.0.15, 6.2.24):**
- **Arity errors:** all 18 rows of the arity table match real 8.0.6 and 7.0.15 byte for byte. The name comes back lower-case however the client sent it, and a bare `XINFO` / `XGROUP` names only the container.
- **Option-list errors:** all 9 syntax-error rows match real 8.0.6 when the key and group exist. That is the `addReplySubcommandSyntaxError` reply, which echoes the subcommand as the client sent it.
- **6.2:** the option-list errors match 6.2.24 through the helper's 6.2 branch. The arity naming on 6.2 is still wrong, but that is the #437 sweep and not part of this PR.

**Tests:**
- `command-errors-subcommand.test.ts` failed on base source (2 failures) and passes on the PR head.
- These pass on the mock backend with the PR applied: `command-errors-subcommand`, `cluster-stream-watch`, `profile-gates`, `tests/compatibility/behavior`, and the ioredis and node-redis stream suites. That is 154 passed, 0 failed, 2 skipped.
- `tsc --noEmit` and eslint on the changed files are clean.

|---|---|---|---|
| medium | test.ts:391 | The comment says "Real Redis looks the key up first". That is wrong for XGROUP, which parses options before the key lookup. It also hides the fact that the mock does not do the key lookup first for XINFO STREAM. | Redis + arch |
| low | xinfo.ts:63 | Pre-existing, out of scope; needs a new issue. Option-list errors are raised at parse time. So `XINFO STREAM <missing> x` gives the syntax error where real Redis says `ERR no such key`, and inside MULTI these errors return EXECABORT where real 7.0+ replies QUEUED and reports the error in the EXEC reply. | Redis + arch |
| low | test.ts:403 | No test sends a lower-case or mixed-case subcommand in the syntax-error table, so echoing the subcommand as the client sent it is not tested. | Redis + arch |
| low | test.ts:406 | `FULL BOGUS 1` repeats `FULL x`. `COUNT n` without FULL is not tested. | arch |
| low | xgroup.ts:77 | The subcommand-to-name ternaries are duplicated in each branch. | arch |
| low | xgroup.ts:68 | The `syntaxError` closure is allocated on every parse, although only some branches use it. | arch |
| low | test.ts:370 | Out of scope. XGROUP/XINFO `HELP` is not implemented, so the `'xgroup|help'` arity row cannot be tested. | Redis |

Out-of-scope follow-ups, all pre-existing and confirmed on 8.0.6:
- `ENTRIESREAD -1` is rejected here but accepted by real Redis.
- `XINFO STREAM k FULL COUNT -1` is rejected here but accepted by real Redis.
- XGROUP on a missing key: real Redis says "requires the key to exist". This server returns NOGROUP or `:0`, or "Invalid stream ID" when the ID is bad.
- `ENTRIESREAD` is not version-gated on the 6.2 profile.

tests-integration/raw-tcp/command-errors-subcommand.test.ts
fatal10110: Fixed in e00c892. I rewrote the comment and checked each claim against real 8.0.6. XGROUP parses its options before it looks up the key, so `XGROUP CREATE <missing> g $ BOGUS` gives the CREATE syntax error. That case is now a row in the table, using a per-run missing key. XINFO STREAM looks up the key first (`XINFO STREAM <missing> x` gives `ERR no such key`). This server checks the options in the parser, so it answers the syntax error instead. The comment now names that as a known divergence and explains that the XINFO rows use a live stream so both orders give the same reply. The PR body lists it as a follow-up. The wording now says "subcommand", not "container".

tests-integration/raw-tcp/command-errors-subcommand.test.ts
fatal10110: Fixed in e00c892. Added mixed-case rows `XINFO sTrEaM <key> x` and `XGROUP cReAtE <key> g2 $ BOGUS`. Real 8.0.6 echoes both exactly as sent (`'sTrEaM'` / `'cReAtE'`), and the mock now matches.

tests-integration/raw-tcp/command-errors-subcommand.test.ts
fatal10110: Fixed in e00c892. Replaced `FULL BOGUS 1` with `XINFO STREAM <key> COUNT 1` (no `FULL`). Real 8.0.6 answers it with the STREAM syntax error. I also rewrote the STREAM option parse to follow real Redis' argument-count check: it accepts nothing, `FULL`, or `FULL COUNT <n>`, and every other shape is the syntax error. That leaves a single throw site.

src/commands/streams/xgroup.ts
fatal10110: Fixed in e00c892. Each branch now computes `name` once (`'create' | 'setid'`, `'createconsumer' | 'delconsumer'`) and uses it for both `xgroup|${name}` and `value.subcommand`.

src/commands/streams/xgroup.ts
fatal10110: Fixed in e00c892. Removed both closures. XGROUP CREATE/SETID now leaves the option loop and throws `subcommandSyntaxError(...)` once if `cursor !== input.length`. XINFO STREAM throws it once from the validity check. Token comparison goes through a module-level `isToken` helper, so nothing is allocated per parse.

src/commands/streams/xinfo.ts
fatal10110: Agreed, out of scope here. It is listed under Follow-ups in the PR body for a new issue: XINFO checks options before the key (`no such key`/`WRONGTYPE` expected), and XINFO/XGROUP option-list errors abort MULTI at queue time where real Redis queues them. The comment above the syntax-error test also names the XINFO key-order divergence.

tests-integration/raw-tcp/command-errors-subcommand.test.ts
fatal10110: Agreed, out of scope. `XGROUP HELP` and `XINFO HELP` are listed under Follow-ups in the PR body. Once they exist, the `xgroup|help` and `xinfo|help` arity rows can be added to this table.

src/commands/streams/xinfo.ts
```diff
@@ -41,35 +48,28 @@ function createXinfoSchema() {
...
-          count = parseNonNegativeInteger(rawCount)
-          cursor += 2
-        }
+        if (!key) throw new WrongNumberOfArgumentsError('xinfo|stream')
 
-        if (cursor !== input.length) {
-          throw new WrongNumberOfArgumentsError(ctx.commandName)
+        // `[FULL [COUNT <count>]]`: nothing, `FULL`, or `FULL COUNT <count>`.
+        const options = input.slice(index + 2)
+        const valid =
```
fatal10110: [low] Optional, not blocking. Rewriting the STREAM option check as three accepted shapes changes one reply, and the new reply is the right one: real Redis checks how many arguments there are before it parses COUNT. The tests do not cover this reply yet.

```
real 8.0.6:  XINFO STREAM s FULL COUNT abc x -> ERR unknown subcommand or wrong number of arguments for 'STREAM'. Try XINFO HELP.
             XINFO STREAM s FULL COUNT abc   -> ERR value is not an integer or out of range
1a61bce:     XINFO STREAM s FULL COUNT abc x -> ERR value is not an integer or out of range   (COUNT was parsed before the trailing check)
e00c892:     XINFO STREAM s FULL COUNT abc x -> STREAM syntax error                            (matches real)
```
Consider adding a `FULL COUNT abc x` row to the syntax table so nobody reverts this by accident. I checked the XGROUP rewrite (loop now uses `break`, one throw after it) and it behaves the same as before, including `ENTRIESREAD abc BOGUS`, which gives the integer error on real 8.0.6 as well.

Lens: Redis wire-compat (checked on 8.0.6).

fatal10110:
## Re-review of e00c892: SHIP

I diffed `1a61bce..e00c892`. None of my earlier findings at high or medium severity is still open.

| earlier finding | status |
|---|---|
| [medium] test comment on key-lookup order | **fixed**: the comment now says XGROUP checks options first and XINFO checks the key first, and names the divergence. The row `XGROUP CREATE <missing> g $ BOGUS` is added. |
| [low] option-list errors raised at parse time (missing key, MULTI) | deferred: recorded under follow-ups in the PR body |
| [low] casing of the echoed subcommand not tested | **fixed**: `sTrEaM` and `cReAtE` rows added |
| [low] `FULL BOGUS 1` duplicates another row / `COUNT` without FULL not tested | **fixed** |
| [low] duplicated subcommand-to-name ternaries | **fixed**: `name` is computed once |
| [low] per-parse closure | **fixed**: the closure is gone |
| [low] XGROUP/XINFO HELP not implemented | deferred: recorded under follow-ups in the PR body |

**New:** one [low] finding, which is optional. The rewrite of the STREAM options now answers `FULL COUNT abc x` with the syntax error, as real 8.0.6 does (before, it gave the integer error). No test pins that yet. The XGROUP rewrite behaves the same as before.

**Checked on real 8.0.6:** every new or changed row, plus the precedence cases listed below. The server was a private container, which I have since stopped.
- `FULL COUNT abc` gives the integer error.
- `FULL COUNT abc x` gives the syntax error.
- `ENTRIESREAD abc BOGUS` gives the integer error.
- `SETID ... ENTRIESREAD 1 MKSTREAM` gives the syntax error.

**Tests (mock):**
- The subcommand-error, stream-watch, profile-gates, compatibility-behavior, and ioredis/node-redis stream suites: 154 passed, 0 failed, 2 skipped.
- `tsc` and eslint are clean.

Issue #370, opened 2026-09-21
fix/refactor: derive COMMAND INFO arity/keys from definitions (mock reports arity -1 for most commands)


fatal10110:
## Problem
A command's metadata has three overlapping sources of truth, and the mock is **already wrong** because of it.

Verified against a real `redis-server` vs the mock (`COMMAND INFO`):

| cmd | real (arity, first key) | mock |
|---|---|---|
| `lpush` | `-3, 1` | `-1, 0` |
| `zadd` | `-4, 1` | `-1, 0` |
| `hset` | `-4, 1` | `-1, 0` |
| `get` | `2, 1` | `2, 1` (has a hand-written block) |

Only commands with a hand-written `introspection` block (~29 blocks in `src/commands`) report correct `arity`/`firstKey`/`lastKey`/`keyStep`/`categories`. Everything else falls back in `createCommandInfoFromIntrospection` (`src/commands/command.ts:349`) to `arity: -1` and empty key specs. Clients that use `COMMAND INFO` for key positions (ioredis, node-redis cluster routing, tooling) get wrong data for most commands.

Meanwhile each definition already declares `schema` (arity/shape), `flags`, and `keys(args)`, so the hand-written block restates information that exists elsewhere.

A related instance of the same smell: `WRITE_EVENT_OVERRIDES` / `DELETE_EVENT_OVERRIDES` / `GENERIC_WRITE_COMMANDS` in `src/state/keyspace-notifier.ts` are name→event tables kept far from the commands they describe.

## Proposal
1. **Derive `arity` from the schema.** Investigate whether `command-schema.ts` nodes can expose min/max positional counts (fixed keys/values, `variadic({min})`, optionals). If a schema is too dynamic (subcommand dispatchers like `CONFIG`/`CLIENT`), keep an explicit `introspection.arity` as an override.
2. **Shrink the hand-written block** to what cannot be derived: `keySpecs`, `docs`, `tips`, `subcommands`. Derive `flags`, `categories` (already partly inferred) and first/last/step from `keySpecs`.
3. Consider a single `keySpecs` declaration that also drives `keys(args)` for simple commands (`first/last/step`) so the two cannot disagree — optional, only if it stays simple.
4. Move notification event naming onto the definition: optional `notifyEvent?: string` (default = command name), delete the override tables in `keyspace-notifier.ts`.

## Acceptance
[...]

Issue #379, opened 2026-09-21
Keyspace notifications: single mutation bus conflates WATCH signal with notification (stream groups silent; last-element removal loses type event)


fatal10110:
Found while ground-truthing PR #375 against real redis-server 7.2.1. **Pre-existing on `main`**, not caused by that PR.

## Root cause
Real Redis keeps two signals independent:
- `signalModifiedKey()` — dirties a WATCH
- `notifyKeyspaceEvent()` — emits a keyspace notification

This mock collapses both onto one `RedisMutationBus`. So there is no way to express *"notify, but do not dirty the WATCH"* — suppressing one suppresses the other. Two high-impact divergences fall out of that.

## Divergence 1 — stream consumer-group commands emit no notifications at all

`markCommitted` on an existing key emits nothing, so the whole consumer-group surface is silent.

Identical command sequence, `notify-keyspace-events KEA`:
```
REAL 7.2.1: xadd key=s / xgroup-create key=s / xgroup-createconsumer key=s /
            xadd key=s / xgroup-createconsumer key=s / xgroup-createconsumer key=s /
            xgroup-setid key=s / xgroup-delconsumer key=s / xgroup-destroy key=s /
            xgroup-create key=ns
MOCK:       xadd key=s / xadd key=s / xgroup key=ns
```
Real Redis confirms these same commands leave **WATCH intact** — so the two signals genuinely differ. (Real Redis emits nothing for XACK; the mock matches there.)

**Failure scenario:** an app subscribes to `__keyevent@0__:xgroup-createconsumer` (or `-delconsumer`/`-destroy`/`-setid`/`-create`) to track consumer-group topology. Against real Redis it receives events; against the mock it receives nothing and the test passes vacuously. Silent false-negative.

## Divergence 2 — the last element removed from a collection loses its type-specific event

The centralized empty-collection branch returns after emitting only `delete`, never calling `emitWrite`. Real Redis emits the type-specific event **and then** `del`.

```
REAL 7.2.1:                    MOCK:
hset key=h / hdel key=h / del key=h    hset key=h / del key=h
rpush key=l / lpop key=l / del key=l   rpush key=l / del key=l
sadd key=st / srem key=st / del key=st sadd key=st / del key=st
zadd key=z / zrem key=z / del key=z    zadd key=z / del key=z
sadd key=sp / spop key=sp / del key=sp sadd key=sp / del key=sp
```
[...]

PR #464, opened 2026-09-23
refactor: parse notify-keyspace-events once into a Set, profile-gate its CONFIG SET failure (#371, #416)

### LINKED ISSUE #371: refactor: represent notify-keyspace-events as a Set instead of 14 booleans + switches
## Problem
`src/state/keyspace-notifier.ts` models `notify-keyspace-events` as a struct of **14 booleans** with three large hand-written switches and a hand-ordered serializer, for what is just a set of characters:

- `KeyspaceNotifyFlags` (14 fields) + `emptyFlags()` (14 lines)
- `parseKeyspaceNotifyFlags` — 14-case `switch`
- `keyspaceNotifyFlagsToString` — 14 `if`s in a specific order, `ALL_CLASS_KEYS` check
- `classEnabled` — another 9-case `switch`
- `classForType` — 6-case `switch`
- `handle()` re-parses the raw config string **on every mutation event** (`server-state.ts:47-55` passes `this.notifyKeyspaceEvents`)

## Proposal
Represent the config as a `Set<string>` of class characters:
- parse: iterate chars, expand `A` to `g$lshzxet` + `d`, throw the existing `CONFIG SET failed…` error on an unknown char;
- serialize: canonical order `'g$lshzxetdn'` filtered by membership → `A` when all classes present, then `K`, `E`, `m` (keep current output order — pin it with the existing tests);
- `classEnabled(flags, cls)` → `flags.has(cls)`; `classForType` → a `Record<type, char>` lookup;
- parse once when `notifyKeyspaceEvents` is set (setter on `RedisServerState`), not per event.

Expected: ~140 lines → ~50, no behaviour change.

## Acceptance
- Existing `CONFIG GET/SET notify-keyspace-events` normalization and keyspace-notification tests pass unchanged (mock + real).
- `CONFIG SET notify-keyspace-events` still rejects invalid chars with the exact current message.
- The name→event override tables are handled in #370 (optional `notifyEvent` on definitions); this issue is only the flags representation.

### LINKED ISSUE #416: CONFIG SET failure wording is gated for proto-max-bulk-len but hardcoded for notify-keyspace-events
Found while reviewing PR #409, and explicitly requested by that PR's author rather than folded into it. PR #409 adds a compatibility gate for `CONFIG SET`'s failure wording; this issue is the other half of the same surface, which is now inconsistent.

## Problem
Real Redis changed the `CONFIG SET` failure message format at 7.0.0. PR #409 gates `proto-max-bulk-len`'s failures correctly via a new `config.set.failure-message` FeatureId, but `src/state/keyspace-notifier.ts:119` still hardcodes the 7.0+ wording. So a single server now emits **two different eras of message depending on which parameter you get wrong**:

```
REDIS_COMPAT=redis-6.2
  CONFIG SET proto-max-bulk-len abc
    -> ERR Invalid argument 'abc' for CONFIG SET 'proto-max-bulk-len' - argument must be a memory value   (correct for 6.2)
  CONFIG SET notify-keyspace-events Xz
    -> ERR CONFIG SET failed (possibly related to argument 'notify-keyspace-events') - Invalid event ...  (7.0 wording)

REAL 6.2.14
  CONFIG SET notify-keyspace-events Xz
    -> ERR Invalid argument 'Xz' for CONFIG SET 'notify-keyspace-events'
```

Note real 6.2 emits **no ` - <detail>` suffix at all** for this parameter — it is not simply the same detail in a different wrapper.

A gate applied to one of two adjacent messages is worse than a gate applied to neither: it makes the inconsistency look intentional.

## Proposal
Route `notify-keyspace-events`' failure through the same `configSetFailed(profile, ...)` helper PR #409 introduces. Having `parseKeyspaceNotifyFlags` return a detail string that `configSetFailed` wraps would close it cheaply, with the 6.2 branch emitting no detail suffix.

Then audit the rest: any other hardcoded `CONFIG SET failed` or `Invalid argument` message in `src/` should go through the same helper. That audit is the actual scope here — the `notify-keyspace-events` case is just the one we can see, because PR #409 put a correctly-gated message next to it.

## Acceptance
- `CONFIG SET notify-keyspace-events <bad>` matches real Redis byte-for-byte on both the `redis-6.2` and `redis-8.0` profiles, pinned by a raw-TCP test.
- No hardcoded CONFIG SET failure wording remains outside the shared helper.

fatal10110:
## Summary
**#371 — flags as a Set, parsed once**
- `KeyspaceNotifyFlags` is now a `ReadonlySet<string>` of flag characters with `A` expanded, replacing the 14-boolean struct, `emptyFlags()`, the 14-case parse switch, the 14-`if` serializer, `classEnabled` and the `classForType` switch. `classForType` became a `Record<type, class>` lookup and `classEnabled` became `flags.has(cls)`.
- `RedisServerState.notifyKeyspaceEvents` now holds the parsed Set. CONFIG SET parses it once. The notifier reads it on each mutation and no longer re-parses the string per event. CONFIG GET renders it back to canonical form.
- Output order is unchanged (`g$lshzxetd n K E m`, collapsed to `A`). The existing normalization table pins it.
- `WRITE/DELETE_EVENT_OVERRIDES`, `resolve()` and the activeNotifyCommand wiring are left alone (they belong to #370/#379/#444).

**#416 — CONFIG SET failure wording through the profile helper**
- `parseKeyspaceNotifyFlags` returns `undefined` on a bad char. `configSet` then throws `configSetFailed(profile, …, INVALID_NOTIFY_FLAG_DETAIL, { legacyDetail: false })`. On redis-6.2 that gives `ERR Invalid argument 'Xz' for CONFIG SET 'notify-keyspace-events'` with no detail suffix (6.2 hand-parses this param via `badfmt`). 7.0+ is unchanged.
- The audit found two more 6.2 divergences in the same surface, and both are fixed:
  - 6.2 echoes the parameter name **as the client sent it** (`'Proto-Max-Bulk-Len'`), while 7.0+ echoes the canonical lower-case name. The helper now receives the raw name. Before this, the mock always lower-cased it.
  - Unknown parameter: 6.2 says `ERR Unsupported CONFIG parameter: <name>`, while 7.0+ says `Unknown option or number of arguments for CONFIG SET - '<name>'`. This was hard-coded to the 7.0+ form and now goes through `configSetUnknownParameter(profile, name)` on the same `config.set.failure-message` gate.
- After this, `src/` has no hard-coded CONFIG SET failure wording outside the two helpers.

Checked against real redis 6.2.14, 7.0.15, 7.2.4, 8.0.0 and valkey 8.0.0 / 9.0.0 (throwaway docker containers).

Closes #371
Closes #416

## Test plan
- [x] New profile-sweep test in `tests-integration/compatibility/profile-gates.test.ts` (bad notify value, mixed-case name, unknown param; 6.2 vs 7.0+ exact bytes). It failed on the redis-6.2 profile before the fix and passes on all 7 profiles after (`npm run test:integration:compatibility:mock`).
- [x] New raw-TCP tests in `tests-integration/raw-tcp/command-errors-config.test.ts` pin the 7.0+ bytes. Green on mock and on real 8.0.
- [x] `tests/keyspace-notifier.test.ts` updated for the Set API. The full normalization table is unchanged and green.
- [x] `npm test` (462 pass), `npm run test:integration:mock` (1251 pass), `npm run build`, `npm run lint`.
- [x] Real backend: keyspace-notifications (ioredis + node-redis), proto-max-bulk-len (ioredis/node-redis/raw), command-errors-config all green.


src/state/keyspace-notifier.ts
```diff
@@ -1,170 +1,58 @@
...
-  'expired',
-  'evicted',
-  'stream',
-  'module',
-] as const satisfies readonly (keyof KeyspaceNotifyFlags)[]
+// The classes 'A' (NOTIFY_ALL) expands to, in Redis' render order: everything
+// except key-miss (m) and new-key (n). Module (d) IS included, so `Ad`
+// collapses to `A` while `g$lshzxet` (no d) stays expanded.
+const ALL_CLASSES = 'g$lshzxetd'
+const VALID_FLAGS = new Set(`${ALL_CLASSES}KEmn`)
```
fatal10110: [medium] The redis-6.2 profile accepts the `n` (new-key) flag, but real Redis 6.2 rejects it. `VALID_FLAGS` is new code in this PR, so the profile has to gate this. `n` was added in Redis 7.0; 6.2 accepts `m` and `d`, but not `n`.

Ground truth from real redis 6.2.24:
```
CONFIG SET notify-keyspace-events KEn   -> -ERR Invalid argument 'KEn' for CONFIG SET 'notify-keyspace-events'
CONFIG SET notify-keyspace-events And   -> -ERR Invalid argument 'And' for CONFIG SET 'notify-keyspace-events'
CONFIG SET notify-keyspace-events n     -> -ERR Invalid argument 'n' for CONFIG SET 'notify-keyspace-events'
CONFIG SET notify-keyspace-events KEnd  -> -ERR Invalid argument 'KEnd' for CONFIG SET 'notify-keyspace-events'
```
The mock under `REDIS_COMPAT=redis-6.2` returns `+OK` for all four. Real 7.0.15, 7.2.1, 8.0.6 and valkey 8.0/9.0 accept `n`.

Fix: gate `n` behind a feature id (redis 7.0.0 / valkey 7.2.0). For example, pass the profile into `parseKeyspaceNotifyFlags`, or have the caller reject `n` when the profile lacks the gate. The failure should go through `configSetFailed(..., { legacyDetail: false })`.

Test gap: the 6.2 sweep in `profile-gates.test.ts` only exercises `Xz`. Add an `n` case, e.g. `KEn`: rejected on 6.2 with the bare `badfmt` wording, `+OK` on 7.0+.

src/commands/config.ts
```diff
@@ -40,23 +45,54 @@ const MEMORY_UNITS = new Map<string, bigint>([
...
+ *   6.2   ERR Invalid argument '<value>' for CONFIG SET '<name>'[ - <detail>]
  *   7.0+  ERR CONFIG SET failed (possibly related to argument '<name>') - <detail>
+ *
+ * `name` is the parameter as the client sent it: 6.2 echoes it verbatim, 7.0+
+ * echoes the canonical lower-case name. 6.2 only appends the detail for
+ * parameters on its typed config table; hand-parsed ones such as
+ * `notify-keyspace-events` fail through a bare `badfmt`, so pass
+ * `legacyDetail: false` for those.
+ *
+ * Every CONFIG SET failure goes through this helper or
```
fatal10110: [medium] This comment overclaims. "Every CONFIG SET failure goes through this helper or configSetUnknownParameter — never a hard-coded message" is not true of the current code. At least two other CONFIG SET failure paths bypass both helpers:
- the odd-argument-count path (`WrongNumberOfArgumentsError('config|set')` at L268-269), whose 7.0+ wording differs from real Redis;
- duplicate-parameter rejection (real 7.0+ replies `- duplicate parameter` through the same template), which is not implemented here.

Both are being fixed in #470, so they are not blocking here. The sentence itself is in scope, though: scope it to what this PR guarantees (e.g. "invalid-value and unknown-parameter failures go through this helper or configSetUnknownParameter"), or drop the "every … never" claim.

src/commands/config.ts
```diff
@@ -40,23 +45,54 @@ const MEMORY_UNITS = new Map<string, bigint>([
  * CONFIG SET's failure wording, which Redis 7.0 changed wholesale when it
  * rewrote the subcommand to accept several parameter pairs:
  *
- *   6.2   ERR Invalid argument '<value>' for CONFIG SET '<name>' - <detail>
+ *   6.2   ERR Invalid argument '<value>' for CONFIG SET '<name>'[ - <detail>]
  *   7.0+  ERR CONFIG SET failed (possibly related to argument '<name>') - <detail>
+ *
+ * `name` is the parameter as the client sent it: 6.2 echoes it verbatim, 7.0+
+ * echoes the canonical lower-case name. 6.2 only appends the detail for
```
fatal10110: [low] A doc-comment precision nit. The code is correct. Real 7.0+ does not echo a *canonical* name: it echoes the name **as sent, lower-cased**, even when that name is an alias. Verified on real 7.0.15 / 8.0.6:
```
CONFIG SET Hash-Max-Ziplist-Entries abc
-ERR CONFIG SET failed (possibly related to argument 'hash-max-ziplist-entries') - ...
```
The canonical name there would be `hash-max-listpack-entries`. Suggest "7.0+ echoes the sent name lower-cased (aliases are not canonicalized)".

src/state/server-state.ts
```diff
@@ -53,11 +53,12 @@ export class RedisServerState {
...
-   * Normalized `notify-keyspace-events` flag string (Redis canonical form, e.g.
-   * `"AKE"`). Empty string disables keyspace notifications. Managed via
-   * CONFIG GET/SET; read by the wired {@link KeyspaceNotifier} on each mutation.
+   * Parsed `notify-keyspace-events` flags; empty disables keyspace
+   * notifications. Parsed once by CONFIG SET (and rendered back to Redis'
+   * canonical string by CONFIG GET), so the wired {@link KeyspaceNotifier}
+   * reads it on each mutation without re-parsing.
    */
-  notifyKeyspaceEvents = ''
+  notifyKeyspaceEvents: KeyspaceNotifyFlags = new Set()
```
fatal10110: [medium] Missing CHANGELOG entry for a breaking `/core` change. `RedisServerState` is exported from `js-redis-server/core` (`src/internal.ts`), and its public mutable field `notifyKeyspaceEvents` changes type from `string` to `ReadonlySet<string>`. CHANGELOG.md says `/core` is published API and marks such changes **BREAKING (`/core`)**. This PR does not touch CHANGELOG.md.

Failure scenario: a `/core` user who enabled notifications with `state.notifyKeyspaceEvents = 'KEA'` now gets a TS compile error. In plain JS it fails at runtime: the mutation-bus subscriber calls `flags.has('K')` on a string and throws `TypeError: flags.has is not a function` inside `RedisMutationBus.emit`, after the write has already committed.

Suggest adding:
- a **BREAKING (`/core`)** Changed entry for the field type;
- a Fixed entry for the #416 6.2-profile wording changes (notify-keyspace-events without a detail suffix, name echoed as sent, `Unsupported CONFIG parameter:`), to match how #410/#414 were recorded.

src/commands/config.ts
```diff
@@ -220,6 +259,7 @@ function configGet(
 type ConfigUpdate =
```
fatal10110: [low] `ConfigUpdate` is not really discriminated by `name`. The apply loop branches on `'bytes' in update` / `'flags' in update`. The catch-all `{ name: string; value: string }` variant also accepts `'notify-keyspace-events'` and `'proto-max-bulk-len'`, so `updates.push({ name: KEYSPACE_NOTIFY_PARAM, value })` still type-checks. It would silently `store.set(...)` a value that CONFIG GET then hides behind its overlay. Nothing does this today, but a real discriminant (`kind: 'store' | 'bulk-len' | 'notify'`) or pushing `apply` closures would make the validate and apply passes impossible to desynchronize.

src/state/keyspace-notifier.ts
```diff
@@ -1,170 +1,58 @@
...
-  hash: boolean // h
-  zset: boolean // z
-  expired: boolean // x
-  evicted: boolean // e
-  stream: boolean // t
-  keyMiss: boolean // m
-  newKey: boolean // n
-  module: boolean // d
-}
+export type KeyspaceNotifyFlags = ReadonlySet<string>
```
fatal10110: [low] `ReadonlySet<string>` is wider than needed. `flags.has('k')` (a typo for `K`) type-checks and is silently always false. A `KeyspaceNotifyFlag` union (`NotifyClass | 'd' | 'K' | 'E' | 'm' | 'n'`), with `VALID_FLAGS` as a type guard in `parseKeyspaceNotifyFlags`, would catch bad reads at compile time for the cost of one guard.

src/state/keyspace-notifier.ts
```diff
@@ -1,170 +1,58 @@
...
-
-/**
- * Validate and normalize a `notify-keyspace-events` value for CONFIG SET.
- * Returns the canonical string; throws {@link RedisCommandError} on bad input.
- */
-export function normalizeKeyspaceNotifyConfig(value: string): string {
-  return keyspaceNotifyFlagsToString(parseKeyspaceNotifyFlags(value))
+  const all = [...ALL_CLASSES].every(cls => flags.has(cls))
+  const order = all ? 'AKEm' : `${ALL_CLASSES}nKEm`
+  return [...order].filter(flag => flag === 'A' || flags.has(flag)).join('')
```
fatal10110: [low] Readability. `flag === 'A' || flags.has(flag)` implicitly relies on two things: `'A'` appearing only in the collapsed order string, and `n` being absent from `'AKEm'` (which is how the "n is dropped under A" rule is enforced). It is correct, and the normalization table pins it. Building the prefix explicitly (`all ? 'A' : <filtered classes + n>`) and then appending the filtered `KEm` would make each Redis rule visible.

tests-integration/raw-tcp/command-errors-config.test.ts
```diff
@@ -168,6 +168,37 @@ describe(`Raw TCP CONFIG errors (${testRunner.getBackendName()})`, () => {
     )
   })
 
+  // CONFIG SET failure wording (#416). The 7.0+ form names the parameter by its
+  // canonical lower-case name whatever casing the client sent; the 6.2 form
+  // (no detail suffix for this parameter, name echoed as sent) is asserted by
+  // the profile sweep in profile-gates.test.ts.
+  test('an invalid notify-keyspace-events class fails with the CONFIG SET template', async () => {
```
fatal10110: [low] Both new raw-tcp tests already pass on `origin/main`. Main lower-cased the name before the notify branch and hard-coded the 7.0+ message, and its unknown-parameter message is byte-identical to the new 7.0+ form. So these are regression pins, not evidence for the #416 fix; only the profile-gates sweep under redis-6.2 goes red before the fix. Keeping them is fine. Consider rewording the comment ("pins the 7.0+ bytes") so nobody reads them as red-before-green proof.

fatal10110:
## Review summary: NOT YET

Reviewed head `bcd39d9`. Two reviewers looked at it: one for Redis wire compatibility, one for Node architecture and code quality. Every "real Redis does X" claim was checked against real redis 6.2.24, 7.0.15, 7.2.1 and 8.0.6, and valkey 8.0.11 and 9.0.6. All six wire claims in the PR description hold byte for byte: the 6.2 bare `badfmt` wording for notify-keyspace-events, the name echoed as sent on 6.2 versus lower-cased on 7.0+, the `Unsupported CONFIG parameter:` wording on 6.2, the 7.0+ templates, and CONFIG GET's canonical rendering.

| Severity | Location | Finding | Reviewer |
|---|---|---|---|
| medium | `src/state/keyspace-notifier.ts:17` | The redis-6.2 profile accepts the `n` flag, but real 6.2 rejects it (`KEn`, `And`, `n`). The 6.2 sweep has no `n` case. | Wire |
| medium | `src/commands/config.ts:57` | The "every CONFIG SET failure goes through the helpers" comment overclaims: the odd-argc and duplicate-parameter paths bypass them. Scope the wording. | Wire + Arch |
| medium | `src/state/server-state.ts:61` | No CHANGELOG entry for the **BREAKING (`/core`)** type change of `RedisServerState.notifyKeyspaceEvents`, or for the #416 6.2 wording fixes. | Arch |
| low | `src/commands/config.ts:52` | 7.0+ echoes the sent name lower-cased; aliases are not canonicalized. Comment nit. | Wire |
| low | `src/commands/config.ts:259` | `ConfigUpdate` is not discriminated by `name`; the catch-all variant admits special names. | Arch |
| low | `src/state/keyspace-notifier.ts:11` | `ReadonlySet<string>` is wider than needed; a flag-char union would catch typos. | Arch |
| low | `src/state/keyspace-notifier.ts:55` | `keyspaceNotifyFlagsToString` relies on implicit rules; make them explicit. | Arch |
| low | `tests-integration/raw-tcp/command-errors-config.test.ts:175` | The new raw-tcp tests already pass on main. They are regression pins, not red-before-green proof. | Arch |

**Out of scope, not blocking (covered by #470):** unknown-name versus bad-value ordering across multiple pairs, duplicate-parameter detection, the 7.0+ odd-arg-count error, 6.2 multi-pair rejection.

**Other out-of-scope follow-ups:** store-backed params accept any value (e.g. `maxmemory abc`). The `m`/`n` classes are accepted but never published. Event naming belongs to #379/#444/#370.

**Tests on PR head (mock backend):**
- `tsc --noEmit`: clean.
- `npm test`: 462/462 pass.
- Compatibility sweep across all 7 profiles: all green (22 tests on 6.2, 20 on each other profile).
- CONFIG, proto-max-bulk-len and keyspace-notification integration tests (raw-tcp, ioredis, node-redis): 98/98 pass.
- Notifier unit tests plus raw-tcp CONFIG errors: 58/58 pass.

The Set refactor itself looks sound. The removed-behavior audit is clean: the empty-flags early return is preserved, class mapping is exhaustive, CONFIG SET is still atomic, and the notifier reads the live field on every event.

src/state/keyspace-notifier.ts
fatal10110: Fixed in 45a7827. Added FeatureId `notify.keyspace.new-key-class` (redis 7.0.0 / valkey 7.2.0). `parseKeyspaceNotifyFlags` takes `{ newKeyClass }`, and `configSet` passes `profile.has(...)`, so on redis-6.2 `n` fails through `configSetFailed(..., { legacyDetail: false })`. Checked against real 6.2.14: `KEn` -> `-ERR Invalid argument 'KEn' for CONFIG SET 'notify-keyspace-events'`, `KEmd` -> OK. New profile-gates test `the n notify-keyspace-events class matches the profile`: `KEn` is rejected on 6.2, and on 7.0+ it gives `+OK` and CONFIG GET returns `nKE`. `KEmd` gives `+OK` on every profile. Also a unit test (`n`/`KEn`/`And`/`KEnd` rejected, `KEmd` accepted) plus a docs/API.md gate row.

src/commands/config.ts
fatal10110: Fixed in 45a7827. The comment now says only that invalid-value failures go through `configSetFailed` and unknown parameters through `configSetUnknownParameter`, and that the odd-argument-count and duplicate-parameter paths are separate (#470).

src/commands/config.ts
fatal10110: Fixed in 45a7827. The comment now says 7.0+ echoes the sent name lower-cased and does not canonicalize aliases (`Hash-Max-Ziplist-Entries` -> `hash-max-ziplist-entries`).

src/state/server-state.ts
fatal10110: Fixed in 45a7827. CHANGELOG has a **BREAKING (`/core`)** Changed entry for the `notifyKeyspaceEvents` type change. It includes a migration snippet and the plain-JS `flags.has is not a function` failure mode. `KeyspaceNotifyFlag`/`KeyspaceNotifyFlags` are now exported from `/core` so callers can type the field. There is also a Fixed entry for the #416 6.2-profile changes: no detail suffix, name echoed as typed, `Unsupported CONFIG parameter:`, and `n` rejected.

src/commands/config.ts
fatal10110: Fixed in 45a7827. `ConfigUpdate` is now discriminated by `kind: 'store' | 'proto-max-bulk-len' | 'notify-keyspace-events'`, and the apply pass is a `switch (update.kind)`.

src/state/keyspace-notifier.ts
fatal10110: Fixed in 45a7827. Added `KeyspaceNotifyFlag = NotifyClass | 'd' | 'm' | 'n' | 'K' | 'E'`, and `KeyspaceNotifyFlags = ReadonlySet<KeyspaceNotifyFlag>`. Parsing narrows through an `isNotifyFlag` type guard, so `flags.has('k')` is now a compile error.

src/state/keyspace-notifier.ts
fatal10110: Fixed in 45a7827. `keyspaceNotifyFlagsToString` now builds the three groups explicitly: classes (`A`, or `g$lshzxetd` then `n`), channels (`K`, `E`), then `m`. The doc comment lists each rule, including that `n` is dropped under `A` and `m` is always last. The normalization table is unchanged and green.

tests-integration/raw-tcp/command-errors-config.test.ts
fatal10110: Agreed. Reworded in 45a7827: the comment now says these two tests pin the 7.0+ bytes as regression guards and already passed before #416, and that the red-before-fix evidence is the redis-6.2 profile-gates sweep.

fatal10110:
## Re-review of `45a7827`: SHIP

This re-review covers the diff `bcd39d9..45a7827`. All 8 findings from the previous review are fixed. I checked each against the changed lines:

| Prior finding | Status |
|---|---|
| [medium] 6.2 profile accepts `n` | **Fixed.** New gate `notify.keyspace.new-key-class` (redis 7.0 / valkey 7.2) threads through `parseKeyspaceNotifyFlags({ newKeyClass })`. A unit test and a profile sweep cover `KEn`/`KEmd` on each profile. Re-verified on my own real redis 6.2.24 container: `KEn`, `n`, `And` and `KEnd` all fail with `-ERR Invalid argument '<v>' for CONFIG SET 'notify-keyspace-events'` (the name is echoed as sent: `'Notify-Keyspace-Events'`); `KEmd`, `Km` and `d` return OK; `CONFIG GET` after `KEmd` returns `dKEm` and after `Adm` returns `Am`. The mock matches. |
| [medium] "every CONFIG SET failure" comment | **Fixed.** Now limited to invalid-value and unknown-parameter failures, with a pointer to #470 for the rest. |
| [medium] CHANGELOG BREAKING (`/core`) entry | **Fixed.** Changed and Fixed entries are in [Unreleased], and the link refs are added. `KeyspaceNotifyFlag(s)` types are exported from `/core`. The migration example `state.notifyKeyspaceEvents = new Set(['K','E',...])` type-checks against the built `dist/core.d.ts`, and a `new Set(['Q'])` control is correctly rejected. |
| [low] 7.0+ alias echo comment | **Fixed.** |
| [low] `ConfigUpdate` discriminant | **Fixed.** It is now `kind` + `switch`. |
| [low] wide `ReadonlySet<string>` | **Fixed.** Narrowed to a `KeyspaceNotifyFlag` union with an `isNotifyFlag` type guard. |
| [low] serializer implicit rules | **Fixed.** The prefix, channels and key-miss parts are now explicit, and the normalization table is unchanged and green. |
| [low] raw-tcp tests described as proving the fix | **Fixed.** The comment now calls them 7.0+ regression pins. |

**New findings:** none in the code.
- [low] The PR description doesn't mention the new `notify.keyspace.new-key-class` gate, i.e. that the redis-6.2 profile now rejects `n`. Please add it to the #416 section. CHANGELOG and docs/API.md already cover it.

**Tests on `45a7827` (mock backend):**
- `tsc --noEmit` and `npm run lint`: clean.
- `npm test`: 463/463 pass.
- `npm run build` (d.ts emit): ok.
- `tests-package`: 8/8 pass.
- Compatibility sweep across all 7 profiles: green (23 tests on 6.2, 21 on each other profile). One run of the 6.2 sweep hit a Node test-runner IPC error (`Unable to deserialize cloned data`) while under parallel load. It passed 3/3 on rerun, and `profile-gates.test.ts` passed on its own.
- Keyspace-notification, CONFIG and proto-max-bulk-len integration tests plus the notifier unit tests: 121/121 pass.

The #470 items (name pre-pass ordering, duplicate params, odd-argc, 6.2 multi-pair) remain out of scope here.

fatal10110:
## Re-review of merge head `b64e127` (#470 resolution): SHIP

Scope: the merge resolution only. That is `45a7827..b64e127` on `src/commands/config.ts`, `src/state`, the profile-gates and raw-tcp CONFIG tests and `docs/API.md`, compared with the #470 side on `origin/main`.

**Resolution looks correct.**
- #470's `checkConfigSetNames` still runs first, unchanged, except that its unknown-name throw now goes through the profile-aware `configSetUnknownParameter`.
- `configSetDuplicateParameter` still echoes the name as typed.
- The old arity check is gone, leaving #470's 6.2 syntax / 7.0+ arity split as the only one.
- The value loop builds `kind`-tagged updates and applies them only after every value validates.
- `src/state` is unchanged since `45a7827`.
- Against main, the test changes only add lines, so no #470 assertion was lost.

**Checked against real Redis on containers I started (identity confirmed by run_id), with the mock on the same profile. 11/11 replies match byte for byte on each version:**
- **redis 6.2.24:** every multi-pair, duplicate and odd-count shape gets `Unknown subcommand or wrong number of arguments for 'SET'. Try CONFIG HELP.` `KEn` gets `Invalid argument 'KEn' for CONFIG SET 'notify-keyspace-events'`. `Notify-Keyspace-Events Xz` is echoed as sent with no detail suffix. `Bogus-Param` gets `Unsupported CONFIG parameter: Bogus-Param`.
- **redis 8.0.6:**
  - `notify-keyspace-events Xz bogus-param 1`: the unknown name is reported before the bad value.
  - `timeout 0 TIMEOUT 0` → `... argument 'TIMEOUT') - duplicate parameter` (as typed). `KEA Notify-Keyspace-Events Xz` → duplicate, not invalid value.
  - An odd pair count → `ERR syntax error`.
  - `Kx proto-max-bulk-len abc` fails, and `CONFIG GET` still reads `AKE`, so the command stays atomic.
  - `KEn` → `OK`, and the 7.0+ failure wording is lower-cased.

**Tests on `b64e127` (mock backend):**
- `tsc` and lint: clean.
- `npm test`: 492/492 pass.
- Compatibility sweep across all 7 profiles: green (24 tests on 6.2, 22 on each other profile).
- raw-tcp CONFIG errors, CONFIG integration (ioredis/node-redis), keyspace notifications and the notifier unit tests: 76/76 pass.

No new findings. The earlier low note is resolved: the PR description now documents the `notify.keyspace.new-key-class` gate. No open findings remain.

PR #465, opened 2026-09-23
fix(node-redis-mock): resolve error classes per class, throw SimpleError (#450)

### LINKED ISSUE #450: node-redis mock: error-class loading is all-or-nothing (redis ≤4.6), and errors are ErrorReply not SimpleError
Found while reviewing PRs #421 and #427. Both issues are in `src/client-mocks/node-redis-mock.ts`.

## 1. Loading the error classes is all-or-nothing, so redis ≤ 4.6 falls back to the stand-in classes
`ensureRedisErrors` (#421) only uses the real classes when **every** expected export exists. `redis` 4.6 and older lack `MultiErrorReply`, and that single missing export discards the real `ClientClosedError`, `DisconnectsClientError` and `WatchError` too. On those versions the documented `instanceof ClientClosedError` teardown check fails. This was reproduced with a stub `redis` that exports exactly what 4.6 exports. The code comment "v4 has no `MultiErrorReply`" is also wrong: `redis` 4.7 exports it.
→ Fall back to the stand-in one class at a time.

## 2. Errors are `ErrorReply`; real node-redis throws `SimpleError`
Real node-redis v6 throws `SimpleError` (a subclass of `ErrorReply`) for server errors. The mock throws `ErrorReply`, so `instanceof ErrorReply` holds on both, but `err.constructor.name` and `instanceof SimpleError` differ. This affects every error.

## Acceptance
With redis 4.6 installed, `instanceof` still holds for the classes that exist. Server errors are `SimpleError` instances where the installed `redis` provides that class.

fatal10110:
## Summary
- **Per-class error loading.** `loadRedisErrors` used the real `redis` error classes only if *every* expected export existed. On a redis without `MultiErrorReply`, that one missing export also threw away the real `ClientClosedError` / `DisconnectsClientError` / `WatchError`, which broke the documented `instanceof ClientClosedError` teardown check. Each class now falls back to its stand-in on its own. The `MultiErrorReply` stand-in is built on top of whichever `ErrorReply` was resolved, so `instanceof ErrorReply` still holds for the aggregate error. Its message and `constructor.name` now match node-redis' own class.
- **Corrected the version claim.** The old comment said "v4 has no `MultiErrorReply`", which is wrong. `MultiErrorReply` was added in @redis/client 1.5.13, which ships with redis **4.6.12**. Only redis ≤ 4.6.11 lacks it. The source comment and `docs/TESTING.md` are fixed.
- **`SimpleError`.** Real node-redis v6 decodes every `-ERR` reply into `SimpleError`, a subclass of `ErrorReply`. That covers top-level replies and replies inside `EXEC`, on both RESP2 and RESP3; I checked this against real Redis on :6399. The facade now does the same in both of its decode paths (`NODE_REDIS_DECODE_OPTIONS.error` and the per-item `EXEC` replies). If `redis` does not export `SimpleError` (v4), it falls back to `ErrorReply`, which is what v4 throws.

Closes #450

## Test plan
- [x] New `tests/client-mocks/node-redis-mock-error-loading-v4.test.ts` seeds `require.cache` with a stub `redis` that exports exactly the error classes of redis 4.6.11 (@redis/client 1.5.12). It asserts that the real `ClientClosedError`, `DisconnectsClientError`, `WatchError` and `ErrorReply` are thrown, and that the `MultiErrorReply` stand-in extends `ErrorReply`. 5 of its 6 tests failed before the fix.
- [x] `node-redis-mock.test.ts` and `decode-redis-value.test.ts` now also assert `SimpleError`, both for top-level errors and for `MultiErrorReply.replies[i]`. These failed before the fix.
- [x] Real node-redis 6.0.0 against real Redis 6399 throws `SimpleError` for top-level errors, unknown commands and `EXEC` items, at RESP2 and RESP3.
- [x] `npm run build`, `npm run lint`, `npm test` (469/469), `npm run test:integration:mock` (1248 pass, 0 fail)


tests/client-mocks/node-redis-mock-error-loading-v4.test.ts
```diff
@@ -0,0 +1,158 @@
...
+        )
+        return true
+      },
+    )
+  } finally {
+    await client.quit()
+  }
+})
+
+test('with no MultiErrorReply, the stand-in still extends the installed ErrorReply', async () => {
```
fatal10110: [medium] This test encodes behaviour that real redis ≤ 4.6.11 does not have. In the release this stub emulates, `exec()` never throws on a failed queued command. It **resolves** with the `ErrorReply` inline. The aggregate throw arrived in the same release as the `MultiErrorReply` class. So on redis 4.6.11 the facade rejects where the real client resolves: `multi().set().incr(str).exec()` gives real `[\"OK\", ErrorReply]` and mock `throw MultiErrorReply-stand-in`. This test, and the TESTING.md line "a stand-in matches the real class in message and shape", lock that divergence in.

Evidence (read from the published tarballs via `npm pack`):
```
@redis/client 1.5.12 (redis 4.6.11) lib/multi-command.js
    transformReplies(rawReplies) {
        return rawReplies.map((reply, i) => {
            const { transformReply, args } = this.queue[i];
            return transformReply ? transformReply(reply, args.preserve) : reply;
        });
    }
@redis/client 1.5.13 (redis 4.6.12)
        const errorIndexes = [], replies = rawReplies.map((reply, i) => {
            if (reply instanceof errors_1.ErrorReply) { errorIndexes.push(i); return reply; }
            ...
        if (errorIndexes.length)
            throw new errors_1.MultiErrorReply(replies, errorIndexes);
```
(1.5.12 `RESP2/decoder.js` `parseError` returns `new errors_1.ErrorReply(message)` as the nested element.)

Suggested fix, at the throw site `node-redis-mock.ts:1027`: when the installed `redis` does not export `MultiErrorReply`, resolve with `replies` (errors inline) instead of throwing a stand-in. For example, keep `MultiErrorReply` `undefined` in `RedisErrorConstructors` and branch in `exec`. Keep the stand-in only for the "redis cannot be required at all" case, or drop it. Then flip this test to assert that `exec()` resolves `[\"OK\", <ErrorReply>]`, and correct the docs paragraph. The throw predates this PR, but the PR adds a test that asserts it as the 4.6.11 behaviour, so it is in scope. (Lens: Redis wire-compat, verified from source by the coordinator.)

tests/client-mocks/node-redis-mock.test.ts
```diff
@@ -206,11 +207,32 @@ describe('createNodeRedisMock (standalone)', () => {
...
         return true
       },
     )
   })
 
+  test('server errors are SimpleError, as real node-redis v6 throws', async () => {
+    const client = await makeClient()
+    await client.set('s', 'notAnInteger')
+    // Real node-redis v6 decodes every `-ERR` reply into a SimpleError (a
+    // subclass of ErrorReply), at RESP2 and RESP3 alike — so both the
```
fatal10110: [medium] The comment says real node-redis v6 decodes `-ERR` into `SimpleError` "at RESP2 and RESP3 alike". That is true (verified below), but this test only exercises the default RESP2 connection, so the RESP3 half of the claim has no test. It also never asserts `err.message`. Please add a RESP3 pass, e.g. `createNodeRedisMock({ RESP: 3 })` for `incr`, an unknown command, and an `EXEC` item, plus a message assertion. The alternative is to drop "and RESP3" from the comment.

Real node-redis 6.0.0 against real Redis 7.2.1, plus the mock at this head, using the same probe:
```
REAL RESP2  incr    SimpleError|name=Error|msg=ERR value is not an integer or out of range
REAL RESP3  incr    SimpleError|name=Error|msg=ERR value is not an integer or out of range
REAL RESP3  unknown SimpleError|msg=ERR unknown command 'NOSUCHCMD', with args beginning with: 'a' 
REAL RESP3  exec    MultiErrorReply|msg=1 commands failed, see .replies and .errorIndexes for more information idx=[1] r0=OK r1=SimpleError
MOCK RESP3  (identical on all four lines)
```
So the mock already matches. The test just does not prove it. (Lenses: Redis + architecture.)

docs/TESTING.md
```diff
@@ -415,12 +415,18 @@ facade deliberately does not reproduce that, and tears everything down.
...
-`MultiErrorReply`). They are resolved lazily, the first time a facade client is
-constructed — importing `js-redis-server` never loads `redis`, so ioredis-only
-users don't pay for it. Only if `redis` genuinely cannot be required (or predates
-one of these classes, as v4 predates `MultiErrorReply`) does the facade fall back
-to local classes that match in message and shape but are not `instanceof` the
-real ones; match on `err.message` if you need to support that.
+`SimpleError` / `MultiErrorReply`). Server errors are `SimpleError` — a subclass
+of `ErrorReply` — exactly as real node-redis v5+ decodes them, so both
+`instanceof ErrorReply` and `instanceof SimpleError` hold; on a `redis` without
+`SimpleError` (v4) they are plain `ErrorReply`, again as that version throws.
```
fatal10110: [low] "on a `redis` without `SimpleError` (v4) they are plain `ErrorReply`, again as that version throws" is not true for the whole `^4.0.0` peer range. redis 4.0.x depends on `@node-redis/client` 1.0.x. That package exports no `ErrorReply`; its `errors.js` has only AbortError, WatchError, ConnectionTimeoutError, ClientClosedError, DisconnectsClientError and SocketClosedUnexpectedlyError, and it decodes replies with `redis-parser`. `ErrorReply` first appears in `@redis/client` 1.0.x, i.e. redis ≥ 4.1.0. On 4.0.x the facade throws the local `FacadeErrorReply`, which is not `instanceof` anything the user can import. Suggest "(redis 4.1–4.x)" plus a note that on 4.0.x you should match on `err.message`. The other version claims are correct: `MultiErrorReply` appears in @redis/client 1.5.13 (redis 4.6.12), `SimpleError` in @redis/client 5.0.0, and the stub export list matches 1.5.12 exactly. (Lens: Redis, verified from npm tarballs.)

tests/client-mocks/node-redis-mock-error-loading-v4.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+function newClient(): NodeRedisMockClient {
+  return new NodeRedisMockClient({
+    state: new RedisServerState({ databaseCount: 1 }),
+    executor: createRedisCommandExecutor(),
+    ownsState: true,
+  })
+}
+
+test('the stub is what the facade loads', () => {
```
fatal10110: [low] This test cannot fail. It reads back, through `facadeRequire`, the cache entry that line 59 just wrote through `facadeRequire`, so it never exercises the facade's own `createRequire(__filename)('redis')`. Either rename it to what it checks (for example "stub is seeded at the facade's resolution path") or delete it. Tests 2–6 already prove that the facade loads the stub: on base `src` they fail 5/6, and this one passes. (Lens: architecture.)

tests/client-mocks/node-redis-mock-error-loading-v4.test.ts
```diff
@@ -0,0 +1,158 @@
...
+      () => client.multi().set('w', 'fromTxn').exec(),
+      (err: unknown) => err instanceof WatchError,
+    )
+  } finally {
+    await other.quit()
+    await client.quit()
+  }
+})
+
+test('with no SimpleError, server errors are the installed ErrorReply', async () => {
```
fatal10110: [low] Several stand-in behaviours this PR changes or adds are untested: the new `MultiErrorReply` stand-in message (`${n} commands failed, see .replies and .errorIndexes for more information`), `[...err.errors()]`, and the "`redis` cannot be required at all" path, where the stand-in extends `FacadeErrorReply` and the local close errors are used. A second stub file that makes `require('redis')` throw would cover the last one. (Lens: architecture.)

src/client-mocks/node-redis-mock.ts
```diff
@@ -112,37 +123,47 @@ let resolvedRedisErrors: RedisErrorConstructors | undefined
...
+  ): RedisErrorConstructors[K] | undefined => {
+    const value = redis?.[name]
+    return typeof value === 'function'
+      ? (value as RedisErrorConstructors[K])
+      : undefined
+  }
+
+  const ErrorReply = exported('ErrorReply') ?? FacadeErrorReply
+  return {
+    WatchError: exported('WatchError') ?? FacadeWatchError,
```
fatal10110: [low] The PR aligns the `MultiErrorReply` stand-in with real node-redis by dropping `this.name = ...`, since real classes leave `.name === 'Error'` (verified: real v6 `MultiErrorReply` and `SimpleError` both report `name=Error`). Its sibling `FacadeWatchError` (line 64, outside the diff) still sets `this.name = 'WatchError'` and ignores the `message?` its constructor type declares. For consistency, drop the name assignment there too. (Lens: architecture.)

src/client-mocks/node-redis-mock.ts
```diff
@@ -112,37 +123,47 @@ let resolvedRedisErrors: RedisErrorConstructors | undefined
...
-    redis.DisconnectsClientError
-    ? {
-        WatchError: redis.WatchError,
-        ErrorReply: redis.ErrorReply,
-        MultiErrorReply: redis.MultiErrorReply,
-        ClientClosedError: redis.ClientClosedError,
-        DisconnectsClientError: redis.DisconnectsClientError,
-      }
-    : FALLBACK_REDIS_ERRORS
+    redis = undefined
```
fatal10110: [low] `redis = undefined` in the `catch` does nothing, because `redis` is already `undefined` when the require throws. A tiny `requireRedis(): Record<string, unknown> | undefined { try { return ... } catch { return undefined } }` would make `redis` a `const` and read top-to-bottom, as AGENTS.md prefers. (Lens: architecture.)

src/client-mocks/node-redis-mock.ts
```diff
@@ -1429,11 +1450,15 @@ export const NODE_REDIS_DECODE_OPTIONS: ClientDecodeOptions = {
...
-  // Surface node-redis' own ErrorReply (so `instanceof ErrorReply` matches the
-  // documented idiom). Falls back to RedisCommandError if `redis` is absent.
+  // Surface node-redis' own server-error class — `SimpleError`, a subclass of
+  // `ErrorReply`, on v5+ — so both `instanceof ErrorReply` (the documented
+  // idiom) and the concrete class match real node-redis. Falls back to
+  // RedisCommandError only before any facade client has resolved the classes.
   error: (text, code) => {
-    const ErrorReply = resolvedRedisErrors?.ErrorReply
-    return ErrorReply ? new ErrorReply(text) : new RedisCommandError(text, code)
+    const SimpleError = resolvedRedisErrors?.SimpleError
```
fatal10110: [low] Optional: `redisErrors()` is synchronous, so this hook could call `new (redisErrors().SimpleError)(text)` directly. The "import must not load `redis`" rule is about import time, not the first decoded error. That would remove the ordering dependency where the hook returns `RedisCommandError` until some facade client has been constructed in the process, which is why `decode-redis-value.test.ts` needs its `before()` hook. The PR already rewrote this comment, so it is cheap to do now. (Lens: architecture.)

fatal10110:
## Review (redis-pr-review) of fb38501. Verdict: **NOT YET**

The core fix is correct. Error classes now resolve one at a time, and server errors are `SimpleError`. I checked the second part live: real node-redis 6.0.0 against real Redis 7.2.1 (RESP2 and RESP3) and the mock produce **identical** output for class, `.name`, message, `MultiErrorReply` message, `errorIndexes` and the class of each reply item, on `INCR`, an unknown command, `WRONGTYPE` and `EXEC`. The version claims also hold, checked against the npm tarballs: `MultiErrorReply` in @redis/client 1.5.13 (redis 4.6.12), `SimpleError` in 5.0.0, and a stub export list that matches 1.5.12 exactly. The red-before-green claim holds as well: on base `src` the new v4 test fails 5 of 6.

|---|---|---|---|
| medium | v4 test:133 (fix at mock:1027) | On redis ≤ 4.6.11, real `exec()` **resolves** with `ErrorReply` inline and never throws. The stand-in throws, and the new test and the docs lock that in as the 4.6.11 behaviour. | R |
| medium | node-redis-mock.test.ts:221 | The comment claims RESP3 coverage but the test only runs RESP2, and it never asserts the message. (The mock does match at RESP3; I verified it.) | R+N |
| low | TESTING.md:421 | "v4 throws plain ErrorReply" is false for redis 4.0.x (`@node-redis/client` has no `ErrorReply`). | R |
| low | v4 test:74 | Tautological first test. | N |
| low | v4 test:113 | No tests for the stand-in message, `errors()`, or the "`redis` cannot be required" path. | N |
| low | mock:158 | `FacadeWatchError` still sets `.name`, unlike the real class. | N |
| low | mock:145 | Redundant `redis = undefined`. | N |
| low | mock:1458 | The `RedisCommandError` fallback could be `redisErrors()`, which removes the ordering dependency. | N |

Tests on the head: 71/71 targeted, 1030/1030 unit plus node-redis integration on the mock backend; `tsc` and eslint are clean.

Out-of-scope follow-ups, not blocking:
- The EXECABORT rejection message differs from real v6. Real v6 rejects with the first queue-time error; the mock rejects with `EXECABORT ...`. This comes from reading the source; I have not captured it live.
- An error nested inside an array reply (e.g. an EVAL returning `redis.error_reply`) is thrown, where real v6 resolves it inline.
- The bare `catch` in `loadRedisErrors` swallows require failures other than `MODULE_NOT_FOUND`.

tests/client-mocks/node-redis-mock-error-loading-v4.test.ts
fatal10110: Fixed in fda48d2. I checked the published source first: @redis/client 1.5.12's `transformReplies` maps the EXEC array unchanged, and the aggregate throw arrived in 1.5.13. `MultiErrorReply` in `RedisErrorConstructors` is now `undefined` when the installed `redis` doesn't export it, and `exec()` only throws when the class exists. Otherwise it resolves with the error inline. The test now asserts the resolved `['OK', ErrorReply]`, where the error's constructor is the installed `ErrorReply` and the message is exact. The TESTING.md paragraph now describes this behaviour. The stand-in `MultiErrorReply` is used only when `redis` can't be required at all.

tests/client-mocks/node-redis-mock.test.ts
fatal10110: Fixed in fda48d2. The test runs twice, at RESP2 and at RESP3 (via `HELLO 3`). Each run covers `incr`, an unknown command and an EXEC item. It checks the exact `message`, `name === 'Error'` and `constructor === SimpleError`, plus the `MultiErrorReply` message and `errorIndexes`. The expected strings match your real 6.0.0 probe.

docs/TESTING.md
fatal10110: Fixed in fda48d2. The doc now says 'redis 4.1–4.x' for the plain-`ErrorReply` case. It also says 4.0.x exports no `ErrorReply` and gets a local look-alike, so you should match on `err.message` there. The source comment now records that `ErrorReply` arrived in 4.1.0.

tests/client-mocks/node-redis-mock-error-loading-v4.test.ts
fatal10110: Removed in fda48d2. Tests 2–6 already prove the stub is loaded. The new redis-absent file keeps a premise guard, and that one can fail: it asserts `require('redis')` throws `MODULE_NOT_FOUND` through the patched resolver.

tests/client-mocks/node-redis-mock-error-loading-v4.test.ts
fatal10110: Added in fda48d2: `tests/client-mocks/node-redis-mock-error-loading-absent.test.ts`. It patches `Module._resolveFilename` so `redis` can't be required. It asserts each stand-in (ClientClosedError, DisconnectsClientError, WatchError, SimpleError→ErrorReply, MultiErrorReply) against the real class from `@redis/client`. Each stand-in must match that class's `constructor.name`, `name === 'Error'` and message, and must not be `instanceof` it. It also covers the stand-in `MultiErrorReply` message for 2 failures and `[...errors()]`.

src/client-mocks/node-redis-mock.ts
fatal10110: Fixed in fda48d2. The stand-ins are now classes with the real names (`WatchError`, `ErrorReply`, `SimpleError`, `MultiErrorReply`, ...). None of them assigns `.name`. `WatchError` takes the optional `message`, defaulting to the real text. The absent-redis test asserts `name === 'Error'` for each.

src/client-mocks/node-redis-mock.ts
fatal10110: Fixed in fda48d2. There's now a `requireRedis(): Record<string, unknown> | undefined` helper. `redis` is a `const`, and there's an early return to the stand-ins when it's absent.

src/client-mocks/node-redis-mock.ts
fatal10110: Done in fda48d2. The hook is now `error: (text) => new (redisErrors().SimpleError)(text)`, so the `RedisCommandError` fallback and its ordering dependency are gone. I also removed the `before()` warm-up in `decode-redis-value.test.ts`, which now passes without it.

fatal10110:
## Re-review of fda48d2 (previously fb38501). Verdict: **SHIP**

All 8 prior findings are fixed. I checked each against the changed lines:

| prior | status |
|---|---|
| [medium] exec() should resolve inline on redis ≤ 4.6.11 | **Fixed.** `MultiErrorReply` is now `undefined` when it isn't exported, and `exec()` returns the replies with errors inline. That matches @redis/client 1.5.12 `transformReplies` and 1.0.0, which behaves the same way. On a WATCH abort, `exec()` still throws `WatchError`, which matches 1.5.12 `handleExecReplies` (`execReply === null → throw new WatchError()`); the v4 stub test covers it. |
| [medium] RESP3 coverage and message assertions | **Fixed.** The test is parameterized over RESP2 and RESP3 and asserts message, `name` and `SimpleError` for INCR, an unknown command and EXEC items. I confirmed `HELLO 3` switches the mock to RESP3 (`ZSCORE` returns `"1.5"` before and `1.5` after). |
| [low] 4.0.x docs | **Fixed.** |
| [low] tautological test | **Removed.** |
| [low] stand-in coverage | **Fixed.** Added the new `…-absent.test.ts`, which blocks `redis` in `_resolveFilename` and checks message, `name`, `constructor.name` and `errors()` on the stand-ins. |
| [low] `.name` on stand-ins | **Fixed.** No stand-in sets `.name` any more. |
| [low] `requireRedis()` helper | **Fixed.** |
| [low] decode-hook ordering | **Fixed.** The hook now uses `redisErrors()` directly, and the test's `before()` hook is gone. |

**EXECABORT on v4:** same as v6 and not introduced here. @redis/client 1.5.12 `multiExecutor` sends MULTI, the commands and EXEC together under `Promise.all`, so it rejects with the first queue-time error. The mock rejects with `EXECABORT …`. This predates the PR and is still an out-of-scope follow-up. The `// Defensive … (shouldn't happen)` comment in `exec()` is inaccurate for that path.

**Tests on fda48d2:** `tsc`, eslint and prettier are clean. 77/77 targeted tests pass, and 1036/1036 unit + node-redis integration tests pass on the mock backend. Red-before-green: against fb38501 `src`, the new absent test fails 3 of 6 and the new inline-exec v4 test fails.

No new findings.

PR #466, opened 2026-09-23
feat: implement ECHO command (#434)

### LINKED ISSUE #434: ECHO is not implemented; the mock replies unknown command
Found by a reviewer while ground-truthing PR #431. The PR's protocol-layer test happens to use `ECHO` and still passes, because the oversized-bulk protocol error is raised **before** command dispatch — so the test is valid, but it incidentally revealed that `ECHO` does not exist in this server at all.

## Problem

`ECHO` returns an unknown-command error. Real Redis has supported it since 1.0.0 and every client library exposes it.

```
real:  ECHO hello  ->  $5\r\nhello\r\n
mock:  ECHO hello  ->  -ERR unknown command 'ECHO', ...
```

## Why it matters

`ECHO` is one of the commands people reach for first when smoke-testing a connection, and several client libraries and connection-pool health checks use it as a liveness probe alongside `PING`. A pool configured to health-check with `ECHO` fails against the mock and succeeds against real Redis.

## Notes for the implementation

- Arity is exactly 2 (`ECHO message`); real Redis returns `wrong number of arguments for 'echo' command` otherwise.
- Binary-safe: the reply is the argument bytes verbatim, so it must go through the Buffer path, not `toString()`.
- Not version-gated; behaviour is identical across 6.2 / 7.x / 8.x.
- Check its behaviour in subscribed mode (RESP2) against a real server — `ECHO` is in the small set of commands allowed there alongside `PING`/`SUBSCRIBE`/`UNSUBSCRIBE`, and the subscribed-mode policy needs to agree.

## Acceptance

- `ECHO <msg>` returns the argument bytes verbatim, including non-UTF-8 input, pinned by a raw-TCP test.
- Arity error wording matches real Redis.
- Subscribed-mode behaviour matches real Redis.
- `COMMAND INFO echo` reports arity and flags matching a real server.

fatal10110:
## Summary
- `ECHO` returned `-ERR unknown command`. Added `echoCommand` in `src/commands/connection.ts`:
  - schema `t.object({ message: t.bulk() })`: exact arity 2, and the reply is the raw argument `Buffer` (binary-safe, no `toString()`).
  - Not version-gated. It exists on every profile (Redis since 1.0.0).
  - It has **no** `subscribed` flag. I checked against real Redis 8.0.6 and 6.2: in RESP2 subscribed mode `ECHO` is rejected with `ERR Can't execute 'echo': only (P|S)SUBSCRIBE / ... are allowed in this context`, and the arity check runs first. In RESP3 subscribed mode it is allowed. The existing subscribed-mode policy (RESP2 only, keyed off the flag) already matches this. The issue text suggested ECHO is allowed in RESP2 subscribed mode, but real Redis rejects it there.
  - Hand-written introspection: `COMMAND INFO echo` gives arity `2`, flags `loading stale fast`, keys `0 0 0`, categories `@fast @connection` (same as Redis 7+/8). `COMMAND DOCS` gives summary/group `connection`.
- `docs/COMMANDS.md`: added ECHO to the Connection section. The README only links to COMMANDS.md, so it needs no change.

Closes #434

## Test plan
- [x] New `tests-integration/raw-tcp/echo.test.ts`: verbatim non-UTF-8/CRLF/NUL bytes, empty message, case-insensitive name, arity errors (0 and 2 args), RESP2 subscribed-mode rejection (and arity-before-subscribed ordering), RESP3 subscribed-mode allowed, MULTI/EXEC and Lua `redis.call('echo')`
- [x] ioredis + node-redis `connection-integration`: `echo`/`echoBuffer` (node-redis via `bufferClient`) binary round-trip, arity errors
- [x] ioredis + node-redis `command-integration`: `COMMAND INFO ECHO` arity/flags/keys/categories
- [x] All new tests red on mock before the fix, green on real Redis 8.0 (standalone 6399 + cluster), green on mock after
- [x] `npm run build`, `npm run lint`, `npm test` (462 pass), `test:integration:mock` (1261 pass), `test:integration:compatibility:mock` (all 7 profiles pass)
- [x] Real: raw-tcp echo/protocol/pubsub-protocol plus ioredis/node-redis connection/command/pubsub suites (68 pass)


tests-integration/ioredis/command-integration.test.ts
```diff
@@ -65,6 +65,22 @@ describe(`COMMAND integration (${testRunner.getBackendName()})`, () => {
     assert.strictEqual(reply[2], null)
   })
 
+  test('COMMAND INFO ECHO reports arity, flags and categories like Redis', async () => {
+    const [info] = (await redisClient?.command('INFO', 'ECHO')) as [
+      CommandInfoReply,
+    ]
+
+    assert.deepStrictEqual(info.slice(0, 7), [
```
fatal10110: [medium] Missing test for a claimed behavior. The PR description says `COMMAND DOCS` gives the summary and group for ECHO, and the PR hand-writes a full docs block (summary, since, group, complexity, arguments). But this test and its node-redis twin (`tests-integration/node-redis/command-integration.test.ts:72`) only assert `info.slice(0, 7)`. So nothing pins the hand-written tips, key specs and subcommands (indices 7-9), and nothing exercises `COMMAND DOCS echo` at all. `since`/`complexity` match Redis today only because the `commandDocs()` defaults happen to be `1.0.0` / `O(1)`. If those defaults change, or the summary drifts, no test fails.

Real Redis 8.0.6, verified:
```
COMMAND INFO echo -> echo | 2 | [loading stale fast] | 0 | 0 | 0 | [@fast @connection] | [] | [] | []
COMMAND DOCS echo -> summary "Returns the given string." | since "1.0.0" | group "connection" | complexity "O(1)" | arguments [name message type string display_text message]
```
Suggested fix: assert the whole `info` array (indices 7-9 are three empty arrays). Add a `COMMAND DOCS ECHO` assertion on summary/since/group/complexity and the argument name/type. It can skip `display_text`, which is a repo-wide gap; see the low comment on connection.ts.

src/commands/connection.ts
```diff
@@ -525,6 +525,30 @@ export const pingCommand = defineCommand({
...
+  flags: ['readonly', 'fast'],
+  introspection: {
+    arity: 2,
+    flags: ['loading', 'stale', 'fast'],
+    firstKey: 0,
+    lastKey: 0,
+    keyStep: 0,
+    categories: ['@fast', '@connection'],
+    keySpecs: [],
+    docs: commandDocs('Returns the given string.', 'connection', [
```
fatal10110: [low] ECHO's `COMMAND DOCS` argument has no `display_text` field. Redis 7.2+, Redis 8 and Valkey 9 all emit one. Byte diff against Redis 8.0: the real argument map is `*6` (RESP3 `%3`) with `$12 display_text $7 message`. The mock's is `*4` (RESP3 `%2`). This gap is repo-wide (`display_text` appears nowhere in `src/`), so it is not blocking here. Filing it as a follow-up is fine.

fatal10110:
## Review summary (reviewed head `23587cd`)

**Verdict: NOT YET.** One medium finding is open. There are no incompatibilities with real Redis.

| Severity | Location | Finding | Lens |
|---|---|---|---|
| medium | `tests-integration/ioredis/command-integration.test.ts:73` (plus node-redis `:72`) | The COMMAND INFO test only checks `slice(0,7)`. `COMMAND DOCS echo` is claimed but has no test. | Redis wire-compat |
| low | `src/commands/connection.ts:544` | The COMMAND DOCS argument has no `display_text` (repo-wide gap) | Redis wire-compat |

**Verified against real Redis (8.0 / 7.2 / 7.0 / 6.2, Valkey 9):**
- Verbatim binary reply, and an empty message gives `$0`.
- Arity errors, and in RESP2 subscribed mode ECHO is rejected with the arity check first. RESP3 subscribed mode allows it.
- The RESP3 reply is a plain bulk string, not verbatim.
- MULTI/EXEC, EXECABORT on an arity error, Lua `redis.call`/`EVAL_RO`, MONITOR output and NOAUTH all behave as in Redis.
- Keyless routing in cluster mode matches.
- COMMAND INFO elements match.

Leaving out the `subscribed` flag is correct.

**Architecture:** no issues. The `readonly`/`fast` internal flags follow `ping`. Hand-written introspection is needed, because the fallback would report arity -1. The command stays pure.

**Tests:** red on the base code and green on the PR (mock). The raw-tcp echo suite passes 7/7 against real Redis 8.0.6. Unit tests pass 462/462. Compatibility suites pass on the redis-6.2 and valkey-8.0 profiles. Typecheck and lint are clean.

**Out-of-scope follow-ups (existing, repo-wide):**
- COMMAND INFO flags and categories are sent as bulk strings, not simple strings or RESP3 sets.
- The Lua arity error wording differs from Redis.
- Per-profile COMMAND INFO (#370).
- The redis-6.2 subscribed-mode error text should say `(P)SUBSCRIBE`.
- CLIENT REPLY is missing.
- MONITOR orders the EVAL line after the inner Lua line.

**Not ground-truthed:** "no MOVED for keyless commands" across several real cluster nodes. The shared cluster is down, so only a single-node cluster-mode server was checked. Mock NOAUTH for ECHO was checked by reading the code, not by running it.

tests-integration/ioredis/command-integration.test.ts
fatal10110: Addressed in c099a12. Both the ioredis and node-redis `COMMAND INFO ECHO` tests now `deepStrictEqual` the whole 10-element reply, including the three empty arrays for tips, key specs and subcommands. A new `COMMAND DOCS ECHO` test (ioredis, plus a node-redis twin that pins the whole field object) asserts summary `Returns the given string.`, since `1.0.0`, group `connection`, complexity `O(1)`, and exactly one argument `message`/`string`. It does not check `display_text`; see the connection.ts thread. Green on mock. Values match Redis 8.0.6 on 6399. The real cluster run is currently blocked: the shared cluster is in `cluster_state:fail` because node 30005 is down, which has nothing to do with this PR.

src/commands/connection.ts
fatal10110: Deferred as a follow-up. `display_text` depends on the version: Redis 6.2 has no `COMMAND DOCS` at all, and the reviewer notes that 7.2+/8/Valkey 9 send it. Adding it correctly means a new `CommandDocumentationArgument.displayText` field, serializer support in `command.ts`, and probably a profile gate. The same gap applies to every command's docs, so a repo-wide change belongs with the #370 introspection work rather than a one-off for ECHO. The new COMMAND DOCS test says so in a comment and does not assert the field, so it won't block that change.

fatal10110:
## Re-review (head `c099a12`, previous review `23587cd`)

**Verdict: SHIP.** No high or medium findings are open.

**Prior findings:**
- **[medium] COMMAND INFO / COMMAND DOCS coverage: fixed.**
  - Both clients now assert the whole `COMMAND INFO ECHO` reply, including the empty tips, key specs and subcommands at indices 7-9.
  - A new `COMMAND DOCS ECHO` test pins summary, since, group, complexity and the argument name and type. The node-redis version also asserts the exact set of doc fields.
  - I spot-checked the expected values read-only against real Redis 8.0.6 on 6399. They match.
- **[low] Missing `display_text`: deferred.** It is a repo-wide gap. The test comment now explains why it is not asserted. This is acceptable as a follow-up.

**New findings:** none. The fix commit changes tests only.

**Tests (mock):**
- The touched suites pass 46/46: ioredis and node-redis `command-integration` and `connection-integration`, plus raw-tcp `echo`.
- Typecheck and lint are clean.
- The command-integration suites could not be run against real Redis, because they need the shared cluster, which is currently down.

PR #467, opened 2026-09-23
fix: clean:redis skips slotless cluster nodes and blames master failures on the master (#458)

### LINKED ISSUE #458: clean:redis: handshake nodes make the result seed-dependent; master failures blamed on replicas
Found while re-reviewing PR #429 (`scripts/flush-redis.ts`, `clean:redis`). Both issues are low severity. Each makes `clean:redis` **fail loudly with a misleading result**; neither makes it report success wrongly.

## 1. A node that is still joining the cluster makes the result depend on the seed
A `handshake` entry in `CLUSTER NODES` fails `clean:redis` on an otherwise healthy cluster, and whether it does depends on which seed port is listed. Pointing `CLUSTER MEET` at a dead port: seed 7321 exits 1, seed 7322 exits 0, same cluster, same moment.
→ Only flush and verify masters that own slots, plus their replicas.

## 2. A master's failure is reported against its replicas
When a master fails its own check after `FLUSHALL`, its replicas are reported as "this run did not flush" with the stale-replica hint. Both the message and the hint point away from the real cause (the master).

## Acceptance
- A handshake/`noaddr` node that owns no slots doesn't fail an otherwise healthy flush, whichever seed is used.
- A master failure is reported on the master, and its replicas' messages refer back to it.

fatal10110:
## Summary
Two misleading-failure bugs in `npm run clean:redis` (`scripts/flush-redis.ts`):

1. **Handshake/`noaddr` nodes made the result seed-dependent.** Every node any seed listed in `CLUSTER NODES` became a target, including a `handshake` entry that exists only on the node that issued the `CLUSTER MEET`. So seed A failed on it and seed B did not. Discovery now covers only **slot-owning masters plus their replicas**. Slotless nodes (`handshake`, `noaddr` placeholders, masters with no slots and their replicas) are skipped, and each skip is logged. Seed views are merged by node id: a node is covered if *any* seed reports it owning slots (or replicating an owner), so a stale view cannot shrink the flush. A slot-owning master listed as `noaddr` is still a failure, as before.
2. **Master failures were blamed on their replicas.** When a master failed (FLUSHALL error or keys left behind, unreachable, or `CLUSTER NODES` unreadable), its replicas were reported as "this run did not flush" with the stale-replica hint. Replicas now say `not verified: its master failed (see <master label>) — fix the master first`, with no hint of their own, so only the master's hint is printed. A master the run never knew about still gets the stale-replica diagnosis.

The pure logic (`parseClusterNodes`, `planCluster`, `replicaMasterProblem`, `HINT`) moved to `scripts/flush-redis-topology.ts` so it can be unit-tested against `CLUSTER NODES` fixtures without touching a real cluster.

Closes #458

## Test plan
- [x] New unit test `tests/flush-redis-topology.test.ts` (11 cases): parsing incl. migration markers, healthy topology, handshake skipped and the same plan from either seed, noaddr slotless skipped, noaddr slot owner still covered, slotless master and its replica skipped, view merging, replica attribution (flushed / failed master / unknown master)
- [x] E2E on a throwaway local Redis 7.2 cluster (ports 47000-47005, torn down afterwards; shared docker-compose cluster not touched):
  - `CLUSTER MEET 127.0.0.1 47999` on 47000: old script exits 1 with seed 47000 and 0 with seed 47001. New script exits 0 with both and logs the handshake node as skipped.
  - `ACL SETUSER default -flushall` on master 47000: old script reports replica 47005 as "this run did not flush" with the stale-replica hint. New script reports it as "its master failed (see cluster node 127.0.0.1:47000)" and prints only the master's hint.
  - Master 47002 shut down: its replica 47004 refers back to the unreachable master.
- [x] `npm run build`, `npm run lint`, `npm test` (473 pass)


scripts/flush-redis-topology.ts
```diff
@@ -0,0 +1,184 @@
...
+      const hostPort = address.split('@')[0]
+      const separator = hostPort.lastIndexOf(':')
+      const slots = rest.slice(4)
+      return {
+        id,
+        host: hostPort.slice(0, separator),
+        port: Number(hostPort.slice(separator + 1)),
+        flags: flags.split(','),
+        masterId: master === '-' ? null : master,
+        ownsSlots: slots.some(slot => slot !== '' && !slot.startsWith('[')),
```
fatal10110: [high] **False success: a slotless master that is the target of an in-progress slot migration holds keys a cluster client can still read. The PR now skips it, where `main` flushed it.**

`ownsSlots` ignores the `[N-<-id]` / `[N->-id]` entries, and `planCluster` covers only slot owners. But during a `MIGRATE`, keys move to the *importing* node before `CLUSTER SETSLOT ... NODE` hands it the slot. A cluster client still reaches those keys through `-ASK`. Real Redis also prints the markers **only on the `myself` line**, so from any other seed the importing node looks like a plain empty `master`. Even when the migrating node is itself the seed, its `[5061->-X]` entry is thrown away.

Reproduced on a private Redis 7.2.1 cluster (ports 47100–47106). 47106 is a fresh master with no slots:
```
$ redis-cli -p 47100 set bar stale-value                     # slot 5061, owned by 47100
$ redis-cli -p 47106 cluster setslot 5061 importing <47100-id>
$ redis-cli -p 47100 cluster setslot 5061 migrating <47106-id>
$ redis-cli -p 47100 migrate 127.0.0.1 47106 bar 0 5000
$ redis-cli -p 47100 cluster nodes | grep myself
... 127.0.0.1:47100@57100 myself,master - 0 ... connected 0-5460 [5061->-0527334151...]
$ redis-cli -p 47101 cluster nodes | grep 47106
0527334151... 127.0.0.1:47106@57106 master - 0 ... 0 connected          # no marker, no slots

$ REDIS_CLUSTER_PORTS=47100 node --import tsx scripts/flush-redis.ts     # this PR
  cluster node 052733415132… 127.0.0.1:47106: skipped, owns no slots and replicates no slot owner (flags: master)
  ...
exit=0
$ redis-cli -p 47106 dbsize        -> 1
$ redis-cli -c -p 47101 get bar    -> "stale-value"      # the next test can read it

# same state, origin/main's scripts/flush-redis.ts:
  cluster node 127.0.0.1:47106 (discovered): flushed, cluster_state:ok
exit=0 ; redis-cli -p 47106 dbsize -> 0
```
So both the header doc in `flush-redis.ts` ("hold no keys a test can reach") and the `planCluster` doc ("Only masters that own slots, plus their replicas, can hold keys a test will read") are wrong for this state.

Suggested fix, and the smallest one that meets #458's acceptance: skip only nodes nobody can connect to (`handshake`, and `noaddr`/port-0 entries with no slots), and keep flushing every other reachable master and its replicas as before. A handshake entry never has a real node behind its id, so that alone removes the seed dependence. If you keep the "slot owners only" rule, at least count a node as an owner when any view names it as a migration destination (`[N->-<id>]`) or shows it importing (`[N-<-…]`). Note that this still misses the case where neither end of the migration is a seed. Add a fixture test for whichever rule you pick.

scripts/flush-redis-topology.ts
```diff
@@ -0,0 +1,184 @@
...
+      }
+      reports.set(node.id, [...(reports.get(node.id) ?? []), node])
+    }
+  }
+
+  const members: ClusterNode[] = []
+  const skipped: ClusterNode[] = []
+
+  for (const [id, nodeReports] of reports) {
+    if (seedIds.has(id)) {
```
fatal10110: [medium] **The result still depends on the seed when a seed is a replica of a slotless master.**

Seeds bypass the plan (`if (seedIds.has(id)) continue`) and are always targets. Take a seed that is a replica of a master `planCluster` skipped. `verifyReplica` finds its master neither flushed nor failed, so it reports `"which this run did not flush"` plus the stale-replica hint. The same cluster seeded from any other node skips that replica and exits 0. The default `REDIS_CLUSTER_PORTS` lists all six nodes, replicas included, so seeding from a replica is the normal setup.

Reproduced on a private Redis 7.2.1 cluster. 47106 is a master with no slots, and 47107 was put under it with `CLUSTER REPLICATE`:
```
$ REDIS_CLUSTER_PORTS=47100 node --import tsx scripts/flush-redis.ts
  cluster node 88085304758d… 127.0.0.1:47107: skipped, owns no slots and replicates no slot owner (flags: slave)
  cluster node 052733415132… 127.0.0.1:47106: skipped, owns no slots and replicates no slot owner (flags: master)
  ...
exit=0

$ REDIS_CLUSTER_PORTS=47107 node --import tsx scripts/flush-redis.ts
  ...
clean:redis failed — 1 of 6 endpoint(s) are not verifiably empty:
  cluster node 127.0.0.1:47107: replica of port 47106, which this run did not flush — nothing emptied it, and it refuses FLUSHALL itself
What to do:
  - A replica is not following a flushed master: wait for it to resync, ...
exit=1
```
This is the same kind of seed dependence that #458 describes, and the diagnosis is wrong: the run chose not to flush that master. Fix options: have `planCluster` also classify the seeds and return the skipped ones, then treat a seed replica of a skipped master as skipped too (or pass `replicaMasterProblem` a "skipped master" set and return a non-failure). Either way, add a fixture test where the seed is such a replica. The finding goes away if the skip rule is narrowed to handshake/noaddr only, as suggested on L66.

scripts/flush-redis.ts
```diff
@@ -300,11 +270,17 @@ async function discoverCluster(
...
+  const { members, skipped } = planCluster(views)
+  for (const node of skipped) {
+    const address = isAddressless(node) ? '' : ` ${node.host}:${node.port}`
+    console.log(
+      `  cluster node ${node.id.slice(0, 12)}…${address}: skipped, owns no slots ` +
+        `and replicates no slot owner (flags: ${node.flags.join(',')})`,
+    )
+  }
+
+  const addressless = members.filter(isAddressless)
```
fatal10110: [low] **An addressless slot-owning master's replicas still get the old "this run did not flush" message and the stale-replica hint.**

The unit test on L155 covers this case: `master,fail,noaddr` owning slots, with its replica R3 still covered. The master goes into `addressless` and becomes an `Unreachable` with **no `masterPort`** (the `addressless.map(...)` below never sets it). A noaddr node's port is also `0`, while the replica's `INFO replication` still reports the real `master_port`. So `failedMasters` never matches, and R3 prints the misleading message plus `HINT.staleReplica`. That is the second bug #458 asks to fix, left open for this one path.

The noaddr state is rare and I could not reproduce it live, so this is low. A cheap fix is to key the attribution by master **id** instead of port: the replica's own CLUSTER NODES line carries `<master-id>`. Or record the replica's `masterId` in the plan and look the failed master up by id.

scripts/flush-redis.ts
```diff
@@ -440,8 +421,24 @@ async function main(): Promise<void> {
     const flushedPorts = new Set(
       masters.filter(t => t.failure === null).map(t => t.port),
     )
+    // Every master this run tried and failed — including the ones it could not
+    // even connect to — so their replicas blame the master, not themselves.
+    const failedMasters = new Map<number, string>()
```
fatal10110: [low] **No automated test covers the wiring that fixes the second bug.**

`replicaMasterProblem` is unit-tested. The inputs it depends on are not: this `failedMasters` construction (unreachable entries via `Unreachable.masterPort` ← `Endpoint.clusterMaster` ← `node.masterId === null`, plus failed master targets). Drop `clusterMaster` in a later refactor and replicas of an unreachable master silently go back to the stale-replica message, with nothing turning red. I confirmed both paths by hand on a private cluster: `ACL SETUSER default -flushall` on a master, and a master shut down.

Consider moving the "which masters failed" computation into `flush-redis-topology.ts` as a pure function (unreachable + targets → `Map<port,label>`) and adding a case per source.

Related nit: the `Endpoint.clusterMaster` doc says "slot-owning master", but the flag is set from `node.masterId === null` on whichever report `planCluster` kept (the last one with an address). That report is not necessarily the one that made the node covered.

tests/flush-redis-topology.test.ts
```diff
@@ -0,0 +1,231 @@
...
+describe('clean:redis cluster topology (#458)', () => {
+  test('parses master, slots, and replica-of from CLUSTER NODES', () => {
+    const nodes = view([
+      ...healthyView(30000),
+      line(
+        id('7'),
+        '127.0.0.1:30006@40006,host-7',
+        'master',
+        '-',
+        '[5461->-' + M2 + ']',
```
fatal10110: [low] **This fixture is not something real Redis prints.** Migration markers appear only on the `myself` line; other nodes never show them. From a private Redis 7.2.1 cluster:
```
$ redis-cli -p 47100 cluster nodes | grep myself
... 127.0.0.1:47100@57100 myself,master - 0 ... 1 connected 0-5460 [100->-29bc5616...]
$ redis-cli -p 47101 cluster nodes | grep myself
... 127.0.0.1:47101@57101 myself,master - 0 ... 2 connected 5461-10922 [100-<-4c7e44ed...]
$ redis-cli -p 47102 cluster nodes | grep 47100
... 127.0.0.1:47100@57100 master - 0 ... 1 connected 0-5460          # no marker
```
The parse assertion still holds. But the file header says these fixtures are "what real Redis 7 prints", and this one hides the fact that matters for the L66 finding: a non-seed importing node looks exactly like an empty master. Suggest putting the marker on a `myself` line and adding an importing (`[N-<-id]`) case. Nit: a real handshake line prints ping-sent `0`, e.g. `548016... 127.0.0.1:47999@57999 handshake - 0 0 0 disconnected`, but the fixture uses `1700000000000`. Harmless.

fatal10110:
## Review of 7626ee5: **NOT YET**

Checked from two angles: Redis wire-compat and Node.js architecture. Every Redis-behaviour claim was reproduced on a private Redis 7.2.1 cluster (ports 47100–47107), which was torn down afterwards. The shared fixtures were not touched.

**What works (confirmed end to end):**
- **Handshake entries:** a `CLUSTER MEET` to a dead port leaves a handshake entry, which is now skipped and logged, and the run exits 0 from either seed.
- **FLUSHALL failure on a master** (via `ACL SETUSER default -flushall`): its replica now reports `its master failed (see cluster node 127.0.0.1:47100 (discovered))`.
- **Master shut down:** its replica points back at it.
- **After failover:** the old `master,fail` entry, which no longer owns slots, is skipped cleanly.
- **Parser against real output:** the parser's field indexing matches real `CLUSTER NODES` output, and a handshake line parses as slotless.

**Findings (inline):**

| Severity | Where | Summary |
|---|---|---|
| high | `scripts/flush-redis-topology.ts:66` | False success, and a regression from `main`: during an in-progress migration, the importing master owns no slots yet but holds migrated keys that a client can reach via `-ASK`. It is now skipped, the run exits 0, and `GET bar` still returns the stale value. `main` flushed it. |
| medium | `scripts/flush-redis-topology.ts:128` | Seed dependence remains. A seed that is a replica of a skipped slotless master fails with "this run did not flush" plus the stale-replica hint (exit 1). The same cluster seeded from a master exits 0. |
| low | `scripts/flush-redis.ts:282` | Replicas of an addressless (`noaddr`) slot-owning master still get the old stale-replica message: no `masterPort`, and the port is `0`. |
| low | `scripts/flush-redis.ts:426` | No automated test covers the `failedMasters` / `clusterMaster` wiring. `Endpoint.clusterMaster` doc is imprecise. |
| low | `tests/flush-redis-topology.test.ts:101` | The migration-marker fixture sits on a non-`myself` line, which real Redis never prints. |

**Out-of-scope follow-ups (not blocking):**
- An unreachable configured seed is reported twice, once as configured and once as `(discovered)`. This predates the PR.
- A FLUSHALL error on a master, such as `NOPERM`, carries no hint, so no "What to do" section is printed.
- `tsconfig.json` includes only `src/**`, so `tsc --noEmit` never type-checks `scripts/` or `tests/`. This predates the PR.

**Tests:** `node --test tests/flush-redis-topology.test.ts` gives 11/11 passing. `tsc --noEmit` and `eslint` on the changed files are clean.

scripts/flush-redis-topology.ts
fatal10110: Fixed in e66c826. I dropped the "slot owners only" rule and took your suggestion: `planCluster` now skips only entries with no server behind them. That means a `handshake` entry, or an addressless (`noaddr`/port 0) entry that serves no slots. It checks this per report and skips a node only when *every* seed's report of it is such an entry. Every other reachable master and its replicas are covered again, as on `main`, slots or not. I corrected the docs in `flush-redis.ts` and `flush-redis-topology.ts`; the `isPhantom` doc now explains the importing-node case and notes that markers appear only on `myself`.

Test: `a slotless master mid-migration is covered from every seed` plans from the migrating node (`[5061->-M4]` on its `myself` line), from the importing node (`[5061-<-M1]`), from a bystander, and from a replica of the importing node. It asserts full coverage and nothing skipped each time.

E2E on a private Redis 7.2.1 cluster (47000-47007; 47006 an empty master): SETSLOT IMPORTING/MIGRATING, then MIGRATE `bar` to 47006, seed 47001.
- 7626ee5: `47006 skipped`, exit 0, `dbsize 1`, and `redis-cli -c -p 47001 get bar` returns `stale-value`.
- e66c826: `127.0.0.1:47006 (discovered): flushed`, exit 0, `dbsize 0`, and GET returns nil.

scripts/flush-redis-topology.ts
fatal10110: Fixed in e66c826, through the narrowed rule from the L66 thread. Skipping now looks at each node on its own, never at its master or at which seed reported it, so a slotless master and its replica are covered from every seed. Test: `a replica of a slotless master is covered whichever seed is used` checks that planning from R4 (a replica of the slotless M4) covers the same nodes as planning from M1.

E2E, same private cluster, 47007 a replica of the slotless master 47006:
- 7626ee5 seeded from 47007: exit 1 with "which this run did not flush" and the stale-replica hint.
- e66c826 seeded from 47007: exit 0 (47006 flushed, 47007 `replica of port 47006, empty`). Seeded from 47000: also exit 0.

scripts/flush-redis.ts
fatal10110: Fixed in e66c826. Masters are now matched by node id. Each target keeps its CLUSTER NODES report: discovered nodes from the plan, seeds from their own `myself` line. `masterOutcomes` records every master under `id:<id>` and `port:<port>`. `replicaMasterProblem(masterPort, masterId, outcomes)` looks the master up by the id from the replica's own line first, and falls back to the port only when no id was ever learned: a standalone server, or a seed whose CLUSTER NODES failed. An addressless master now carries its node on the `Unreachable` entry, so its replicas point back at it.

Test: `a replica is judged by its master id first` covers a noaddr master (port 0) whose replica still reports `master_port` 30002. The replica gets "its master failed (see cluster node cccccccccccc…)" and no hint.

scripts/flush-redis.ts
fatal10110: Fixed in e66c826. The failed-master bookkeeping is now the pure `masterOutcomes(endpoints)` in `flush-redis-topology.ts`. `main()` only maps unreachable entries and targets onto it: `{ label, port, node, role?, failed }`. Test: `sorts masters from every source into flushed and failed` covers a flushed master, a master whose FLUSHALL failed, an unreachable addressless master, an unreachable discovered master, a seed master known only by port, and replicas both reachable and not, plus an unanswering seed of unknown role (neither of those counts as a master). If the discovered-node wiring is dropped again, masters stop landing under `id:` keys and the attribution tests fail.

`Endpoint.clusterMaster` is gone. It was replaced by `Endpoint.node`, documented as "this node's CLUSTER NODES report: set up front for a discovered node, and from its own `myself` line for a seed once discovery has read it." It no longer claims anything about slots.

tests/flush-redis-topology.test.ts
fatal10110: Fixed in e66c826. The fixtures are built by `viewFrom(seed, specs, myselfSlots)`, which prints migration markers only on the seed's `myself` line. The parse test now uses an importing marker (`[5061-<-M1]`) on `myself`, and the migration test covers both `->` and `-<-` from each end plus a bystander. The handshake fixture now prints ping-sent `0` (`... handshake - 0 0 0 disconnected`).

scripts/flush-redis.ts
```diff
@@ -367,26 +361,30 @@ async function flushMaster(target: Target): Promise<string> {
...
 
-  if (!flushedPorts.has(replication.masterPort)) {
-    throw new FlushFailure(
-      `replica of port ${replication.masterPort}, which this run did not flush — ` +
-        `nothing emptied it, and it refuses FLUSHALL itself`,
-      HINT.staleReplica,
-    )
+  const problem = replicaMasterProblem(
+    replication.masterPort,
+    target.node?.masterId,
```
fatal10110: [low] **For a *discovered* replica, `target.node` is not "the replica's own CLUSTER NODES line".**

Only seeds get `node` from their own `myself` line (L278). A discovered replica's `node` is the report `planCluster` kept from some seed's view: the last one that has an address. `replicaMasterProblem` checks the id before the port, so a seed's stale `<master>` id wins over the replica's own `INFO replication` `master_port`. That can happen in the gossip window right after a failover. The replica then follows the new, flushed master, but a seed still lists it under the old, dead one, so the replica is reported as `its master failed (see <old master>)`.

This is not a false success. Either that run fails anyway, because the dead old master is covered and unreachable, or the DBSIZE poll catches a replica that really follows an unflushed master. So I rate it low. Still, the `replicaMasterProblem` `@param masterId` doc and the PR description ("taken from the replica's own `CLUSTER NODES` line") overstate what it is. Either fix the wording, or read the id from the replica's own `myself` line: you already hold a connection to it, and one `CLUSTER NODES` call per discovered replica is cheap. That would also make id and port come from the same source.

fatal10110:
## Re-review of e66c826: **SHIP**

I re-ran every scenario on a private Redis 7.2.1 cluster (ports 47120–47127, confirmed free with `lsof` first, torn down afterwards). The shared fixtures were not touched.

**Prior findings:**

| Prior | Status | Evidence |
|---|---|---|
| [high] slotless importing master skipped (false success) | **fixed** | `MIGRATE bar` to slotless 47126 mid-migration: 47126 is now `flushed`, `dbsize` is 0, and the run exits 0. |
| [medium] seed-replica of a slotless master fails the run | **fixed** | Seeding from 47127 (replica of slotless 47126) exits 0, with `replica of port 47126, empty`. |
| [low] noaddr slot-owning master's replicas get the stale-replica hint | **fixed** | Masters are now matched by id first. Unit-tested; I could not create a live `noaddr` node. |
| [low] failed-master wiring untested / `clusterMaster` doc | **fixed** | New pure `masterOutcomes` with a test that covers every source. `Endpoint.node` replaces `clusterMaster`. |
| [low] migration-marker fixture on a non-`myself` line | **fixed** | Markers are now only on `myself` lines. There is an importing case too. |

**Also confirmed end to end:**
- **Handshake entry:** after `CLUSTER MEET` to the dead port 47199, the entry is logged as `skipped, no server behind this entry` and the run exits 0.
- **Master FLUSHALL denied** (`ACL SETUSER default -flushall`): its replica reports `its master failed (see cluster node 127.0.0.1:47120 (discovered))`.
- **Master shut down:** its replica points back at it.

**Keeping dead nodes a failure (the implementer's call):** I agree with it. After a failover, the old `master,fail` entry has no slots and `cluster_state:ok`, but it still fails the run with the `clusterUnhealthy` hint. This matches `main` and is seed-independent, because every view lists a known failed node. It errs toward a loud failure, never a false success. Skipping it would need a `fail`-flag special case, which is only safe when the node owns no slots and has no importing state. Nobody can observe that importing state for a dead node, so the conservative choice is right.

**New finding:**
- [low] `scripts/flush-redis.ts:383`: a discovered replica's master id comes from a seed's view, not from the replica's own `myself` line, so the docs and PR text overstate it. The worst case is a misleading message in the seconds after a failover, never a false success.

**Out-of-scope follow-ups (unchanged):**
- An unreachable configured seed is reported twice. This predates the PR.
- A FLUSHALL error such as `NOPERM` carries no hint, so the "only the master's hint is printed" line in the description prints no hint at all in that case.
- `tsconfig.json` only type-checks `src/**`.

**Tests:** `node --test tests/flush-redis-topology.test.ts` gives 14/14 passing. `tsc --noEmit` and `eslint` on the changed files are clean.

Issue #449, opened 2026-09-23
Lua: typed tables ({double}/{big_number}/{map}) dropped without redis.setresp(3); setresp(3) redis.call replies convert differently


fatal10110:
Found while reviewing PR #427, and verified against real Redis 7.2.1 and 8.0.6. **Pre-existing on `main`, in `src/core/lua-runtime.ts` and the Lua engine.**

## 1. Typed Lua tables are dropped unless the script calls `redis.setresp(3)`
Real Redis converts `{double=…}`, `{big_number=…}` and `{map=…}` tables no matter what protocol level the script uses. The mock returns `[]`.
```
EVAL "return {big_number='12345678901234567890'}" 0
  real RESP2: $20\r\n12345678901234567890    real RESP3: (12345678901234567890
  mock:       [] at both protocols
EVAL "return {double=2.5}" 0     real: "$3\r\n2.5" / ",2.5"    mock: []
```
Only **booleans** actually depend on `redis.setresp(3)`. `docs/TESTING.md` (as updated by #427) describes the rest as a mock gap.

## 2. With `redis.setresp(3)`, replies read through `redis.call` convert differently than in real Redis
```
redis.setresp(3); return redis.call('HGETALL','h')   on RESP3: real -> object, mock -> flat array
redis.setresp(3); return redis.call('ZSCORE','z','b') on RESP3: real -> number 2.5, mock -> string "2.5"
```

## Acceptance
Both cases match real Redis at RESP2 and RESP3, pinned by client-level EVAL tests.

fatal10110:
Part 1 and the related RESP3 null gap need engine changes, now filed upstream: fatal10110/lua-redis-wasm#29 (typed tables encode as [] without setresp(3)) and fatal10110/lua-redis-wasm#30 (RESP3 null should reach the script as nil). The 6.2 pcall position prefix is fatal10110/lua-redis-wasm#28.

Issue #460, opened 2026-09-23
waitUntilGone default deadline is too loose at the four 5ms-TTL sites


fatal10110:
Follow-up from the re-review of PR #423. The PR's agent could not be resumed to make this change.

## Problem
`waitUntilGone()` (`tests-integration/utils.ts`) defaults to a 5000ms deadline. At the four sites that wait on a **5ms** TTL, that default is loose enough that a TTL 1000× too long still passes. That is exactly what you'd see if HPEXPIRE treated milliseconds as seconds, the bug the old `delay(40)` caught. The reviewer copied the loop and ran it on real Redis 8.0.6: a 5-second field TTL passed 17 of 18 runs.

Sites:
- `tests-integration/ioredis/hash/field-expire.test.ts`: 74, 119
- `tests-integration/node-redis/hash/field-expire.test.ts`: 54, 95

## Fix
Pass `{ timeoutMs: 1000 }` at those four sites, and keep the 5000ms default for the 500ms-TTL sites. Correct the helper comment's "not a weaker assertion" claim: the deadline must be chosen per site relative to the TTL.

PR #468, opened 2026-09-23
fix: hand redis.call replies to setresp(3) scripts as RESP3 typed tables (#449)

fatal10110:
## Summary
- **Part 2 of #449, fixed here.** After `redis.setresp(3)`, real Redis hands a script RESP3 replies from `redis.call`. A map arrives as `{map=…}`, a double as `{double=…}`, a set as `{set=…}`, a boolean as a Lua boolean, a big number as `{big_number=…}`, and a verbatim string as `{verbatim_string=…}`. The mock always flattened replies to their RESP2 shapes, so after `setresp(3)` HGETALL gave a flat array instead of a map and ZSCORE gave the string `"2.5"` instead of the number 2.5. `RedisLuaRuntime` now records the script's protocol through the engine's `onSetResp` hook and resets it on every eval. `redisValueToLuaReply` builds the RESP3 typed tables to match `encodeResp3`. The RESP2 path is unchanged, and so are `renderScriptError` and the double formatter.
- **Part 1 of #449, not fixable here.** Returning `{double=…}`, `{big_number=…}`, `{map=…}` or `{set=…}` without `setresp(3)` still gives `[]`. The engine's C encoder skips typed tables unless the script is at RESP3: in `lua-redis-wasm` `wasm/src/runtime.c`, `encode_table` only calls `encode_resp3_marker` under `redis_resp_version() == 3`. The fix is to call it unconditionally; only the `LUA_TBOOLEAN` branch should keep the RESP-version check. By the time the host sees the reply, the table has already been encoded as an empty array, so glue code can't recover it.
- **Related engine gap.** `decode_reply` turns `REPLY_NULL` into Lua `false`. Real Redis at RESP3 gives the script `nil`, so `redis.setresp(3); return redis.call('GET', missing)` returns `false` (`#f` / `:0`) instead of null. The fix is to push `nil` when `redis_resp_version() == 3`.
- New client-level EVAL tests use ioredis at RESP2 and node-redis at RESP2 and RESP3. They pin real Redis behavior for both parts. The engine-gap cases are marked `todo` on the mock backend, so they will show up once a new `lua-redis-wasm` ships the fix.
- `docs/TESTING.md` now describes the `setresp(3)` reply shapes and names both remaining engine gaps.

Refs #449. Part 2 is fixed. Part 1 needs a `lua-redis-wasm` release plus a dependency bump, so this PR deliberately does not auto-close the issue.

## Test plan
- [x] New tests `tests-integration/{ioredis,node-redis}/scripts-typed-replies.test.ts` failed on mock before the fix: HGETALL map, ZSCORE and ZINCRBY double, ZRANGE WITHSCORES pairs.
- [x] Both files pass against real Redis 8.0 (cluster), all 25 cases including the ones that are `todo` on mock.
- [x] With the fix they pass on mock. 11 cases are `todo` on mock: 10 engine gaps, plus the Lua-side `{set=…}` check, because the mock's SMEMBERS replies with an array rather than a RESP3 set.
- [x] `npm run build`, `npm run lint` and `npm test` (462/462) pass.
- [x] `test:integration:mock` passes: 1262 pass, 0 fail, 11 todo.
- [x] Script suites run directly against real Redis (ioredis/node-redis `scripts*`, raw-tcp `script-resp3-replies`): 70/70 pass.


docs/TESTING.md
```diff
@@ -277,11 +277,17 @@ A Lua script can return two more of these kinds. A `{big_number=…}` table is a
...
+hands the script RESP3 replies — `HGETALL` as a `{map=…}` table, `ZSCORE` as a
+`{double=…}` table — so returning one gives the client a map or a double.
+
+Known mock gaps here, both in the bundled Lua engine (`lua-redis-wasm`), not in
+Redis: the `setresp(3)` requirement is Redis's rule for booleans only — real
+Redis converts `{big_number=…}`, `{double=…}`, `{map=…}` and `{set=…}` tables
+without it, while the mock's engine returns `[]` for them unless the script
+calls `redis.setresp(3)` first. And after `redis.setresp(3)` a missing value
+(`GET` of an absent key) reaches the script as `false`, not `nil`, so returning
+it gives `false` rather than a null reply.
```
fatal10110: [medium] The documented null gap misdescribes what clients actually get, and leaves out the array case.

The engine gap itself is real: `redis_api.c` `decode_reply`, `REPLY_NULL` → `lua_pushboolean(L, 0)`. But the doc says "returning it gives `false` rather than a null reply", and that is only true for a RESP3 client. At `setresp(3)` the engine encodes the Lua `false` as `REPLY_BOOL`, so a RESP2 client such as ioredis receives **integer 0**. The larger effect goes unmentioned: in real Redis a `nil` inside a reply ends the Lua array, while the mock keeps a `false` element.

Checked against real Redis 7.2.1:
```
redis.setresp(3); return redis.call('GET','nope')
   real: (nil) at RESP2 and RESP3        mock: (integer) 0 / (false)
redis.setresp(3); return redis.call('HMGET','h','f','nope','g')
   real: 1) "v"                          mock: 1) "v" 2) (integer) 0 3) "w"
```
The same happens with MGET, ZMSCORE, XREAD with no data, and so on. Suggested wording: "…reaches the script as `false`, not `nil`. Returning it gives `false` to a RESP3 client and `0` to a RESP2 client, not a null reply, and inside an array it doesn't end the array the way a Lua `nil` does in Redis." It would also help to pin the HMGET case next to the existing `ENGINE_NULL_GAP` todo tests.

tests-integration/node-redis/scripts-typed-replies.test.ts
```diff
@@ -0,0 +1,191 @@
...
+    })
+
+    test('ZRANGE WITHSCORES is member/double pairs', async () => {
+      const script =
+        "redis.setresp(3); return redis.call('ZRANGE', KEYS[2], 0, -1, 'WITHSCORES')"
+      assert.deepStrictEqual(await evalAt(2, script), [['b', '2.5']])
+      assert.deepStrictEqual(await evalAt(3, script), [['b', 2.5]])
+    })
+
+    test('SMEMBERS is a set', async () => {
```
fatal10110: [medium] This test can't tell a set from an array, so on the mock it passes without proving what its title says.

node-redis turns a RESP3 set (`~`) into a plain JS array by default, so `['x']` at both protocols is what it reads for either shape. The mock's SMEMBERS still sends an array: the tests' own `SET_REPLY_GAP` says so, and no command in `src/` builds a `RedisValue.set`. Checked against real Redis 7.2.1:
```
real RESP3 client:  1~ "y"  2~ "x"
mock RESP3 client:  1) "x"  2) "y"
real Lua (setresp 3): SMEMBERS => {set={x=true,y=true}}   mock Lua: {1="x",2="y"}
```
A reader would take the `{set=…}` path to be covered on the mock, when the only test that checks it is `todo`. Either rename the test to what it actually checks (for example "SMEMBERS members reach the client"), or make it tell the two shapes apart, for example with node-redis type mapping (`RESP_TYPES.SET` → `Set`) on the RESP3 client, and mark it `todo: SET_REPLY_GAP` like the Lua-side test.

src/core/lua-runtime.ts
```diff
@@ -360,6 +392,50 @@ function redisValueToLuaReply(value: RedisValue): ReplyValue {
...
+      return {
+        verbatim_string: {
+          format: Buffer.from(value.format),
+          string: value.value,
+        },
+      }
+    case 'set':
+      return { set: value.items.map(toLua) }
+    case 'map':
+    case 'map-pairs':
```
fatal10110: [medium] Tests cover only part of the RESP3 conversion this PR adds.

- **`map-pairs` → `{map=…}` has no test.** XREAD and XREADGROUP are the only commands that return `map-pairs`. Real Redis 7.2.1: `redis.setresp(3); redis.call('XREAD','STREAMS','st','0')` returns `{map={st={{"1-1",{"a","1"}}}}}`. The mock matches today, but nothing pins it, so a refactor that sent `map-pairs` down the pairs path would pass the suite.
- **The `setresp(2)` switch back and the per-eval reset have no tests.** `hostState.resp` is set in three places: reset before the eval, the `onSetResp` hook, and reset in `finally`. Real Redis resets per script and honours `setresp(3)` followed by `setresp(2)`. The mock does too (`setresp(3); setresp(2); return type(redis.call('HGETALL',KEYS[1]).map)` gives `"nil"` on both). A client-level test for each would pin this bookkeeping.
- **`verbatim`, `set`, `boolean` and `big-number` can't be reached by any command.** No command in `src/` builds these kinds, so those four branches never run through `redis.call`. On real Redis, INFO, MEMORY DOCTOR and LOLWUT reach Lua as `{verbatim_string=…}`, while the mock hands the script a plain string. That's acceptable as forward-looking code that mirrors `encodeResp3`, but the JSDoc above and the PR description present them as working behaviour. Either say they're unreachable until a command emits those kinds, or drop them until one does.

docs/TESTING.md
```diff
@@ -277,11 +277,17 @@ A Lua script can return two more of these kinds. A `{big_number=…}` table is a
...
-`{map=…}` tables without it; the mock's Lua engine currently returns `[]` for
-those unless the script calls `redis.setresp(3)` first — a gap in the mock, not
-Redis behaviour.)
+integer `1` and `false` to nil. After `redis.setresp(3)`, `redis.call` also
+hands the script RESP3 replies — `HGETALL` as a `{map=…}` table, `ZSCORE` as a
+`{double=…}` table — so returning one gives the client a map or a double.
+
+Known mock gaps here, both in the bundled Lua engine (`lua-redis-wasm`), not in
+Redis: the `setresp(3)` requirement is Redis's rule for booleans only — real
+Redis converts `{big_number=…}`, `{double=…}`, `{map=…}` and `{set=…}` tables
```
fatal10110: [low] This list of engine gaps is incomplete in two places.
- `{verbatim_string=…}` is missing. Real Redis 7.2.1 converts it without `setresp(3)`: `return {verbatim_string={format='txt',string='hi'}}` gives `"hi"` at RESP2 and a verbatim `hi` at RESP3. The mock's engine returns `(empty array)` for it, the same as the other four.
- The tests pin a third mock gap, `SET_REPLY_GAP` (SMEMBERS sends an array, not a RESP3 set, so Lua never sees `{set=…}` on the mock). That gap is in this repo, not the engine, and it isn't mentioned here. The sentence "both in the bundled Lua engine" therefore doesn't cover everything the tests mark `todo`.

src/core/lua-runtime.ts
```diff
@@ -311,7 +324,26 @@ export function renderScriptError(value: ReplyValue): ReplyValue {
...
 }
 
-function redisValueToLuaReply(value: RedisValue): ReplyValue {
+/**
+ * Convert a `redis.call`/`redis.pcall` reply into the value the script sees.
+ * Like real Redis, the shape follows the protocol the script selected with
+ * `redis.setresp()`: at RESP2 every reply is flattened to what a RESP2 client
+ * would read, while at RESP3 maps, doubles, sets, booleans, big numbers and
+ * verbatim strings reach Lua as their typed tables (`{map=…}`, `{double=…}`,
+ * …) — the same shapes `encodeResp3` writes to the wire.
```
fatal10110: [low] "the same shapes `encodeResp3` writes to the wire" isn't accurate for null. `encodeResp3` writes `_`, but Lua gets `false` because the engine decodes every null that way. `resp3TypedLuaReply`'s own JSDoc says so correctly. Suggest adding "(except null, see below)", or pointing to that note, so the two comments agree.

tests-integration/node-redis/scripts-typed-replies.test.ts
```diff
@@ -0,0 +1,191 @@
...
+const SET_REPLY_GAP = mockGap('mock SMEMBERS replies an array, not a set')
+
+describe(`Lua typed replies per protocol (node-redis, ${testRunner.getBackendName()})`, () => {
+  const tag = `{lua449:${RUN}}`
+  const hashKey = `${tag}:h`
+  const zsetKey = `${tag}:z`
+  const setKey = `${tag}:s`
+  const missingKey = `${tag}:missing`
+  let cluster: RedisClusterType
+  const clients: Record<2 | 3, RedisClientType> = {} as never
```
fatal10110: [low] `{} as never` turns off type checking for `clients`. If `before` fails partway, `clients[3]` is `undefined` while its type says `RedisClientType`. The `?.` in `after` is a sign the type is lying. Suggest `Partial<Record<2 | 3, RedisClientType>>` with a non-null read in `evalAt`, or building both clients with `Promise.all` and assigning once. `tests-integration/utils.ts` already has `connectToNodeRedisEndpoint`. It could take an optional `RESP` so this file doesn't hand-roll `createClient` and the `as unknown as` double cast.

fatal10110:
## Review summary: **NOT YET**

Reviewed head: `43a8776c0fa696bd60d17e73d9b6479d3dfb8ab3`. Two lenses: Redis wire-compat, with claims checked against a local real Redis 7.2.1, and code/architecture.

**What works.** After `redis.setresp(3)`, `redis.call` replies now reach Lua in the same shapes real Redis uses, and the RESP2 path is unchanged. Covered: HGETALL `{map}`, ZSCORE, ZINCRBY and ZADD INCR `{double}`, WITHSCORES as `[member,{double}]` pairs, ZPOPMIN (flat without a count, pairs with one), ZMPOP, ZRANK WITHSCORE, ZRANDMEMBER WITHSCORES, XREAD and XINFO `{map}`, and integers. The engine resets its own RESP version at every eval (`runtime.c` `redis_reset_resp_version()`). It calls `onSetResp` only after checking the argument (`redis_api.c` `l_redis_setresp`), so the host's `hostState.resp` stays in step with the engine.

**The engine-side claims check out.** Both can only be fixed in `lua-redis-wasm`:
- `runtime.c` `encode_table` only calls `encode_resp3_marker` when `redis_resp_version() == 3`. The table is already encoded as an empty array before the host sees it.
- `redis_api.c` `decode_reply` turns `REPLY_NULL` into `lua_pushboolean(L, 0)`. `ReplyValue` has no way to represent a Lua `nil`.

The PR is right not to close #449.

**Tests.** Red before green: with main's `lua-runtime.ts`, 8 of the new cases fail. On the PR, the mock run of the new files plus the `scripts` and `raw-tcp/script-resp3-replies` suites gives 59 pass, 0 fail, 11 todo. `tsc --noEmit` is clean. Unit tests pass 462/462. ESLint is clean on the changed files.

### Findings (inline)
| sev | where | summary |
|---|---|---|
| medium | docs/TESTING.md:290 | The null-gap text says clients get `false`, but a RESP2 client gets `0`. It also leaves out that a `false` element stays in arrays: HMGET/MGET return `[v, 0, w]` on the mock and `[v]` on Redis. |
| medium | node-redis/scripts-typed-replies.test.ts:144 | "SMEMBERS is a set" can't tell a set from an array, so it passes on the mock, which sends an array. |
| medium | src/core/lua-runtime.ts:422 | No tests for `map-pairs`→`{map}` (XREAD), for `setresp(2)` switching back, or for the per-eval reset. The `verbatim`/`set`/`boolean`/`big-number` branches can't be reached by any command. |
| low | docs/TESTING.md:286 | The gap list leaves out `{verbatim_string=…}` and the repo-side `SET_REPLY_GAP`. |
| low | src/core/lua-runtime.ts:333 | The JSDoc says the shapes match `encodeResp3` "on the wire", which isn't true for null. |
| low | node-redis/scripts-typed-replies.test.ts:42 | `{} as never` hides that `clients` can be partly filled. The file also hand-rolls `createClient` when a helper exists. |

### Out-of-scope follow-ups (not blocking)
- The mock marks HRANDFIELD as `noscript` (`src/commands/hashes.ts`). Real Redis allows it in scripts.
- LCS isn't implemented, so it's unknown from scripts.
- Several RESP3 reply kinds are missing from the mock, both on the wire and in Lua:
  - SMEMBERS, SINTER, SUNION and SDIFF should be sets.
  - COMMAND INFO flags should be a set.
  - INFO, MEMORY DOCTOR and LOLWUT should be verbatim strings.
  - MEMORY STATS should be a map.
- For |score| ≥ 1e17, `formatRedisDouble` gives `100000000000000000000` where Redis gives `1e+20`. This predates the PR.

docs/TESTING.md
fatal10110: Fixed in be3ed8e. TESTING.md now says the value arrives as `false` to a RESP3 client and `0` to a RESP2 client, and that it does not end an array the way a Lua `nil` does. It includes your HMGET example (`["v"]` on real Redis, `["v", 0, "v"]` on the mock). Both test files now have a `todo: ENGINE_NULL_GAP` test named "a missing value ends an array reply" that runs `HMGET KEYS[1] f nope f`. Both pass against real Redis 8.0.

tests-integration/node-redis/scripts-typed-replies.test.ts
fatal10110: Fixed in be3ed8e. The RESP3 check now reads the reply with `withTypeMapping({ [RESP_TYPES.SET]: Set })` and asserts `new Set(["x"])`, so it can tell `~` from `*`. It is marked `todo: SET_REPLY_GAP` on the mock and passes on real Redis. The Lua-side `{set=…}` check stays a separate `todo` test.

src/core/lua-runtime.ts
fatal10110: Addressed in be3ed8e.
- map-pairs: both files now have an XREAD test covering the client view at RESP2 and RESP3, plus the Lua view `r.map[KEYS[5]][1][1] == "1-1"`.
- Protocol bookkeeping: the new tests are "redis.setresp(2) switches redis.call back to RESP2 shapes" and "every EVAL starts at RESP2 again". The second one also checks that the protocol resets after a script that errors following `setresp(3)`.
- Unreachable kinds: the JSDoc on `resp3TypedLuaReply` now says only double, map, map-pairs and flat-pairs are reachable through `redis.call` today, and that the set, boolean, big-number and verbatim branches mirror `encodeResp3` for when a command emits them. The PR description says the same. `redisValueToLuaReply` is now exported for tests, and `tests/lua-reply-conversion.test.ts` covers those branches at both protocols and checks that conversion recurses.

docs/TESTING.md
fatal10110: Fixed in be3ed8e. TESTING.md now has a three-item gap list. It adds `{verbatim_string=…}` to the engine typed-table gap, and adds the SMEMBERS set-reply gap, which it notes is in this repo rather than the engine. Both test files also have a `todo` test for `{verbatim_string=…}` without `setresp(3)`, and it passes on real Redis.

src/core/lua-runtime.ts
fatal10110: Fixed in be3ed8e. The JSDoc now says "the shapes `encodeResp3` writes to the wire, except null (see resp3TypedLuaReply)".

tests-integration/node-redis/scripts-typed-replies.test.ts
fatal10110: Fixed in be3ed8e. The `{} as never` cast and the `as unknown as` double cast are both gone. Clients come from a local `connectAt(port, RESP)`, whose return type is inferred. Both are connected with `Promise.all` and assigned once to a `Record<2 | 3, Client> | undefined`, and `client()` asserts they are set before use. I did not change `connectToNodeRedisEndpoint` in `tests-integration/utils.ts`: a parallel PR (#460/#453) is reworking the integration helpers, and I wanted to avoid conflicting with it. The suite also now runs on a standalone server (`setupRawStandalone`), since Lua semantics do not depend on cluster mode.

fatal10110:
## Re-review: **SHIP**

Reviewed head: `be3ed8e0513ebdfdcf692a15eeb897654880eb64` (diffed against `43a8776`).

### Prior findings
| sev | finding | status |
|---|---|---|
| medium | TESTING.md null gap: RESP2 gives `0`, and a null doesn't end arrays | **fixed.** The text is corrected, and HMGET `todo` tests were added in both suites. On the mock they fail with `['v', 0, 'v']` as the docs describe. |
| medium | "SMEMBERS is a set" couldn't tell a set from an array | **fixed.** The test now uses `withTypeMapping({[RESP_TYPES.SET]: Set})` and is `todo: SET_REPLY_GAP`. It fails on the mock and passes on real Redis. |
| medium | Missing tests for map-pairs, setresp(2) and the per-eval reset; unreachable branches | **fixed.** XREAD `{map}`, `setresp(3)`→`setresp(2)`, and a reset-per-EVAL test (it would fail if `hostState.resp` leaked, because the engine decodes `REPLY_MAP` at any version) were added. The set, boolean, big-number and verbatim branches are now covered by `tests/lua-reply-conversion.test.ts` and documented as unreachable today. |
| low | Gap list was missing verbatim and SET_REPLY_GAP | **fixed** |
| low | JSDoc "on the wire" claim didn't hold for null | **fixed** |
| low | `{} as never` / hand-rolled client casts | **fixed.** Clients are now `Record | undefined` plus an assert, created with `Promise.all`, with no casts. |

### New code
- Exporting `redisValueToLuaReply` doesn't widen the public API. `src/internal.ts` (the `./core` entry) re-exports a named list from `lua-runtime` that doesn't include it.
- Moving the tests to a standalone server is fine, since Lua semantics don't depend on cluster mode. Keys are per-run random and are deleted in `after`.
- No new findings.

### Tests
- Mock: the new unit test plus the ioredis/node-redis `scripts` and `scripts-typed-replies` suites and `raw-tcp/script-resp3-replies` give 82 tests: 66 pass, 0 fail, 16 todo. All 16 todo cases really fail on the mock, so none are stale.
- Real Redis 8.0.6 (my own container): both `scripts-typed-replies` files pass 34/34, including every case that is `todo` on the mock.
- `tsc --noEmit` and ESLint on the changed files are clean.

#449 still needs the `lua-redis-wasm` fixes (typed tables without `setresp(3)`, and RESP3 null decoding as nil), as the PR says.

Issue #390, opened 2026-09-21
defineCommand silently strips prototype and identity; decide whether it should accept non-literal definitions


fatal10110:
Surfaced while reviewing PR #377. **Pre-existing on `main`**, not introduced there. Filed as an API-shape question rather than a straight bug report — the fix is a design call.

## Problem
`defineCommand` (`src/core/command-definition.ts`) returns `{ ...definition, name: definition.name.toLowerCase() }`. That spread:

- **strips the prototype** — a class-based `CommandDefinition` loses its `keys`/`execute` methods, silently at definition time and fatally at first invocation;
- **returns a different object** — so a caller holding the argument no longer has the registered definition, breaking `WeakMap` keying and any identity comparison.

`CommandDefinition` is an `interface`, so a class instance type-checks fine. And `docs/API.md:177-187` advertises `defineCommand` as *the* extension point for the published `js-redis-server/core` subpath. So the documented public API has a footgun that compiles cleanly and crashes on first use.

Note `CommandRegistry.register()` does **not** have this problem — it stores by reference. Registering a class instance directly works today, and PR #377 added a regression test pinning exactly that.

## What this issue should decide
Not which patch to apply, but whether `defineCommand` should accept non-literal inputs at all. Two directions, and they are not equivalent:

1. **Preserve the prototype** — e.g. `Object.assign(Object.create(Object.getPrototypeOf(d)), d, { name })`. Fixes the crash but **not** the identity break: the result is still a copy, so `WeakMap` keying still misses. Half a fix, and arguably worse for being less obviously broken.
2. **Narrow the parameter type** so anything other than an object literal fails to compile. Closes the hole at the right layer and costs nothing at runtime.

Direction 2 looks better, but it is a public-API decision and deserves a deliberate call.

## Coupling to watch
`register()` deliberately does **not** normalize the name it carries — a directly-registered mixed-case definition keeps its casing, which then reaches clients through `COMMAND INFO` and arity errors (real Redis lowercases). That split is currently intentional: normalizing inside `register()` is exactly what caused the prototype-stripping regression found in PR #377. If `defineCommand` ever became the enforced entry point, that decision needs revisiting at the same time.

## Acceptance
A decision recorded in this thread, then: either the parameter type rejects non-literals at compile time, or `defineCommand` preserves both prototype and identity — not a partial fix — with a test covering a class-based definition either way.

PR #469, opened 2026-09-23
fix: fold command and subcommand names ASCII-only (#382)

### LINKED ISSUE #382: Command-name folding uses Unicode toLowerCase; Redis folds ASCII only (mock executes KEYS * for Kelvin-sign spelling)
Found while reviewing PR #377, verified against real redis-server 7.2.1. **Pre-existing on `main`** (the PR touches the surrounding code but does not introduce this).

## Problem
Command-name normalization uses JavaScript's `toLowerCase()`, which is **Unicode-aware**. Redis folds **ASCII only** (`tolower()` per byte).

So characters that Unicode folds onto ASCII letters resolve to a real command in the mock, where Redis rejects them:

```
Wire name: "Ks *"   (U+212A KELVIN SIGN + "s *")
REAL 7.2.1: -ERR unknown command ...
MOCK:       executes KEYS *
```

**Failure scenario:** a test suite asserting that a malformed/hostile command name is rejected passes against real Redis and silently *executes* against the mock. The Kelvin sign is the classic case; U+017F LATIN SMALL LETTER LONG S (ſ → s) is another.

## Proposal
Replace `toLowerCase()` with an explicit ASCII fold at the registry/executor boundary — map only `A-Z` (0x41–0x5A) to `a-z`, leave every other byte untouched. One small helper, used everywhere a wire command name is normalized.

## Acceptance
- `Ks *` and `ſet k v` are rejected as unknown commands, matching real Redis byte-for-byte.
- Existing case-insensitivity for ASCII (`GET`/`get`/`Get`/`eVaL`) is unaffected — `tests/command-registry.test.ts` covers this.

fatal10110:
## Summary
- Command and subcommand names were matched with JS `toLowerCase()`/`toUpperCase()`, which fold Unicode. U+212A KELVIN SIGN lowercases to `k` and U+017F LONG S uppercases to `S`, so the mock ran KEYS for `Keys *`, SET for `ſet k v`, and XINFO STREAM for `XINFO ſTREAM`. Real Redis rejects all of these because it folds ASCII only (`tolower`/`strcasecmp`).
- Added `src/core/ascii-case.ts` with `asciiLowerCase`/`asciiUpperCase`, which map `A-Z` <-> `a-z` and nothing else.
- The new helpers now handle every wire-name normalization point:
  - `CommandRegistry.register`/`get`. The executor's `plan()` goes through `get`, and so do Lua `redis.call`/`pcall`, `ACL DRYRUN` and `COMMAND GETKEYS`.
  - The `defineCommand` name fold. Its spread/copy behaviour is unchanged (#390 left alone).
  - The EXEC-arity check in `CommandExecutor`.
  - Subcommand dispatch in COMMAND, CONFIG, CLIENT, ACL, SLOWLOG, SCRIPT, FUNCTION, PUBSUB, CLUSTER, XINFO and XGROUP.
  - `COMMAND INFO`/`DOCS` target lookup, and the `equalsAscii` helpers in `command.ts`/`connection.ts`.
- Not changed: the unknown-command echo. Real Redis echoes the name's raw bytes (`'Keys'`), while the mock hex-escapes non-printable names (`'0xe284aa657973'`). That's #384 part 2, so the new test checks only the `-ERR unknown command '` prefix.
- Not changed: option/keyword parsing (`t.keyword`, per-command `toUpperCase()` option matching), which still folds Unicode. For example, `ſ` uppercases to `S`. These are option tokens, not command or subcommand names, so they're outside #382.

Closes #382

## Test plan
- [x] New `tests-integration/raw-tcp/command-name-folding.test.ts` (exact wire bytes) failed on mock before the fix (4/6 red) and passes against real Redis 8.0.6 (6/6).
- [x] It passes on mock after the fix. ASCII case-insensitivity (`sEt`, `GeT`, `CoMmAnD gEtKeYs`, `XiNfO StReAm`) is still covered.
- [x] `tests/command-registry.test.ts` has a new non-ASCII case, and its existing ASCII case-insensitivity tests still pass.
- [x] `npm run build`, `npm run lint`, `npm test` (463 pass)
- [x] Full mock integration suite (1254 pass)
- [x] Real backend, run directly without `clean:redis`: all raw-tcp tests (155 pass), plus the ioredis command/config/connection/pubsub/scripts/multi/stream suites (102 pass)


tests/command-registry.test.ts
```diff
@@ -35,9 +35,24 @@ describe('CommandRegistry', () => {
...
+    const keys = makeCommand('keys')
+    const set = makeCommand('set')
+    registry.registerAll([keys, set])
+
+    assert.strictEqual(registry.get('KeYs'), keys)
+    // U+212A KELVIN SIGN lowercases to 'k' under Unicode folding.
+    assert.strictEqual(registry.get('\u212Aeys'), undefined)
+    assert.strictEqual(registry.get('\u212AEYS'), undefined)
+    // U+017F LATIN SMALL LETTER LONG S uppercases to 'S'.
+    assert.strictEqual(registry.get('\u017Fet'), undefined)
```
fatal10110: [medium] The long-s (U+017F) cases on the command-name path don't test this fix, and the PR description overstates the bug.

`registry.get` lowercases, and `'ſ'.toLowerCase() === 'ſ'` (ſ is already lowercase). So `registry.get('ſet')` returned `undefined` on `origin/main` as well. The comment above it ("uppercases to 'S'") describes a path `get` never takes. The raw-tcp test `a long-s spelling of SET is an unknown command and writes nothing` (command-name-folding.test.ts:78) has the same problem.

Evidence: I copied both PR test files onto an `origin/main` tree and ran them on mock.
```
not ok 1 - a Kelvin-sign spelling of KEYS is an unknown command
ok 2 - a long-s spelling of SET is an unknown command and writes nothing   <- green before the fix
not ok 3 - COMMAND INFO / DOCS / GETKEYS do not resolve a non-ASCII fold
not ok 4 - subcommand dispatch does not resolve a non-ASCII fold
not ok 5 - redis.pcall from Lua does not resolve a non-ASCII fold
ok 6 - ASCII case-insensitivity is unaffected
```
That matches the "4/6 red" in the test plan. But it contradicts the Summary's claim that the mock ran "SET for `ſet k v`". It never did. The only character JS `toLowerCase()` folds onto an ASCII letter is U+212A (→ `k`). U+017F (→ `S`) and U+0131 dotless ı (→ `I`) only matter on the **uppercasing** dispatch paths (XINFO/XGROUP), and those are covered correctly.

Suggested fix:
- Correct the PR description.
- Here, either drop the `ſet` assertion or label it an explicit regression guard ("never resolved: ſ is already lowercase; kept because #382's acceptance lists it").
- Do the same for the raw-tcp `ſet` test's name/comment. Keeping it as a wire regression guard is fine, since real 7.2.1 and 8.0.6 both reject it (the test passes against both with `TEST_BACKEND=real`). It just shouldn't read as proof of the fix.

tests-integration/raw-tcp/command-name-folding.test.ts
```diff
@@ -0,0 +1,156 @@
...
+ * ```
+ *
+ * The *unknown command* replies are asserted by prefix only: real Redis echoes
+ * the name's raw bytes, while the mock hex-escapes non-printable names
+ * (`'0xe284aa...'`). That rendering is a separate fidelity gap, out of scope
+ * here.
+ */
+const testRunner = new TestRunner()
+
+const KELVIN = 'K'
```
fatal10110: [low] In an editor, diff view or grep result, `'K'` on this line looks exactly like ASCII `K`, and `'ſ'` is easy to misread. A formatter, a copy-paste or an NFKC-normalizing tool could also quietly turn them into `K`/`s`. The test would then *send real `KEYS *` to the shared real backend*, and still pass on the mock after this fix, because it would just run KEYS.

Use escapes, as `tests/command-registry.test.ts` already does:
```ts
const KELVIN = 'K'
const LONG_S = 'ſ'
```

src/core/ascii-case.ts
```diff
@@ -0,0 +1,22 @@
...
+ * rejects (#382). Every other character passes through unchanged.
+ */
+export function asciiLowerCase(value: string): string {
+  return value.replace(/[A-Z]/g, char =>
+    String.fromCharCode(char.charCodeAt(0) + 0x20),
+  )
+}
+
+/** Upper-case counterpart of {@link asciiLowerCase}: folds `a-z` only. */
+export function asciiUpperCase(value: string): string {
```
fatal10110: [low] Nothing unit-tests `asciiLowerCase`/`asciiUpperCase` directly. They're only exercised indirectly, through the registry and the raw-tcp dispatch tests. The doc comment also leaves out the third hazard for `toUpperCase()`: U+0131 LATIN SMALL LETTER DOTLESS I uppercases to `I`, which matters for `XGROUP SETID`.

I verified that this PR already handles it correctly:
```
real 8.0.6:  XGROUP SETıD k g 0  -> -ERR unknown subcommand 'SETıD'. Try XGROUP HELP.
mock @7667504: identical bytes
```
Suggest a small `tests/ascii-case.test.ts`:
- `asciiLowerCase('K')`, `asciiUpperCase('ſ')` and `asciiUpperCase('ı')` each come back unchanged.
- `A-Z`/`a-z` round-trip.
- `@`/`[`/`` ` ``/`{` (the bytes next to the letter ranges) are left alone.

Also mention ı in this doc comment.

src/commands/command.ts
```diff
@@ -592,7 +593,7 @@ function expectArgCount(
 }
 
 function equalsAscii(value: Buffer, expected: string): boolean {
-  return value.toString().toLowerCase() === expected
+  return asciiLowerCase(value.toString()) === expected
```
fatal10110: [low] This `equalsAscii` duplicates the one in `src/commands/connection.ts:501` byte for byte. The name was inaccurate before this PR, since it folded Unicode; now it's right. Consider exporting a single `asciiEqualsIgnoreCase(value: Buffer, expected: string)` from `src/core/ascii-case.ts` and using it in both files, so the next folding change touches one place. Non-blocking.

fatal10110:
## Review of 7667504: not ready to merge yet

**Verdict: NOT YET.** One medium finding is still open. It's about docs and test accuracy; the fix itself is correct.

| Severity | Location | Finding |
|---|---|---|
| medium | tests/command-registry.test.ts:49 (+ raw-tcp test:78) | The `ſet` cases never failed before this fix, because `'ſ'.toLowerCase()` is `'ſ'`. The PR description's claim that the mock ran "SET for `ſet k v`" is false. Correct the description and label those cases as regression guards. |
| low | tests-integration/raw-tcp/command-name-folding.test.ts:36 | Use `'K'`/`'ſ'` escapes instead of raw characters that look like ASCII. |
| low | src/core/ascii-case.ts:18 | No direct unit test for the helpers. The doc comment should mention that U+0131 ı uppercases to `I`. |
| low | src/commands/command.ts:596 | `equalsAscii` duplicates the one in connection.ts. Move it into ascii-case.ts. |

**Checked against real Redis (7.2.1 local, 8.0.6 docker):**
- The new raw-tcp test passes 6/6 against both versions with `TEST_BACKEND=real`.
- I ran the new tests on an `origin/main` tree: the raw-tcp file had 4/6 failing and the registry test had 2 new cases failing (the Kelvin cases). They are green at the PR head.
- RESP3 `COMMAND INFO hKeys` returns `*1 _` and `COMMAND DOCS hKeys` returns `%0` on both real and mock.
- A Kelvin-spelled command inside MULTI gets the same EXECABORT on both. The only difference is how the unknown name is echoed, which is #384.
- `XGROUP SETıD`, `CLIENT KILL`, `CONFIG REſETSTAT` and `ACL DRYRUN default hKeys` give byte-identical replies on real 8.0.6 and on mock.

**Completeness:** every place that dispatches on a command or subcommand name now folds ASCII only. The `toLowerCase`/`toUpperCase` calls that remain handle option or mode tokens (out of scope, known follow-up), error text built from definition names that are already ASCII, or the SHA in the script cache.

**Tests at the PR head:** `tsc --noEmit` clean; unit tests 463/463; mock raw-tcp + ioredis integration 674/674.

**Out of scope, for follow-up:**
- #384: the unknown-command echo is hex-escaped (`'0xe284aa657973'` vs real `'Keys'`).
- Option and keyword tokens still fold Unicode (`keywordMatches` in command-schema.ts, per-command `toUpperCase()` on options).
- Unrelated to #382: `SCRIPT HELP` replies with bulk strings (`*13`, `$63…`), while real 8.0.6 replies with simple strings (`*17`, `+…`).

tests/command-registry.test.ts
fatal10110: Fixed in 6064c19. Agreed: U+017F lowercases to itself, so the long-s command-name cases never failed. Both the unit test and the raw-tcp test now call them regression guards, against the lookup ever growing an upper-casing fold. The PR description is corrected too: the real long-s exposure was the upper-casing XINFO/XGROUP dispatch, and the SET claim is gone.

tests-integration/raw-tcp/command-name-folding.test.ts
fatal10110: Fixed in 6064c19. KELVIN and LONG_S are now built with String.fromCodePoint(0x212a) and String.fromCodePoint(0x017f), with a comment explaining why. The header doc now writes them as <U+212A>/<U+017F>. The file contains no raw look-alike bytes any more, so an editor or formatter can't turn one into a real KEYS on the shared backend.

src/core/ascii-case.ts
fatal10110: Fixed in 6064c19. Added tests/core-ascii-case.test.ts, which tests asciiLowerCase, asciiUpperCase and equalsAscii directly. It covers U+212A, U+017F, U+0131 (dotless i), U+0130, accented letters and the ASCII boundary characters around A-Z/a-z. It also pins the premise that JS really does fold those look-alikes onto ASCII. The doc comment now lists U+0131.

src/commands/command.ts
fatal10110: Fixed in 6064c19. There's now one exported equalsAscii in src/core/ascii-case.ts. The two local copies in command.ts and connection.ts are gone, and both files import it.

fatal10110:
## Re-review of 6064c19 (since 7667504): ready to merge

**Verdict: SHIP.** All four earlier findings are fixed and I found no new issues.

| Prior finding | Status |
|---|---|
| [medium] The `ſet` cases never failed before the fix, and the PR description claimed they did | **Fixed.** The registry test, the raw-tcp test and the file header now call the long-s cases regression guards. The PR description now says `ſet` was never executed as SET, and splits the lower-casing and upper-casing paths. |
| [low] Characters that look like ASCII were written as raw literals | **Fixed.** They are now built with `String.fromCodePoint(0x212a / 0x017f)`, and the header uses `<U+212A>` placeholders. |
| [low] No direct test for the helpers, and ı was undocumented | **Fixed.** New `tests/core-ascii-case.test.ts` covers K, ſ, ı and İ, the characters just outside the letter ranges (`@[`{`), accented letters, the empty string, and `equalsAscii`, and pins the Unicode behaviour the tests depend on. The doc comment lists all three characters. |
| [low] `equalsAscii` was duplicated | **Fixed.** There is one exported `equalsAscii` in `src/core/ascii-case.ts`, and both call sites import it. The behaviour is identical. |

**Tests at 6064c19:**
- `tsc --noEmit` clean; eslint and prettier clean on the changed files.
- Unit tests: 467/467 (the 4 new ones included).
- Mock raw-tcp + ioredis integration: 674/674.
- `command-name-folding.test.ts` against a real redis-server (7.2.1): 6/6.

The out-of-scope follow-ups from the first review still apply: #384 (echoing the unknown-command name), Unicode folding of option and keyword tokens, and the `SCRIPT HELP` reply encoding.

PR #470, opened 2026-09-23
fix: gate multi-pair CONFIG SET on 7.0 and reject duplicate parameters (#419)

### LINKED ISSUE #419: CONFIG SET: gate multi-pair arity on the 7.0 boundary, then add duplicate-parameter detection

fatal10110:
## Summary

Implements #419 in the order the issue requires: first gate `CONFIG SET`'s multi-pair form on the 7.0 boundary, then add duplicate-parameter detection on 7.0+ profiles only.

- **New gate `config.set.multi-pair`** (`redis: 7.0.0`, `valkey: 7.2.0`). This is a sibling of `config.set.failure-message`, not a reuse of it. The boundary is the same 7.0 rewrite, but this gate controls which argument shapes are accepted, not the error wording.
- **`redis-6.2` profile:** any `CONFIG SET` that is not exactly one pair returns the legacy subcommand syntax error, which echoes the subcommand as sent (`'SET'`, `'set'`, `'SeT'`). This covers zero pairs, a dangling name, several pairs and repeated pairs, and matches real 6.2's `c->argc == 4` dispatch.
- **7.0+ profiles:**
  - Zero pairs gives `wrong number of arguments for 'config|set' command`.
  - A dangling name after the first pair gives `ERR syntax error`. Before this change it was a `config|set` arity error, which was wrong on 7.0+ too.
  - A repeated parameter gives `CONFIG SET failed (possibly related to argument '<name>') - duplicate parameter`. Names match case-insensitively, and the error echoes the repeat exactly as typed (`'TIMEOUT'`).
- **Name checks run before value checks,** as in real `configSetCommand`. A new pass checks every name before any value is parsed, and the first unknown or duplicate name in argument order is the error. So a bad value no longer hides a later unknown or duplicate name: `CONFIG SET proto-max-bulk-len bad nosuch 1` now reports `nosuch`, not the memory-value error.

Ground truth came from dedicated containers: redis 6.2.24, 7.0.15, 8.0.6 and Valkey 7.2.14. All three 7.0+ servers reply identically on every case.

### Expected conflict with #464

#464 (#371 + #416) rewrites the same `configSet` loop. It routes the unknown-parameter throw through `configSetUnknownParameter`, which has 6.2 wording, and changes `configSetFailed` to take the typed name and lowercase it on 7.0+. It also edits `profile-gates.test.ts` and `command-errors-config.test.ts`. Whichever PR lands second resolves the conflict like this:

- Change the unknown-name throw in the new `checkConfigSetNames` pre-pass to `configSetUnknownParameter(profile, raw)`. On 6.2 the pre-pass only ever sees exactly one pair, so its wording is right there too. Remove #464's `!store.has(name)` check from the value loop; the pre-pass already covers it.
- Keep `configSetDuplicateParameter` separate from `configSetFailed`. Real Redis echoes the name as typed (`argv`) for a duplicate but the canonical name for value failures. #464's lowercasing would turn `'TIMEOUT'` into `'timeout'`, and the raw-TCP test pins that difference.

Closes #419

## Test plan

- [x] TDD: the new raw-TCP tests in `tests-integration/raw-tcp/command-errors-config.test.ts` failed on mock before the fix (5 of 7 new tests), and the new profile-gate test failed on the `redis-6.2` mock profile.
- [x] Raw-TCP suite green against real 8.0.6 (shared `6399`), real 7.0.15 and Valkey 7.2.14 (`REDIS_STANDALONE_PORT`). Every `CONFIG SET` it sends either fails or sets a parameter to its current value, and `proto-max-bulk-len` is restored in `finally`.
- [x] Byte-for-byte differential over 17 cases: mock `redis-6.2` vs real 6.2.24 (0 diffs), and mock `redis-8.0` vs real 8.0.6 (0 diffs).
- [x] `npm run test:integration:compatibility:mock`: all 7 profiles green.
- [x] `npm run build`, `npm run lint`, `npm test` (462 pass), `npm run test:integration:mock` (1256 pass).
- [x] Config suites (raw-tcp config + proto-max-bulk-len, ioredis + node-redis config-integration) green against real 8.0. Run directly with node, not through `test:integration:real`, so shared servers were not flushed.
- [x] Docs: `docs/API.md` gate-table row and `docs/COMMANDS.md` CONFIG SET note.


src/commands/config.ts
```diff
@@ -221,15 +227,76 @@ type ConfigUpdate =
...
+  subcommand: Buffer,
+  args: readonly Buffer[],
+  ctx: RedisExecutionContext,
+): RedisResult {
   const store = getConfigStore(ctx)
+  checkConfigSetNames(subcommand, args, ctx, store)
+
   const updates: ConfigUpdate[] = []
   for (let i = 0; i < args.length; i += 2) {
     const name = args[i].toString().toLowerCase()
```
fatal10110: [low] The value loop now depends on a check that lives in another function. It decodes and lowercases every name a second time (`args[i].toString().toLowerCase()`, already done at L274-275), and the `!store.has(name)` guard is gone from here. So `store.set(name, value)` (L337) only avoids creating new keys because `checkConfigSetNames` ran first, and nothing in the types says so. If that call is dropped or its normalization drifts (the #464 rebase is the obvious first chance), unknown names would silently land in the per-server config map and show up in `CONFIG GET`. Suggestion: have the pre-pass return the resolved `{ name, value }[]` (or the lowercased names) and loop over that, so names are decoded once and the dependency shows in the types. (architecture lens)

src/commands/config.ts
```diff
@@ -221,15 +227,76 @@ type ConfigUpdate =
...
+/**
+ * CONFIG SET's shape and name checks, run before any value is looked at.
+ *
+ * Redis 6.2 dispatches SET only for exactly one pair; every other shape falls
+ * through to the legacy subcommand syntax error. 7.0+ (`config.set.multi-pair`)
+ * takes any number of pairs, and — like its `configSetCommand` — resolves every
+ * name in one pass before validating a value: the first unknown or repeated
+ * name in argument order is the error, so a bad value never masks either.
+ */
+function checkConfigSetNames(
```
fatal10110: [low] `checkConfigSetNames` also does the profile-gated shape checks: the 6.2 legacy syntax error, the `config|set` arity error, and `syntax error` (L261-270). Someone looking for where the CONFIG SET arity errors come from won't find them under this name. It also takes the whole `ctx` but only reads `ctx.server.profile`. Suggestion: rename it (e.g. `parseConfigSetPairs` / `resolveConfigSetNames`) or split it into a shape check and name resolution, and pass `profile` instead of `ctx`. (architecture lens)

src/commands/config.ts
```diff
@@ -221,15 +227,76 @@ type ConfigUpdate =
...
+ * CONFIG SET's repeated-parameter failure. The `CONFIG SET failed` wrapper of
+ * {@link configSetFailed}, but only ever reachable on 7.0+ profiles (6.2 never
+ * accepts a second pair), and it echoes the repeat exactly as the client sent
+ * it — real Redis reports `argv` here, not the canonical name its value
+ * failures report. Captured from 7.0.15, 8.0.6 and Valkey 7.2.14:
+ *
+ *   CONFIG SET timeout 0 TIMEOUT 0
+ *     -> ERR CONFIG SET failed (possibly related to argument 'TIMEOUT') - duplicate parameter
+ */
+function configSetDuplicateParameter(name: string): RedisCommandError {
```
fatal10110: [low] The JSDoc calls this "the `CONFIG SET failed` wrapper of configSetFailed", but it copies the template instead of sharing it. `CONFIG SET failed (possibly related to argument '<name>') - <detail>` is now built in three places: `configSetFailed` (L52-66), here, and `src/state/keyspace-notifier.ts`. A shared message builder that both paths call would stop the wording drifting. The duplicate path can skip the profile check, since it is only reachable with `config.set.multi-pair` on. Keeping the argv echo separate from the canonical-name echo, as the PR description explains, is correct: real 7.0.15, 8.0.6 and Valkey 7.2.14 all answer `CONFIG SET timeout 0 TIMEOUT 0` with `ERR CONFIG SET failed (possibly related to argument 'TIMEOUT') - duplicate parameter`. (architecture lens)

src/core/compatibility/feature-gates.ts
```diff
@@ -12,6 +12,13 @@ export const FEATURE_GATES: Record<FeatureId, VersionGate> = {
   // `Invalid argument '<value>' for CONFIG SET '<name>' - <detail>` to
   // `CONFIG SET failed (possibly related to argument '<name>') - <detail>`.
   'config.set.failure-message': { redis: '7.0.0', valkey: '7.2.0' },
+  // The same 7.0 rewrite let CONFIG SET take several parameter/value pairs.
+  // 6.2 dispatches SET only for exactly one pair and answers every other shape
+  // with the legacy subcommand syntax error; 7.0+ splits that into a
+  // `config|set` arity error (no pair) and `syntax error` (dangling name), and
```
fatal10110: [low] "arity error (no pair)" is incomplete: a lone name with no value also gets the arity error, not `syntax error`. Real 7.0.15, 8.0.6 and Valkey 7.2.14 answer `CONFIG SET timeout` with `ERR wrong number of arguments for 'config|set' command`, and the code (`args.length < 2`) and the `nameOnly` test agree. Suggest "(fewer than one complete pair)". (architecture lens, checked against real servers)

fatal10110:
## Review summary: **SHIP**

Reviewed head `45fb0b7dd5cfa1aa54067da2a9bb3169f59b6b87` against `origin/main`. There are no high or medium findings and 4 low nits, all posted inline.

### Checked against real servers (6.2.24, 7.0.15, 8.0.6, Valkey 7.2.14)
I compared the mock and the real servers on 17 `CONFIG SET` shapes for both profiles: mock `redis-8.0` vs real 8.0.6, and mock `redis-6.2` vs real 6.2.24. **Every direct reply matched exactly.** That covers:
- zero pairs, a name with no value, a dangling name, and several pairs
- a repeat in different case, with the name echoed as typed
- which error wins when unknown and repeated names both appear, and in which order
- a bad value followed by an unknown or repeated name
- `notify-keyspace-events` repeats
- the mixed-case subcommand echo on 6.2

The RESP3 replies are identical too. Real 7.0, 8.0 and Valkey 7.2 give the same answer on every shape, so the 7.0 gate boundary is right. The `valkey-8.0+` docs row matches the gate: there is no `valkey-7.2` profile, and other 7.2.0 gates are documented the same way.

### Tests
- Red before green: the PR's tests fail on origin/main's `src` (5 raw-tcp failures, plus 1 profile-gate failure under `redis-6.2`).
- Green on the PR head:
  - raw-tcp CONFIG suite on mock, and on real 7.0.15, 8.0.6 and Valkey 7.2.14 (my own disposable containers)
  - `profile-gates` on redis-6.2, redis-7.0 and valkey-8.0
  - config and raw-tcp integration on mock: 182 pass
  - `npm test`: 462 pass
  - `tsc --noEmit` and eslint: clean

### Inline nits (low)
1. `config.ts:302`: the value loop only avoids adding unknown names to the store because the pre-pass ran first. Return the resolved names from the pre-pass instead.
2. `config.ts:255`: `checkConfigSetNames` also does the shape and arity checks. Rename it or split it, and pass `profile` instead of `ctx`.
3. `config.ts:240`: the `CONFIG SET failed (...)` message is now built in three places. Share one message builder.
4. `feature-gates.ts:18`: "(no pair)" should say fewer than one complete pair, because a name with no value also gets the arity error.

### Out-of-scope follow-ups (not caused by this PR)
- **MULTI arity timing:** `MULTI` / `CONFIG SET timeout` / `EXEC`. Real 8.0 rejects it when the command is queued (`ERR wrong number of arguments for 'config|set' command`, then `EXECABORT`). The mock replies `QUEUED` and returns the error from `EXEC`. The same thing happened before this PR, because `config|set` subcommand arity is only enforced inside `execute`.
- **Wording gate:** `keyspace-notifier.ts` hard-codes the 7.0+ `CONFIG SET failed` wording, so the `redis-6.2` profile gets the wrong message there.
- **Aliases:** `CONFIG_DEFAULTS` has no aliases such as `hash-max-ziplist-entries`. Real 7.0, 8.0 and Valkey 7.2 accept `CONFIG SET hash-max-ziplist-entries 128 hash-max-listpack-entries 128` with `OK`: an alias plus its canonical name is *not* flagged as a duplicate.
- **Immutable parameters:** real 7.0+ rejects them with `can't set immutable config` (e.g. `daemonize`). The mock doesn't model this.
- **Case matching:** names are lowercased with JS `toLowerCase()`, which folds some Unicode characters to ASCII letters. Redis only folds ASCII (`strcasecmp`).

PR #471, opened 2026-09-23
fix(demo): typecheck browser demo, drop elliptic from Pages bundle (#396, #393)

### LINKED ISSUE #393: Browser demo ships elliptic with known advisories in the Pages bundle; 'dev dependency' framing is incorrect for a bundler plugin
Found during a post-merge review of PR #391. Not urgent, but the reasoning that dismissed it was wrong, so recording it properly.

## Problem
`examples/browser-demo` carries 9 `npm audit` advisories (`elliptic`, `qs`, `nanoid`, `postcss`). These are conventionally waved through as "dev dependencies" — but that framing does not hold for a **bundler plugin**, which inlines its dependency graph into the shipped artifact.

`elliptic` is verifiably present in the deployed Pages bundle: 9 occurrences plus real curve tables (`secp256k1`, `p384`, `curve25519`). It arrives through `vite-plugin-node-polyfills` → `node-stdlib-browser` → `crypto-browserify`.

## Why it is not urgent
The demo calls `createHash` only — for Lua script SHA1 — and never exercises ECDSA verification, which is where the `elliptic` advisories live. So the vulnerable code paths ship but are not reachable through the demo's UI.

## Why it is still worth fixing
- "Dev dependency, not shipped" is the justification that will be reached for next time, and for a bundler plugin it is simply false. Worth correcting in the repo's own understanding.
- The bundle is served from GitHub Pages under the project's name. Shipping known-vulnerable crypto code, reachable or not, is a bad look for a project whose audience is testing infrastructure.
- Bundle size: the polyfill chain is a large part of the demo's 1,325 kB JS.

## Options
1. Check whether a newer `node-stdlib-browser` / `crypto-browserify` chain clears the advisories.
2. Narrow the polyfill set — the demo needs `Buffer`, `process`, `events`, `stream` and `crypto.createHash`. `vite-plugin-node-polyfills` supports selecting individual polyfills, which may drop `elliptic` entirely.
3. Replace the SHA1 path with `SubtleCrypto` (available in every target browser) and drop the `crypto` polyfill.

Option 2 or 3 would cut bundle size as well.

## Acceptance
Either the advisories clear, or `elliptic` is no longer in the built bundle (`grep` the artifact), or a recorded decision explaining why it stays — not an implicit "dev deps don't count".

### LINKED ISSUE #396: Browser demo is not type-checked; Reply union omits Buffer that the client can actually return
Split out of PR #394, where typechecking the browser demo was attempted and deliberately deferred because it surfaces pre-existing errors. Recording it so the deferral does not evaporate on merge.

## Problem
`examples/browser-demo` is not type-checked by anything. Adding a `tsconfig` plus a `tsc --noEmit` step surfaces **3 pre-existing type errors**:

- `format.ts`'s `Reply` union omits `Buffer`, while `InMemoryRedisClient.command()` can return one.
- `formatReply` has no `Buffer` branch.

So the demo's reply formatter is typed against a narrower contract than the client actually provides.

## Why it was deferred
Fixing it changes demo behavior — the formatter would need to decide how a `Buffer` reply renders in the console — and PR #394 was scoped to closing CI gaps, not changing what the demo shows. Shipping the typecheck step with 3 known failures would have made the new CI job red on arrival; shipping it with the errors suppressed would have been a partial gap-closure disguised as a complete one.

## What to decide
How a `Buffer` reply should render. Real `redis-cli` prints binary-safe output with escapes (e.g. `"\xe9\xe9"`); the demo currently has no such path. Options: widen `Reply` and add a `Buffer` branch that mirrors `redis-cli` escaping, or have the client decode to string at the boundary for demo purposes and narrow the return type instead.

The first is more faithful to what this project is for. Note PR #391's verification leaned on exactly this path — `APPEND bin éé` → `4` proves a real `Buffer` is in play, so the demo can already produce values the formatter is not typed for.

## Acceptance
- `examples/browser-demo` has a tsconfig and is covered by `tsc --noEmit` in CI.
- The three errors are fixed rather than suppressed.
- A `Buffer` reply renders deliberately, with a test in `format.test.ts` (now wired into CI by #394) pinning the chosen representation.

fatal10110:
## Summary

Three commits, all confined to `examples/browser-demo` and `.github/workflows/deploy-demo.yml`. No `src/` changes.

### 1. Typecheck the demo (#396)
- Adds `examples/browser-demo/tsconfig.json` (strict, `moduleResolution: bundler`, DOM + `node` + `vite/client` types) and a **Typecheck demo** step (`npx tsc -p examples/browser-demo`) in `deploy-demo.yml`. It uses the root TypeScript install and also covers the demo's `../../src` imports.
- The tsconfig pins `useDefineForClassFields: true`, following the warning in the #396 comment. **Verified:** with only the tsconfig added, the bundle is byte-identical to the root-config build (`cmp`, same `index-C5MYV7rP.js`, 1,329,131 B).
- The 3 `TS2322` errors in `backend.ts` are fixed, not suppressed. `Reply` now includes `Buffer`, and `formatReply` gets a `Buffer` branch that follows `redis-cli`'s `sdscatrepr`:
  - printable ASCII is printed as-is
  - `\\` and `\"` are escaped
  - `\n`, `\r`, `\t`, `\a` and `\b` are escaped
  - every other byte becomes lowercase `\xNN`
  - the result is wrapped in quotes

  The branch checks `instanceof Uint8Array`, so it needs no `Buffer` global, and it sits before the map branch.
- `format.test.ts` pins that output. The expected string was taken from real `redis-cli --no-raw` (Redis on :6399) for `a\b"c\n\r\t\a\bé\0\x7f~ `, which printed `"a\\b\"c\n\r\t\a\b\xc3\xa9\x00\x7f~ "`.
- Note: the demo's client doesn't set `returnBuffers`, so today `GET` of `éé` still comes back as a string. That is how `APPEND bin éé` → `GET bin` → `"éé"` behaves in the browser. The `Buffer` branch makes that part of the contract explicit and tested, rather than something the UI can currently reach.

### 2. Fix: the demo on `main` renders an empty terminal (found while verifying #393)
`src/core/clock.ts` (#388 / PR #410) calls `process.hrtime.bigint()` at module load. The browser `process` shim has no `hrtime`, so the page throws `Cannot read properties of undefined (reading 'bigint')` before the UI starts. I confirmed this on an unmodified `main` build served locally. A new demo-local `process-shim.ts` wraps the plugin's shim and adds `hrtime` / `hrtime.bigint` based on `performance.now()`. `vite.config.ts` points the injected `vite-plugin-node-polyfills/shims/process` import at it. This is a demo-only fix: the library stays Node-targeted.

### 3. Get `elliptic` out of the Pages bundle (#393)
- Calling `nodePolyfills()` with no `include` bundled all of node-stdlib-browser. The demo's `src` graph needs only the `buffer` module; the `Buffer`/`process`/`global` globals are injected regardless of `include`. `events` and `stream` turned out not to be in the demo's graph: the build emits no "externalized for browser compatibility" warnings. The old `vm-browserify` build warning is gone as well.
- Crypto's only use is a **synchronous** `createHash('sha1')`, in `script-cache.ts` (script SHAs) and `hyperloglog.ts` (element hash, `.digest().readBigUInt64BE(0)`). Both call sites are sync, which rules out SubtleCrypto. `node:crypto` / `crypto` are now aliased to a pure-JS SHA-1 in `crypto-shim.ts` (~100 lines). Any other algorithm throws.
- `crypto-shim.test.ts` checks the shim against the FIPS "abc" vector, against `node:crypto` for every length from 0 to 200 bytes (covering the padding edges), for UTF-8 string input and chained updates, and that the raw digest is a `Buffer` with a matching `readBigUInt64BE`.
- A new CI step, **Check bundle has no elliptic**, fails the build if `elliptic` or `secp256k1` shows up in `dist/assets/*.js` again.

| | main | this PR |
|---|---|---|
| `index-*.js` | 1,329,131 B | 583,402 B |
| gzip (vite report) | ~353 kB | 150.4 kB |
| `elliptic` / `secp256k1` / `p384` / `curve25519` / `bn.js` / `crypto-browserify` occurrences | 9 / 6 / 6 / 1 / 1 / – | 0 / 0 / 0 / 0 / 0 / 0 |

The `npm audit` advisories still exist in the demo's lockfile, because the plugin still depends on node-stdlib-browser. Per the #393 acceptance criteria, what matters is that that code no longer ships. I didn't bump any dependencies, to stay clear of Dependabot #400/#401.

Closes #396
Closes #393

## Test plan
- [x] `npx tsc -p examples/browser-demo` passes. Before the fix it reported exactly the 3 `backend.ts` TS2322 errors, plus the `.css` side-effect import, which `vite/client` types now cover.
- [x] Adding only the tsconfig gives a byte-identical bundle (`cmp` against the pre-change build).
- [x] `npm run test:demo`: 22 pass (format + crypto shim).
- [x] `npm run lint`, `npx tsc --noEmit` (root), and `npx prettier --check "**/*.{js,ts,json}"` are clean.
- [x] `npm run build` in the demo is clean. `grep -E 'elliptic|secp256k1' dist/assets/*.js` finds no matches (the same grep matches the `main` build).
- [x] Tested `vite preview` in a browser. It boots with no console errors, and these all work: `SET`/`GET`, `SCRIPT LOAD "return 1"` → `e0e1f9fabfc9d4800c877a703b823ac0578ff8db` (the correct SHA-1), `EVAL` via jsDelivr WASM, `EVALSHA`, `PFADD hll a b c d e` → 1, `PFCOUNT` → 5, and `MONITOR` in a second tab showing microsecond timestamps.


examples/browser-demo/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
+  // Typecheck-only config for the demo (`tsc -p examples/browser-demo`), run in
+  // CI by deploy-demo.yml. Vite also picks this up as the nearest tsconfig for
+  // its transform of the demo and ../../src/**, displacing the root one — so it
```
fatal10110: [medium] This comment says Vite uses this file for its transform of the demo **and `../../src/**`**, displacing the root tsconfig. It doesn't. Vite resolves the nearest `tsconfig.json` per file, walking up from that file's directory. For `src/**` the nearest one is still the root `tsconfig.json`. This file only governs `examples/browser-demo/*.ts`.

How I checked: I built with `"useDefineForClassFields": false` here and `--minify false`, then diffed the result against the `true` build. All 31 changed lines are class fields in demo files: `SingleConnection`/`ClusterConnection` (`conn`, `conns`, `byAddr`, `streamingNodes = []`, …) from `backend.ts`, and the terminal/tab-manager classes from `tabs.ts` (`line = ""`, `history = []`, `tabs = []`, `active = -1`). No class from `src/` changed. (Building with this tsconfig removed gives the same `index-DPz7YATV.js` as with it, so the byte-identical claim does hold.)

The same wrong premise appears in `deploy-demo.yml`:
- L16-19 say the root `tsconfig.json` stays in the push `paths` filter only because `tsx` loads it. It is still a Vite transform input for every `src/**` file, so that rationale invites someone to drop it from `paths` and miss redeploys.
- L87-88 say `../../src` is typechecked "under the browser-oriented config the bundle is built with". `tsc -p` does check src under this config, but the bundle builds src under the root config.

Suggest rewording all three to say that this config applies to the demo's own files, and that `src/**` keeps the root tsconfig for the transform.

examples/browser-demo/process-shim.ts
```diff
@@ -0,0 +1,36 @@
...
+// first line of UI code runs and the page renders an empty terminal.
+
+// Imported by path rather than by package name: the bare specifier is aliased
+// to this very file.
+// @ts-expect-error TS7016 — the plugin's shim ships no type declarations.
+import shim from './node_modules/vite-plugin-node-polyfills/shims/process/dist/index.js'
+
+const process = shim as NodeJS.Process
+
+if (typeof process.hrtime !== 'function') {
```
fatal10110: [medium] Nothing tests this shim, and it is the fix for the most visible regression in the PR: the page rendering an empty terminal because `src/core/clock.ts` calls `process.hrtime.bigint()` at module load. `test:demo` runs under Node, where the real `process` has `hrtime`, and CI has no browser smoke test. If the alias in `vite.config.ts` gets dropped or the shim changes, CI stays green and the Pages build ships blank again.

A unit test is cheap here. The relative import on L14 resolves to the plugin's browser `process` object, which has no `hrtime`, so this branch also runs under `node:test`. A `process-shim.test.ts` could check that `process.hrtime.bigint()` returns a `bigint`, that it is non-decreasing over repeated calls, and that `nextTick` still exists.

Also, the `prev` branch (L26-28) is speculative. `src/` only ever calls `hrtime.bigint()` (clock.ts L19/54/67). Either drop `prev` or cover it in the same test.

(For the record: I smoke-tested `vite preview` of this build in a browser. It boots with no console errors, `SCRIPT LOAD "return 1"` → `"e0e1f9fabfc9d4800c877a703b823ac0578ff8db"`, `PFADD hll a b c d e` → 1, `PFCOUNT` → 5, `EVALSHA` → 1.)

.github/workflows/deploy-demo.yml
```diff
@@ -82,16 +81,34 @@ jobs:
...
         working-directory: examples/browser-demo
 
+      # vite-plugin-node-polyfills inlines what it polyfills into the shipped
+      # artifact, so "it's only a dev dependency" does not hold for it (#393).
+      # The demo polyfills only `buffer` and shims crypto's SHA-1; this fails
+      # the build if a config change lets crypto-browserify's `elliptic` back
+      # into the Pages bundle.
+      - name: Check bundle has no elliptic
+        run: |
+          if grep -l -E 'elliptic|secp256k1' dist/assets/*.js; then
```
fatal10110: [low] `if grep -l …; then fail; fi` treats grep's exit 2 the same as a clean pass. Exit 2 happens when the glob matches nothing (a literal `dist/assets/*.js` gives "No such file") or on a read error. The build step failing first makes this unlikely today. Still, a guard that passes vacuously is worth closing: first assert that at least one `dist/assets/*.js` exists, or capture `rc=$?` and fail on anything other than 1.

I checked that the guard does go red when it should. With `nodePolyfills()` restored (no `include`), the bundle is 1,329.76 kB and `grep -l -E 'elliptic|secp256k1'` matches it. With this PR's config, the bundle is 583.40 kB and has no matches.

examples/browser-demo/format.ts
```diff
@@ -106,3 +113,34 @@ function formatList(items: Reply[]): string {
 function quote(value: string): string {
```
fatal10110: [low] Follow-up, not blocking. `quoteBytes` matches `redis-cli --no-raw` byte for byte. I checked all 256 byte values, the empty string, and the test's pinned string against real redis-cli 7.2 and found no mismatches. The problem is that it is unreachable today, because the demo client doesn't set `returnBuffers`. The path users actually hit is this `quote()`, and it only escapes `\` and `"`.

Real redis-cli:
```
$ redis-cli set s $'é\nx'
$ redis-cli --no-raw get s
"\xc3\xa9\nx"
```
The demo prints `"é` followed by a raw newline and `x"`. This predates the PR. Routing `quote(value)` through `quoteBytes(new TextEncoder().encode(value))` would make the two agree. Replies that aren't valid UTF-8 will still have been lossy-decoded upstream.

fatal10110:
## Review of 5150ab4: **NOT YET** (2 medium, 2 low)

|---|---|---|---|
| medium | `tsconfig.json:4` (+ `deploy-demo.yml:16-19, 87-88`) | The comments say the demo tsconfig drives Vite's transform of `../../src/**`. It doesn't: `src/**` still uses the root tsconfig. That also makes the rationale for keeping the root tsconfig in the `paths` filter misleading. | arch |
| medium | `process-shim.ts:18` | Nothing tests the hrtime shim, which is the fix for the blank-page regression. The `prev` branch is unused. | arch |
| low | `deploy-demo.yml:106` | The elliptic guard passes vacuously when grep exits 2, for example when the glob matches nothing. | arch |
| low | `format.ts:113` | Follow-up: the reachable string path `quote()` still doesn't match redis-cli escaping (`"\xc3\xa9\nx"`). | wire |

**Verified:**
- `quoteBytes` matches real `redis-cli --no-raw` (7.2.1) for all 256 bytes, the empty string and the pinned test string.
- The shim's SHA-1 matches `SCRIPT LOAD` for 11 scripts: empty, 55/56/63/64/65/119/120 bytes, UTF-8/emoji, and 1.4 KB.
- The shim's HLL `readBigUInt64BE` matches `node:crypto` on 5,000 random elements.
- Nothing else in the demo's graph imports `crypto`.
- `npx tsc -p examples/browser-demo` is clean and covers 127 `src/` files. Swapping `main`'s `format.ts` back in gives exactly the 3 TS2322 errors in `backend.ts` (plus the new tests' Buffer errors).
- `test:demo` passes 22/22.
- The build gives `index-DPz7YATV.js` at 583.40 kB with no elliptic/secp256k1. It is byte-identical with or without the demo tsconfig.
- The CI guard goes red when `nodePolyfills()` is unrestricted (1,329.76 kB, matches).
- `vite preview` smoke test: boots with no console errors, `SCRIPT LOAD`/`EVALSHA`/`PFADD`/`PFCOUNT` are correct.

Both lenses were applied by the coordinator directly, because the specialist agents couldn't be spawned (concurrency limit).

examples/browser-demo/tsconfig.json
fatal10110: Fixed in fc359d1. You were right, and thanks for the class-field diff. I rewrote the tsconfig header to say that Vite resolves the nearest tsconfig per file, so this one covers only the demo's own `*.ts` (the class-field emit in `backend.ts` and `tabs.ts`), while `src/**` is still transformed under the root tsconfig. In `deploy-demo.yml`, the `paths` comment now names the root tsconfig as a build input for every `src/**` file (and also the `tsx` config for `test:demo`). The Typecheck step comment now says `tsc -p` checks src under the browser config while the bundle builds src under the root config.

examples/browser-demo/process-shim.ts
fatal10110: Fixed in 285f5c5. I added `process-shim.test.ts`, which imports `./process-shim` the same way the bundle does, so it runs against the plugin's browser `process`. I confirmed `typeof hrtime` is `undefined` on that raw object under Node, so without the patch these tests fail. The tests check that the object is not Node's `process` and still has `nextTick`, that `hrtime.bigint()` exists and returns a bigint, that it never decreases over 1000 samples, and that it counts nanoseconds (a 5 ms sleep reads as 1e6 to 5e9 ns). I dropped the speculative `hrtime(prev)` diff form. I kept the bare no-arg `[s, ns]` tuple so `hrtime` has a sensible shape, and that is tested too. `test:demo` now passes 26 tests.

examples/browser-demo/format.ts
fatal10110: Agreed. I'm leaving it out of this PR as you suggested, and it's now listed under **Follow-up** in the PR body: the `quote()` string path prints `é` and newlines raw, where `redis-cli --no-raw` prints `"\xc3\xa9\nx"`. The fix is to route `quote(value)` through `quoteBytes(new TextEncoder().encode(value))`. Thanks for checking `quoteBytes` against all 256 bytes.

.github/workflows/deploy-demo.yml
fatal10110: Fixed in fc359d1. The step now uses `shopt -s nullglob` and fails if `dist/assets/*.js` matches no file. It captures grep's exit code with `|| rc=$?` so it works under Actions' `bash -e`, and it passes only when grep exits 1. Exit 0 fails with the elliptic error, and any other code fails with `grep failed (exit N)`. I checked grep's exit codes locally: 1 on this PR's build (583,356 B) and 0 on the `main` build. I couldn't run the step as a standalone bash script in my sandbox, so the full step will first run for real in CI on this push.

examples/browser-demo/process-shim.ts
```diff
@@ -0,0 +1,35 @@
...
+  // `.bigint()` — the only form ../../src uses — as monotonic nanoseconds on
+  // top of performance.now(), plus the bare no-arg `[s, ns]` tuple so the
+  // function isn't a half-shaped hrtime. (Node's `hrtime(prev)` diff form is
+  // not implemented: nothing calls it.) Browsers coarsen performance.now() to
+  // microseconds or worse, which clock.ts tolerates — it only needs a
+  // monotonic source to offset a Date.now() anchor.
+  const nowNanos = () => BigInt(Math.round(performance.now() * 1e6))
+  const hrtime = (): [number, number] => {
```
fatal10110: [low] Non-blocking. With `prev` gone, `hrtime(prev)` still type-checks because the object is typed as `NodeJS.Process`, but it now silently returns the absolute time instead of the diff. It's unreachable today (src only calls `.bigint()`). If you want it to fail loudly rather than return a wrong answer, throw when an argument is passed, e.g. `(...args: unknown[]) => { if (args.length) throw new Error('hrtime(prev) not shimmed'); … }`. The comment above already documents it, so this is optional.

fatal10110:
## Re-review of 285f5c5 (since 5150ab4): **SHIP**

Prior findings:
- [medium] tsconfig/workflow comments claimed the demo tsconfig covers `src/**`: **fixed**. The new wording matches what I observed: flipping `useDefineForClassFields` only changes the classes in `backend.ts` and `tabs.ts`, and the root tsconfig is correctly described as a build input for `src/**`.
- [medium] no test for the hrtime shim, unused `prev`: **fixed**. `process-shim.test.ts` imports the plugin's browser `process`, which has no hrtime, so it tests the patch itself. It asserts that the object isn't Node's `process`, that `bigint()` returns a bigint, is non-decreasing and counts in nanoseconds, and that bare `hrtime()` returns a tuple. `prev` is dropped.
- [low] elliptic guard passed vacuously on grep exit 2: **fixed**. I ran the step under `bash --noprofile --norc -eo pipefail`, the way Actions runs it:
  - real dist → exit 0 (pass)
  - bundle containing "elliptic" → grep rc 0 → exit 1
  - no `*.js` (nullglob) → exit 1
  - unreadable file → grep rc 2 → exit 1 ("grep failed (exit 2)")
- [low] `quote()` escaping: deferred to follow-ups, as agreed.

New: one [low] inline comment, non-blocking (`hrtime(prev)` type-checks but silently ignores `prev`).

Checks on 285f5c5:
- `npx tsc -p examples/browser-demo` is clean.
- `npm run test:demo`: 26/26 pass.
- `vite build` gives `index-CjgSjUpb.js` at 583.35 kB. The hash changed from the previous `index-DPz7YATV.js` only because the `prev` branch was dropped.
- `grep -E 'elliptic|secp256k1'` finds no matches.

PR #472, opened 2026-09-23
fix(resp2): match Redis multibulk, bulk-terminator and inline-cap framing (#441)

### LINKED ISSUE #441: RESP2 decoder: pre-7.0 multibulk bound is ungated, the unterminated-bulk error is invented, and the 64KB inline cap is missing
Split out of the review of #431, which fixed the adjacent `proto-max-bulk-len`
bulk-length check but deliberately left these two alone — both are pre-existing,
unrelated to `proto-max-bulk-len`, and one needs a compatibility gate.

Ground-truthed against redis-server **6.2.24** and **7.2.16**.

## 1. The multibulk element-count bound is version-specific and only half applied

`src/core/transports/resp2/decoder.ts` now rejects a count above `INT_MAX`,
which is the bound 7.x and 8.x use. 6.2 is stricter:

```
*1048577   (1M+1)   6.2.24 -> -ERR Protocol error: invalid multibulk length, close
                    7.2.16 -> accepted (waits for elements)
*1048576   (1M)     both   -> accepted
*2147483648         both   -> -ERR Protocol error: invalid multibulk length, close
```

Redis tightened `processMultibulkBuffer` from `ll > 1024*1024` to `ll > INT_MAX`.
Modelling the pre-7.0 bound needs a `FeatureId` in
`src/core/compatibility/profile.ts` + a `VersionGate` in `feature-gates.ts`,
plus profile-boundary tests — the pattern `config.set.failure-message` uses.

No memory hazard in the meantime: the decoder allocates nothing from the count,
it just reports `incomplete` until the elements arrive.

## 2. `Protocol error: bulk string not terminated` is not a real Redis error

Real Redis never validates the CRLF after a bulk payload. It reads exactly `ll`
bytes and skips two, so a bulk whose terminator is wrong simply dispatches:

```
*1\r\n$3\r\nfooXX   6.2.24 -> -ERR unknown command `foo`, with args beginning with:
                   7.2.16 -> -ERR unknown command 'foo', with args beginning with:
```

The mock raises a protocol error and closes instead. Matching Redis means
deleting the check, which changes framing for malformed input and has no
coverage anywhere today, so it wants its own change plus a raw-tcp test. The
divergence is marked with a comment in the decoder.

(The sibling invented string, `Protocol error: expected bulk string`, was fixed
in #431 to Redis' real `Protocol error: expected '$', got '%c'` wording, since
that one needed no behaviour change.)

## 3. The 64KB inline-request cap is not modelled

Added after the second review of #431. Real Redis caps an inline request at
`PROTO_INLINE_MAX_SIZE` (64KB). The mock has no bound at all:
`parseInlineFrame`'s `readLine` scans for `\r\n` indefinitely, so a client can
grow the buffer without limit and park the connection with no reply.

The cap bounds an **unterminated** inline buffer, not an inline command's
length as such. Ground-truthed on redis 6.2.24, 7.2.16 and 8.0.6 (identical):

```
'PING ' + 'a'*70000            (no CRLF)  -> -ERR Protocol error: too big inline request, close
'PING ' + 'a'*60000 + '\r\n'              -> $60000 ... (served)
'ECHO ' + 'a'*70000 + '\r\n'   (one write) -> $70000 ... (served)
```

The last line is the subtle one: Redis checks the cap only when a read leaves
no newline in the buffer, so a >64KB inline line that arrives complete is
served. Whether a given oversized line trips the cap therefore depends on how
the bytes are split across reads — worth modelling as "no `\n` in the first
64KB buffered", and worth testing only in that deterministic form (write the
unterminated prefix, assert the error and the close).

This also means "inline commands are exempt from `proto-max-bulk-len`" is
vacuous upstream: the smallest settable `proto-max-bulk-len` is 1048576, far
above this cap. #431 renamed its inline test accordingly; it now pins only
that an inline command with a 60KB argument is still served.

fatal10110:
## Summary

Fixes three RESP2 request-decoder gaps (`src/core/transports/resp2/decoder.ts`). Each was checked against real servers: redis 6.2.24, 6.2.14, 7.0.15, 7.2.4 and 8.0, and valkey 7.2.14, 8.0.0, 8.0.11 and 9.0.0.

1. **The multibulk element-count bound is now gated.** Redis 6.2 refuses a count above `1024*1024`; 7.0+ refuses one above `INT_MAX`.
   - New `FeatureId` `protocol.multibulk-count-int-max` with gate `{ redis: '7.0.0', valkey: '7.2.0' }`. valkey 7.2.14 already accepts `*1048577`, so every Valkey profile gets `INT_MAX`.
   - `Resp2SessionAdapter` passes `session.server.profile` to the decoder through a new optional `profile` option. Without it, the decoder uses the 7.0+ rules.
   - Same line: a negative count (`*-5`) is now skipped like `*0`/`*-1` on every version, instead of raising `invalid multibulk length`. This matches Redis' `ll <= 0` check.
2. **Removed the invented `Protocol error: bulk string not terminated`.** Redis reads exactly `ll` bytes and skips the next two without checking them. So `*1\r\n$3\r\nfooXX` now dispatches `foo` (`unknown command`), and framing continues normally after it.
3. **The 64KB inline cap (`PROTO_INLINE_MAX_SIZE`) is now modelled.** When more than 64KB is buffered with no `\n`, the server sends `-ERR Protocol error: too big inline request` and closes. Commands earlier in the same write are still answered first.
   - Inline requests now end at `\n`, with an optional `\r` before it, as in `processInlineBuffer`. A bare-LF `PING hi\n` used to hang; it is now served.

### Note on the inline-cap rule

I implemented Redis' literal check: refuse when the buffer holds more than 64KB and **no** `\n` at all. The issue's other reading ("no `\n` within the first 64KB") would refuse a case every real version serves:
- Write `PING ` + 65531 bytes, which is exactly 64KB with no newline.
- Pause, then write `a\r\n`.
- Result: served on 6.2, 7.0, 8.0 and valkey 7.2.

An oversized line that arrives complete in one write does not behave the same on every version, because it depends on how the reads split. 100KB is refused by 6.2.24 and 8.0 but served by 7.0.15 and valkey 7.2.14. So the tests cover only cases that behave the same everywhere.

### Not modelled (noted in code)

- Later valkey patch releases (7.2.14, 8.0.11) **do** check the bulk terminator: `-ERR Protocol error: invalid CRLF in request`, then close. valkey 8.0.0 and 9.0.0 do not. No preset profile is one of those patch releases, and one `VersionGate` can't express a fix backported to several branches, so this is left out.

Closes #441

## Test plan
- [x] New `tests-integration/raw-tcp/resp2-decoder-framing.test.ts` (10 cases): red on mock before the fix (4 failures, and the inline-cap cases hung). Green against real Redis 8.0 (`REDIS_STANDALONE_PORT=6399`) and green on mock after the fix.
- [x] New `tests-integration/compatibility/multibulk-count-gate.test.ts`: `*1048577` and `*2147483647` followed by a `+` element tell the 6.2 bound (`invalid multibulk length`) apart from 7.0+ (`expected '$', got '+'`). Red on `redis-6.2` before the fix. `npm run test:integration:compatibility:mock` is green on all 7 profiles.
- [x] New `tests/core/resp2-decoder.test.ts`: covers what a socket can't show deterministically, such as exactly 64KB with no newline still waiting while one byte more is refused. Also covers gate boundaries per profile and the two skipped bytes. The profile gate is asserted in `tests/compatibility/profile.test.ts`.
- [x] `npm run build`, `npm run lint`, `npm test` (474 pass), `npm run test:integration:mock` (1262 pass, 2 skipped)
- [x] Raw-tcp protocol suites (`protocol`, `protocol-errors`, `proto-max-bulk-len-protocol`, `resp2-decoder-framing`) against real standalone Redis: 34/34. The cluster-backed raw suites failed in this run because the shared real cluster was returning `CLUSTERDOWN`. That's environmental; on the real backend, real Redis parses the requests, so this change can't affect those results.
- [x] Docs: `docs/API.md` has a new gate row; `docs/ARCHITECTURE.md` has a sentence on the profile passed to the decoder and the inline cap.


src/core/transports/resp2/decoder.ts
```diff
@@ -131,14 +155,11 @@ export class Resp2CommandDecoder {
...
     }
 
-    // Redis bounds the element count as well as each bulk. The upper bound is
-    // version-specific: 7.2.16 rejects above INT_MAX, while 6.2.24 rejects
-    // anything above 1024*1024 (`*1048577` errors on 6.2 and is accepted on
-    // 7.2). Only the INT_MAX bound is applied here, because it is the one every
-    // supported profile agrees on; the tighter pre-7.0 bound needs a
-    // compatibility gate and is tracked in #441.
+    // Redis bounds the element count as well as each bulk, but only from above:
+    // `ll <= 0` (including `*-5`) is skipped like `*0`, never an error. The
```
fatal10110: [medium] `ll <= 0` is only half of Redis' rule — the count is parsed with `string2ll`, which rejects leading zeros and `-0`. `parseLength`'s `/^-?\d+$/` does not, so widening the negative branch from `count < -1` to "skip anything `<= 0`" regresses `*-05` (main errored, now silently skipped), and `*-0` / `*01` stay wrong.

Ground truth (raw TCP, redis 6.2, 7.0, 8.0, valkey 7.2.14 — identical on all four), each followed by a `PING`:
```
*-05\r\n   -> -ERR Protocol error: invalid multibulk length\r\n   CLOSED
*-0\r\n    -> -ERR Protocol error: invalid multibulk length\r\n   CLOSED
*01\r\n... -> -ERR Protocol error: invalid multibulk length\r\n   CLOSED
*-5\r\n    -> +PONG                                                 (skip, as the PR says)
*-2147483649\r\n -> +PONG                                           (skip)
```
This PR on mock (redis-6.2 and redis-8.0 profiles): `*-05` -> `+PONG`, `*-0` -> `+PONG`, `*01` -> executes the frame.

Fix: make `parseLength` accept only what `string2ll` does, e.g. `/^(0|-?[1-9]\d*)$/` (this also fixes `$01` / `$-0` on the bulk side, which string2ll turns into `invalid bulk length`). Add `*-05`, `*-0`, `*01` to the decoder unit test.

src/core/transports/resp2/decoder.ts
```diff
@@ -175,27 +196,19 @@ export class Resp2CommandDecoder {
         throw new Resp2ParseError('Protocol error: invalid bulk length')
       }
 
+      // Redis reads exactly `length` bytes and then skips two *without looking
+      // at them*: a wrong terminator is not a protocol error, so
+      // `*1\r\n$3\r\nfooXX` dispatches `foo`. Verified on redis 6.2.24,
+      // 7.0.15, 7.2.4 and 8.0, and valkey 8.0.0 / 9.0.0. (Later valkey patch
+      // releases — 7.2.14, 8.0.11 — refuse it with `invalid CRLF in request`;
```
fatal10110: [medium] The Valkey caveat here (and in the PR body) understates it: it is not just "later 7.2/8.0 patch releases" — the current **9.0** line checks the terminator too. Only the `x.0.0` releases skip it.

Ground truth, `*1\r\n$3\r\nfooXX` then `PING` on the same connection:
```
valkey 7.2.14 -> -ERR Protocol error: invalid CRLF in request\r\n   CLOSED
valkey 8.0.11 -> -ERR Protocol error: invalid CRLF in request\r\n   CLOSED
valkey 9.0.6  -> -ERR Protocol error: invalid CRLF in request\r\n   CLOSED
valkey 8.0.0  -> -ERR unknown command 'foo', with args beginning with: \r\n+PONG\r\n   open
valkey 9.0.0  -> -ERR unknown command 'foo', with args beginning with: \r\n+PONG\r\n   open
redis 6.2 / 7.0 / 8.0 -> unknown command + PONG, open (matches this PR)
```
So every maintained Valkey line (7.2.x, 8.0.x, 9.0.x at latest patch) refuses, and a `valkey-8.0`/`valkey-9.0` user comparing against a current Valkey release will see the mock accept what Valkey rejects. Keeping the Redis behaviour for the pinned `.0.0` presets is defensible, but please (a) correct this comment to say that current patch releases of every Valkey line, 9.0 included, refuse with `invalid CRLF in request`, and (b) open a follow-up issue to model it for the valkey flavor (e.g. a valkey-only gate, since no maintained Valkey line lacks the check).

tests-integration/raw-tcp/resp2-decoder-framing.test.ts
```diff
@@ -0,0 +1,214 @@
...
+      await expectThenClose(
+        conn,
+        '-ERR Protocol error: invalid multibulk length\r\n',
+      )
+    })
+
+    // Accepting a count is otherwise invisible — the server just waits for the
+    // elements. A non-'$' element prefix makes it observable: past the count
+    // check the parser reaches the element and names the bad byte.
+    test('a count above 1024*1024 is accepted on 7.0+', async () => {
```
fatal10110: [medium] This test is named "on 7.0+" but is not gated, and the file header (lines 12-15) says every assertion holds on 6.2.24 and valkey 7.2.14 "except where noted" — that is not true, and nothing is noted.

Ran this file against real servers I started (`TEST_BACKEND=real REDIS_STANDALONE_PORT=...`) and against the mock with a 6.2 profile:
```
redis 8.0            10/10 pass
redis 7.0            10/10 pass
redis 6.2            8/10 — fails "a count above 1024*1024 is accepted on 7.0+" (gets invalid multibulk length)
                             and "a bulk with a wrong terminator..." (6.2 wording is: unknown command `foo`, with backticks)
valkey 7.2.14        8/10 — both "bytes after a bulk payload" tests fail (-ERR Protocol error: invalid CRLF in request)
mock REDIS_COMPAT=redis-6.2   9/10 — this test fails
```
Fix: skip this test when `activeProfile === 'redis-6.2'` (or drop it; `compatibility/multibulk-count-gate.test.ts` already covers both sides of the bound per profile), and make the header accurate: state which assertions are 7.0+/Redis-only (6.2 unknown-command wording, Valkey CRLF check).

src/core/transports/resp2/decoder.ts
```diff
@@ -175,27 +196,19 @@ export class Resp2CommandDecoder {
...
+      // not modelled, since no preset profile is one of them.)
       const valueStart = bulkHeader.nextIndex
       const valueEnd = valueStart + length
       const lineEnd = valueEnd + 2
       if (this.buffered.length < lineEnd) {
```
fatal10110: [low] Non-blocking, pre-existing (not introduced here), raised because the issue asks about decoder memory/hang risk: when a multibulk frame is incomplete, `next()` starts parsing it again from index 0 on every chunk. It copies every finished element again with `Buffer.from`, and `push()` concatenates the whole buffer again (lines 92 and 118). The cost is quadratic, and it runs synchronously on the event loop, so it stalls every connection. Measured on this branch with elements fed through the decoder in 64KB chunks (`*N` followed by `$1\r\na\r\n` for each element):
```
 50000 elements (350KB)  ->  86 ms
100000 elements (700KB)  -> 242 ms
200000 elements (1.4MB)  -> 923 ms
```
Redis avoids this by keeping `multibulklen`/`bulklen` across reads. This PR only makes the bound tighter (on 6.2), so there is no regression. Suggest a follow-up issue for incremental parse state.

src/core/transports/resp2/decoder.ts
```diff
@@ -209,22 +222,35 @@ export class Resp2CommandDecoder {
...
-    const line = readLine(this.buffered, index)
-    if (!line) {
+    // An inline request ends at `\n`; a `\r` just before it is optional
+    // (networking.c, `processInlineBuffer`).
+    const newline = this.buffered.indexOf(0x0a, index)
+    if (newline === -1) {
+      // Redis' 64KB inline cap bounds an *unterminated* buffer, not a line's
+      // length: it is checked only when no newline has arrived at all, so a
+      // line longer than 64KB whose newline turns up in the same read is still
+      // served. Verified on redis 6.2.24, 7.0.15, 8.0 and valkey 7.2.14.
```
fatal10110: [low] I confirmed the literal rule is the right choice: the mock matches every deterministic case on real 6.2, 7.0, 8.0 and valkey 7.2.14. That covers 64KB+1 with no newline (refused and closed), exactly 64KB (waits), 64KB then a pause then `a\r\n` (served), `PING\r\n` followed by 70KB with no newline (PONG, then refused), 66KB of pipelined `PING\r\n` followed by 64KB+1 with no newline (all PONGs, then refused), and a bare `PING hi\n` (served).

Nit on wording: "a line longer than 64KB whose newline turns up in the same read is still served. Verified on redis 6.2.24 ... 8.0" reads as if a single client write is safe. On real 6.2 and 8.0, a 100KB line sent in **one** write is refused (`-ERR Protocol error: too big inline request`, then close), because Redis reads 16KB at a time. 7.0 and valkey 7.2 serve it, and so does the mock on every profile, because Node hands it 64KB chunks. You already covered this in the PR body. Suggest one sentence here too, saying the mock settles this read-split-dependent case by serving the line.

fatal10110:
Review of head `8a5c01b` against `origin/main`. Checked through two lenses: Redis wire-compat and Node architecture. The specialist subagents were unavailable, so the coordinator applied both lenses directly.

**Verdict: NOT YET.** 3 medium, 2 low.

|---|---|---|---|
| medium | `decoder.ts:159` | `parseLength` is looser than `string2ll`. `*-05` is now skipped where it used to be an error (a regression). `*-0` and `*01` are still accepted. Real servers answer all three with `invalid multibulk length` and close | Redis |
| medium | `decoder.ts:203` | The Valkey CRLF caveat leaves out valkey **9.0.6**, which also refuses with `invalid CRLF in request`. So does the latest patch of every maintained Valkey line. Fix the comment and open a follow-up | Redis |
| medium | `resp2-decoder-framing.test.ts:141` | The "7.0+" test is not gated, and the header claim that it holds on every version is false. Fails on real 6.2, on valkey 7.2.14, and on the mock with the redis-6.2 profile | Redis/Arch |
| low | `decoder.ts:208` | Pre-existing: an in-progress multibulk is re-parsed from zero on every chunk, which is quadratic (200k elements take about 0.9s on the event loop). Follow-up, not blocking | Arch |
| low | `decoder.ts:232` | Wording on the inline cap: on 6.2 and 8.0, a single write of 100KB is refused | Redis |

Checked against real redis 6.2, 7.0, 8.0 and valkey 7.2.14, 8.0.0, 8.0.11, 9.0.0 and 9.0.6, using servers started just for this review. Confirmed: the multibulk bound per version (`*1048577` and `*2147483647` refused only on 6.2, `*2147483648` refused everywhere), `*-5` skipped, the bulk terminator not checked on Redis, and every deterministic inline-cap case, all matching the mock byte for byte. The literal rule (buffer over 64KB and no `\n` anywhere) is correct. The issue's "first 64KB" reading would refuse the 64KB, pause, `a\r\n` case, which every server serves.

Tests on the PR head: unit 474/474, raw-tcp mock suite 167/167, `multibulk-count-gate` 4/4 on the redis-6.2, redis-7.0, valkey-8.0 and valkey-9.0 profiles. `tsc` and `eslint` are clean.

src/core/transports/resp2/decoder.ts
fatal10110: Fixed in 59e4517. `parseLength` now accepts only what `string2ll` does: `/^(0|-?[1-9]\d*)$/` plus an int64 range check, returning a bigint. `*-05`, `*-0`, `*01`, `*00` and `*+1` now give `invalid multibulk length` and close. `$01`, `$-0`, `$04` and `$+4` now give `invalid bulk length`. `*-2147483649` and `*-9223372036854775808` are still skipped, and `*-9223372036854775809` is refused. I checked all of this on real redis 6.2, 8.0 and valkey 7.2.14. Covered by raw-tcp tests in `resp2-decoder-framing.test.ts` (new "length format" group) and unit tests in `tests/core/resp2-decoder.test.ts`.

src/core/transports/resp2/decoder.ts
fatal10110: Fixed in 59e4517. The comment now says Valkey patch releases on every current line (7.2.14, 8.0.11, 9.0.6) check the terminator, and only the `x.0.0` releases (the 8.0.0 and 9.0.0 presets) skip it. It also explains why a single `VersionGate` cannot model this. The PR body lists patch-level gating for this as a follow-up.

tests-integration/raw-tcp/resp2-decoder-framing.test.ts
fatal10110: Fixed in 59e4517:
- The "accepted on 7.0+" test is skipped when `activeProfile === 'redis-6.2'`. `multibulk-count-gate.test.ts` covers both sides of the limit for each profile.
- The unknown-command reply is now matched with a regex that accepts either backticks or single quotes, since 6.2 uses backticks.
- The header now lists the real exceptions: Valkey patch releases refuse a bad terminator, 6.2 quotes the name with backticks, and the count limit is version-specific.

Results: mock `redis-8.0` 20/20, mock `REDIS_COMPAT=redis-6.2` 19 pass + 1 skipped, real redis 8.0 20/20, real redis 6.2 19 pass + 1 skipped.

src/core/transports/resp2/decoder.ts
fatal10110: Agreed. This predates this PR, and fixing it properly means keeping partial-frame parse state and growing the buffer instead of concatenating it on every `push()`. That touches every framing path, so I kept it out of this PR. It is listed as a follow-up in the PR body.

src/core/transports/resp2/decoder.ts
fatal10110: Reworded in 59e4517. The comment now lists the deterministic cases, explains that a >64KB line sent in one write depends on how the reads are split (6.2 and 8.0 refuse a 100KB line; 7.0 and valkey 7.2 serve it), and says the mock serves it because Node delivers 64KB chunks.

fatal10110:
Re-review of `59e4517`, diffed against the previously reviewed `8a5c01b`.

**Verdict: SHIP.** All 3 medium findings are fixed. There are no new findings.

| prior | status |
|---|---|
| medium `decoder.ts` `parseLength` looser than `string2ll` | **fixed**: `/^(0\|-?[1-9]\d*)$/` plus an int64 range check, done in BigInt |
| medium Valkey CRLF comment left out 9.0.6 | **fixed** |
| medium framing test "7.0+" not gated, header claim false | **fixed**: skipped on `redis-6.2`, header corrected, unknown-command quote style matched with a regex |
| low quadratic multibulk re-parse | deferred to a follow-up (pre-existing, not blocking) |
| low inline-cap read-split wording | **fixed** |

I ran the same malformed-length probes against real redis 6.2, redis 8.0 and valkey 9.0.0, on servers I started for this review, and against this head on mock. All three real servers gave identical results, and the mock matches them for every length:
- **Skipped (connection stays open):** `*-9223372036854775808` and `*-2147483649`.
- **`invalid multibulk length`, then close:** `*-9223372036854775809`, `*9223372036854775807`, `*9223372036854775808`, `*00`, `*+1`, `*` (empty), `* 1`, `*1 `, `*-0000000000000000001`.
- **`invalid bulk length`, then close:** `$-1`, `$-0`, `$04`, `$+4`, `$` (empty), `$9223372036854775807`, `$9223372036854775808`.
- The earlier inline-cap and bulk-terminator probes still match.

Tests: unit 488/488, raw-tcp on mock 177/177. The framing file passes 20/20 on mock redis-8.0 and on real 8.0. It passes 19/19 with 1 skipped on mock redis-6.2 and on real 6.2. `multibulk-count-gate` passes 4/4 on the redis-6.2 and valkey-9.0 profiles. `tsc` and `eslint` are clean.

One note outside this PR's scope, covered by the header-line follow-up: a NUL inside the `*` header (`*1\x00\r\n`) makes redis 6.2 and 8.0 wait, because they search for the newline with `strchr`. valkey 9.0.0 and the mock both answer `invalid multibulk length`.

Issue #439, opened 2026-09-22
An unknown container subcommand called from Lua should fail command lookup on redis-7.0+, not reach the container


fatal10110:
Found while reviewing #430 (the #413 wording consolidation).

## Problem

From Redis 7.0 a script's `redis.call` / `redis.pcall` resolves the command through the command table, which for a container means resolving `container|subcommand`. An unknown subcommand therefore fails **lookup**, and the script gets the unknown-command error — it never reaches the container at all.

`EVAL "return redis.pcall('<C>', 'BOGUS')" 0`, captured from real servers:

| container | real 6.2.24 | real 7.0.15 and 8.0.6 |
|---|---|---|
| `SLOWLOG`, `COMMAND`, `PUBSUB`, `XGROUP`, `XINFO` | the container's own error, e.g. `ERR Unknown subcommand or wrong number of arguments for 'BOGUS'. Try PUBSUB HELP.` | `ERR Unknown Redis command called from script` |
| `CONFIG`, `CLIENT`, `ACL`, `SCRIPT` | `@user_script: 1: This Redis command is not allowed from scripts` — refused before the subcommand is looked at | `ERR Unknown Redis command called from script` |
| `FUNCTION` | `@user_script: 1: Unknown Redis command called from Lua script` (no such command in 6.2) | `ERR Unknown Redis command called from script` |

On 7.0+ every container behaves the same. On 6.2 only the five in the first row reach the container, and `XINFO` only when called without a key: with one (`redis.pcall('XINFO', 'BOGUS', 'k')`) 6.2 checks the key first and answers `ERR no such key` or `WRONGTYPE` (see #436).

This server dispatches the container on every profile, so on 7.0+ it answers with the container's unknown-subcommand reply instead of the lookup failure:

```
EVAL "return redis.pcall('PUBSUB','BOGUS')" 0 -> -ERR unknown subcommand 'BOGUS'. Try PUBSUB HELP.   (redis-8.0 profile)
```

So this is version-specific and belongs behind a gate, not a flat fix.

A known subcommand with bad arguments is unaffected: `redis.pcall('PUBSUB','CHANNELS','a','b')` reaches the container on every version and gives the same `addReplySubcommandSyntaxError` reply as direct dispatch. This server matches it byte-for-byte on 8.0.6, through both `redis.pcall` and a failing `redis.call` (pinned in `tests-integration/raw-tcp/command-errors-subcommand.test.ts`).

## Related

[...]

PR #473, opened 2026-09-23
fix: use the 6.2 script-abort decoration on the redis-6.2 profile (#442)

### LINKED ISSUE #442: Script abort errors use the Redis 7.0 'script: <sha>' decoration on the redis-6.2 profile
Found while addressing review on #430.

## Problem

When a script aborts — a failing `redis.call`, or a Lua runtime error such as `error(...)` — Redis wraps the error with a script-position decoration. That decoration changed at 7.0, and this server always emits the 7.0 form, including on the `redis-6.2` profile.

Captured from real servers, `EVAL "return redis.call('PUBSUB', ARGV[1])" 0 <arg>`:

```
6.2.24  -ERR Error running script (call to f_<sha>): @user_script:1: ERR Unknown subcommand or wrong number of arguments for '<arg>'. Try PUBSUB HELP.
8.0.6   -ERR Unknown Redis command called from script script: <sha>, on @user_script:1.
```

and `EVAL "error(ARGV[1])" 0 <arg>`:

```
6.2.24  -ERR Error running script (call to f_<sha>): @user_script:1: user_script:1: <arg>
8.0.6   -ERR user_script:1: <arg> script: <sha>, on @user_script:1.
```

This server, `redis-6.2` profile:

```
-ERR Unknown subcommand or wrong number of arguments for '<arg>'. Try PUBSUB HELP. script: <sha>, on @user_script:1.
-ERR user_script:1: <arg> script: <sha>, on @user_script:1.
```

So on `redis-6.2` the body is right (and, since #430, byte-exact) but the wrapper is the 7.0 one: prefix vs. suffix, `f_<sha>` vs. `<sha>`, and the doubled `@user_script:1: user_script:1:` position for runtime errors.

The wrapper is built in one place, `renderScriptError` in `src/core/lua-runtime.ts`, so this is a single gated branch there. The first unrelated difference to watch: 7.0+ also adds `ERR ` to a `redis.error_reply()` body that lacks a code (`-ERR <arg>` on 8.0.6 against `-<arg>` on 6.2.24), which may want the same gate.

## Related

- #439 — what the *body* is for an unknown container subcommand from a script on 7.0+ (lookup failure rather than the container's error). Independent of the wrapper.
- `tests-integration/compatibility/profile-gates.test.ts` asserts only the body segment of the 6.2 `redis.call` reply for this reason; it can assert the whole frame once this is fixed.

## Acceptance

`EVAL "return redis.call('PUBSUB', ARGV[1])" 0 x` and `EVAL "error(ARGV[1])" 0 x` match real Redis byte-for-byte on both the `redis-6.2` and `redis-8.0` profiles.

fatal10110:
## Summary
- On `redis-6.2`, script aborts (failing `redis.call`, Lua runtime errors, engine-classified errors like nonexistent-global reads) now use 6.2's wrapper, `-ERR Error running script (call to f_<sha>): @user_script:<line>: <error>`, instead of the 7.0 suffix `<error> script: <sha>, on @user_script:<line>.`.
- There is one gated branch in `renderScriptError` (`src/core/lua-runtime.ts`) behind a new `script.abort-error-suffix` FeatureId (`redis: 7.0.0`, `valkey: 7.2.0`). Valkey 7.2.14 and 8.0 were checked and answer the 7.0 form.
- The 6.2 body is the raw Lua error string. A failing `redis.call`'s code is put back into the body (`@user_script:2: WRONGTYPE Operation ...` under `-ERR`), and a runtime error keeps its own text, which is why the position appears twice (`@user_script:1: user_script:1: x`). The engine gives both kinds of error the same default `ERR` code, so `RedisLuaRuntime` now records the last error `redis.call` returned, and `raisedByRedisCall()` checks whether the abort is that exact error. A `redis.call` error caught with `pcall` and raised again counts as a runtime error, as it does in 6.2.
- `profile-gates.test.ts` now checks the whole 6.2 frame for the `redis.call('PUBSUB', ARGV[1])` case, where it used to check only the body. It also has a new test for every profile that covers a runtime error with raw bytes, `WRONGTYPE` from `redis.call`, a caught-and-re-raised call error, an error inside a nested function (line 2), and a nonexistent-global read. All expected bytes were captured from real redis 6.2, 7.0, 8.0 and valkey 7.2 / 8.0.
- Updated the `docs/API.md` gate table and the CHANGELOG.

### `redis.error_reply()` `ERR ` prefix: not done here
It is confirmed against real servers: 6.2 answers `-x` to `redis.error_reply('x')`, and 7.0+ answers `-ERR x`. The host can't fix it, though. The engine (`l_redis_error_reply` in lua-redis-wasm `redis_api.c`) adds `ERR ` before anything reaches the host, so `error_reply('x')` and `error_reply('ERR x')` arrive as the same `{err:'x', code:'ERR'}`. This needs a profile-aware `error_reply` in lua-redis-wasm.

### Known remaining 6.2 difference
`error('ERR x', 0)` (level 0, so no position) renders `@user_script:1: x` where 6.2 has `@user_script:1: ERR x`. The engine splits the leading code off a runtime error string, so the host can't recover it. The same engine split also affects 7.0+: real Redis answers `-ERR ERR x script: ...`.

Closes #442

## Test plan
- [x] The tightened and new tests in `tests-integration/compatibility/profile-gates.test.ts` failed on mock `redis-6.2` before the fix and pass after it. They are mock-only, like the rest of that file; every expected frame was compared by hand against real 6.2 / 7.0 / 8.0 / valkey 7.2 / valkey 8.0 containers.
- [x] A raw probe diff of mock `redis-6.2` against real redis 6.2 is byte-identical for all the cases above.
- [x] `npm run test:integration:compatibility:mock`: all 7 profiles pass.
- [x] `npm test`: 462/462 pass.
- [x] Full mock integration: 1249 pass, 0 fail.
- [x] `npm run build`, `npm run lint` and prettier are clean.
- [x] Real standalone: `raw-tcp/command-errors-subcommand.test.ts` and `profile-gates.test.ts` pass. The real `ioredis/scripts` and `node-redis/scripts` suites could not run because the shared real cluster was in `cluster_state:fail` during this work, which has nothing to do with this change. The 7.0+ output path is byte-for-byte unchanged.


src/core/lua-runtime.ts
```diff
@@ -311,6 +356,36 @@ export function renderScriptError(value: ReplyValue): ReplyValue {
...
+ * (the reply itself is always `-ERR`). A Lua runtime error keeps its text as
+ * is; the engine reports those under a default `ERR` code, so only a code it
+ * split off the message itself (`error('WRONGTYPE x', 0)`) is restored.
+ */
+function legacyScriptErrorBody(
+  body: Buffer,
+  code: Buffer | undefined,
+  raisedByRedisCall = false,
+): Buffer {
+  if (!code || (!raisedByRedisCall && code.toString() === 'ERR')) {
```
fatal10110: [medium] **6.2 branch: host-side `redis.call` rejections get `ERR ` folded in, but real 6.2 renders them through `luaPushError` (inner `@user_script: N: ` position, no code, 6.2 wording).**

`raisedByRedisCall` is true for every error the `redisCall` host callback returns, including the script-level rejections produced in `runRedisCommand` itself (`ScriptUnknownCommandError`, `ScriptNotAllowedCommandError`, `ScriptCallNoCommandError`), so this branch prepends `ERR `. Real 6.2 raises these via `luaPushError`, which adds its own `@user_script: <line>: ` (with a space) and no code. The engine-classified `command-arg-type` body (composed right here in `renderScriptError`) has the same gap.

Real redis 6.2.24 vs this PR's mock on `--compat redis-6.2` (redis-cli, same sha):
```
EVAL "return redis.call('nosuchcmd')" 0
 real 6.2: ERR Error running script (call to f_4f59…dd95): @user_script:1: @user_script: 1: Unknown Redis command called from Lua script
 mock    : ERR Error running script (call to f_4f59…dd95): @user_script:1: ERR Unknown Redis command called from script
EVAL "return redis.call('multi')" 0
 real 6.2: … @user_script:1: @user_script: 1: This Redis command is not allowed from scripts
 mock    : … @user_script:1: ERR This Redis command is not allowed from script
EVAL "return redis.call()" 0
 real 6.2: … @user_script:1: @user_script: 1: Please specify at least one argument for redis.call()
 mock    : … @user_script:1: ERR Please specify at least one argument for this redis lib call
EVAL "return redis.call('get', {})" 0
 real 6.2: … @user_script:1: @user_script: 1: Lua redis() command arguments must be strings or integers
 mock    : … @user_script:1: Lua redis lib command arguments must be strings or integers
```
(7.0.15 / 8.0.6 / valkey 7.2.14 all answer the suffix form with the 7.0 wording, e.g. `ERR Unknown Redis command called from script script: <sha>, on @user_script:1.`, which the mock already matches.)

The CHANGELOG entry says "a failing `redis.call` … now carries Redis 6.2's decoration", which isn't true for this family. Either render them on 6.2 (the host knows these are its own `Script*Error` rejections, so it could tag them instead of recording them as command errors, and the `command-arg-type` case is composed right here), or narrow the CHANGELOG/PR wording and list them as a known remaining 6.2 difference with a follow-up issue.

src/core/lua-runtime.ts
```diff
@@ -32,15 +32,43 @@ export class RedisLuaRuntime {
...
       redisPcall: args => this.runRedisCommand(args),
       log: () => {},
     })
   }
 
+  /**
+   * Whether `reply`, the result of the last `eval`, is a script abort raised by
+   * a failing `redis.call` (as opposed to a Lua runtime or engine error).
+   */
+  raisedByRedisCall(reply: ReplyValue): boolean {
```
fatal10110: [low] **Stateful side channel with temporal coupling.** `raisedByRedisCall(result)` is only correct if it is called after `eval()` returns and before the next `eval()` starts, and `lastRedisCallError` keeps the last error buffers alive until the next script runs. That holds today because `eval` is synchronous and `runLuaScript` calls the method right after it, so there's no cross-script or cross-session leak on the shared runtime. The coupling is still implicit, though. Consider having `eval` return the flag with the reply (e.g. `{ reply, raisedByRedisCall }`), or computing it inside `eval` before the `finally` and clearing `lastRedisCallError` there. That way a future caller can't query stale state.

tests-integration/compatibility/profile-gates.test.ts
```diff
@@ -466,41 +472,142 @@ describe(
...
+            Buffer.from(
+              `-ERR Error running script (call to f_${sha}): @user_script:2: WRONGTYPE Operation against a key holding the wrong kind of value\r\n`,
+            ),
+          current: sha =>
+            Buffer.from(
+              `-WRONGTYPE Operation against a key holding the wrong kind of value script: ${sha}, on @user_script:2.\r\n`,
+            ),
+        },
+        {
+          // A redis.call error caught by pcall and re-raised is a runtime
```
fatal10110: [low] **A level-0 re-raise is classified as a `redis.call` abort, and no test covers it.** The comment here (and the PR body) says a pcall-caught `redis.call` error that is raised again "is a runtime error". That's only true for `error(e)` (level ≥ 1, which adds a position). `error(e, 0)` re-raises the identical bytes, so `raisedByRedisCall()`'s equality check returns true and the code is folded back in. That happens to be exactly right, because 6.2's `pcall(redis.call, …)` yields the raw string and re-raising it is indistinguishable from the original call:
```
EVAL "local ok, e = pcall(redis.call, 'incr', 's')\nerror(e, 0)" 0
 real 6.2 : ERR Error running script (call to f_eddc…1c48): @user_script:2: ERR value is not an integer or out of range
 mock 6.2 : (identical)
EVAL "local ok, e = pcall(redis.call, 'incr', 's')\nerror('ERR value is not an integer or out of range', 0)" 0
 real 6.2 / mock 6.2: … @user_script:2: ERR value is not an integer or out of range (identical)
```
Suggest adding an `ERR`-coded `error(e, 0)` case to this table so the equality heuristic's behavior is locked in. The WRONGTYPE cases would pass without the flag, and the ERR-coded PUBSUB test is the only one that currently needs it. Also reword the comment to say "re-raised with a position (`error(e)`)".

src/core/lua-runtime.ts
```diff
@@ -257,23 +286,27 @@ export function luaReplyToRedisValue(value: ReplyValue): RedisValue {
...
-    typeof value !== 'object' ||
-    Array.isArray(value) ||
-    Buffer.isBuffer(value) ||
-    !('err' in value)
-  ) {
+export function renderScriptError(
+  value: ReplyValue,
+  options: {
+    /** Picks the decoration; without one, the 7.0+ suffix form is used. */
+    profile?: CompatibilityProfile
```
fatal10110: [low] `profile` is optional and a missing one silently picks the 7.0+ form. There is a single caller (`runLuaScript`), which always has `ctx.server.profile`, so making `profile` required (and `raisedByRedisCall` required with it) would stop a future caller, e.g. a function/FCALL-specific renderer, from quietly emitting the 7.0 frame on the 6.2 profile. This is a nit.

fatal10110:
## Review of 12df78e: **NOT YET** (1 medium open)

Two lenses: Redis wire-compat and Node.js architecture. The specialist subagents couldn't be spawned because the session hit its concurrent-agent limit, so the coordinator applied both lenses directly. Every Redis claim below was checked against real redis 6.2.24, 7.0.15, 8.0.6 and valkey 7.2.14 / 8.0.11.

|---|---|---|---|
| medium | `src/core/lua-runtime.ts:371` | On 6.2, host-side `redis.call` rejections (unknown command, noscript, no args, and the `command-arg-type` body) get `ERR ` folded in. Real 6.2 renders them through `luaPushError`: `@user_script:1: @user_script: 1: Unknown Redis command called from Lua script`. The CHANGELOG claims the 6.2 decoration for every failing `redis.call`. | Redis |
| low | `src/core/lua-runtime.ts:54` | `raisedByRedisCall()` is a stateful query that has to run between `eval`s. It's correct today (sync eval, single caller) and there's no cross-session leak, but the coupling is implicit. | Arch |
| low | `profile-gates.test.ts:552` | The level-0 re-raise `error(e, 0)` is classified as a `redis.call` abort. The output is correct (identical to real 6.2), but the case isn't tested and the comment overgeneralizes. | Both |
| low | `src/core/lua-runtime.ts:304` | `profile` is optional and silently defaults to the 7.0 form. Make it required. | Arch |

**Verified:** every frame in the new test table and the PUBSUB frame is byte-identical to real 6.2 and to 7.0/8.0/valkey 7.2/valkey 8.0, and valkey 7.2.14 does use the suffix form, so `valkey: '7.2.0'` is correct. The `raisedByRedisCall` heuristic also matches real 6.2 for `pcall(redis.call)` followed by `error(e)`, `error(e,0)`, `error('ERR …',0)` and `error('boom')`, for returning the pcall'd error, for a later failing `redis.call`, and for EVALSHA. The 7.0 path is unchanged. The state is reset at the start of each eval and read synchronously, so it can't leak across scripts or sessions.

**Tests (mock, PR head):** `profile-gates.test.ts` passes on all 7 profiles, and the unit suite passes 462/462. `raw-tcp/command-errors-subcommand.test.ts` fails 10 tests under `REDIS_COMPAT=redis-6.2`. That file was already not profile-aware, so the failures aren't caused by this PR.

**Out-of-scope follow-ups (existed before this PR, not blocking):**
- Compile errors render as a runtime abort on every profile. The mock on 6.2 gives `Error running script (call to f_…): @user_script:1: user_script:1: unexpected symbol near '+'` and on 7.0 gives `… script: <sha>, on @user_script:1.`, while real 6.2, 7.0 and 8.0 all give `ERR Error compiling script (new function): user_script:1: unexpected symbol near '+'`.
- `redis.call` arity: the mock says `ERR wrong number of arguments for 'get' command`, where real 7.0+ says `Wrong number of args calling Redis command from script` and 6.2 says `… From Lua script`.
- `error(nil)` / `error({err=…})` give the engine's `script execution failed ` (with a trailing space). Real 7.0+ gives `ERR nil …` / `WRONGTYPE x script: …`.
- Already documented as out of scope: the `redis.error_reply` `ERR` prefix and the `error('ERR x', 0)` code strip.

src/core/lua-runtime.ts
fatal10110: Addressed in c10b541 (options a, b and c).
(a) The scripting layer's own rejections are now built with a separate `scriptRejection()` helper and never recorded as the last `redis.call` error, so the 6.2 body has no `ERR ` folded in. That covers: no command, unknown command, top-level wrong arity from `plan()`, noscript, and a write from a read-only script. The `command-arg-type` kind already had no code added. These now render as `…: @user_script:1: Unknown Redis command called from script`. The inner `@user_script: N:` position and the 6.2 wording are not modelled: the engine doesn't expose the calling Lua line.
(b) The CHANGELOG and PR body now say the frame is exact for errors a command returns and for Lua runtime errors, and that script-level rejections still differ (#439).
(c) New 6.2-only test 'script-level redis.call rejections on 6.2 (known gap: …)' pins the current frames. Its comment lists the real 6.2 frames, which I re-captured from redis 6.2.

src/core/lua-runtime.ts
fatal10110: Addressed in c10b541. The stateful getter is gone. `evalScript()` returns `{ reply, raisedByRedisCall }`, computing the flag before the `finally`, and the `finally` clears `lastRedisCallError` so no buffers outlive the eval. `eval()` is kept, still returns only the reply, and delegates to it (it's part of the `internal` export and used by `tests/lua-runtime-isolation`). `runLuaScript` now calls `evalScript`.

tests-integration/compatibility/profile-gates.test.ts
fatal10110: Addressed in c10b541. The comment now says `error(e)` (level 1) and points to the level-0 case. I added a 6.2-only test for `pcall(redis.call, 'LPUSH', …)` followed by `error(e, 0)`, pinned to real 6.2.24's `@user_script:2: WRONGTYPE Operation …`. It isn't asserted on 7.0+: real Redis answers `-ERR WRONGTYPE … script: …` there, but the engine splits the leading code off the Lua error string, so this server answers `-WRONGTYPE …`. That's the same engine-side issue as `error('WRONGTYPE x', 0)`, listed as a follow-up.

src/core/lua-runtime.ts
fatal10110: Addressed in c10b541. `renderScriptError` now takes a required `{ profile: CompatibilityProfile; raisedByRedisCall: boolean }`, so there's no silent 7.0 fallback.

src/core/lua-runtime.ts
```diff
@@ -73,15 +118,19 @@ export class RedisLuaRuntime {
...
     let plan: CommandPlan
     try {
       plan = ctx.executor.plan(args[0], args.slice(1))
     } catch (err) {
       if (err instanceof UnknownRedisCommandError) {
-        return redisErrorToLuaReply(new ScriptUnknownCommandError())
+        return scriptRejection(new ScriptUnknownCommandError())
+      }
+
+      if (err instanceof WrongNumberOfArgumentsError) {
```
fatal10110: [low] **This treats a container *subcommand* arity failure as a scripting-layer rejection, but 6.2 only rejects on the top-level arity.** In 6.2, `luaRedisGenericCommand` checks only `cmd->arity`. `XINFO` is `-2`, so `redis.call('xinfo', 'stream')` reaches the container, and the container's own reply is a command error with its code folded in. `executor.plan()` in this server also throws `WrongNumberOfArgumentsError` for subcommand arity, so on the 6.2 profile that case is now marked as a rejection and loses the `ERR` code:
```
EVAL "return redis.call('xinfo', 'stream')" 0
 real 6.2.24 : ERR Error running script (call to f_9092…1f83): @user_script:1: ERR Unknown subcommand or wrong number of arguments for 'stream'. Try XINFO HELP.
 mock 6.2    : ERR Error running script (call to f_9092…1f83): @user_script:1: wrong number of arguments for 'xinfo' command
 real 7.0/8.0: ERR Wrong number of args calling Redis command from script script: 9092…1f83, on @user_script:1.
```
The body was already wrong before c10b541, since it's the #439/#435 container-resolution issue, so this doesn't block. When #439 is picked up, it's worth limiting this branch to the top-level command arity.

src/core/lua-runtime.ts
```diff
@@ -48,21 +56,58 @@ export class RedisLuaRuntime {
     ctx: RedisExecutionContext,
     options?: { readOnly?: boolean },
   ): ReplyValue {
+    return this.evalScript(script, keys, args, ctx, options).reply
```
fatal10110: [low] `eval()` has no callers left. `runLuaScript` is the only user of the runtime and it now calls `evalScript()`, and nothing in `src/`, `tests/` or `tests-integration/` calls `runtime.eval`. Consider deleting it, or folding `evalScript` back into `eval` with the richer return type, so there's one entry point and nobody can get the reply without the flag. This is a nit.

CHANGELOG.md
```diff
@@ -336,6 +336,19 @@ so the PR body is not a durable home for a breaking-change note.
...
+  `-ERR Error running script (call to f_<sha>): @user_script:<line>: <error>`,
+  instead of the 7.0 suffix `<error> script: <sha>, on @user_script:<line>.`
+  ([#442]). As in 6.2 the reply is always `-ERR`: a failing command's own code
+  (`WRONGTYPE ...`) is folded into the body, and a runtime error shows its
+  position twice. Gated as `script.abort-error-suffix` (Redis 7.0 / Valkey 7.2).
+  The frame is exact for errors a *command* returns and for Lua runtime errors;
+  rejections raised by the scripting layer itself (unknown or not-allowed
+  command, wrong arity, no arguments, bad argument type) still differ on 6.2,
+  which words them differently and adds an inner `@user_script: <line>: `
+  position the engine does not expose ([#439]).
```
fatal10110: [low] #439 doesn't really track this gap. It's about container subcommands failing command *lookup* on 7.0+, and it mentions the 6.2 noscript wording only in passing. The rest of the known gap listed here and in the KNOWN GAP test isn't tracked anywhere: 6.2's `luaPushError` wording for unknown command, no arguments, argument type and arity (`… called from Lua script`, `… for redis.call()`, `Lua redis() command arguments …`, `… From Lua script`), plus the inner `@user_script: <line>: ` position, which needs the engine to expose the call line. Suggest opening a dedicated follow-up issue, and pointing this CHANGELOG line and the test comment at it.

fatal10110:
## Re-review of c10b541: **SHIP** (no open high/medium)

Prior findings (from 12df78e):

| sev | finding | status |
|---|---|---|
| medium | script-level rejections on 6.2 got `ERR ` folded in | **Fixed.** These rejections are now kept out of `lastRedisCallError`. On the 6.2 profile, unknown command, noscript, no args, arg type and arity no longer carry `ERR `. The remaining difference from real 6.2 (wording plus the inner `@user_script: N: `) is documented in the CHANGELOG, and a KNOWN GAP test pins today's output. |
| low | stateful `raisedByRedisCall()` getter | **Fixed.** `evalScript()` returns `{ reply, raisedByRedisCall }` and clears the stored error in `finally`. |
| low | level-0 re-raise untested, comment overgeneralized | **Fixed.** There's a new 6.2 `error(e, 0)` test, byte-identical to real 6.2.24. The comment is corrected, and its 7.0+ note (real `-ERR WRONGTYPE …` vs mock `-WRONGTYPE …`) matches 7.0.15 and 8.0.6. |
| low | optional `profile` in `renderScriptError` | **Fixed.** `profile` and `raisedByRedisCall` are now required. |

New findings, all low and posted inline:
- `lua-runtime.ts:132`: a container subcommand arity failure (`redis.call('xinfo','stream')`) is classified as a scripting-layer rejection. Real 6.2 reaches the container, and the body was already wrong because of #439.
- `lua-runtime.ts:59`: `eval()` has no callers left.
- `CHANGELOG.md:350`: #439 doesn't track the 6.2 `luaPushError` wording and inner-position gap. It needs a dedicated issue.

**Tests (mock, c10b541):** `profile-gates.test.ts` passes on all 7 profiles (6.2 passes 19/19). The ioredis, node-redis and raw-tcp script suites pass 45/45 on redis-8.0. Unit tests pass 462/462, and `tsc --noEmit` is clean.

Under `REDIS_COMPAT=redis-6.2`, the client script suites fail 12 tests, against 3 on `origin/main`. They hardcode 7.0+ frames and CI doesn't run them per profile, so this is the expected change and doesn't block.

PR #474, opened 2026-09-23
fix: refuse CLIENT, RESET and QUIT from Lua scripts (#452)

### LINKED ISSUE #452: CLIENT is missing the noscript flag: real Redis refuses it from scripts
Found while re-reviewing PR #430, and verified against real Redis 6.2.24 and 8.0.6. **Pre-existing on `main`.**

## Problem
`CLIENT` has no `noscript` flag (`src/commands/connection.ts`). Real Redis refuses it from scripts:
```
EVAL "return redis.pcall('CLIENT','GETNAME')" 0
  real 6.2 / 8.0: This Redis command is not allowed from scripts
  mock:           nil   (runs the command)
```
#439's acceptance criteria assumed `CLIENT` already has this refusal; that issue's wording has been corrected.

## Acceptance
`CLIENT` (every subcommand) is refused from scripts with real Redis's wording on every profile. Audit the other container commands that real Redis flags `noscript` (`CONFIG`, `ACL`, `SCRIPT`, `SLOWLOG`?) against a real server at the same time.

fatal10110:
## Summary
- `CLIENT` had no `noscript` flag, so `EVAL "return redis.pcall('CLIENT','GETNAME')" 0` ran the command. Real Redis refuses it. `CLIENT` now carries the flag.
- **noscript audit.** I diffed every `noscript` command/subcommand from `COMMAND` on real redis 6.2, 7.0, 7.2 and 8.0 against the mock's registry, then probed with `EVAL` on redis 6.2/7.0/8.0 and valkey 7.2/8.0/9.0. Two more commands were missing the flag:
  - `RESET` is refused on every version.
  - `QUIT` is refused on 7.0+. 6.2 has no `QUIT` table entry, so its scripts get "unknown command" there. The mock's QUIT would otherwise run and close the connection from inside a script.
  - `CONFIG`, `ACL` and `SCRIPT` already had the flag. `SLOWLOG` is correctly not `noscript` on any version.
- **HELP exemption, gated by profile.** From 7.0, `noscript` is set per subcommand, and no container's `HELP` subcommand has it. So `redis.pcall('CLIENT'|'ACL'|'SCRIPT'|'CONFIG'|'FUNCTION','HELP')` returns the help text on 7.0+ (checked on redis 7.0/8.0 and valkey 7.2/8.0/9.0), while 6.2 refuses the whole container. New `FeatureId` `script.per-subcommand-noscript` (redis 7.0 / valkey 7.2), checked in `isRefusedFromScript` in `lua-runtime.ts`. The mock was refusing `ACL/SCRIPT/FUNCTION HELP` on every profile before this change.
- Docs: gate row in `docs/API.md`, CLIENT note in `docs/COMMANDS.md`, one line in `docs/ARCHITECTURE.md`.

Closes #452

### Deferred / not in this PR
- **6.2 refusal wording (#439).** 6.2 answers `@user_script: <line>: This Redis command is not allowed from scripts`. The prefix needs the calling Lua line, and the engine's `redisCall`/`redisPcall` host callback only receives args, so this needs `lua-redis-wasm` support. The mock emits the 7.0+ wording on every profile. The same line-prefix gap affects 6.2's unknown-command and wrong-arity wordings from scripts.
- **Valkey 9.0 wording.** Valkey 9.0 says `This Valkey command is not allowed from script`. All Valkey profiles also say `Unknown command called from script` and `Wrong number of args calling command from script`, with no "Redis". This is a broader Valkey script-wording gap.
- **Commands the mock refuses but real Redis allows (the other side of the audit):**
  - `SPOP`, `SRANDMEMBER` and `HRANDFIELD` are flagged `noscript` in the mock, but real Redis allows them from scripts on 6.2 and 8.0.
  - The blocking commands (`BLPOP`, `BRPOP`, `BLMOVE`, `BLMPOP`, `BZPOPMIN/MAX`, `BZMPOP`, `BRPOPLPUSH`) and `WAIT` are `noscript` only through 7.0. From 7.2 they run from scripts (non-blocking).
- `CONFIG HELP` is not implemented in the mock at all (`ERR unknown subcommand 'HELP'`, even when called directly), so it is left out of the HELP test.

## Test plan
- [x] New `tests-integration/ioredis/scripts-noscript.test.ts` and its node-redis twin (standalone, keyless EVAL). Both fail on the mock before the fix (5/6 each) and pass against real Redis 8.0 (`REDIS_STANDALONE_PORT=6399`).
- [x] Both pass on the mock after the fix.
- [x] `profile-gates.test.ts`: new mock-only test, `<container> HELP` is refused on `redis-6.2` and returns an array on 7.0+. Passes on all 7 profiles.
- [x] `npm run build`, `npm run lint`, `npm test` (462 pass), `npm run test:integration:mock` (1261 pass, 0 fail).
- [ ] Real-backend cluster suites (`ioredis/scripts.test.ts` etc.) could not be re-run: the shared docker cluster is currently `cluster_state:fail` (nodes 30002/30005 are down), so every cluster-based real test times out. That is not caused by this change.


src/core/lua-runtime.ts
```diff
@@ -108,6 +108,35 @@ export class RedisLuaRuntime {
...
+  if (!profile.has('script.per-subcommand-noscript')) {
+    return true
+  }
+
+  const subcommand = plan.rawArgs[0]?.toString().toLowerCase()
+  const helpName = `${definition.name}|help`
+  const hasHelp = definition.introspection?.subcommands?.some(
+    sub => sub.name === helpName,
+  )
+  return !(subcommand === 'help' && hasHelp)
```
fatal10110: [medium] **6.2 branch: host-side `redis.call` rejections get `ERR ` folded in, but real 6.2 renders them through `luaPushError` (inner `@user_script: N: ` position, no code, 6.2 wording).**

`raisedByRedisCall` is true for every error the `redisCall` host callback returns, including the script-level rejections produced in `runRedisCommand` itself (`ScriptUnknownCommandError`, `ScriptNotAllowedCommandError`, `ScriptCallNoCommandError`), so this branch prepends `ERR `. Real 6.2 raises these via `luaPushError`, which adds its own `@user_script: <line>: ` (with a space) and no code. The engine-classified `command-arg-type` body (composed right here in `renderScriptError`) has the same gap.

Real redis 6.2.24 vs this PR's mock on `--compat redis-6.2` (redis-cli, same sha):
```
EVAL "return redis.call('nosuchcmd')" 0
 real 6.2: ERR Error running script (call to f_4f59…dd95): @user_script:1: @user_script: 1: Unknown Redis command called from Lua script
 mock    : ERR Error running script (call to f_4f59…dd95): @user_script:1: ERR Unknown Redis command called from script
EVAL "return redis.call('multi')" 0
 real 6.2: … @user_script:1: @user_script: 1: This Redis command is not allowed from scripts
 mock    : … @user_script:1: ERR This Redis command is not allowed from script
EVAL "return redis.call()" 0
 real 6.2: … @user_script:1: @user_script: 1: Please specify at least one argument for redis.call()
 mock    : … @user_script:1: ERR Please specify at least one argument for this redis lib call
EVAL "return redis.call('get', {})" 0
 real 6.2: … @user_script:1: @user_script: 1: Lua redis() command arguments must be strings or integers
 mock    : … @user_script:1: Lua redis lib command arguments must be strings or integers
```
(7.0.15 / 8.0.6 / valkey 7.2.14 all answer the suffix form with the 7.0 wording, e.g. `ERR Unknown Redis command called from script script: <sha>, on @user_script:1.`, which the mock already matches.)

The CHANGELOG entry says "a failing `redis.call` … now carries Redis 6.2's decoration", which isn't true for this family. Either render them on 6.2 (the host knows these are its own `Script*Error` rejections, so it could tag them instead of recording them as command errors, and the `command-arg-type` case is composed right here), or narrow the CHANGELOG/PR wording and list them as a known remaining 6.2 difference with a follow-up issue.

tests-integration/ioredis/scripts-noscript.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  test('RESET and QUIT are refused from scripts', async () => {
+    for (const command of ['RESET', 'QUIT']) {
+      await assert.rejects(
+        () => redis.eval(`return redis.pcall('${command}')`, 0),
+        errorWithMessage(NOT_ALLOWED),
+        command,
+      )
+    }
+    // Neither ran: the connection is still open and usable.
+    assert.strictEqual(await redis.ping(), 'PONG')
```
fatal10110: [low] **Stateful side channel with temporal coupling.** `raisedByRedisCall(result)` is only correct if it is called after `eval()` returns and before the next `eval()` starts, and `lastRedisCallError` keeps the last error buffers alive until the next script runs. That holds today because `eval` is synchronous and `runLuaScript` calls the method right after it, so there's no cross-script or cross-session leak on the shared runtime. The coupling is still implicit, though. Consider having `eval` return the flag with the reply (e.g. `{ reply, raisedByRedisCall }`), or computing it inside `eval` before the `finally` and clearing `lastRedisCallError` there. That way a future caller can't query stale state.

src/commands/connection.ts
```diff
@@ -528,7 +528,9 @@ export const pingCommand = defineCommand({
 export const quitCommand = defineCommand({
   name: 'quit',
   schema: t.object({}),
-  flags: ['readonly', 'fast', 'subscribed'],
+  // noscript: real 7.0+ refuses QUIT from a script (6.2 has no QUIT command
+  // entry at all, so its script sees an unknown command instead).
+  flags: ['readonly', 'fast', 'subscribed', 'noscript'],
```
fatal10110: [low] On the `redis-6.2` profile, QUIT from a script now returns the noscript refusal. Real 6.2 has no QUIT table entry and answers with the unknown-command error:

```
redis 6.2.24: EVAL "return redis.pcall('QUIT')" 0
@user_script: 1: Unknown Redis command called from Lua script
mock redis-6.2 (this PR):
ERR This Redis command is not allowed from script
```

The `@user_script: 1:` prefix belongs to #439. The error *class* (unknown vs not-allowed) is a separate difference, and it could be gated cheaply, e.g. return `ScriptUnknownCommandError` for `quit` when `!profile.has('script.per-subcommand-noscript')`. Fine to defer, but then please record it in the #439 follow-up and don't leave it only in the PR description.

docs/COMMANDS.md
```diff
@@ -42,6 +42,12 @@ shapes; see the gate matrix in [Compatibility Profiles](API.md#compatibility-pro
 - [x] `CLIENT KILL [ID client-id] [MAXAGE seconds] [SKIPME YES|NO]` - Close matching client connections; `MAXAGE` is accepted for Redis 7.4+ / Valkey 9.0+ profiles
 - [x] `CLIENT NO-EVICT ON|OFF` - Toggle the current connection's no-eviction flag
 - [x] `CLIENT HELP` - Return subcommand help
+
+`CLIENT` is flagged `noscript`: every subcommand is rejected from Lua with
```
fatal10110: [low] This paragraph splits the CLIENT checklist in two: the `- [ ] CLIENT PAUSE`/`GETREDIR`/`TRACKINGINFO` bullets below now render as a separate list after the prose. Suggest moving the paragraph after the last CLIENT bullet.

docs/API.md
```diff
@@ -160,6 +160,7 @@ Current profile gates:
 | Cluster `SELECT` for non-zero databases | unsupported | `valkey-9.0` |
 | `SORT`/`SORT_RO` cluster `BY`/`GET` patterns compared by slot (and the longer `...may be in different slots.` error wording) instead of refused outright | `redis-7.4+` | `valkey-8.0+` |
 | `SORT`/`SORT_RO` cluster `GET '#'` exempt from that slot comparison | `redis-7.4+` (7.4.2) | `valkey-9.0+` (8.0.2) |
+| `<container> HELP` callable from a Lua script although the container (`CLIENT`, `ACL`, `SCRIPT`, `CONFIG`, `FUNCTION`) is `noscript` — 6.2 refuses every subcommand | `redis-7.0+` | `valkey-8.0+` |
```
fatal10110: [low] This row lists `CONFIG`, but the mock has no `CONFIG HELP`. On 7.0+ profiles `redis.pcall('CONFIG','HELP')` returns `ERR unknown subcommand 'HELP'. Try CONFIG HELP.` (mock redis-7.0 / valkey-9.0), whereas real 7.0/7.2/8.0/valkey return the help array. The exemption does apply to CONFIG, but the row reads as if the call works. Suggest a short note ("`CONFIG HELP` itself is not implemented yet") or dropping CONFIG from the list until it is. Everything else in the row checks out: I verified every other subcommand of CLIENT/ACL/SCRIPT/CONFIG/FUNCTION is `noscript` on real 7.0.15 and 8.0.6 via `COMMAND INFO`, with only `<container>|help` lacking it.

fatal10110:
## Review of 2fed88b: **NOT YET** (1 medium, 4 low)

Reviewed from two angles: fidelity to real Redis, checked against live redis 6.2.24 / 7.0.15 / 7.2.1 / 8.0.6 and valkey 8.0.11 / 9.0.6, and code/test quality.

| sev | where | finding |
|---|---|---|
| medium | `src/core/lua-runtime.ts:137` | On 7.0+, `redis.pcall('CLIENT','NOPE')` returns the noscript refusal. Real 7.0+ returns `ERR Unknown Redis command called from script` (valkey: `ERR Unknown command called from script`). |
| low | `tests-integration/*/scripts-noscript.test.ts` (ping after RESET/QUIT) | The ping passes before the fix too: the mock's script QUIT never closed the socket. The observable side effect is RESET clearing the client name, so assert on `CLIENT GETNAME`. |
| low | `src/commands/connection.ts:533` | On the redis-6.2 profile, QUIT from a script is refused with "not allowed". Real 6.2 answers with the unknown-command error. |
| low | `docs/COMMANDS.md:46` | The paragraph splits the CLIENT checklist. |
| low | `docs/API.md:163` | The row lists CONFIG, but the mock has no `CONFIG HELP`. |

**Verified correct:**
- Every CLIENT subcommand is refused on all versions.
- RESET is refused on all versions.
- QUIT is refused on 7.0+.
- `<container> HELP` runs from a script on 7.0+ and valkey, and is refused on 6.2. On real 7.0 and 8.0, `COMMAND INFO` shows every CLIENT/ACL/SCRIPT/CONFIG/FUNCTION subcommand except `|help` is `noscript`.
- The decorated `redis.call` error matches.
- The gate value is right.

**Tests:**
- Mock, all 7 profiles: the new tests plus profile-gates are green (27 pass).
- Real servers: the new ioredis and node-redis suites pass against real redis 7.0, redis 8.0 and valkey 8.0 (12/12 each).
- Red check: 10 of 12 fail against `origin/main` src.
- Existing script suites (ioredis, node-redis, unit) are green.

**Out-of-scope follow-ups:**
- Script arity wording (`ERR wrong number of arguments for 'client|help' command` vs `ERR Wrong number of args calling Redis command from script`; the same gap exists for `GET`).
- `CONFIG HELP` is not implemented.
- Mock `COMMAND INFO` subcommand entries (e.g. `client|getname`) lack `noscript`.
- On the 6.2 profile, `COMMAND INFO quit` returns an entry where real 6.2 returns nil.

src/core/lua-runtime.ts
fatal10110: Addressed in b78421e. Note: this comment's body (`raisedByRedisCall`, CHANGELOG) seems to belong to the #442 script-error PR; nothing in this PR has that code. I fixed the finding the coordinator relayed for this line instead.

**Fix:** on 7.0+ profiles, `noscriptRefusal` now looks up `container|subcommand` in the definition's `introspection.subcommands`, filtered by the profile's subcommand gates (the map moved to `core/compatibility/subcommand-gates.ts`, shared with `COMMAND`). The outcomes:
- A missing entry returns `ScriptUnknownCommandError` (`ERR Unknown Redis command called from script`). This covers `CLIENT/ACL/CONFIG/SCRIPT/FUNCTION NOPE`, and `CLIENT SETINFO` on `redis-7.0`, which I verified is unknown on real 7.0.15.
- `HELP` runs. Everything else is refused.
- 6.2 still refuses the whole container, as real 6.2 does.

**Valkey wording:** Valkey says `Unknown command called from script`. There's no existing helper for that, so it isn't modelled; it's noted in the code and the PR.

**Tests:** new ioredis and node-redis cases for `<container> NOPE`, green against real 8.0 and red against the previous commit's runtime. `profile-gates.test.ts` covers the per-profile split.

tests-integration/ioredis/scripts-noscript.test.ts
fatal10110: Addressed in b78421e. This comment's body (the `raisedByRedisCall` side channel) seems to belong to the #442 PR; this file has no such code. I fixed the finding the coordinator relayed for this line: the PING after the RESET/QUIT loop proved nothing.

Both suites now set a client name first and assert `CLIENT GETNAME` still returns it after the refused `RESET`/`QUIT`. On main, RESET from a script would have cleared it. I also corrected the PR description: on main, QUIT from a script replied `+OK` and left the socket open, not closed.

src/commands/connection.ts
fatal10110: Gated in b78421e: a new `command.quit-table-entry` FeatureId (redis 7.0 / valkey 7.2). On `redis-6.2`, a script calling `QUIT` now gets `ScriptUnknownCommandError`, the same error class as real 6.2. The `@user_script: 1:` prefix and 6.2's `from Lua script` wording are still #439's line-number gap. `profile-gates.test.ts` asserts the per-profile split.

docs/COMMANDS.md
fatal10110: Fixed in b78421e. The paragraph now follows the last CLIENT bullet (`CLIENT TRACKINGINFO`), so the checklist renders as one list. I also extended it to cover the 7.0+ unknown-subcommand lookup and QUIT's 6.2 behaviour.

docs/API.md
fatal10110: Qualified in b78421e. The row now covers the whole 7.0+ lookup (HELP runs, unknown subcommand gives the unknown-command error) and notes that the mock has no `CONFIG HELP` yet, so that one call still errors. I also added a separate row for the `QUIT` gate.

src/core/lua-runtime.ts
```diff
@@ -108,6 +110,59 @@ export class RedisLuaRuntime {
...
+  if (!subcommands || plan.rawArgs.length === 0) {
+    return new ScriptNotAllowedCommandError()
+  }
+
+  const name = `${definition.name}|${plan.rawArgs[0].toString().toLowerCase()}`
+  const known = subcommands.some(
+    sub => sub.name === name && subcommandSupported(name, profile),
+  )
+  if (!known) {
+    return new ScriptUnknownCommandError()
```
fatal10110: [medium] **New regression from the lookup fix.** On 7.0+ the lookup checks the name against the mock's `introspection.subcommands`, which lists only the subcommands the mock *implements*. Real subcommands the mock has not implemented yet therefore now fail lookup as unknown commands. Real Redis refuses them with `noscript`. At 2fed88b, and on `main`, the mock returned the refusal for these, which was correct.

`EVAL "return redis.pcall(<args>)" 0`:

| args | real redis 7.0.15 / 8.0.6, valkey 8.0.11 | mock redis-7.0 / redis-8.0 / valkey-8.0 @ b78421e |
|---|---|---|
| `'ACL','CAT'` | `ERR This Redis command is not allowed from script` | `ERR Unknown Redis command called from script` |
| `'ACL','LOG'` | `ERR This Redis command is not allowed from script` | `ERR Unknown Redis command called from script` |
| `'CLIENT','PAUSE','1'` | `ERR This Redis command is not allowed from script` | `ERR Unknown Redis command called from script` |
| `'CLIENT','TRACKINGINFO'` | `ERR This Redis command is not allowed from script` | `ERR Unknown Redis command called from script` |

The full diff against real 8.0 `COMMAND INFO` gives 18 affected subcommands:
- ACL: `cat`, `deluser`, `genpass`, `getuser`, `list`, `load`, `log`, `save`, `setuser`, `users`
- CLIENT: `caching`, `getredir`, `no-touch`, `pause`, `reply`, `tracking`, `trackinginfo`, `unblock`, `unpause`

This fixes `NOPE` but regresses many more real names. (My earlier suggestion to use `introspection.subcommands` caused this, sorry.) The check needs the *real* table's names, not the implemented ones. One way is a per-container list of real subcommand names in `subcommand-gates.ts`, reusing the existing gates like `client|setinfo`. Anything in that list but not implemented would get the `noscript` refusal, and only names outside it would get the unknown error. Please add a test for one unimplemented-but-real subcommand, e.g. `ACL CAT` or `CLIENT PAUSE`.

src/core/lua-runtime.ts
fatal10110: [medium] *Reposted. The body first posted here belonged to another PR's review, a mix-up in shared drafts, and was deleted. b78421e has since addressed this finding; see the re-review summary.*

On 7.0+ profiles an unknown `CLIENT` subcommand called from a script got the `noscript` refusal. Real 7.0+ answers `ERR Unknown Redis command called from script` because it resolves `container|subcommand` first, and valkey answers `ERR Unknown command called from script`. 6.2 refuses the whole container. Checked with `EVAL "return redis.pcall('CLIENT','NOPE')" 0` on redis 6.2.24, 7.0.15, 7.2.1 and 8.0.6, and valkey 8.0.11 and 9.0.6.

tests-integration/ioredis/scripts-noscript.test.ts
fatal10110: [low] *Reposted. The body first posted here belonged to another PR's review, a mix-up in shared drafts, and was deleted. b78421e has since addressed this finding.*

The `PING` after the RESET/QUIT loop proved nothing. On a mock built from `origin/main` src, the script `QUIT` replied `+OK` and left the socket open, because the Lua runtime drops `close: true`. The script `RESET` silently cleared the client name (`CLIENT GETNAME` came back `$-1`). The fix is to assert `CLIENT GETNAME` instead. The same applies to the node-redis twin.

fatal10110:
## Re-review of b78421e: **NOT YET** (1 new medium)

Housekeeping: two of my first-round comments, at `lua-runtime.ts:137` and `ioredis/scripts-noscript.test.ts:83`, had been posted with another PR's review text (a mix-up in shared draft files). I deleted them and reposted the correct #474 text, marked as reposted.

**Prior findings**

| sev | finding | status |
|---|---|---|
| medium | unknown `CLIENT` subcommand from a script on 7.0+ | **fixed.** `CLIENT NOPE` → `Unknown Redis command called from script`; `CLIENT SETINFO` on redis-7.0 → unknown, as on real 7.0.15 |
| low | the PING after RESET/QUIT proved nothing | **fixed.** The test now asserts `CLIENT GETNAME`, in both twins |
| low | 6.2 QUIT from a script | **fixed.** New `command.quit-table-entry` gate; mock redis-6.2 → `ERR Unknown Redis command called from script` |
| low | COMMANDS.md split the checklist | **fixed** |
| low | API.md CONFIG HELP | **fixed** (qualified) |

**New finding**
- medium, `src/core/lua-runtime.ts:158`. The lookup uses the mock's *implemented* subcommand list. So 18 real ACL/CLIENT subcommands the mock doesn't implement (e.g. `ACL CAT`, `ACL LOG`, `CLIENT PAUSE`, `CLIENT TRACKINGINFO`) now return `Unknown Redis command called from script` from a script. Real 7.0, 8.0 and valkey 8.0 return `This Redis command is not allowed from script` for all of them. This is a regression against both 2fed88b and `main`. Details inline.

**Tests at b78421e**
- Mock, noscript + scripts + compat profile-gates suites:
  - redis-7.0, 7.2, 7.4 and 8.0: 68 pass each.
  - valkey-8.0 and 9.0: 66 pass each.
  - redis-6.2: 3 failures in `scripts.test.ts`. They fail the same way on `origin/main`, so they come from `main`, not this PR.
- Unit tests: 462 pass. tsc and eslint are clean.
- Real servers, new suites: redis 7.0 and redis 8.0 pass 14/14; valkey 8.0 passes 12/12 with 2 skips.

**Out-of-scope follow-ups**
- Script arity is checked after `noscript`. Real 7.0+ checks it before: `CLIENT GETNAME x` → `Wrong number of args calling Redis command from script`, where the mock gives the refusal.
- 6.2 and valkey script wordings (#439).
- `CONFIG HELP` is not implemented.

src/core/lua-runtime.ts
fatal10110: Fixed in 6418e2e. The lookup now checks a table of the **real** subcommands of each `noscript` container, not the mock's implemented subset. The table lives in `NOSCRIPT_CONTAINER_SUBCOMMANDS` in `src/core/compatibility/subcommand-gates.ts`, with a version gate per subcommand where profiles differ. I captured it from `COMMAND INFO` on redis 7.0.15, 7.2.16, 7.4.11, 8.0.6 and valkey 7.2.14, 8.0.11, 9.0.6. The per-profile differences:
- `CLIENT NO-TOUCH` and `CLIENT SETINFO` start at redis 7.2 / valkey 7.2.
- `CLIENT CAPA` and `SCRIPT SHOW` are valkey 8.0+ only.
- `CLIENT IMPORT-SOURCE` is valkey 9.0+ only.

Only a name missing from that table (for that profile) gets `Unknown Redis command called from script`. Real-but-unimplemented subcommands such as `ACL CAT` and `CLIENT PAUSE` are refused as not allowed again. The `COMMAND` subcommand map move from b78421e is reverted, since it's no longer needed.

**Tests**
- New ioredis and node-redis cases: `ACL CAT/LOG/USERS` and `CLIENT PAUSE/TRACKINGINFO/GETREDIR` from a script give not-allowed. They fail on b78421e and pass on the mock and on real 8.0.
- `profile-gates.test.ts` covers the per-profile split on all 7 profiles: `ACL CAT`, `CLIENT PAUSE`, `CLIENT NO-TOUCH` (unknown on `redis-7.0`), `CLIENT CAPA` and `SCRIPT SHOW` (unknown on redis 7.0+), `CLIENT IMPORT-SOURCE` (unknown except on `valkey-9.0`), plus `CLIENT NOPE`. I checked every expectation against the real servers above.

src/core/compatibility/subcommand-gates.ts
```diff
@@ -0,0 +1,112 @@
...
+    id: ALWAYS,
+    'import-source': { valkey: '9.0.0' },
+    info: ALWAYS,
+    kill: ALWAYS,
+    list: ALWAYS,
+    'no-evict': ALWAYS,
+    'no-touch': REDIS_72,
+    pause: ALWAYS,
+    reply: ALWAYS,
+    setinfo: REDIS_72,
```
fatal10110: [low] The version facts for `client|setinfo` (7.2), `client|no-evict` and `acl|dryrun` (7.0 / valkey 7.2) now exist twice: here as `REDIS_72`/`ALWAYS`, and in `FEATURE_GATES` as `client.setinfo`, `client.no-evict` and `acl.dryrun`. `COMMAND INFO` and direct calls use the `FEATURE_GATES` copy. If one of them is later corrected, `COMMAND INFO client` and a script's lookup would disagree about whether the subcommand exists on that profile.

The two copies agree today, so this is not blocking. Suggestion: reference the existing gates, e.g. `setinfo: FEATURE_GATES['client.setinfo']`, or add a unit test asserting they match.

I diffed this table against `COMMAND INFO` on real redis 7.0.15, 7.2.4, 7.4 and 8.0.6, and valkey 8.0.0, 8.0.11, 9.0.0 and 9.0.6. It has 0 mismatches, including a `nope` probe.

fatal10110:
## Re-review of 6418e2e: **SHIP** (1 new low, no open high/medium)

**Prior findings**
- medium, `lua-runtime.ts`: real-but-unimplemented subcommands failed lookup as unknown. **Fixed.** I compared every real CLIENT/ACL/CONFIG/SCRIPT/FUNCTION subcommand from a script against real servers, mock vs real. Profiles: redis-7.0 vs redis 7.0.15, redis-8.0 vs redis 8.0.6, and valkey-9.0 vs valkey 9.0.6. The only differences left are:
  - `CONFIG HELP`, which the mock doesn't implement (documented);
  - the arity-before-noscript ordering (out of scope);
  - valkey 9.0's `Unknown command` wording (out of scope).
- All earlier low findings remain fixed.

**Real-subcommand table.** I diffed `NOSCRIPT_CONTAINER_SUBCOMMANDS` against `COMMAND INFO` on real redis 7.0.15, 7.2.4, 7.4 and 8.0.6, and valkey 8.0.0, 8.0.11, 9.0.0 and 9.0.6. It has **0 mismatches**. The version gates are right:
- `capa` and `script|show` exist from valkey 8.0.0;
- `import-source` exists from valkey 9.0.0;
- `no-touch` and `setinfo` exist from redis 7.2.
The `COMMAND` subcommand-list move is fully reverted; `command.ts` has no net diff against `main`.

**New finding**
- low, `subcommand-gates.ts:57`: the setinfo/no-evict/dryrun version facts are duplicated with `FEATURE_GATES` and could drift apart.

**Tests at 6418e2e**
- Mock, noscript + scripts + compat profile-gates + COMMAND suites:
  - redis-7.0, 7.2, 7.4 and 8.0: 78 pass each.
  - valkey-8.0 and 9.0: 76 pass each.
  - redis-6.2: 9 failures in `scripts.test.ts` and `command-integration.test.ts`. They fail the same way on `origin/main`, so they come from `main`, not this PR.
- Unit tests: 462 pass. tsc, eslint and prettier are clean.
- Real servers, new suites: redis 7.0, 7.2 and 8.0 pass 16/16; valkey 8.0 passes 14/14 with 2 skips.
- Valkey 9.0 fails only on its `This Valkey command` wording. The PR defers that wording gap.

Issue #461, opened 2026-09-23
Docs/comment inaccuracies left after wave-1 PRs


fatal10110:
Small documentation and comment inaccuracies left by the wave-1 PRs after their final reviews. None of them blocked a merge.

- [ ] **#428** `docs/TEST-INTEGRATION.md:169` says the init script waits for nodes to answer `PING`. It now only sends `INFO cluster`.
- [ ] **#431** `docs/COMMANDS.md:139` says reads past a raised `proto-max-bulk-len` return the bit-offset error. They return `0`, like Redis. The CHANGELOG already has the right wording.
- [ ] **#424** `socket-connection-transport.ts:131` and `CHANGELOG.md:104`: "the shipped paths do not hit this" is false for TCP (see #459).
- [ ] **#424** `session-adapter.ts:71`: the comment says the final `close()` is a no-op when the client ends first. Since the round-7 deferral it performs the actual half-close, and removing it would bring back the `SUBSCRIBE a b c` + `end()` bug.
- [ ] **#422** CHANGELOG and the `tagNotifyCommand` docblock say the parked-command tag leak is "tracked separately" / "as a follow-up". Link #444.
- [ ] **#425** the ALPHA-tie note says tied elements "keep the load order". That's only true for ascending sorts; under DESC the mock reverses them (see #443).
- [ ] **#429** the `utils.ts` docstring and the PR description still cite the flush-async-sync suites as exact-DBSIZE coverage.

Do this after the PRs are merged, as one small docs PR.

PR #475, opened 2026-09-23
refactor: remove ResponseStream; deliver SUBSCRIBE/MONITOR frames via the session push queue (#366)

### LINKED ISSUE #366: refactor: remove ResponseStream; deliver SUBSCRIBE/MONITOR frames via the session push queue

### LINKED ISSUE #455: Multi-channel SUBSCRIBE confirmations can interleave with a pipelined command's reply
Found while re-reviewing PR #424, verified against real Redis 7.2.1. **Pre-existing on `main`.**

## Problem
The confirmation frames for a multi-channel `SUBSCRIBE a b c` are sent from a background drain (a `ResponseStream`). If the client sends another command in the same write, that command's reply can be **interleaved between the confirmations**. Real Redis sends all confirmations first.

## Acceptance
Confirmation frames and a pipelined command's reply arrive in real Redis's order, pinned by a raw-TCP test. #366 (which removes `ResponseStream` and delivers these frames through the session push queue) likely fixes this; verify it there.

fatal10110:
## Summary

This removes `ResponseStream`. Server-initiated frames now go through one channel: the session push queue.

- **SUBSCRIBE family** (option (a) from the issue). A multi-target `(UN)SUBSCRIBE` / `P…` / `S…` returns a single `RedisResult.preEncoded(...)` holding every confirmation frame, encoded for the session's RESP version. The adapter writes it in one piece, so a command pipelined behind it can no longer land between confirmations (#455). `EXEC` now pre-encodes whenever an item carries its own bytes, so a queued `SUBSCRIBE a b` is embedded in the `EXEC` array. That matches real Redis byte for byte (checked on Redis 6.2, 7.0 and 8.0 and on Valkey 7.2). Before, it replied `Streaming command is not allowed in transaction`. `value` holds the first frame for callers that read values.
- **MONITOR** replies `+OK`. `ClientSession.startMonitor()` subscribes to the feed and enqueues lines as pushes. The lines are held with `deferPushesUntilAfterReply()` until `+OK` is written. `close()` and `RESET` call `stopMonitor()`, which replaces `registerResponseStreamCleanup` / `resetResponseStreams`. I checked two cases against real Redis and matched them:
  - `MONITOR` inside `MULTI` now gives `-ERR MONITOR isn't allowed for DENY BLOCKING client`. This is the same text on every version I checked, so it is not gated.
  - A repeated `MONITOR` gets no reply and does not double the feed.
- **Removed:**
  - `response-stream.ts`, `isResponseStream` and `ExecutorResult`
  - the executor's stream branches (async path, `ensureNonThenableStream`, and the Lua `assertSyncCommandResult` stream branch)
  - the transaction stream-close branch
  - the adapter's `spawnStreamDrain` / `drainStream` / `activeStreams`
  - the two-iterator merge in `InMemoryRedisClient.pushes()`, which is now a thin wrapper over `session.readPushes`
  - the node-redis mock's `drainSubscribeAck`
- `InMemoryRedisClient.command()` now runs `afterReply`, as the network path does. MONITOR's held-back lines need this.
- #461 (only the `session-adapter.ts` item): rewrote the stale comment on the final `transport.close()`. It is the real half-close whichever end finishes first, and it still delivers `SUBSCRIBE a b c` + EOF. I'm leaving #461 open for its other items.
- **BREAKING (`/core`)**: CHANGELOG entry added under Removed, with a migration table; fixes are listed under Fixed. AGENTS.md, ARCHITECTURE.md and COMMANDS.md are updated.

Net diff: +425 / −579 (in `src/`: +136 / −416).

One trade-off: an in-memory client that sends a multi-target `SUBSCRIBE` gets the first confirmation from `command()`. The rest no longer show up in `pushes()`, which now carries only messages and MONITOR lines. The CHANGELOG documents this.

Closes #366
Closes #455

## Test plan
- [x] New raw-TCP tests in `tests-integration/raw-tcp/pubsub-protocol.test.ts`, each run in RESP2 and RESP3:
  - multi-target SUBSCRIBE/PSUBSCRIBE/SSUBSCRIBE confirmations all arrive before a pipelined `PING`, byte for byte (#455)
  - `EXEC` embeds every confirmation of a queued multi-channel SUBSCRIBE
- [x] New raw-TCP tests in `monitor-protocol.test.ts`: MONITOR inside MULTI, a repeated MONITOR (no reply, no duplicate lines), and `+OK` arriving before any feed line while another client is busy
- [x] The new tests failed on the mock before the fix: #455 interleaving reproduced, EXEC/MONITOR-in-MULTI/double MONITOR all red. They pass against real Redis 8 (standalone :6399).
- [x] After the fix: unit tests 459/459; full mock integration run 1255 pass, 0 fail (2 skipped); raw pubsub + monitor pass on mock and real
- [x] `npm run build` and `npm run lint` are clean


CHANGELOG.md
```diff
@@ -105,6 +105,31 @@ so the PR body is not a durable home for a breaking-change note.
...
+  ```
+  SUBSCRIBE a b c (any (UN)SUBSCRIBE family) -> one RedisResult pre-encoded
+                                                with every confirmation frame;
+                                                `value` is the first frame
+  MONITOR                                    -> +OK, then feed lines as session
+                                                pushes (ClientSession.startMonitor)
+  ExecutorResult                             -> RedisResult
+  ```
+
+  A custom command that returned a stream should enqueue its frames as pushes
```
fatal10110: [medium] The migration advice can't be followed by a custom command as written, and the teardown hook it relies on is gone without a replacement.

- A command only receives `ctx.session: RedisClientSession` (src/core/redis-context.ts). That interface has `deferPushesUntilAfterReply` but **not** `enqueuePush`, so "enqueue its frames as pushes (`ClientSession.enqueuePush` …)" only works with a cast to the concrete `ClientSession`. `deferPushesUntilAfterReply` is not usable on its own either, because nothing on the interface lets you enqueue.
- `registerResponseStreamCleanup` was the only general way for a long-lived producer to be torn down on `RESET` or disconnect. After this change, only MONITOR has that (`startMonitor`/`stopMonitor`). A third-party command that subscribes to a feed and pushes frames now leaks its subscription across `RESET`, and across close unless it hooks `ctx.signal` itself.

Fix, either way:
(a) add `enqueuePush` plus a generic cleanup registration (e.g. `onReset(cleanup)`) to `RedisClientSession`, called from both `RESET` and `close()`; or
(b) say plainly in this entry that custom long-lived push producers are no longer supported through the public interface, and drop the `enqueuePush` pointer.

src/in-memory-client.ts
```diff
@@ -111,9 +101,11 @@ export class InMemoryRedisClient {
    * to its native reply. Throws a {@link RedisCommandError} for `-ERR` replies,
    * mirroring what a real client surfaces.
    *
-   * Streaming commands (MONITOR / SUBSCRIBE / …) resolve to their *immediate*
+   * Push-mode commands (MONITOR / SUBSCRIBE / …) resolve to their *immediate*
    * reply (MONITOR's `OK`, the subscribe confirmation); their server-initiated
-   * frames are delivered through {@link pushes}.
+   * frames are delivered through {@link pushes}. A multi-target (UN)SUBSCRIBE
+   * resolves to its first confirmation — on the wire the rest follow it in the
```
fatal10110: [medium] This is a behavior regression in the **root** public API, and the CHANGELOG (L128) files it under `BREAKING (/core)`.

`InMemoryRedisClient` is exported from `src/index.ts` (the root entry), not only `/core`. Verified on this head:

```
inmem SUBSCRIBE a b -> ["subscribe","a",1]
inmem pushes()      -> [["message","b","x"], <timeout>]   // ["subscribe","b",2] is gone
```

On main, `pushes()` yielded `["subscribe","b",2]` before any message. A consumer that waits for N confirmations in `pushes()`, which is the natural way to know a multi-channel subscribe has settled, now hangs. The value returned is also the *first* confirmation, whose count (`1`) under-reports the subscription.

Suggested fix: keep the information. For example, carry the per-frame values on the multi-target result (an optional field next to `encoded`) and have `command()` either return all of them or enqueue frames 2..N into the session push queue. Either restores the main behavior at low cost. If you keep the trade-off, at least move the note to a root-level BREAKING entry, not `/core`, and say the returned count is the first channel's.

tests-integration/raw-tcp/monitor-protocol.test.ts
```diff
@@ -353,6 +353,81 @@ describe(`Raw TCP MONITOR protocol (${testRunner.getBackendName()})`, () => {
...
+    await actor.readRawFrame()
+
+    const lines = await collectMonitorLines(monitor, `${marker}:`, 2)
+    assert.deepStrictEqual(
+      lines.map(line => line.slice(line.indexOf('"PING"'))),
+      [`"PING" "${marker}:1"`, `"PING" "${marker}:2"`],
+    )
+  })
+
+  test('MONITOR replies +OK before any monitor line', async () => {
```
fatal10110: [medium] This test passes with the mechanism it guards removed, so the deferral in `monitor.ts` has no test coverage.

Mutation check on this head: I replaced `const flushPushes = ctx.session.deferPushesUntilAfterReply()` (src/commands/monitor.ts:49) with `const flushPushes = () => {}` and ran this test on the mock 3 times: `# pass 1 / # fail 0` each time. The adapter appends the `+OK` write to `writeChain` in the continuation right after `execute` returns. That is before any other session's command can finish and publish a feed event, so the race this test aims at never opens over TCP.

Suggest a deterministic session-level test instead. Run `monitor` on session A, then, **before** invoking `result.options.afterReply`, execute a command on session B. Assert that A's push queue is still empty (for example, `readPushes` with a short signal yields nothing). Then call `afterReply()` and assert the line arrives. Keep this raw-TCP test as an end-to-end guard if you like, but don't cite it as proof of the ordering guarantee.

src/core/transports/resp2/session-adapter.ts
```diff
@@ -83,67 +79,25 @@ export class Resp2SessionAdapter {
...
-        // we return.
-        await Promise.allSettled(this.activeStreams)
       } finally {
-        // The read loop is over, so close our side as Redis does. If the
-        // client ended first, the transport has already torn down (before the
-        // drains above, so they cannot wait on a client that stopped reading)
-        // and this is a no-op. Otherwise the server ended first and this is
-        // the half-close. Last, so pending drains still get to write; in a
-        // finally, so a throw from session.close() cannot skip it.
+        // The read loop is over, so close our side as Redis does: a half-close
```
fatal10110: [low] Stale comments left behind by the drain removal (out of the diff, anchored here). `src/core/transports/socket-connection-transport.ts` `dropOnPeerEof` docs (~L117-131) still refer to "the adapter's pending drains" and "(not a background drain)". `tests/core/virtual-connection.test.ts:438-440` says the EOF teardown "must not cut the background drain off after the first". No background drain exists after this PR: the confirmations are one pre-encoded reply. Reword them to match this comment.

src/client-mocks/node-redis-mock.ts
```diff
@@ -1308,8 +1307,8 @@ export class NodeRedisMockCluster extends CommandRunner {
 /**
  * Execute a tokenised command on a session and return its raw {@link RedisValue}
  * (callers decode it to the node-redis-correct shape). Throws when the client is
- * closed, and rejects streaming commands the facade can't deliver as a reply
- * (pub/sub uses the dedicated push-draining session instead).
+ * closed. A multi-target (UN)SUBSCRIBE's value is its first confirmation;
```
fatal10110: [low] `runOnSession` ignores `result.options.afterReply`, while `InMemoryRedisClient.command()` now runs it. `sendCommand(['MONITOR'])` on the node-redis mock now resolves `"OK"` (verified; on main it hung forever in `drainSubscribeAck` over an endless stream). But the main session's `deferredPushes` is never released, so every later feed line piles up in an unreachable array for the client's lifetime. `CLIENT KILL` on self (`afterReply: disconnect`) is skipped the same way. This is not a regression, but for parity either call `result.options?.afterReply?.()` here or reject MONITOR on the facade (nothing reads the main session's pushes anyway).

fatal10110:
## Review of 379270e: **NOT YET**

Reviewed through two lenses: Redis wire compatibility (checked against real Redis) and Node.js architecture. The wire behavior matches Redis everywhere I tested. What is left is two API and docs problems and one test gap.

### Checked against real Redis
I compared the mock with real Redis 7.2.1, 6.2.24 and 8.0.6, byte for byte, in RESP2 and RESP3:

| Scenario | Result |
|---|---|
| Multi-target SUBSCRIBE / UNSUBSCRIBE (no args), then a pipelined PING | matches; unsubscribe-all order differs only by dict order, which is already documented as not reproducible |
| `MULTI; SUBSCRIBE a b; PING; EXEC; PING` | the EXEC array header is `*2` and embeds 3 frames, the same as Redis |
| `MULTI; SUBSCRIBE a b; UNSUBSCRIBE a b; EXEC` and `MULTI; UNSUBSCRIBE; EXEC` | matches |
| `MULTI; MONITOR; EXEC` | `*1\r\n-ERR MONITOR isn't allowed for DENY BLOCKING client` on every version |
| MONITOR, then RESET | `+OK`, the feed line, `+RESET`, then no more lines |
| RESP3 MONITOR | lines go out as `+` simple strings, not `>` push frames |
| A repeated MONITOR | no reply |
| Lua `redis.call('subscribe'/'monitor')` | still rejected, via the `noscript` flag |
| A PUBLISH from another client right after a multi-channel SUBSCRIBE | the message arrives after every confirmation |

The PR's new raw tests (8) pass against real 7.2.1 and 8.0.6. On 6.2 the MONITOR tests pass; the pubsub file times out there, most likely because 6.2 has no SSUBSCRIBE (I did not pin down the cause).

### Findings (inline)
| Severity | Where | What |
|---|---|---|
| medium | CHANGELOG.md:122 | The migration advice points to `enqueuePush`, which is not on `RedisClientSession`. With `registerResponseStreamCleanup` gone, a custom producer has no hook for teardown on RESET or close. |
| medium | src/in-memory-client.ts:107 | A multi-target SUBSCRIBE now silently drops confirmations 2..N. `InMemoryRedisClient` is a root export, but the change is filed under `BREAKING (/core)`. It is a regression, not just a trade-off. |
| medium | monitor-protocol.test.ts:410 | The "+OK before feed" test still passes when `deferPushesUntilAfterReply` is replaced by a no-op, so the deferral has no test. |
| low | session-adapter.ts:83 | Comments in socket-connection-transport.ts and virtual-connection.test.ts still talk about "background drain" and "pending drains". |
| low | node-redis-mock.ts:1310 | `runOnSession` does not run `afterReply`. `sendCommand(['MONITOR'])` now returns OK, but its lines pile up in a buffer nothing reads. Before this PR it hung forever. |

### Out-of-scope follow-ups (not blocking)
- `executeTransaction` drops the inner results' `afterReply`. RESP3 PUBLISH to a channel the session is subscribed to, inside EXEC: the TCP mock sends the message *before* the EXEC reply, while real Redis sends it after. On the in-memory client, with SUBSCRIBE and PUBLISH in the same EXEC, no later pushes arrive. This predates the PR.
- `CommandCapabilities.pushOnly` is unused.

### Tests on this head
- `tsc` is clean.
- Unit tests: 459 of 459 pass.
- Mock integration: 1255 pass, 0 fail, 2 skipped.

CHANGELOG.md
fatal10110: Fixed in 5d5bfb8 (took the preferred option). `RedisClientSession` now has `enqueuePush`, `onReset(cleanup)` (runs once on RESET or close; returns an unregister) and `resetPushProducers()`. RESET and `ClientSession.close()` call the last one. MONITOR uses the same hook. The CHANGELOG entry now maps `registerResponseStreamCleanup` to `onReset` and `resetResponseStreams` to `resetPushProducers`, and includes a short worked example that uses only `ctx.session`.

src/in-memory-client.ts
fatal10110: Fixed in 5d5bfb8, so the tail frames are kept. A multi-target (UN)SUBSCRIBE result now has `options.trailingFrames`: confirmations 2..N as values, alongside the pre-encoded bytes the wire writes. `InMemoryRedisClient.command()` enqueues them as pushes right after the reply, so `SUBSCRIBE a b c` returns `["subscribe","a",1]` and `pushes()` yields `b`/2 and `c`/3 before any message. The new test in `tests/in-memory-streaming.test.ts` covers that and `UNSUBSCRIBE a c`; it failed on the previous head. The CHANGELOG no longer calls this out as a regression.

tests-integration/raw-tcp/monitor-protocol.test.ts
fatal10110: Fixed in 5d5bfb8. New session-level test in `tests/commands-foundation.test.ts`: "MONITOR holds feed lines back until its +OK is delivered". It runs MONITOR without running `afterReply`, has a second session run `PING early`, and asserts the monitoring session has no push after a macrotask turn. Then it runs `afterReply` and expects the line. Mutation check: with `deferPushesUntilAfterReply()` in monitor.ts swapped for `() => {}`, the test fails. With the real code it passes. The raw-TCP test stays as a wire-level regression guard.

src/core/transports/resp2/session-adapter.ts
fatal10110: Fixed in 5d5bfb8. I reworded the `dropOnPeerEof` doc in socket-connection-transport.ts: the "pending drains" line now refers to the adapter write chain and push writer, and the "(not a background drain)" aside is gone. I also reworded the virtual-connection.test.ts comment to match: the confirmations are one reply that is already being written at EOF. The "shipped paths do not hit this" wording belongs to the other #461 item and is left for that PR.

src/client-mocks/node-redis-mock.ts
fatal10110: Fixed in 5d5bfb8. `runOnSession` now runs `result.options?.afterReply?.()`, as the wire and the in-memory client do. That releases the deferral, and a self `CLIENT KILL` now takes effect. The docblock notes that nothing reads a non-pub/sub facade session push queue, so `sendCommand([\"MONITOR\"])` resolves `OK` but its lines are never surfaced (the facade has no `monitor()`). Full mock integration still passes: 1255 pass, 0 fail.

src/core/client-session.ts
```diff
@@ -767,7 +794,7 @@ export class ClientSession implements RedisClientSession {
   close(): void {
     this.unregisterClientSession?.()
     this.unregisterClientSession = undefined
-    this.resetResponseStreams()
+    this.resetPushProducers()
```
fatal10110: [low] `onReset` is now public API that third-party commands are told to use (CHANGELOG example), but a cleanup that throws breaks teardown. `resetPushProducers()` runs first in `close()` and has no guard. One throwing hook skips every hook after it, and also `signalSource.abort()`, `unwatch()`, `resetPubSub()` and `closePushQueue()`. The session's broker subscriptions and WATCH listeners then outlive the connection. Under RESET, the throw escapes as a non-Redis error, and the adapter turns that into writing an error and closing the connection.

Suggest running each hook in its own `try/catch` (log and continue) inside `resetPushProducers` (L687), or calling `resetPushProducers()` last in `close()`. A one-line test would pin it: register an `onReset` that throws, `close()`, then assert `pubsubSubscriptionCount === 0` and that the signal is aborted.

src/in-memory-client.ts
```diff
@@ -128,19 +121,14 @@ export class InMemoryRedisClient {
...
-      this.activeStream = result
-      this.streamFrames = result
-        .frames(this.lifetime.signal)
-        [Symbol.asyncIterator]()
-      const first = await this.streamFrames.next()
-      return first.done ? null : this.decode(first.value.value)
+    // The reply has been "delivered": the rest of a multi-target SUBSCRIBE's
+    // confirmations follow it, then whatever the command held back until now
+    // (MONITOR's first feed lines, a RESP3 PUBLISH's own message).
+    for (const frame of result.options?.trailingFrames ?? []) {
```
fatal10110: [low] `trailingFrames` fixes a bare multi-target SUBSCRIBE, but not one queued in a transaction. `executeTransaction` keeps only each item's `value` and `encoded`, so `trailingFrames` is dropped. Verified on 5d5bfb8:

```
MULTI; SUBSCRIBE a b; EXEC  -> [["subscribe","a",1]]
pushes()                    -> <timeout>   // ["subscribe","b",2] never arrives
```

This is not a regression, since main replied `Streaming command is not allowed in transaction`, and it is niche. Either forward the items' trailing frames onto the EXEC result's `trailingFrames`, or say in the CHANGELOG / `command()` doc that the in-memory client can't represent the extra frames Redis embeds in an EXEC array.

fatal10110:
## Re-review of 5d5bfb8: **SHIP**

All three medium findings from the review of 379270e are fixed.

| Prior finding | Status |
|---|---|
| [medium] CHANGELOG migration advice pointed to `enqueuePush`, which isn't on the interface, and there was no teardown hook | **Fixed**: `RedisClientSession` gains `enqueuePush`, `onReset` and `resetPushProducers`, RESET and `close()` both run them, and the CHANGELOG has a working example |
| [medium] The in-memory client dropped confirmations 2..N of a multi-target SUBSCRIBE | **Fixed** by `trailingFrames`, with a test. An in-memory race (a concurrent PUBLISH during a 3-channel SUBSCRIBE) still delivers every confirmation before the message |
| [medium] The MONITOR "+OK first" test passed with the deferral removed | **Fixed**: with `deferPushesUntilAfterReply` replaced by a no-op, the new session test fails 0/1; restored, it passes 1/1 |
| [low] Stale "background drain" comments | **Fixed** |
| [low] `runOnSession` skipped `afterReply` | **Fixed**. The node-redis mock is not affected: its `CLIENT KILL ID <self>` returns 0 there, and its SUBSCRIBE paths don't use `trailingFrames` |

What else I checked:
- `onReset` runs once, including across RESET followed by close. After RESET, a new MONITOR works again.
- `trailingFrames` never reaches the wire: the adapter writes `encoded` only.
- The only implementer of the widened interface is `ClientSession`, and `tsc` is clean.
- On a real Redis 8.0.6 container (mine, checked by run_id), these match the mock byte for byte in RESP2 and RESP3: SUBSCRIBE a b followed by PING, EXEC containing SUBSCRIBE a b, MONITOR queued in MULTI, and MONITOR → RESET → MONITOR.

New findings (low, posted inline, not blocking):
- A throwing `onReset` cleanup cuts `close()` teardown short.
- The in-memory client still drops trailing confirmations when SUBSCRIBE runs inside EXEC.

Tests on this head:
- `tsc` and lint are clean.
- Unit tests: 461 of 461 pass.
- Mock integration: 1255 pass, 0 fail, 2 skipped.

Issue #443, opened 2026-09-23
SORT: four divergences from real Redis (ALPHA+DESC ties, multiple BY, int-only set order, OBJECT ENCODING)


fatal10110:
Found while reviewing PR #425. All four divergences were verified against real Redis 8.0.6. **They all exist on `main`**, and #425 changes none of them. They belong with the SORT rework in #417.

## 1. `ALPHA` + `DESC` reverses elements that tie
When elements compare equal, real Redis keeps them in load order even under `DESC`. The mock reverses them.
```
list: real  c a b e d
      mock  d e b a c     (same on origin/main)
```

## 2. More than one `BY`
Real Redis stops sorting as soon as it sees a constant `BY` pattern. The mock uses the **last** `BY` given.

## 3. A set containing only integers
The mock returns it in insertion order. Real Redis returns it sorted, because an intset is stored in sorted order.

## 4. `OBJECT ENCODING` after SORT
On real Redis, SORT converts a small zset from `listpack` to `skiplist`. The mock doesn't.

## Related, lower priority
`readSortSource()` reads the key twice (type, then contents). If the key expired between the two reads, the code would crash. The window is about 1ms and this is pre-existing.

Note: the tie order for `SORT z BY <glob> ALPHA` (every weight missing) is **undefined** in real Redis. It changes between process restarts because the zset dict's hash seed is random. #425 deliberately leaves it untested; don't try to pin it.

## Acceptance
Each of cases 1–4 matches real Redis, pinned by tests that fail on the current code.

Issue #117, opened 2026-06-09
Admin: OBJECT ENCODING/REFCOUNT/IDLETIME/FREQ/HELP


fatal10110:
Part of #28.

## Scope
- `OBJECT ENCODING key`
- `OBJECT REFCOUNT key`
- `OBJECT IDLETIME key`
- `OBJECT FREQ key`
- `OBJECT HELP`

## Suggested minimal implementation
For `OBJECT ENCODING`, compute a plausible encoding string from the value's type/size: `int` for small string integers, `embstr` for short strings, `raw` for long strings, `listpack` for small hashes/sets/zsets, `hashtable`/`skiplist` for large ones.

PR #476, opened 2026-09-23
fix: scan SORT options at run time, guard patterns in the scan (#417, #443)

### LINKED ISSUE #417: SORT cluster guard runs on a parsed plan: wrong option reported, syntax error preempts denial, fires at MULTI queue time

fatal10110:
## Summary

**#417: cluster guard order.** `ClusterPolicy` checked SORT's BY/GET patterns in `beforeExecute`, against a plan that was already fully parsed. By that point the argument order was gone and any syntax error had already been thrown. Now:

- SORT/SORT_RO parse only the source key when planned (arity still fails at plan time). The options are scanned when the command runs, left to right, as `sortCommand()` does.
- As the scan reaches each BY glob or GET pattern, it calls `assertSortPatternAllowed()` in the new `src/core/sort-cluster-guard.ts`. That module owns the cluster knowledge: the cluster-mode check, the `patternHashSlot()` port, both profile gates and all four wordings.
- The result:
  1. The first offending option in argument order is reported.
  2. A denied pattern is reported before a later token that fails to parse. Junk *before* the pattern still gives `ERR syntax error` / `ERR value is not an integer or out of range`.
  3. Inside MULTI, every SORT option error (the cluster denial, a syntax error, a bad LIMIT integer) replies `+QUEUED` and appears in the EXEC array. Real Redis does the same, and the ground-truth run below confirms it for the syntax errors too.
- `SORT k GET BY BY GET` still returns three nils. The scan is positional, so option names in value position are just values.
- Routing keys come from a port of `sortGetKeys()`. That port does its own light scan: it skips LIMIT's two arguments and BY's or GET's one, and the last STORE wins. `SORT_RO` reports only the source key, matching its key spec.
- `ClusterPolicy` no longer knows about SORT. MOVED/CROSSSLOT/CLUSTERDOWN are still decided there, before the command runs, so a stale slot map still gets a MOVED it can follow.

**Design call (the layering question in #417).** I went with the cheapest option that keeps cluster knowledge out of the command. SORT only says "I'm about to use pattern P for option BY/GET"; the helper in `src/core` decides everything else.

- A policy-level hook can't work here. The check has to interleave with a scan that can fail partway through, and it has to run at EXEC time rather than at queue time. `beforeExecute` offers neither.
- The helper detects cluster mode from `ctx.server.clusterTopology.nodes.length`. `HELLO`/`INFO` already use the same test (`isClusterMode` in `connection.ts`).
- No new context plumbing and no new policy API.

**#443: SORT divergences.**
1. **ALPHA + DESC ties.** `sortCompare()` negates its result for DESC instead of reversing the vector, so ties keep load order in both directions. The mock now uses a stable sort on the negated comparison. While here I found that under ALPHA a missing BY weight (NULL) orders before an empty-string weight; that is fixed too.
2. **Multiple BY.** A constant BY sets `dontsort` for good, and otherwise the *last* glob BY is the one looked up.
3. **Integer-only set.** A set whose members are all canonical int64 values loads in ascending numeric order (intset order). Any other set keeps insertion order (listpack order). Documented known difference: the mock doesn't model the one-way encoding conversion.
4. **OBJECT ENCODING after SORT.** Not fixable yet: the mock has no `OBJECT` command. It is tracked as #117 and noted in `docs/COMMANDS.md`.
- **Double read.** `readSortSource()` now does a single `db.get()`.

Also addresses the #461 item about #425's ALPHA-tie note: the comment now says ties keep load order in both directions, and it records the remaining known differences. Those are pqsort instability with BY + partial LIMIT beyond six elements, and the undefined zset dict order, which is still deliberately untested.

Closes #417
Refs #443 — items 1–3 and the double read are fixed here. Item 4 depends on `OBJECT ENCODING` (#117), so #443 stays open (or item 4 can move to #117).

## Test plan
- [x] New ioredis tests in `tests-integration/ioredis/sort-integration.test.ts` (first-offender order, denial before trailing syntax/LIMIT error, MULTI→EXEC errors, `GET BY BY GET`, ALPHA DESC ties, NULL-before-empty weight, multiple BY, intset order, `sortGetKeys` STORE detection). Red on mock before the fix (7 failures), green on real Redis 8.0.
- [x] node-redis twins for all of them in `tests-integration/node-redis/sort-integration.test.ts`. Red on mock before the fix (7 failures), green on real. That file's `before` no longer FLUSHALLs the shared cluster, since its keys are per-run.
- [x] `tests-integration/compatibility/sort-cluster-gates.test.ts`: scan order, trailing syntax error and MULTI checks under each profile's wording. Red on redis-7.2/redis-8.0 before the fix, green on all profiles after.
- [x] Real Redis ground truth (8.0.6): every new assertion was checked by hand, including `SORT l BY w_* BADARG` → BY denied, `SORT l BADARG BY w_*` → syntax error, and MULTI queuing syntax errors.
- [x] `npm run build`, `npm run lint`, `npm test` (unit), `npm run test:integration:mock`, `npm run test:integration:compatibility:mock` all pass.
- [x] SORT and MULTI suites (ioredis + node-redis) against real Redis 8.0 (88/88 + 8/8). The shared compose cluster had a master and its replica (30002/30005) down during this run, so I used a private 3-master `redis:8.0` cluster on ports 31000-31002 via `REDIS_CLUSTER_PORTS`. I didn't touch the shared containers.


docs/COMMANDS.md
```diff
@@ -254,9 +254,21 @@ with `GT` or `LT`.
...
-  it while `valkey-8.0` (pinned at 8.0.0) still refuses it. Hash-field
-  dereference patterns such as
+  it while `valkey-8.0` (pinned at 8.0.0) still refuses it. As in Redis, the
+  guard runs inside SORT's own left-to-right option scan when the command
+  executes: the first offending option is the one reported, a later syntax
+  error is never reached, and inside `MULTI` the command queues and the error
+  surfaces in `EXEC`. Hash-field dereference patterns such as
   `object_*->field` are not modeled.
+- `SORT` loads a set of canonical 64-bit integers in ascending numeric order
+  (as an intset is stored) and any other set in insertion order (as a small
```
fatal10110: [medium] "any other set in insertion order" does not hold for a set that **started** as an intset (Redis wire-compat lens)

A set whose first `SADD` member is an integer is created as an intset (`setTypeCreate`), so its integer members are stored sorted. When a non-integer arrives later, `setTypeAddAux` converts the intset to a listpack by walking it in sorted order, then appends the new member. The integer prefix stays sorted, so this is not insertion order. The mock's `setLoadOrder()` sees a non-integer member and falls back to insertion order.

Ground truth, real Redis 8.0.6 (docker `redis:8.0`) vs this PR's mock (`--compat redis-8.0`):

```
SADD s 3 1 a            ; SORT s BY nosort
  real: 1 3 a           mock: 3 1 a
SADD s2 3 1 ; SADD s2 a ; SORT s2 BY nosort
  real: 1 3 a           mock: 3 1 a
SADD s3 a 3 1           ; SORT s3 BY nosort
  real: a 3 1           mock: a 3 1   (listpack from the start, matches)
```

So the claims in docs/COMMANDS.md here, in the `setLoadOrder()` doc comment (`src/commands/keys.ts:733-740`) and in the CHANGELOG are wrong for a common shape (`SADD s 3 1 a`). The one documented known difference covers only the reverse direction (never converting back to an intset).

Fix, minimum: document this as a second known difference in all three places, e.g. "a set created from an integer keeps its integers sorted, ahead of later non-integer members; the mock uses insertion order".

Preferred, can be a follow-up: the order is a property of the set's encoding history, and SORT can't reconstruct that from the current members. It belongs in the state layer, with the set value tracking an intset flag. That would also fix `SMEMBERS`, which diverges today in the same way: `SADD zz 3 1; SMEMBERS zz` gives `1 3` on real Redis and `3 1` on the mock. It would also let SORT drop its read-time re-derivation.

src/commands/keys.ts
```diff
@@ -885,30 +965,51 @@ function expandSortPattern(
...
 
 export const sortRoCommand = defineCommand({
   name: 'sort_ro',
   since: { redis: '7.0.0', valkey: '7.2.0' },
-  schema: sortSchema(false),
+  schema: sortSchema(),
   flags: ['readonly'],
-  keys: sortRoutingKeys,
-  execute: (args, ctx) => runSort(args, ctx),
+  // SORT_RO has no getkeys procedure: its key spec names the source key only.
```
fatal10110: [low] This comment is inaccurate (Redis wire-compat lens). SORT_RO *does* have a getkeys procedure: `sortROGetKeys()` in `db.c` returns only position 1, flagged RO|ACCESS. That is also why real Redis marks it `movablekeys`:

```
$ redis-cli -p <8.0.6> COMMAND INFO sort_ro
sort_ro / -2 / readonly movablekeys / 1 / 1 / 1 ...
$ redis-cli COMMAND GETKEYS SORT_RO l STORE d
l
```

The behavior (`[args.key]`) is correct. Suggested wording: "sortROGetKeys() reports the source key only: SORT_RO has no STORE."

src/core/sort-cluster-guard.ts
```diff
@@ -0,0 +1,112 @@
...
+ *   returns the element itself and reads no other key, from that comparison.
+ */
+export function assertSortPatternAllowed(
+  ctx: RedisExecutionContext,
+  option: 'BY' | 'GET',
+  pattern: Buffer,
+  key: Buffer,
+): void {
+  const topology = ctx.server.clusterTopology
+  if (topology.nodes.length === 0) {
```
fatal10110: [low] Cluster-mode detection is now a copy of `isClusterMode()` from `src/commands/connection.ts:77`, and its standalone branch has no test (Node.js architect lens).

Before this PR, "standalone never sees the SORT guard" held structurally: `ClusterPolicy` is only installed on cluster nodes. Now it depends on this `nodes.length === 0` check. If the check regresses, standalone `SORT l BY w_*` starts replying `denied in Cluster mode`, and no test would notice. Every new SORT guard test runs against a cluster, and the standalone raw-tcp SORT tests (`tests-integration/raw-tcp/command-errors-list.test.ts`) never use a BY glob or a GET pattern.

Suggest:
1. Move the predicate somewhere both callers can share, e.g. an `isClusterEnabled` getter on the topology or server state, mirroring `server.cluster_enabled`.
2. Add one standalone assertion, e.g. in raw-tcp: `SORT l BY w_* GET n_*` succeeds. It does work today; I checked by hand against the mock's `--single` server.

src/core/execution-policies/cluster-policy.ts
```diff
@@ -77,9 +70,10 @@ export function createClusterPolicy(
         transactionSlots.delete(ctx.session)
       }
 
-      // Redirection is decided in processCommand(), before sortCommand() ever
-      // parses BY/GET — so a stale slot map must still get a MOVED it can
-      // follow rather than a terminal "denied in Cluster mode".
+      // Redirection is decided here, before a command runs. SORT's BY/GET
```
fatal10110: [low] There is a stale comment outside the diff, in `tests/client-mocks/node-redis-mock.test.ts:699-704` (Node.js architect lens). It says SORT's routing keys "deliberately include its BY/GET *patterns* (see sortRoutingKeys ...) so ClusterPolicy can run its own check on them".

Neither half is true anymore. `sortRoutingKeys` returns only the source key plus STORE, and with this PR `ClusterPolicy` no longer checks SORT at all. The test itself still passes and is still useful, because it proves the node-redis facade forwards the command rather than refusing it client side. Please reword the comment to point at `src/core/sort-cluster-guard.ts` and SORT's option scan.

fatal10110:
## Review of 63827af: **NOT YET**

The #417 core is right, and I checked it against real Redis. One medium finding blocks: a doc claim about set load order that real Redis disproves. Three low nits.

| Severity | Where | Finding | Lens |
|---|---|---|---|
| medium | docs/COMMANDS.md:264 (also the `setLoadOrder` comment in keys.ts and the CHANGELOG) | A set created from an integer that later gains a non-integer keeps its integers sorted. `SADD s 3 1 a; SORT s BY nosort` gives `1 3 a` on real 8.0.6 and `3 1 a` on the mock. The docs say "insertion order". At minimum, document it as a known difference. The proper fix is in the state layer, which would also fix `SMEMBERS`. | Redis |
| low | src/commands/keys.ts:1010 | SORT_RO *has* a getkeys procedure (`sortROGetKeys`, `movablekeys` in COMMAND INFO). The behavior is correct; the comment isn't. | Redis |
| low | src/core/sort-cluster-guard.ts:45 | Duplicates `isClusterMode()`. No standalone test proves that SORT BY/GET aren't guarded. | Node |
| low | cluster-policy.ts:73 → tests/client-mocks/node-redis-mock.test.ts:699 | A stale comment still says ClusterPolicy checks SORT patterns through the routing keys. | Node |

### Checked against real Redis (redis:8.0.6 and redis:7.2 single-node clusters, plus 8.0.6 standalone)
Every probe below matches the PR's mock character for character, in both the 7.4+ and the pre-7.4 wordings:
- which option is reported first (`GET n_* BY w_*` vs `BY w_* GET n_*`)
- `BY nosort BY w_*`
- a denied pattern before a trailing `BADARG` or a bad `LIMIT`
- a leading syntax or LIMIT error still winning
- `GET #` (7.2 denies it, 8.0 allows it)
- `GET BY BY GET`
- MULTI queuing a denial, a syntax error and `LIMIT 0`, which all appear in the EXEC array, while `SORT` with no arguments gets EXECABORT
- errors from `EVAL` via call and pcall
- `SORT_RO … STORE` giving a syntax error
- a wrong-type key vs a syntax error, in both orders
- COMMAND GETKEYS for the `sortGetKeys` port, including `STORE by STORE d2` giving `[l, by]`
- ALPHA DESC ties, a missing vs an empty weight, multiple BY, and the intset order

Full-range `qsort` ties are stable on the docker image (glibc). The `pqsort` instability with BY plus a partial LIMIT reproduces too (`z w e r t` on real vs `q w e r t` on the mock, with 20 tied elements). It is documented as a known difference and correctly left untested.

The layering is fine. SORT only says "about to use pattern P for BY or GET"; the helper in `src/core` owns the cluster mode check, the gates and the wording. MOVED/CROSSSLOT still come from ClusterPolicy before the command runs.

### Tests
- tsc and eslint on the changed files: clean.
- Mock: unit tests (cluster-policy, seed, registry) plus the ioredis and node-redis SORT suites, set core and raw-tcp errors: 136/136. node-redis-mock: 47/47. Compatibility sort gates and profile gates under all 7 profiles: all pass.
- Real Redis 8.0.6 (private 1-node cluster): ioredis and node-redis SORT suites, 88/88.
- Red before green: the new tests run against origin/main's `src` fail 16 (8 per client). `GET BY BY GET` already passed there, as the PR says.

docs/COMMANDS.md
fatal10110: Fixed in f95608b (minimum fix, as suggested). This is now documented as a second known set-order difference in all three places: the `setLoadOrder()` doc comment, docs/COMMANDS.md and the CHANGELOG. The example used is `SADD s 3 1 a`, which gives `1 3 a` on Redis and `3 1 a` on the mock. The notes also say that both differences come from encoding history the mock doesn't track, and that SMEMBERS has the same intset-order gap. The state-layer fix (an intset flag on the set value) is left as a follow-up.

src/commands/keys.ts
fatal10110: Fixed in f95608b: the comment now reads `sortROGetKeys() reports the source key only: SORT_RO has no STORE.`

src/core/sort-cluster-guard.ts
fatal10110: Fixed in f95608b.
1. Cluster-mode detection is now one getter, `RedisServerState.clusterEnabled` (mirrors `server.cluster_enabled`). `HELLO`/`INFO` in connection.ts and `assertSortPatternAllowed()` both use it, and the private `isClusterMode()` copy is gone.
2. Added a standalone raw-tcp test in `tests-integration/raw-tcp/command-errors-list.test.ts`: `SORT`/`SORT_RO l BY <prefix>:w_* GET <prefix>:n_* GET #` returns the sorted projection. It passes on mock and on real standalone 8.0. As a mutation check, I forced the guard to run on standalone and the test fails.

src/core/execution-policies/cluster-policy.ts
fatal10110: Fixed in f95608b: the comment in `tests/client-mocks/node-redis-mock.test.ts` now says the BY/GET patterns aren't keys and the cluster-mode error comes from SORT's own option scan on the node (`src/core/sort-cluster-guard.ts`). So the facade must route by the source key and not refuse the command client side.

fatal10110:
## Re-review of f95608b: **SHIP**

All 4 earlier findings are fixed, and I found nothing new.

| Earlier finding | Status |
|---|---|
| [medium] intset-origin set order was undocumented | Fixed. It is now documented in `setLoadOrder`, COMMANDS.md and the CHANGELOG. Both documented differences match real 8.0.6: `SADD s 3 1 a` gives `1 3 a` on real Redis and `3 1 a` on the mock; `SADD s a 3 1; SREM s a` gives `3 1` on real Redis and `1 3` on the mock. The `SMEMBERS` gap is also real (`1 3` vs `3 1`). |
| [low] SORT_RO comment | Fixed. The comment now names `sortROGetKeys()`. |
| [low] duplicated cluster-mode predicate, no standalone test | Fixed. There is now one `RedisServerState.clusterEnabled` getter with the same `nodes.length > 0` logic, and no other callers are left. The new raw-TCP test passes on the mock and on real standalone 8.0.6. |
| [low] stale node-redis-mock comment | Fixed. |

**HELLO/INFO are unchanged.** In standalone, the mock and real 8.0.6 agree on `HELLO 3` mode (`standalone`), `redis_mode:standalone` and `cluster_enabled:0`. On a cluster node they agree on `mode cluster`, `redis_mode:cluster` and `cluster_enabled:1`. The mock's extra `cluster_state`/`cluster_slots_assigned` lines in INFO predate this PR.

**Tests**
- tsc and eslint (`src` plus the changed tests): clean.
- Mock: HELLO/INFO/connection/cluster suites, raw-tcp, node-redis-mock, cluster-policy unit tests and both SORT suites: 267/267.
- SORT cluster gates and profile gates under all 7 profiles: all pass.
- Real Redis 8.0.6 (my own standalone and single-node cluster): raw-tcp command-errors-list plus the ioredis and node-redis SORT suites, 91/91.

PR #477, opened 2026-09-23
fix: spell double replies per compatibility profile (#451)

### LINKED ISSUE #451: RESP double formatting is not profile-aware: 6.2 / 7.0 print %.17g
Found while reviewing PR #427. The review fuzzed doubles against real Redis 6.2.14, 7.0.15, 7.2.1, 8.0.6 and Valkey 8.0/9.0.

## Problem
The RESP encoder formats doubles one way on every compatibility profile. Real Redis formats them differently by version:
- **7.2+ and Valkey:** `d2string()` / `fpconv_dtoa` (Grisu2).
- **6.2 and 7.0:** `%.17g`, e.g. `0.1` → `0.10000000000000001`.

Mismatches over about 36k edge-weighted values:

| formatter | 7.2+ / Valkey | 6.2 / 7.0 |
|---|---|---|
| `main` (JS `toString`) | 4,442 | ~23,765 |
| #427 rewrite (7.2 layout) | 112 | ~23,928 (456 values newly wrong) |

So on the `redis-6.2` and `redis-7.0` profiles the mock sends double text real Redis never sends. That affects ZSCORE, ZINCRBY, INCRBYFLOAT, WITHSCORES and every RESP2 double.

## Also
Even after #427's rewrite, about 75 in 30k values differ on 7.2+: Grisu2 picks a different digit string than JS's shortest round-trip, though both parse back to the same number. Matching those means porting `fpconv_dtoa`.

## Acceptance
- Double formatting is profile-aware: `%.17g` on 6.2/7.0, `d2string` on 7.2+.
- There is a single shared formatter; `lua-runtime.ts` currently has its own copy.
- A fuzz test is pinned against captured real-Redis output for each version.

Depends on the decision in #427 about whether the formatter rewrite lands there or separately.

fatal10110:
## Summary
Real Redis spells a double differently depending on the version. Redis 6.2 and 7.0 print `%.17g`. Redis 7.2+ and every Valkey print `d2string()`: every digit of an integer within ±2^62, and Redis's Grisu2 `fpconv_dtoa` for anything else. The mock used JS `toString()` on every profile, and `lua-runtime.ts` had its own copy of that formatter.

| score | redis-6.2 / 7.0 | redis-7.2+ / valkey | before (every profile) |
|---|---|---|---|
| `0.1` | `0.10000000000000001` | `0.1` | `0.1` |
| `0.0000123` | `1.2300000000000001e-05` | `1.23e-5` | `0.0000123` |
| `2^62` | `4.6116860184273879e+18` | `4611686018427387904` | `4611686018427388000` |
| `1e20` | `1e+20` | `1e+20` | `100000000000000000000` |
| `4.8911660955712037e-5` | `4.8911660955712037e-05` | `4.8911660955712037e-5` (Grisu2 is not shortest) | `0.00004891166095571204` |

- **`src/core/double-format.ts`** holds the one shared `formatRedisDouble(value, profile)`.
  - `%.17g`: the digits come from the double's exact decimal expansion (BigInt), because glibc rounds a tie to even and `toPrecision(17)` rounds it up. For example, `1234567890123456.25` prints `…456.2` on 6.2, where `toPrecision` gives `…456.3`. The C `%g` layout rules follow: exponent form when X < -4 or X ≥ 17, trailing zeros stripped, and at least two exponent digits.
  - `d2string` / `fpconv_dtoa`: a faithful BigInt port of `deps/fpconv/fpconv_dtoa.c`. The cached powers of ten are computed exactly, and I checked them against `fpconv_powers.h`: 87/87 identical. **Nothing is deferred.** The ~75/30k Grisu2 digit-choice mismatches noted on the issue are gone.
- **New gate `reply.double-fpconv`** (`redis: 7.2.0`, `valkey: 7.2.0`) in `FeatureId` / `FEATURE_GATES`, with a row in the `docs/API.md` gate table.
- **Plumbing.** `RespEncodeOptions` and `DecodeRedisValueOptions` gain an optional `profile`; when it is missing, the default profile's spelling is used. It is passed in by:
  - `session-adapter.ts`
  - `ClientSession`'s transaction encoding
  - `in-memory-client.ts`
  - the node-redis facade (`sendCommand` / `eval` / `multi().exec()`)
  - the Lua bridge (`redisValueToLuaReply`)
  - `ZSCAN`'s `scoreBuffer`, which was also a `toString()` copy
- **Covers** ZSCORE, ZMSCORE, ZINCRBY, every WITHSCORES reply, RESP3 `,` doubles, `redis.call` score reads and ZSCAN.

### Not covered
- **INCRBYFLOAT / HINCRBYFLOAT** are not on this path. Real Redis formats them with `ld2string(…, LD_STR_HUMAN)`, which is `%.17Lf` over a `long double`, and that is the same on every version. For example, `INCRBYFLOAT k 0.1` then `0.2` returns `0.3` on 6.2 through 8.0, while the mock returns `0.30000000000000004`. Matching that means emulating 80-/128-bit long-double arithmetic, which is architecture-dependent, so it is left as a separate issue.
- **ZSCAN on a skiplist-encoded zset** (more than 128 members, or long members) prints `%.17g` (`LD_STR_AUTO`) on every version. The mock does not model encodings, so ZSCAN follows the listpack (`d2string`) spelling.
- **`ZADD k -0 m`** stores `-0` on 6.2 but `0` on 7.0+. The mock normalizes to `0` on every profile. That is a zset storage difference, not a formatting one.

Closes #451

## Test plan
- [x] `tests/core/double-format.test.ts` replays `tests/fixtures/redis-double-format.json` against every server. The fixture holds 1,310 values, including every Grisu2 non-shortest case and every `%.17g` tie found in 30k probes. They were captured by `scripts/capture-double-format-fixture.ts` from real redis-server 6.2.14, 7.0.15, 7.2.4, 7.4.4, 8.0.6 and valkey-server 8.0.0 and 9.0.0. Result: 0 mismatches on every server. An extra 60k-value fuzz against live 6.2 / 7.0 / 7.2 / 8.0 / valkey-9.0 also gave 0 mismatches.
- [x] New `tests-integration/{ioredis,node-redis}/zset/double-format.test.ts` covers ZSCORE, ZMSCORE, ZINCRBY, WITHSCORES, ZPOPMAX and a Lua `redis.call` read. They failed on mock before the fix and passed against real Redis 8.0.6; they now pass on mock (8/8). The node-redis twin notes that node-redis parses RESP3 doubles, so it reads the text through a script, and it pins the number node-redis parses `,1.23e-5` to.
- [x] New `tests-integration/compatibility/double-format.test.ts` checks the raw RESP2 and RESP3 bytes plus Lua, per profile. It passes on all 7 profiles, and the expected bytes were read off the real servers.
- [x] `npm run build`, `npm run lint`, `npm test` (477 pass), full mock integration (1258 pass), `test:integration:compatibility` for all 7 profiles, `test:package` and `test:demo`.
- [x] Real: zset, scan, geo, string and scripts suites for ioredis and node-redis ran against a private redis 8.0.6 cluster plus standalone (444 pass). The shared docker-compose cluster was in `cluster_state:fail` during this work (nodes 30002/30005 down), so I did not use it.


src/core/double-format.ts
```diff
@@ -0,0 +1,425 @@
...
+ *
+ * Two spellings, by profile (#451):
+ *  - Redis 6.2 / 7.0: `%.17g` — `0.1` is `0.10000000000000001`.
+ *  - Redis 7.2+ and every Valkey (`reply.double-fpconv`): `d2string()` —
+ *    exact integer digits within ±2^62, `fpconv_dtoa` (Grisu2) otherwise.
+ *
+ * `inf` / `-inf` / `nan` and `-0` are spelled the same by both. Without a
+ * profile, the default profile's spelling is used.
+ */
+export function formatRedisDouble(
```
fatal10110: [medium] GEO coordinates still go through `toString()` and are never a `double` reply. They are the one double-emitting path left outside this formatter.

`src/commands/geo/geopos.ts:19-20` and `src/commands/geo/search-core.ts:372-373` (WITHCOORD for GEOSEARCH / GEORADIUS*) still build `RedisValue.bulkString(Buffer.from(lon.toString()))` on every profile and every protocol. Real servers differ from that in three ways (checked on private containers):

**RESP3: every version sends a `,` double, and the mock sends `$`.**
```
redis 8.0.6   HELLO 3 / GEOPOS g Palermo -> *1 *2 ,13.361389338970184 ,38.1155563954963
redis 6.2.24  HELLO 3 / GEOPOS g Palermo -> *1 *2 ,13.36138933897018433 ,38.11555639549629859
mock (any)    HELLO 3 / GEOPOS g Palermo -> *1 *2 $18 13.361389338970184 $16 38.1155563954963
```

**redis-8.0, the default profile: GEO uses `d2string()` (this PR's fpconv path), but the mock uses JS `toString()`.** I stored 20k random coordinates on real 8.0.6 and the mock, then read them back with RESP2 GEOPOS. 6,680 of 40,000 values were the same number spelled differently:
```
real=1.5288591384887695e-4   mock=0.00015288591384887695
real=-4.559755325317383e-5   mock=-0.00004559755325317383
```

**6.2 / 7.0 / 7.2 / 7.4 / Valkey 8.0 / 9.0 print `addReplyHumanLongDouble`, which is `%.17Lf` with trailing zeros stripped**, e.g. `13.36138933897018433`. The coordinate is a `double` that gets promoted exactly to a `long double`. So this does not need the long-double arithmetic the INCRBYFLOAT follow-up needs: it is `exactDecimal()` rounded half-even to 17 fraction digits.

Fix: send coordinates as `RedisValue.double` and format them per profile. That is `d2string` on 8.0, which this PR already has, and a `%.17Lf`-human variant on the older profiles and Valkey. If you would rather keep this PR's scope, add GEO to "Not covered" and the CHANGELOG, and file a follow-up. Right now the CHANGELOG / API.md wording reads as though every double reply is handled.

Separate, pre-existing, not blocking: on 8.0 many coordinates also differ as numbers (`real=-3.421873565107584e-5 mock=-0.00003421873565656597`). The mock's geohash decode does not reproduce Redis's. That belongs in its own issue.

tests-integration/compatibility/double-format.test.ts
```diff
@@ -0,0 +1,101 @@
...
+
+      const small = usesFpconv ? '1.23e-5' : '1.2300000000000001e-05'
+      assert.strictEqual(
+        await send('ZRANGE', key, '0', '0', 'WITHSCORES'),
+        `*2\r\n${bulk('small')}${bulk(small)}`,
+      )
+      await send('DEL', key)
+    })
+
+    test('RESP3 doubles and scores read inside Lua', async () => {
```
fatal10110: [medium] Several plumbing sites the PR claims to fix have no test that fails without them.

The profile-sensitive tests are this file, `tests/core/double-format.test.ts` and the default-profile client twins. Together they cover RESP2 ZSCORE / ZINCRBY / ZRANGE WITHSCORES, one RESP3 ZSCORE, and `redis.call`. By inspection, none of the following is checked on a `%.17g` profile. That profile is the only place where the new spelling differs from what the old code would print for these inputs:

- **ZSCAN** (`scan.ts:144`, `scoreBuffer(entry.score, ctx.server.profile)`). The only ZSCAN assertions (`ioredis/zset/core.test.ts:215`) run on the default profile with `inf` / `-inf` scores, which print the same way on every profile.
- **EXEC reply encoding** (`client-session.ts:467`, `profile: this.server.profile`). No test reads a double out of an EXEC on redis-6.2 / 7.0.
- **Socketless decoding** (`in-memory-client.ts:241`, `profile: this.session.server.profile`). `tests/in-memory-client.test.ts` has a `redis-6.2` client but never reads a double through it.
- **RESP3 WITHSCORES** (`flat-pairs` → `*2 [$member ,score]`). Only RESP3 ZSCORE is asserted here.

I probed all four against real 6.2.24 / 8.0.6 and against the mock with `--compat redis-6.2` / `redis-8.0`, and they all match today. For example, 6.2 `MULTI / ZSCORE / EXEC` gives `*.. $19 0.10000000000000001` on both. So this is a regression-guard gap, not a bug. But if any one of those lines lost its `profile` argument, every test would still pass.

Suggested fix: extend this compat test with a ZSCAN, a `MULTI`/`ZSCORE`/`EXEC`, and a RESP3 `ZRANGE … WITHSCORES` assertion, using the same `usesFpconv ? … : …` pattern. Also add one `createInMemoryClient({ compatibility: 'redis-6.2' })` ZSCORE assertion to `tests/in-memory-client.test.ts`.

src/core/double-format.ts
```diff
@@ -0,0 +1,425 @@
...
+
+const G17_PRECISION = 17
+
+/**
+ * `snprintf("%.17g", value)` for a finite, non-zero double, as glibc prints
+ * it: correctly rounded, ties to even. `toPrecision(17)` would round a tie
+ * *up* (`1234567890123456.25` → `…456.3`, where glibc prints `…456.2`), so the
+ * digits come from the double's exact decimal expansion instead.
+ */
+function formatPercentG17(value: number): string {
```
fatal10110: [low] Performance: both BigInt paths cost 10–100× more than the `toString()` they replace.

I measured this on PR head (node 22, per call):

| input set | `toString` | `%.17g` (6.2) | `d2string` (8.0) |
|---|---|---|---|
| typical scores (2 dp, 0..1000) | 1.2 µs | 5.9 µs | 9.9 µs |
| `Math.random()` | 0.3 µs | 9.6 µs | 11.6 µs |
| random bit patterns | 1.0 µs | 39 µs | 15 µs |
| subnormals / 1e-300..1e-250 | 0.1–0.2 µs | **152–156 µs** | 12–33 µs |

So a 100k-member `ZRANGE … WITHSCORES` now spends about 1 s just formatting scores. AGENTS.md puts correctness first, so this does not block the PR. There is a cheap exact fast path for `%.17g`, though. `toPrecision(17)` is already correctly rounded, and it differs from glibc only on an exact tie. A tie needs the exact expansion to end at the 18th significant digit. After you strip the mantissa's trailing zero bits, that is impossible once `oddMantissa · 5^k` has more than 18 digits (roughly k > 26 when `exp < 0`). You could use BigInt only when a tie is possible, or at least memoize `5n ** k`. The Grisu2 port could also use two 32-bit `Number` halves instead of BigInt. Correctness is fine: I fuzzed 113k fresh values against all 7 servers and got 0 mismatches.

src/core/lua-runtime.ts
```diff
@@ -320,35 +327,38 @@ function redisValueToLuaReply(value: RedisValue): ReplyValue {
     case 'integer':
       return value.value
     case 'double':
-      return Buffer.from(formatNumber(value.value))
+      return Buffer.from(formatRedisDouble(value.value, profile))
```
fatal10110: [low] This PR didn't cause it, but the Lua bridge still loses RESP3 doubles in both directions. It's worth a follow-up issue, because this line is now the Lua half of "the one shared formatter".

```
                                               real 6.2.24                real 8.0.6       mock (6.2 / 8.0)
EVAL "return {double=0.1}" 0          RESP2   $19 0.10000000000000001    $3 0.1           *0
                                      RESP3   ,0.10000000000000001       ,0.1             *0
EVAL "redis.setresp(3); return redis.call('ZSCORE',KEYS[1],'b')" 1 k
                                      RESP3   ,1.2300000000000001e-05    ,1.23e-5         $7 1.23e-5 (bulk)
EVAL "redis.setresp(3); return tostring(redis.call('ZSCORE',KEYS[1],'b')['double'])" 1 k
                                              $8 1.23e-05                $8 1.23e-05      $3 nil
```

The bridge has two gaps:
- It ignores the script's `setresp` level. Under `setresp(3)`, a `double` should reach the script as `{ double = n }`, not as the formatted bulk from this line.
- A script-built `{double=…}` table comes back as an empty array, so `luaReplyToRedisValue`'s `'double' in value` branch (line 229) never gets hit.

It doesn't block this PR, but it should be tracked.

src/client-mocks/node-redis-mock.ts
```diff
@@ -1445,10 +1464,12 @@ export const NODE_REDIS_DECODE_OPTIONS: ClientDecodeOptions = {
 function decodeReply(
   value: RedisValue,
   respVersion: RespVersion,
+  profile?: CompatibilityProfile,
```
fatal10110: [low] `profile` is optional here, but every caller passes one: `sendCommand`, `eval`, and both `multi().exec()` sites. Make it required, so a future `decodeReply` call can't silently fall back to the default (redis-8.0) spelling on a server running another profile.

The public `RespEncodeOptions.profile` / `DecodeRedisValueOptions.profile` can stay optional for API compatibility. This one is module-private.

scripts/capture-double-format-fixture.ts
```diff
@@ -0,0 +1,236 @@
...
+  values.forEach((v, i) => {
+    const g17 = g17Server.replies[i]
+    const fpconv = fpconvServer.replies[i]
+    // Integers within ±2^62 print every digit (`ll2string`), so their digits
+    // differ from the shortest round-trip by design — not a Grisu2 case.
+    const integerPath = Number.isInteger(v) && Math.abs(v) <= 2 ** 62
+    const interesting =
+      (!integerPath && significantDigits(fpconv) !== shortestDigits(v)) ||
+      significantDigits(g17) !== toPrecisionDigits(v)
+    if (i < KEEP || interesting) {
```
fatal10110: [low] The fixture has 1,310 rows but only 1,203 distinct inputs. The `interesting` rows repeat values the generator draws more than once, for example `-7.17e+21` appears twice at the end. You could dedupe by input here, e.g. with a `Set` of seen inputs, so the file only holds what it needs.

Reproducibility is fine otherwise. I re-ran this script against my own redis 6.2.24 / 7.0.15 / 7.2.16 / 7.4.11 / 8.0.6 and valkey 8.0.11 / 9.0.6. It produced a byte-identical `cases` array, and only the server patch versions in `servers` changed. So the committed fixture is authentic and reproducible.

fatal10110:
## Review of PR #477 at `2801bcd`: **NOT YET**

I looked at this through two lenses: Redis wire compatibility (R) and Node.js architecture (N). I checked it against private redis 6.2.24 / 7.0.15 / 7.2.16 / 7.4.11 / 8.0.6 and valkey 7.2.14 / 8.0.11 / 9.0.6.

**The formatter itself is correct.** I fuzzed it independently against all 7 preset servers with 113,107 fresh values and got **0 mismatches on every server**. The values included random bit patterns, 3k random subnormals, every power of 2 from 2^-1074 to 2^1023, every power of 10 from 1e-323 to 1e308, ±40 ulp around 2^53 / 2^62 / 2^63 / 2^64 / 1e16–1e19, 30k `n/2^k` tie candidates, 40k "human" decimals, ±inf and ±MAX_VALUE. Valkey 7.2.14 also prints the fpconv spelling, which confirms the `valkey: 7.2.0` gate. Re-running the capture script against my servers produced a byte-identical `cases` array, so the fixture is authentic and reproducible. I also compared raw bytes, real vs mock (`--compat redis-6.2` / `redis-8.0` / `valkey-9.0`), for ZSCORE, ZMSCORE, ZINCRBY, ZADD INCR, ZRANGE / ZREVRANGEBYSCORE / ZUNION / ZINTER / ZDIFF / ZRANDMEMBER WITHSCORES, ZPOPMAX, BZPOPMIN, ZSCAN, EXEC, `redis.call`, and inf / -inf, over RESP2 and RESP3. All identical. Socketless (`createInMemoryRedis`, 6.2 and 8.0) and node-redis RESP3 parsing (real vs mock) also match.

|---|---|---|---|
| medium | `double-format.ts:29` (fix site `geo/geopos.ts:19`, `geo/search-core.ts:372`) | GEO coordinates still use `toString()` bulk strings. RESP3 should be a `,` double on every version, and 8.0 spells them with `d2string` (6,680 / 40k fuzz values differ). Fix, or scope out explicitly with a follow-up | R |
| medium | `compatibility/double-format.test.ts:75` | No test fails if the ZSCAN, EXEC-encoding or socketless-decode `profile` plumbing is dropped. RESP3 WITHSCORES is also unasserted | N |
| low | `double-format.ts:66` | BigInt paths cost 6–15 µs per double (150 µs for tiny `%.17g` values) vs about 1 µs; a fast path is possible | N |
| low | `lua-runtime.ts:330` | Pre-existing: Lua `setresp(3)` doubles and `return {double=…}` are wrong (`*0`). Follow-up issue | R |
| low | `node-redis-mock.ts:1467` | Private `decodeReply` `profile` should be required | N |
| low | `capture-double-format-fixture.ts:210` | 107 duplicate rows in the fixture | N |

**Tests on PR head:** `npm test` 477/477 pass. The full mock integration suite passes: 1258 passed, 2 skipped, 0 failed. `compatibility/double-format` passes on all 7 profiles, and `tsc --noEmit` is clean. **Red before green:** with `src/` reset to `origin/main`, the new ioredis / node-redis tests fail 8/8, and the compat test fails 2/2 on redis-6.2 and 1/2 on redis-8.0.

**Not checked against real Redis:** the new ioredis / node-redis *cluster* tests, because the shared cluster is down. I checked the same behavior on standalone servers instead.

These are not re-flagged because they are documented follow-ups: INCRBYFLOAT long-double, skiplist ZSCAN `%.17g`, and `ZADD -0` on 6.2.

src/core/double-format.ts
fatal10110: Fixed in 44edb53. I kept it in this PR because it stayed small: about 60 lines of src and one new gate.

- GEOPOS and WITHCOORD (GEOSEARCH / GEORADIUS*) now reply with `RedisValue.double`. That makes them a `,` double on RESP3 and a bulk string on RESP2, as on every real version.
- New gate `geo.coord-d2string` (`redis: 8.0.0`, no Valkey version):
  - redis-8.0 uses `d2string()`, e.g. `13.361389338970184`.
  - Every other profile uses `formatGeoCoordinate`'s `%.17Lf` human form. That is the double's exact decimal value rounded half-even to 17 fraction digits, with trailing zeros stripped and `-0` → `0`, e.g. `13.36138933897018433`.
- `RedisValue.double(value, text?)` carries that exact text. The encoder, the decoders and the Lua bridge use it before falling back to `formatRedisDouble`.

How I verified it, on my own containers (redis 6.2.14 / 7.0.15 / 7.2.4 / 7.4.4 / 8.0.0 / 8.0.6, valkey 8.0.0 / 9.0.0):
- I stored 20k random coordinates and read them back with RESP2 GEOPOS. Result: 0 mismatches on every server. The raw RESP3 bytes also match, `,13.36138933897018433` on 6.2 and `,13.361389338970184` on 8.0.
- 6.2 and 7.0 were only checked where their own text round-trips (|v| ≥ 0.1), because their geohash decode differs from 7.2+.
- 8.0.0 already has the d2string spelling, so the gate sits at 8.0.0.

GEODIST and WITHDIST stay `%.4f` bulk strings on RESP3 as well, which matches every real version.

I left the separate numeric decode difference you noted alone. The CHANGELOG entry now says so explicitly.

tests-integration/compatibility/double-format.test.ts
fatal10110: Fixed in 44edb53. The compat test now asserts, on all 7 profiles, with bytes read off real 6.2.14 and 8.0.0:
- **ZSCAN**
- **EXEC across a HELLO**: `MULTI / ZSCORE / HELLO 3 / ZSCORE / EXEC`. This is the case that actually reaches `ClientSession`'s own encoder. A plain EXEC is encoded by the session adapter.
- **RESP3 `ZRANGE … WITHSCORES`**
- **GEOPOS / GEOSEARCH WITHCOORD**, on RESP2 and RESP3

`tests/in-memory-client.test.ts` now checks ZSCORE and GEOPOS through a `redis-6.2` client and a default one.

Mutation check: I dropped the profile at each site, one at a time, and each drop turns a test red:
- `scan.ts`
- `client-session.ts`
- the RESP3 flat-pairs arm in `resp-encoder.ts`
- `in-memory-client.ts`
- `geopos.ts`
- `search-core.ts`
- `session-adapter.ts`
- `lua-runtime.ts`

src/core/double-format.ts
fatal10110: `%.17g` is addressed in 44edb53. The fast path uses `toExponential(16)`, which is correctly rounded, and falls back to the exact BigInt expansion only when an exact tie is possible. That means an odd mantissa × 2^-k with k ≤ 25 (the expansion `m·5^k` has more than 18 digits once k > 25), or an integer ≥ 2^53.

Measured per call (node 22, 20k values, after warm-up):

| input set | `toString` | `%.17g` | `d2string` |
|---|---|---|---|
| typical 2-dp scores | 0.11 µs | 0.79 µs | 1.16 µs |
| `Math.random()` | 0.09 µs | 0.58 µs | 2.09 µs |
| 1e-300-scale | 0.09 µs | 0.59 µs | 2.03 µs |
| random bits | 0.12 µs | 1.02 µs | 2.21 µs |

The subnormal and tiny-value case that took about 150 µs is gone. A fresh 80k-value fuzz against 6.2 / 7.0 / 7.2 / 8.0.6 / valkey-9.0 still gives 0 mismatches; it includes a generator arm aimed at small-k tie candidates. The fixture replay passes as well.

The Grisu2 port stays on BigInt. At about 1–2 µs warm, I've left a Number-halves rewrite as a possible follow-up rather than risk the exactness here.

src/core/lua-runtime.ts
fatal10110: Agreed, this predates this PR and is tracked by #468. Both gaps are in scope there: `setresp(3)` delivering `{double=n}` to the script, and a script-built `{double=…}` coming back. I've left the code alone here apart from threading the profile through. With 44edb53 the bulk-string branch also honours an explicit `text`, so GEO coordinates read through `redis.call` keep their per-profile spelling.

src/client-mocks/node-redis-mock.ts
fatal10110: Done in 44edb53: `decodeReply`'s `profile` is now required. That surfaced one more caller, the pub/sub `dispatchPush` path, which now passes `pubsub.session.server.profile`. The public `RespEncodeOptions.profile` and `DecodeRedisValueOptions.profile` stay optional.

scripts/capture-double-format-fixture.ts
fatal10110: Done in 44edb53. The script now keeps each input once through a `Set`. I regenerated the fixture on my own containers: 1,203 distinct rows, and redis 8.0.0 (the preset version) is now captured alongside 8.0.6. All 8 servers match one of the two columns.

fatal10110:
## Re-review of PR #477 at `44edb53`: **SHIP**

I re-checked every finding from the `2801bcd` review against the changed lines on my own containers, confirmed by run_id: redis 6.2.24 / 7.0.15 / 7.4.11 / 8.0.6 and valkey 8.0.11 / 9.0.6.

| prior finding | status | how I checked |
|---|---|---|
| [medium] GEO coordinates used `toString()` bulk strings | **fixed** | Real vs mock raw bytes (`--compat` 6.2 / 7.4 / 8.0 / valkey-9.0) are now identical for GEOPOS, GEOSEARCH WITHCOORD, GEORADIUS WITHCOORD, and `redis.call('GEOPOS')`, over RESP2 and RESP3. I also ran 20k random coordinate pairs through GEOADD → GEOPOS on each of 6 servers: `formatGeoCoordinate(Number(text), profile) === text` held for 240,000 / 240,000 values. That confirms the `geo.coord-d2string` split: redis 8.0 uses d2string, and 6.2–7.4 plus both Valkeys use `%.17Lf` trimmed. |
| [medium] profile plumbing had no failing test | **fixed** | Spot-checked two sites. Reverting ZSCAN (`scan.ts` + `helpers.ts`) to main turns `ZSCAN, EXEC across a HELLO, and RESP3 WITHSCORES` red on redis-6.2. Reverting `geopos.ts` to `2801bcd` turns `GEOPOS / WITHCOORD coordinates` (redis-8.0) and the new in-memory-client test red. |
| [low] BigInt cost | **fixed** | `%.17g` now takes about 0.6–1 µs per value, including subnormals (was up to 156 µs). |
| [low] Lua `setresp(3)` / `{double=…}` | deferred to #468 | Pre-existing. Not blocking. |
| [low] `decodeReply` profile optional | **fixed** | It is now required, and the pub/sub caller passes the profile. |
| [low] duplicate fixture rows | **fixed** | 1,203 rows, 1,203 distinct inputs. 8.0.0 is now captured. |

**The new `%.17g` fast path is correct.**
- Against real servers: 85,396 values, including 2,519 exact 17-digit ties, `n/2^k` for k = 1–30 (either side of the `k ≤ 25` tie bound), 6k subnormals, carry cases near `9.999…eN`, and integers ≥ 2^53. **0 mismatches** on redis 6.2, 7.0, 7.4, 8.0 and valkey 9.0.
- Offline: 2.3M values compared the fast path plus fallback against an independent exact-BigInt `%.17g`, and `formatGeoCoordinate` against an independent `%.17Lf` reference. 0 mismatches.

**`RedisValue.double(value, text?)` does not leak.** `text` is set only by the two GEO builders. The encoder (RESP2 and RESP3), the RESP2 decoder and the Lua bridge are the only readers. Values without `text` keep their old shape, so existing `deepStrictEqual` comparisons are unaffected.

**Tests on 44edb53:**
- `tsc --noEmit` and lint are clean.
- `npm test`: 482/482 pass.
- Full mock integration suite: 1260 passed, 2 skipped, 0 failed.
- The compatibility suite passes on all 7 profiles. One redis-6.2 run failed once while it was running in parallel with the full suite, and it passed on 8/8 reruns.

No new findings.

Issue #454, opened 2026-09-23
Test runs can hang with no named failure; add --test-timeout and chase the node-redis mock stall


fatal10110:
Found during the reviews of PR #424.

## Problem
The unit and mock-integration runs sometimes **hang** instead of failing, with no test named.
- `unit (node 22.6)` in CI stalled for more than 12 minutes in `tests/client-mocks/node-redis-mock.test.ts`, and the rerun passed in about 60s. That file and its source are byte-identical to `main`, and none of the 117 modules it loads is a transport module. It only surfaced because #424 added the 22.6 job.
- `tests-integration/node-redis/sort-integration.test.ts` hung once, with the server showing no open connections. That points at the node-redis client.
- One mock-integration run hung in what looked like an ioredis cluster slot-refresh loop (`tests-integration/ioredis/key/keyspace-empty-cleanup.test.ts`). The "loop" is most likely ioredis's periodic 5s slot refresh ticking while some other await hangs, since the harness sets a huge `slotsRefreshTimeout`. It passed 60/60 when run in isolation.

## Proposal
1. Add `--test-timeout` (e.g. 60s per test) to the `test` and `test:integration:*` scripts, so a hang becomes a **named failure** instead of a stalled job.
2. Then investigate whichever test the timeout names.

## Acceptance
No script can hang indefinitely. The node-redis mock stall is reproduced by name and fixed, or filed separately with its cause.

PR #478, opened 2026-09-23
chore: bound test runs with --test-timeout (#454)

fatal10110:
## Summary
- Adds `--test-timeout 60000` to `test`, `test:integration:mock`, `test:integration:mock:*`, `test:integration:raw:mock`, `test:integration:compatibility:mock`, `test:package` and `test:demo`. The `*:real` scripts already had `--test-timeout 30000` and are unchanged, including their `clean:redis` step. CI always goes through these scripts.
- Adds `timeout-minutes` to every CI job as an outer backstop: 10 for lint, 15 for unit and package, and 25 for integration, which also covers the 420s Redis startup budget. Current jobs take 1–2 minutes.
- AGENTS.md now explains how the timeout behaves. The stale "`npm test` has no timeout" comment in `node-redis-mock.test.ts` is updated.

### How `--test-timeout` behaves (checked on 22.6, 22.12 and 24.20)
- **Node 22:** the timeout bounds each test **file**. The child process is killed and only the file is named. This is what turns the 22.6 unit stall into a failure.
- **Node 24:** the timeout applies to each **test** and names it. There is no per-file bound, so if a timed-out test leaks a ref'd handle, its file's process can stay alive. `timeout-minutes` is the backstop for that case.
- I left out `--test-force-exit` on purpose. On Node 22 under macOS it drops whole files from the report: `npm test` reported 373–442 tests instead of 462, with exit 0.
- The slowest single test in unit and mock integration takes about 4s, and the slowest file about 4.4s. 60s leaves room for slow CI runners, including the per-file limit on 22.6.

### The node-redis mock stall: a Node/V8 deadlock, not our code
- **Reproduced by name on Node 22.6.** The run stalls in `tests/client-mocks/node-redis-mock.test.ts` at `EVAL routes by its declared KEYS` (about 1 in 40–160 runs). That test's 3-master cluster makes 3 concurrent `lua-redis-wasm` `load()` calls.
- **Native stack of the hung child (macOS `sample`):**
  - The main thread is blocked in `node::NodePlatform::DrainTasks → uv_cond_wait`.
  - A Maglev background compile job is blocked in `LocalHeap::PerformCollectionAndAllocateAgain → CollectionBarrier::AwaitCollectionBackground`. It is waiting for a GC that only the main thread can run.
  - This is a V8 platform deadlock. It has the same signature as nodejs/node#54918.
- **A 10-line script reproduces it with no test runner:** `Promise.all([load(), load(), load()])` under `--import tsx` hangs about 3–5% of the time on 22.6.
- **By version:**
  - 22.6: 5/100 and 3/200 script runs hung; 1/160 test-file runs hung.
  - 22.8: 1/160 test-file runs hung (not sampled).
  - 22.12, 22.22 and 24.20: 0/240 test-file runs hung.
  - With `--no-maglev`, 22.6 had 0/360 hangs. That flag is not allowed in `NODE_OPTIONS`.
- **With this PR on 22.6:** 3/300 runs failed after the timeout with `tests/client-mocks/node-redis-mock.test.ts … 'test timed out'` and exit 1, instead of hanging.
- **Not fixed here.** The real fix is outside this repo. Options: raise the CI unit floor and/or `engines.node` from 22.6 to ≥22.12, or accept the occasional named failure on 22.6.

### `node-redis/sort-integration.test.ts`
- Not reproduced on the mock backend: 0/120 runs on 22.12, 0/120 on 24 and 0/200 on 22.6.
- I did not loop it against real Redis, because its `before()` calls `flushAll` on the shared servers.

Refs #454

## Test plan
- [x] `npm run build`, `npm run lint`
- [x] `npm test`: 462/462 on Node 22.12, and 462/462 on 22.6 with the new flag
- [x] `npm run test:integration:mock`: 1248 pass, 2 skipped
- [x] `test:integration:compatibility:mock`, `test:integration:raw:mock`, `test:package`, `test:demo` all green
- [x] On 22.6, a looped `node-redis-mock.test.ts` stall now ends as a named file timeout (exit 1) instead of hanging


AGENTS.md
```diff
@@ -26,6 +26,8 @@ node --enable-source-maps --import tsx --no-warnings --test ./tests/path/to/test
 TEST_BACKEND=real node --enable-source-maps --import tsx --no-warnings --test-concurrency 1 --test ./tests-integration/**/*.test.ts
 ```
 
+Every `test*` npm script passes `--test-timeout` (60s; 30s for the real backend), so a hung test becomes a failure instead of a stalled run (#454). On Node 24 the timeout applies to each test and names it. On Node 22 it applies to each test *file*: the file's child process is killed and only the file is named, so give a test its own `{ timeout }` if you want a hang to name it. On Node 24, a timed-out test that leaks a socket or timer can still keep its file's process alive, and the CI job's `timeout-minutes` is the backstop for that. Don't add `--test-force-exit`: on Node 22 under macOS it drops whole files from the report.
```
fatal10110: [medium] **The Node 24 half of this paragraph overstates what `--test-timeout` guarantees.**

On Node 24 the timeout is a timer inside the test file's own process. So it only fires if that process's main thread is still running. Two cases escape it:

1. **Main-thread hang (a sync loop, or the V8/Maglev platform deadlock this PR diagnoses).** The timer never fires, nothing is named, and the run stalls until the job's `timeout-minutes`.
2. **A ref'd handle left open.** The test is named as failed, but the run still stalls. The paragraph does cover this case.

This matters because nodejs/node#54918 is still open, and an Aug 2026 comment there reports the same `CollectionBarrier::AwaitCollectionBackground` cycle on stock **v24.18.1**. On Node 24, which runs every integration job, package, demo, release and the unit-24 leg, `timeout-minutes` is the only guard against the exact failure class that motivated #454.

Repro (3 s timeout, one good file alongside):

```js
// sync.test.mjs
test('sync hang', () => { while (true) {} })
test('sync next', () => {})
// ref.test.mjs
test('ref hang', async () => { setInterval(() => {}, 1000); await new Promise(() => {}) })
test('ref next', () => {})
```

```
v22.6.0  sync -> ✖ sync.test.mjs 'test timed out after 3000ms'   exit=1 after 3s
v22.12.0 sync -> ✖ sync.test.mjs 'test timed out after 3000ms'   exit=1 after 3s
v24.20.0 sync -> (nothing reported)                             killed by outer 25s alarm
v24.20.0 ref  -> ✖ ref hang (3001ms), ✔ ref next                killed by outer 25s alarm
```

Suggested fix: rewrite the Node 24 sentences along these lines. "On Node 24 the timeout is enforced inside the test process. A timed-out test is named, but if it leaks a ref'd handle, or if the main thread itself is blocked (a sync loop or a V8 deadlock), the process never exits, and the CI job's `timeout-minutes` is the only backstop." Also soften "a hung test becomes a failure instead of a stalled run", because on Node 24 that is not guaranteed.

.github/workflows/ci.yml
```diff
@@ -12,6 +12,7 @@ permissions:
 jobs:
   lint:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
```
fatal10110: [medium] **Not every CI job gets `timeout-minutes`.** The PR body says it adds `timeout-minutes` "to every CI job", but it only touches `ci.yml`. Two other workflows run test scripts on Node 24 and still use GitHub's default of 360 minutes:

- `.github/workflows/release.yml`, `publish` job: runs `npm test` and `npm run test:package` before `npm publish`. A hang there holds a release runner for 6 hours.
- `.github/workflows/deploy-demo.yml`, `build` job: runs `npm run test:demo`. Its own comment calls this "the only place `test:demo` runs", and it runs on every PR.

The PR's own analysis, confirmed in the AGENTS.md comment, is that on Node 24 `--test-timeout` cannot end a run whose process stays alive. `timeout-minutes` is the backstop for that case, and these two jobs don't have it.

Suggested fix: add `timeout-minutes: 15` to `release.yml` → `publish` and `deploy-demo.yml` → `build`. Optionally add a small value to `deploy-demo.yml` → `deploy` too. Or narrow the PR description to say "every job in ci.yml".

.github/workflows/ci.yml
```diff
@@ -35,6 +36,9 @@ jobs:
   unit:
     name: unit (node ${{ matrix.node-version }})
     runs-on: ubuntu-latest
+    # Outer backstop only: the npm test scripts pass --test-timeout, so a hung
+    # test fails by name long before this (#454).
+    timeout-minutes: 15
```
fatal10110: [low] **I reproduced the diagnosis, with one caveat for the "raise the floor to ≥22.12" decision.** No code change is needed. This is about how the PR description frames the fix.

I ran the PR's own script, `Promise.all([load(), load(), load()])` under `--import tsx`, 20s cap per run, with all three versions running at the same time on macOS arm64:

```
v22.6.0   runs=500 hung=14  (2.8%)
v22.12.0  runs=300 hung=0
v24.20.0  runs=300 hung=0
```

With 22.6 as a live control hanging at ~3%, getting 0/300 on 22.12 by chance has a probability of about e^-10. So this trigger really does not occur on 22.12+ or 24.

The caveat: that result is empirical, not a known upstream fix.
- nodejs/node#54918 is **still open**.
- The mitigation, nodejs/node#58047 ("src: only block on user blocking worker tasks"), landed in **v24.0.0** and was **not** backported to v22.x. Only its debug-log and test commits appear in 22.16.0's changelog.
- An Aug 2026 comment on the issue reports the same Maglev + `AwaitCollectionBackground` cycle on stock **v24.18.1**.

So raising the floor to 22.12 removes this particular trigger. It does not remove the deadlock class, and Node 24 is not immune. It would help if the PR description said this, so the owner's pending floor decision rests on the right basis. This isn't blocking.

.github/workflows/ci.yml
```diff
@@ -35,6 +36,9 @@ jobs:
   unit:
     name: unit (node ${{ matrix.node-version }})
     runs-on: ubuntu-latest
+    # Outer backstop only: the npm test scripts pass --test-timeout, so a hung
+    # test fails by name long before this (#454).
```
fatal10110: [low] **This comment is not accurate on either matrix leg.** It says a hung test "fails by name long before this".

- **22.6:** the failure names only the *file*. The PR's own AGENTS.md text and the updated `node-redis-mock.test.ts` comment both say so.
- **24:** a ref'd-handle hang is named, but the job still runs until this 15-minute cap. A main-thread hang, such as a V8 deadlock, is never named. See the AGENTS.md comment for the repro.

Suggested wording: "Outer backstop: on Node 22, `--test-timeout` kills a hung test file after 60s. On Node 24 it is enforced in-process, so a hang that keeps the process alive only ends here (#454)."

.github/workflows/ci.yml
```diff
@@ -90,6 +95,8 @@ jobs:
   integration:
     name: ${{ matrix.name }}
     runs-on: ubuntu-latest
+    # Covers the Redis startup budget (--wait-timeout 420 below) plus the run.
+    timeout-minutes: 25
```
fatal10110: [low] **A job-level timeout cancels the job, so the `#398` diagnostics step won't run.** When `timeout-minutes` fires at the job level, GitHub cancels the job. The later `Dump Redis container state and logs on failure` step is gated on `if: failure() && matrix.needsRedis`, and `failure()` is false for a cancelled job. So a real-backend run that hangs mid-test and hits this cap loses the Redis logs that might explain the hang. Only the `always()` teardown runs.

Startup can't reach this cap: it is bounded at 420s and *fails* rather than cancels. So this only affects hangs during the test phase.

Suggested fix: also put a step-level `timeout-minutes` (e.g. 15) on `Run integration tests (${{ matrix.name }})`. A step timeout fails the step, so the dump still runs, and the 25-minute job cap stays as the outer bound. Alternatively, gate the dump on `failure() || cancelled()`.

AGENTS.md
```diff
@@ -26,6 +26,8 @@ node --enable-source-maps --import tsx --no-warnings --test ./tests/path/to/test
 TEST_BACKEND=real node --enable-source-maps --import tsx --no-warnings --test-concurrency 1 --test ./tests-integration/**/*.test.ts
```
fatal10110: [low] **The copy-paste commands above don't match the new sentence.** The paragraph below says every `test*` npm script has a timeout, but the two raw commands in this block have none. The "Run integration tests sequentially (needed for real Redis backend)" command drops the `--test-timeout 30000` that `test:integration:real` uses, and the "Run a single test file" command has no timeout either. Someone who pastes these gets exactly the unbounded run that #454 is about.

Suggested fix: add `--test-timeout 30000` to the real-backend line and `--test-timeout 60000` to the single-file line, so the block matches the scripts.

fatal10110:
## Review of `be55697`: **NOT YET**

I checked every test entry point. `ci.yml`, `release.yml`, `deploy-demo.yml` and the husky `pre-push` hook all go through the npm scripts, and every non-real `test*` script now passes `--test-timeout 60000`. The real-backend scripts keep 30000. The `timeout-minutes` values are reasonable next to the observed 20s–1m47s job times, and the Node 22 description in the PR is accurate. The two open medium findings are both about Node 24, where `--test-timeout` can't be relied on to end a run.

| Sev | Where | Finding | Lens |
|---|---|---|---|
| medium | AGENTS.md:29 | On Node 24 the timeout runs inside the test process. A main-thread hang (a sync loop, or the V8 deadlock itself) is never timed out and never named, and a ref'd-handle hang is named but still stalls the run. The docs only mention the second case, and "a hung test becomes a failure" doesn't hold on 24. | architect |
| medium | ci.yml:15 | The PR says "timeout-minutes on every CI job", but `release.yml` (`publish`: `npm test`, `test:package`) and `deploy-demo.yml` (`build`: `test:demo`) still have the 360-minute default. Both run Node 24, where `timeout-minutes` is the only backstop. | architect |
| low | ci.yml:41 | The diagnosis reproduces, but the 22.12+ result is empirical, not an upstream fix. nodejs/node#54918 is still open, #58047 went into 24.0.0 and was never backported to 22.x, and there is a report on v24.18.1. This is not blocking; the floor decision is the owner's. | architect |
| low | ci.yml:40 | The comment "fails by name long before this" is wrong on both legs: on 22 it names only the file, and on 24 it may not fire at all. | architect |
| low | ci.yml:99 | A job-level timeout cancels the job, so the `failure()`-gated Redis log dump (#398) is skipped. A step-level timeout on the test step would keep it. | architect |
| low | AGENTS.md:26 | The copy-paste raw `node --test` commands in the same block have no `--test-timeout`. | architect |

Wire-compat lens: N/A. There are no protocol or command changes.

**What I checked:**
- **`--test-timeout` behaviour (3s timeout):** on 22.6 and 22.12, both a sync hang and a ref'd hang end with `✖ <file>.test.mjs 'test timed out'` and exit 1 after 3s. On 24.20, the ref'd hang is named but the run was only killed by an outer 25s alarm, and the sync hang reported nothing before that alarm.
- **Stall repro** (`Promise.all([load(),load(),load()])` under tsx, run concurrently across versions): 22.6 hung 14/500, 22.12 hung 0/300, 24.20 hung 0/300.
- **PR head test runs:**
  - `npm test`: 462/462 on 24.20, and exit 0 on 22.6.
  - `test:integration:mock`: 1248 pass and 2 skipped on 24.20, and exit 0 on 22.12.
  - `test:package`: 8/8. `test:demo`: 12/12.

**Not verified:**
- The claim that `--test-force-exit` drops files on macOS Node 22.
- That GitHub marks a job-level timeout as cancelled (not failed), so `failure()` is false. This is from the docs, not a live run.

AGENTS.md
fatal10110: Fixed in 4c83b4c. The paragraph is now split by Node version. On Node 24 it says the timeout runs inside the test process: a timed-out test is named, but a leaked ref'd handle or a blocked main thread (a sync loop, or a V8 deadlock like nodejs/node#54918) keeps the process alive, and the job's `timeout-minutes` is the only backstop. I removed the unconditional "a hung test becomes a failure" wording. The Node 22 bullet now also notes that a per-test `{ timeout }` runs in-process, so it can't name a blocked main thread.

.github/workflows/ci.yml
fatal10110: Fixed in 4c83b4c. `release.yml` `publish` now has `timeout-minutes: 20` because it also builds and publishes. `deploy-demo.yml` `build` has 15 and `deploy` has 10. Every job in every workflow now has a timeout.

.github/workflows/ci.yml
fatal10110: Agreed. There was no matching text in ci.yml, so I fixed it in the PR description. It now says that 0 hangs on 22.12+ is an observation, not an upstream fix: #54918 is still open, #58047 shipped in 24.0.0 without a 22.x backport, and the deadlock has been reported on 24.18.1. So raising the floor removes this trigger but not the deadlock class. I left the floor decision to the owner.

.github/workflows/ci.yml
fatal10110: Fixed in 4c83b4c. I used your wording: on Node 22 the timeout kills a hung test file after 60s; on Node 24 it runs in-process, so a hang that keeps the process alive only ends at this cap.

.github/workflows/ci.yml
fatal10110: Fixed in 4c83b4c. The `Run integration tests` step now has its own `timeout-minutes: 15`, below the job's 25, with a comment explaining why. A step timeout fails the step, so the `failure()`-gated Redis log dump still runs.

AGENTS.md
fatal10110: Fixed in 4c83b4c. The single-file command now has `--test-timeout 60000` and the real-backend command has `--test-timeout 30000`, matching the npm scripts.

.github/workflows/ci.yml
```diff
@@ -90,6 +97,9 @@ jobs:
   integration:
     name: ${{ matrix.name }}
     runs-on: ubuntu-latest
+    # Outer bound: the Redis startup budget (--wait-timeout 420 below), plus
+    # the test step's own 15-minute cap, plus the log dump and teardown.
+    timeout-minutes: 25
```
fatal10110: [low] **The 25-minute job cap is close to its own worst case.** The comment says this bound covers checkout, `npm ci`, the image pull, the Redis wait, the test step, the dump and teardown. Adding those up in the worst case:

- checkout, setup-node and `npm ci`: about 1.5 min
- `docker compose up` image pull: up to about 1 min. This comes before `--wait-timeout` starts counting.
- Redis wait: up to 7 min (420s)
- test step: up to 15 min
- dump and teardown: about 1 min

That totals roughly 25 to 25.5 minutes. So in the one case where the dump matters most (a slow cluster formation followed by a hung test run), the job-level cancel can still land before or during the `failure()`-gated dump. This needs two failures at once and never happens in normal runs (the real jobs take about 1.5 min), so it is not blocking.

Suggested fix: raise the job cap to 30, or lower the step cap to 12, so the comment's ordering holds with some margin.

fatal10110:
## Re-review of `4c83b4c`: **SHIP**

Prior findings (reviewed at `be55697`):

| Sev | Where | Status |
|---|---|---|
| medium | AGENTS.md:29 | **Fixed.** The Node 24 bullet now covers both ways a run stalls (a leaked ref'd handle, and a blocked main thread / #54918). `timeout-minutes` is named as the only backstop. |
| medium | ci.yml:15 | **Fixed.** release `publish` has 20m, deploy-demo `build` 15m and `deploy` 10m. `grep runs-on\|timeout-minutes` shows all 8 jobs in the 3 workflows now have a cap. |
| low | ci.yml:41 | **Fixed.** The PR body now has a "Caveat on 22.12+" section covering #54918 being open, #58047 being 24-only, and the v24.18.1 report. |
| low | ci.yml:40 | **Fixed.** The comment is reworded per Node version. |
| low | ci.yml:99 | **Fixed.** The test step has a 15m `timeout-minutes`, below the job's 25m, so the `failure()` dump can still run. |
| low | AGENTS.md:26 | **Fixed.** The copy-paste commands now carry 60000 / 30000. |

New finding:

- low | ci.yml:102 | The 25m job cap is only about 0 to 0.5 min above its own worst-case sum (setup, image pull, the 420s wait, the 15m step, then dump and teardown). Suggest 30. Not blocking.

The step cap is not too tight: 15m against observed real-job runs of 44s to 1m47s.

Checks:
- actionlint is clean.
- `package.json`, the tests and `src/` are unchanged since `be55697`, so the earlier test results still apply: `npm test` 462/462 on 24.20 and exit 0 on 22.6, `test:integration:mock` 1248 pass / 2 skipped, `test:package` 8/8, `test:demo` 12/12.

Still unverified: I have not seen a live GitHub run confirm that a step-level timeout fails the step (rather than cancelling it). That is taken from the docs.

The Node 22.12 floor is still an open owner decision and does not block this.

.github/workflows/ci.yml
fatal10110: Fixed in 3f9098e. The integration job's `timeout-minutes` is raised from 25 to 30, which leaves headroom over setup + image pull + the 420s Redis wait + the 15-minute test step + the log dump. I also updated the test step's comment to say "30-minute cap".

